PR #98, opened 2013-05-17
Using multiple DSNs as part of a failover mechanism

kaiprice:
...ailover list. Upon successfully transmitting data, it will not attempt to proceed through the list. If it were to fail then it would not proceed.

Also suppressing the UDP socket_sendto in send_remote so that it behaves more like send_http, where it will collect the last error and return false if it fails.


stayallive:
Since this is very old I am closing this PR.

Feel free to open a new PR or an issue so we can address it if it's still relevant.

PR #125, opened 2013-11-09
fields cvv, cardNumber and expirationDate added to sanitization + tests ...

gfilipiak:
...+ phpStorm gitignore cofniguration

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.lock
 package.xml
 /vendor
+.idea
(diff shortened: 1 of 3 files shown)


ste93cry:
As this is a pretty old issue and I think that it would be better to not duplicate the logic to handle sanitization on each SDK, I'm going to close it. With version `2.0` (which is still in development at the time of writing) we have introduced the concept of middlewares, so you can customize the data being sent before it leaves your application. Also, you could have a look at the [Relay](https://docs.sentry.io/relay/?platform=javascript) project which is a proxy which stands between the SDKs and the Sentry server and does the sanitization

Issue #108, opened 2013-07-28
Investigate curl_multi for async


dcramer:
It seems like we could use the curl_multi stuff to do async http requests with fairly large compatibility

http://us3.php.net/manual/en/function.curl-multi-init.php

ste93cry:
Closing this as it's been fixed in #132

PR #132, opened 2014-02-06
Async (non-blocking) option via executing curl through exec()

kyle-johnson:
This is one way of addressing #108 and #96.

We use this across many sites logging upwards of 20k events per hour to Sentry. If there is interest in this patch, I will update it for automatic merging.

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -16,7 +16,7 @@
 
 class Raven_Client
 {
-    const VERSION = '0.7.1-exec';
+    const VERSION = '0.7.1';
     const PROTOCOL = '3';
 
     const DEBUG = 'debug';


adamlc:
@kyle-johnson thanks for writing this. I'm not sure if its easy to fix but this seems to leave loads of zombie processes that aren't being cleaned up properly. I'm using this in a ReactPHP event loop.

```
root      9209  0.0  0.0      0     0 ?        Z    11:02   0:00 [curl] <defunct>
root      9253  0.0  0.0      0     0 ?        Z    11:02   0:00 [curl] <defunct>
root      9454  0.0  0.0      0     0 ?        Z    11:03   0:00 [curl] <defunct>
root     10208  0.0  0.0      0     0 ?        Z    11:07   0:00 [curl] <defunct>
root     10239  0.0  0.0      0     0 ?        Z    11:07   0:00 [curl] <defunct>
root     10248  0.0  0.0      0     0 ?        Z    11:07   0:00 [curl] <defunct>
root     10319  0.0  0.0      0     0 ?        Z    11:08   0:00 [curl] <defunct>
```

Not sure if its an easy fix or not? It might be nice for the Raven library to allow for pluggable transports so that we can write our own

kyle-johnson:
@adamlc I wonder if the behavior depends on how PHP is run. I'm using this with php-fpm, and while I have max_requests set to 500, I don't see zombies in the process tree (although I do see a fairly steady stream of errors being logged to Sentry).

Looking at the PHP docs, there doesn't appear to be a way to modify exec's behavior, but I think you could use the pcntl functions (http://php.net/manual/en/book.pcntl.php) to launch the curl process with a bit more control. Perhaps PHP is not ignoring SIGCHILD (that would certainly result in the child processes sticking around as zombies)?

adamlc:
@kyle-johnson thats probably the case. I think it might be better to do a pluggable transport interface so that people can make their own, shouldn't be too difficult. I'll have a stab at it this week!

popstas:
@kyle-johnson, I compared performance of async and exec methods and I see that both methods works well with long response DSN URLs:

``` php
$time = microtime(true);

$client = new Raven_Client('http://user:pass@slow.test.dsn.ru/slow.php', [
    'curl_method' => 'exec'
]);

$event_id = $client->getIdent($client->captureMessage('my log message'));

$time = microtime(true) - $time;

echo "time: $time\n";
```

URL http://slow.test.dsn.ru/slow/index.php just sleeps for 3 seconds and returns 200 with empty body.

With method `async` script return ~3 ms.
With method `exec` script return ~5 ms.

What's the advantage of exec over async?

Issue #168, opened 2014-08-22
Raven_ErrorHandler::handleFatalError() handles E_STRICT messages


raebbar:
Raven_ErrorHandler::handleFatalError() handles E_STRICT errors and not only fatal errors. 

Only errors of the types E_ERROR | E_PARSE | E_CORE_ERROR | E_CORE_WARNING | E_COMPILE_ERROR | E_COMPILE_WARNING should be handled

justiniso:
+1 can we get this change merged into the official library?

madssj:
A solution to this was implemented in bc88ec3a9fa14df4f36ff9ca08b0396b855db4fd and f89ecb3c3a58d5dce8812d5adaeadab8c56c7918 and released in `0.12.0`.

madssj:
Also see #181.

jamieburchell:
I've just run in to this one. I'm getting strict notices reported from the handleFatalError function, but I've set up the client like so:

```
$handler = new Raven_ErrorHandler($client, false, null, E_ALL & ~E_NOTICE & ~E_STRICT);
$handler->registerExceptionHandler();
$handler->registerErrorHandler();
$handler->registerShutdownFunction();
```

Is my understanding of how this should work incorrect, or should the client only report on those types passed in to the constructor?

dcramer:
@jamieburchell it looks like handleFatalError doesn't correctly deal w/ the changes to the newer types. Frankly, we need to go back over this, remove the old option, and ensure all things are using the new "these really are the only error codes I want to capture" bits.

dcramer:
@jamieburchell short term work around, try passing that same value where you have null (the third arg)

jamieburchell:
I've just run in to this issue again and stumbled back here :)

Error reporting is actually turned off in my case (`error_reporting(0)`) but I'm still getting the strict/info messages appearing in Sentry. `handleFatalError` is not factoring in the current error reporting level like `handleError` is. For now I got around it using:

```
$handler = new Raven_ErrorHandler($this->_client, false, [
	E_ERROR,
	E_PARSE,
	E_CORE_ERROR,
	E_CORE_WARNING,
	E_COMPILE_ERROR,
	E_COMPILE_WARNING
]);
```

PR #181, opened 2014-12-07
Added an option to configure error types for registerShutdownFunction

pauloschilling:
The default error list includes E_STRICT, which is not something that everyone will need to use.

There's also an issue opened for it (see #168), so basically I've kept the current list as default and added an extra parameter on registerShutdownFunction to overwrite it.

### DIFF
--- a/lib/Raven/ErrorHandler.php
+++ b/lib/Raven/ErrorHandler.php
@@ -30,9 +30,11 @@ class Raven_ErrorHandler
     private $reservedMemory;
     private $send_errors_last = false;
     private $error_types = -1;
+    private $error_types_shutdown;
 
     public function __construct($client, $send_errors_last = false)
     {
+        $this->error_types_shutdown = E_ERROR | E_PARSE | E_CORE_ERROR | E_CORE_WARNING | E_COMPILE_ERROR | E_COMPILE_WARNING | E_STRICT;
         $this->client = $client;
         register_shutdown_function(array($this, 'detectShutdown'));
         if ($send_errors_last) {
@@ -71,9 +73,7 @@ class Raven_ErrorHandler
 
         unset($this->reservedMemory);
 
-        $errors = E_ERROR | E_PARSE | E_CORE_ERROR | E_CORE_WARNING | E_COMPILE_ERROR | E_COMPILE_WARNING | E_STRICT;
-
-        if ($lastError['type'] & $errors) {
+        if ($lastError['type'] & $this->error_types_shutdown) {
             $e = new ErrorException(
                 @$lastError['message'], @$lastError['type'], @$lastError['type'],
                 @$lastError['file'], @$lastError['line']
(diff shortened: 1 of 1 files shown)


madssj:
Hey @dcramer, this would be really useful to have merged. Is there any technical reason for not merging it into the official client?

@pauloschilling care to update the PR to resolve the merge conflicts? :)

dcramer:
This gets a bit more complex with our previous change to allow you to configure error types in general. Do we really still need different types for shutdown?

madssj:
That'd be the changes made in bc88ec3a9fa14df4f36ff9ca08b0396b855db4fd and f89ecb3c3a58d5dce8812d5adaeadab8c56c7918.

I'm currently using `0.11.0` as it was the version composer installed, but the changes from before should solve the same problem, possibly better/cleaner which have been released as of `0.12.0`.

bc-shusson:
Any updates on this? In particular being able to exclude E_STRICT from being handled and sent.

madssj:
I ended up writing my own handler for fatal errors, as the proposed changes from above cannot remove an error from being reported, it can only add errors.

Here's my temporary handler:

```php
// global reserved memory chunk for fatal errors
$my_sentry_reserved_memory = str_repeat('x', 1024 * 10);

function my_sentry_handle_fatal_error() {
	if (null === $lastError = error_get_last()) {
		return;
	}
	unset($my_sentry_reserved_memory);

	$errors = E_ERROR | E_PARSE | E_CORE_ERROR | E_CORE_WARNING | E_COMPILE_ERROR | E_COMPILE_WARNING;

	if ($lastError['type'] & $errors) {
		$e = new ErrorException(
			@$lastError['message'], @$lastError['type'], @$lastError['type'],
			@$lastError['file'], @$lastError['line']
		);
		$sentry_client->getIdent($sentry_client->captureException($e, null, null, null));
	}
}
register_shutdown_function('my_sentry_handle_fatal_error');
```

@dcramer I really think the error types need to be fully customisable. It shouldn't be up to sentry to decide what errors I have to see in this context.

dcramer:
Sorry we've been distracted with getting a long-awaited release of the new platform out the door. I agree that we shouldn't dictate everything, but have sane defaults. That said, I dont have time to review this until we get some other things off our plates.

dcramer:
Actually just skimmed it since its small.

My biggest complaint here is that we're changing configuration of the handler every time we register the shutdown handler.

I'd like to see either:

1. The shutdown_error_types are specified as construction time of the ErrorHandler
2. The shutdown hook is handled as its own isolated, contained instance

madssj:
@pauloschilling are you planning to update the PR, or should I create a new one?

pauloschilling:
Sorry, we moved our projects to Rails, and I'm not using raven-php anymore, so I don't think I'll have any free time to update this PR or even to contribute on this discussion. :(

Please, @madssj @dcramer, feel free to create a new PR with a more appropriate implementation if you think this will be useful. :)

Best!

madssj:
I'm not using `raven-php` anymore either, but I still feel like this would be worth contributing back to @dcramer. I'll have a look at a new PR within a couple of weeks. Feel free to close this one.

NanneHuiges:
Is there any update on this? #168 points to this as well. 

My current 'assessment' is:
* there isn't a good way to remove error types from getting handled (you can only add them)
* Currently there is a function called `handleFatalError` and by default it handles all sorts of non-fatal errors. This is weird :)

If there is any work to be done I wouldn't mind doing it, but I kinda want to know if there is something in the air somewhere allready?

dcramer:
It's possible to override it today by subclassing the handler and overriding getErrorTypesToProcess

NanneHuiges:
Yeah, I will, but I thought I might give some effort back to the project :)
But you're right, a simple subclass would do the trick, I'll try that then.

ste93cry:
This should be (hopefully) fixed in the `2.0` version, however if it's not the case we may rediscuss it in a new issue. Closing this for now as it's quite old and as far as I know there are no plans to do it in the `1.x` series

PR #183, opened 2014-12-09
Update dev branch alias

piotrantosik:


marten-cz:
It will be good to merge this change. It doesn't make sense to developing 0.8 when 0.11 is out. It's messing with dependencies via composer.
I'm not sure if you can remove the section completely and it will work as you want or not, but I'll have a look. For now, it will be nice to merge this, it will not damage anything...

PR #185, opened 2014-12-30
Handle CLI scripts

iw2:
Detect cli scripts and avoid undefined index.
Fix issues : 
- https://github.com/getsentry/raven-php/issues/113 
- https://github.com/getsentry/raven-php/issues/136 
- https://github.com/getsentry/raven-php/issues/97


dcramer:
I'm going to close this as it isn't entirely valid still

PR #186, opened 2015-01-12
Create a backwards compatible way to override Raven_Processor sub-class ...

dkoston:
There's probably a more elegant way to do this but I wanted to provide backwards compatibility for those already passing in $options = array('processors'=>array($x));

Create a backwards compatible way to override Raven_Processor sub-class properties so that SanitizeDataProcessor can be modified via $options when creating a new Raven_Client instance:

	lib/Raven/Client.php:
		add PHPDoc comments for many functions
		abstract out setting of processors to setProcessorsFromOptions()
		allow for Raven_Processor->setProcessorOptions() to be called if defined in the Raven_Processor() sub-class
		allow one to set $this->processors with setProcessors()
	lib/Raven/Processor.php:
		make class abstract as it has no functionality
		make process abstract to ensure that sub-classes implement it
	lib/Raven/SanitizeDataProcessor.php:
		convert static $mask,$fields_re,$values_re to const MASK,FIELDS_RE,VALUES_RE
		allow overriding $this->fields_re and $this->values_re with setProcessorOptions()
		add getters/setters for $fields_re and $values_re class variables
	test/Raven/Tests/ClientTest.php:
		Use processor Raven_SanitizeDataProcessor in testDoesRegisterProcessors() now that Raven_Processor is abstract
	test/Raven/Tests/SanitizeDataProcessorTest.php:
		Add testSettingProcessorOptions() to test new setProcessorOptions() function
		Add testOverrideOptions() to test setProcessorOptions() when called via __construct of Raven_Client
		Add testOverridenSanitize() to test that overriding sanitizers work
	README.rst:
		Add processors and processorOptions to docs

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -199,6 +199,25 @@ Defaults to 1024 characters.
 
 This value is used to truncate message and frame variables. However it is not guarantee that length of whole message will be restricted by this value.
 
+``processors``
+~~~~~~~~~~~~~~~~~
+
+An array of classes to use to process data before it is sent to Sentry. By default, Raven_SanitizeDataProcessor is used
+
+``processorOptions``
+~~~~~~~~~~~~~~~~~
+Options that will be passed on to a setProcessorOptions() function in a Raven_Processor sub-class before that Processor is added to the list of processors used by Raven_Client
+
+An example of overriding the regular expressions in Raven_SanitizeDataProcessor is below:
+
+.. code-block:: php
+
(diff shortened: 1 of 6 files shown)


dcramer:
I think at some point we're going to change how processors work so that we move away from the complexities of building classes, and ideally just provide a list of functions, but as I dont have a better solution for this it seems like a reasonable fix.

PR #187, opened 2015-01-22
- adding option for forcing ipv4

lechbaczynski:
Was having problem with timeout when sending message. Found that it was due to curl using ipv6. 
Forcing use of ip v4 resolving like this:

      curl_setopt($curl, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V4 );

near line 606-610 in Client.php, just before:

        $options = $this->get_curl_options();
        $ca_cert = $options[CURLOPT_CAINFO];
        unset($options[CURLOPT_CAINFO]);
        curl_setopt_array($curl, $options);
        curl_exec($curl);

fixed the problem. But it is more elegant to use an option for this, like:

       $client = new Raven_Client('https://.....',array('curl_ipv4'=>true));

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -69,6 +69,7 @@ class Raven_Client
         $this->send_callback = Raven_Util::get($options, 'send_callback', null);
         $this->curl_method = Raven_Util::get($options, 'curl_method', 'sync');
         $this->curl_path = Raven_Util::get($options, 'curl_path', 'curl');
+        $this->curl_ipv4 = Raven_util::get($options, 'curl_ipv4', false);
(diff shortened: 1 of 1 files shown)


dcramer:
Thanks!

lechbaczynski:
My pleasure :)

lechbaczynski:
If I understand correctly, you also merged the latest change that sets forcing ipv4 by default. 
Was it your real intention?

dcramer:
@lechbaczynski I definitely merged the entire request. Maybe we want to backout the default IPV4 change. I personally dont have a lot of experience with IPV6, so I don't have a strong opinion on the importance.

PR #189, opened 2015-01-29
Handle empty HTTP_HOST values

thomasbachem:
The `Host` HTTP header is optional in HTTP 1.0 and can therefore be empty, meaning the current code raised some `Notice: Undefined index: HTTP_HOST` errors on my installation. I adjusted `get_current_url()` to fall back on `LOCAL_ADDR` (IIS) or `SERVER_ADDR` if `HTTP_HOST` is empty.


dcramer:
A couple months late here, but thanks!

PR #191, opened 2015-02-27
Added verify_ssl option

Kras4ooo:
Added a `verify_ssl` as in Version 4.1.0 of the raven: [Raven 4.1.0] (http://raven.readthedocs.org/en/latest/changelog/#version-4-1-0)


dcramer:
Thanks!

PR #193, opened 2015-03-10
sanitize $_SERVER['PHP_AUTH_PW']

pschultz:
The http basic auth credentials are not sanitized in the _SERVER array.


dcramer:
Thanks!

PR #194, opened 2015-03-24
Pass the return value of the original error handler on

mcuelenaere:
As specified at http://php.net/manual/en/function.set-error-handler.php,
PHP reacts differently whether false is returned or not. This gives the
original error handler the possibility to influence this value.


mcuelenaere:
Seems like the unit tests are failing even without this commit, so Travis output should be ignored.

dcramer:
Thanks!

mcuelenaere:
Any idea when this will get released?

dcramer:
We just tagged 0.11.0 which includes this change

PR #197, opened 2015-04-03
[1.0] Move to PSR-4 for version 1.0 and drop PHP 5.2 support

ameliaikeda:
Making the pull request now (despite barely even being started) just for some kind of review process as I go. I can also make a public todo list of sorts.

This pull request:
- [ ] Removes PHP 5.2 hacks
  - [x] Removed Raven_Compat entirely
  - [x] Refactored Raven_Util into Raven\get
  - [x] Use composer for file loading exclusively
  - [ ] Other unseen hacks are todo
- [x] Moves to PHPUnit 4.5
- [ ] Uses interfaces for all major classes that we'd like to test (so we can mock them)
  - [x] `Raven\Contracts\Client`
  - [ ] `Raven\Contracts\Handler` (CurlHandler, GuzzleHandler, etc)
- [ ] Uses docblocks literally everywhere (think of the IDEs ;_;)
- [ ] Uses PSR-1, PSR-2 and PSR-4

I also want to allow for other methods of transport to be used. For example, I could pull in `raven/raven-guzzle` (etc), And use the Guzzlehttp library instead of `Raven\Handlers\CurlHandler`.

This is a work in progress. Feel free to completely rip anything apart in the comments or on diffs.


localheinz:
@ameliaikeda 

:+1: for dropping PHP5.2 support!

Had a look at the PR since @dcramer suggested it in https://github.com/getsentry/raven-php/pull/202#issuecomment-100559504, and have a few thoughts!

What do you think about breaking the PR apart into a bunch of smaller PRs - it would have the advantage of being easier to review, and thus, faster to integrate into `master`!

Here's what I would do (each of these could be a separate PR)

* drop PHP5.2 from the build matrix (along with removing the newly added conditions in `.travis.yml`)
* deprecate classes/methods used to hack around missing features in PHP5.2
* remove deprecated classes/methods (would need a tag)
* switch to PHP5.3 namespaces
* switch to PSR-4 autoloading

Basically, all of these changes should be possible without having to make a major release yet, see http://semver.org:

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

With all of these things above done, I would probably move forward and tag with `1.0.0`. 

> Version 1.0.0 defines the public API. The way in which the version number is incremented after this release is dependent on this public API and how it changes.

composer.json
```diff
@@ -12,10 +12,11 @@
         }
     ],
     "require-dev": {
-        "phpunit/phpunit": "3.7.*"
+        "phpunit/phpunit": "4.5.*"
```
localheinz: See #201.

composer.json
```diff
@@ -24,9 +25,12 @@
         "bin-dir": "bin"
     },
     "autoload": {
-        "psr-0" : {
-            "Raven_" : "lib/"
-        }
+        "psr-4" : {
+            "Raven\\" : "lib/"
+        },
```
localheinz: :+1:

ameliaikeda:
@localheinz I agree with pretty much all of this.

I've been pretty busy the last few weeks (deadlines are fun), but i'll do them in smaller steps. I had to make some pretty giant changes to make everything fully mockable, so I'll start with new pull requests and reference them here.

I haven't kept up with the travis issue, so I didn't realise it was conflicting now. I'll fix that separately when redoing the travis.yml build matrix.

Thanks~

ameliaikeda:
@dcramer would it be possible to use Symfony's `FlattenException::create(Exception $e)` method to serialize exceptions? It flattens all of the arguments for an exception down and adds a `toArray()` method to the exception.

https://github.com/symfony/Debug/blob/master/Exception/FlattenException.php

dcramer:
@ameliaikeda it's likely we could pull it in, though we'd want to rip out anything not useful, as well as ensure it conforms to the Sentry upstream interface. I wouldn't want it to be a 3rd party dep though.

ameliaikeda:
I'll make a sentry-specific version of it; it's MIT licensed so it can be used however we like provided we attribute it.

nokitakaze:
I think this is obsolete now.

PR #200, opened 2015-05-09
Fix: Remove bin-dir configuration

localheinz:
This PR

* [x] removes the useless `bin-dir` configuration

See https://getcomposer.org/doc/04-schema.md#config.

> `bin-dir`: Defaults to `vendor/bin`. If a project includes binaries, they will be symlinked into this directory.

:bulb: Leaving it to the default value has the advantage that you don't need to add rules to `.gitignore`, as the otherwise configured directory is the same directory in which the `raven` binary is provided in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@
 package.xml
 /vendor
 .idea
-bin/phpunit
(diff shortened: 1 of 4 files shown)


localheinz:
On a separate note, if you're really lazy and want to make it easier to run tests, then just create an alias like this:

```bash
alias p="vendor/bin/phpunit"
```

PR #202, opened 2015-05-09
Fix: Use the same version of phpunit/phpunit on Travis as required

localheinz:
This PR updates `.travis.yml` 

* [x] to install the dependencies as required in `composer.json` in the `install` section
* [x] to run the required version of `phpunit/phpunit` in the `script` section
* [x] falls back to using the pre-installed version of `phpunit/phpunit` when running on PHP5.2

:bulb: Using the same version decreases the risk of unexpected results because of using different versions locally and on Travis. The current version used on Travis is `4.5.0` (see https://travis-ci.org/getsentry/raven-php/jobs/61914157#L97), the currently required version is `3.7.*`.

Related to #201.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,16 @@ php:
   - 5.5
   - 5.6
   - hhvm
- 
+
 matrix:
     allow_failures:
         - php: hhvm
+
+install:
+  - composer install
+
+script:
+  - bin/phpunit --configuration phpunit.xml.dist
 
 notifications:
   irc:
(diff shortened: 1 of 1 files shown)


localheinz:
Uh, I see - you're still supporting PHP 5.2. Closing.

localheinz:
@dcramer 

Re-opened with a fallback for PHP5.2.

Do you have any plans for dropping PHP5.2 support?

dcramer:
There's some work being done in GH-197 which would drop 5.2 support (though I'm not sure the status on it)

PR #204, opened 2015-05-09
Fix: Update readme in regard to installing with Composer

localheinz:
This PR

* [x] updates `README.rst` in regard to installing with Composer

:bulb: You shouldn't need to touch `composer.json`, unless you want @rdohms to RELEASE THE KRAKEN.


dcramer:
Thanks!

PR #209, opened 2015-05-10
Drop PHP 5.2 from the travis build matrix and add 7.0

ameliaikeda:
PHP 7.0 builds are allowed to fail, but it's nice to have them in there for the future; PHP is aiming for a Q3/4 2015 release.

Ideally, the tool used for tests should be versioned and used exclusively (as if anything were to ever change in these tools on travis, we'd not get devs running tests, having them pass, and then having travis report failures/errors).

This is part of the changes to Raven for an eventual 1.0.0 release, see #197.


ameliaikeda:
It's interesting how PHP 7.0 and HHVM both have the same results for the exact same tests; should maybe have a look at that when fixing.

dcramer:
Thanks!

PR #211, opened 2015-05-12
Fix: Use composer for autoloading resources in tests

localheinz:
This PR
- [x] updates the test bootstrapping to use Composer autoloading

### DIFF
--- a/test/bootstrap.php
+++ b/test/bootstrap.php
@@ -13,5 +13,4 @@ error_reporting(E_ALL | E_STRICT);
 
 session_start();
 
-require_once dirname(__FILE__).'/../lib/Raven/Autoloader.php';
-Raven_Autoloader::register();
+require_once __DIR__ . '/../vendor/autoload.php';


dcramer:
One concern I have with this is it no longer tests our autoloader

localheinz:
Good point. 

Maybe switch to Composer autoloading when removing the Raven autoloader?

localheinz:
@dcramer 

Do you have any plans for totally dropping PHP5.2 support? If so, I'd be happy to chime in with some work!

dcramer:
The autoloader is required when you're not using composer still unless I'm mistaken? We don't enforce composer as a requirement so we don't want to cripple installs where it's not being used.

dcramer:
#197 has some effort towards dropping 5.2

ste93cry:
I think it makes no sense anymore to not use Composer to install dependencies in a project, so I vote for merging this PR. Also, in doing it we should drop entirely the custom autoloader. This will make things easier when updating to PSR-4 standard.

localheinz:
@dcramer 

I opened this PR almost two years ago. 

Do you have any intention to move forward here?

For reference, please also take a look at:

* http://php.net/supported-versions.php
* https://seld.be/notes/php-versions-stats-2016-2-edition
* https://seld.be/notes/php-versions-stats-2016-1-edition
* https://seld.be/notes/php-versions-stats-2015-edition
* https://seld.be/notes/my-view-of-php-version-adoption

localheinz:
@dcramer 

Also, composer has been around for almost 5 years. If anyone isn't using composer in a PHP project, then it's easier to small, or headed nowhere.

ste93cry:
I have to admit that I know a few people that do not use Composer... Also, removing the custom autoloader in a minor version is a BC break. But nothing prevents us to use the Composer autoloading for tests

ste93cry:
Closing this as it has been fixed in `2.x` branch

localheinz:
Ha, nice - took only two years.

PR #214, opened 2015-05-17
Error Types processed by ErrorHandler can be expaned

lskupnjak:
I'm using Sentry for one of the projects at work and I needed to process deprecated errors after upgrading to higher version of PHP. To implement that, I had to extend the ErrorHandler and copy-paste some of its logic to proccess additional two types of errors. That is messy and I think this way would be a lot better.

### DIFF
--- a/lib/Raven/ErrorHandler.php
+++ b/lib/Raven/ErrorHandler.php
@@ -31,6 +31,16 @@ class Raven_ErrorHandler
     private $send_errors_last = false;
     private $error_types = -1;
 
+    private $defaultErrorTypes = array(
+        E_ERROR,
+        E_PARSE,
+        E_CORE_ERROR,
+        E_CORE_WARNING,
+        E_COMPILE_ERROR,
+        E_COMPILE_WARNING,
+        E_STRICT,
+    );
+
     public function __construct($client, $send_errors_last = false)
     {
         $this->client = $client;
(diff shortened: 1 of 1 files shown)


dcramer:
This seems reasonable. Thanks!

madssj:
@dcramer This has been implemented in a way so it's impossible to remove an error type, such as `E_STRICT` from being reported. I don't see any way of fixing this without braking backward compatibility with `0.12.0` and the `getAdditionalErrorTypesToProcess` protected method.

I would however propose that the currently implemented functionality be refactored into something like this:

```php
protected function getErrorTypesToProcess()
{
     return $this->defaultErrorTypes;
}
```

So you can control what errors need to be processed by subclassing. A simpler approach would be to give `errorTypes` in the constructor, which would remove the need for subclassing and make it simpler to use.

There's no apparent reason that the handler should dictate what errors it needs to handle.

lskupnjak:
@madssj @dcramer My goal with this implementation was to leave current error handling as-is, and give an easy way to insert additional error to process.

If this is something that needs to be implemented, my vote goes to injecting errorTypes in constructor.

Other solutions could be:
1. setting $defaultErrorTypes as protected you can override it in child class
2. setting getErrorTypesToProcess() as protected you can override it in child class to do whatever you want

(#1 is better)

However, with this approach there will be some copy-paste needed to get all defaults except some and that is something I wanted to avoid :)

PR #215, opened 2015-05-18
CURLOPT_CONNECTTIMEOUT_MS not defined in some PHP versions

nicktacular:
Some versions of PHP do not correctly defined this constant.

 * http://stackoverflow.com/questions/9062798/php-curl-timeout-is-not-working/9063006#9063006
 * http://php.net/manual/en/function.curl-setopt.php#107314

This addresses by defining correctly if not already set.

image


dcramer:
Thanks!

PR #223, opened 2015-08-07
When request method is POST, PUT, or PATCH.

yjinchun:
http://php.net/manual/en/reserved.variables.post.php
http://php.net/manual/en/wrappers.php.php

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -349,10 +349,16 @@ class Raven_Client
             'query_string' => $this->_server_variable('QUERY_STRING'),
         );
 
-        // dont set this as an empty array as PHP will treat it as a numeric array
+        // don't set this as an empty array as PHP will treat it as a numeric array
         // instead of a mapping which goes against the defined Sentry spec
-        if (!empty($_POST)) {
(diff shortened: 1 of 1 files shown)


dcramer:
Won't this exhaust the input stream so that no one else can read it?

yjinchun:
Yes, you're right. It also needs to be changed.
I modified the reason here is that our team are using angular, transfer data to the back-end, not with a "form-data", nor is "form-urlencoded", but "request payload". This prevented us see the POST data in the Sentry.
So, if our set "Content-Type:application/x-www-form-urlencoded" able to solve this problem?
And, If it is a file? How to deal with more reasonable?
Thanks your answer.

dcramer:
In certain situations I dont believe there will be a way to capture form data. I think ideally what we'd do here is having a special flag that says "form data not available" and show that in the UI, with a link to why. A basic requirement of Sentry is that it doesnt interfere with the app, and risking exhausting a stream to show debug information would be a problem. AFAIK there's no way around it.

yjinchun:
I need to think more. 
Thanks.

PR #228, opened 2015-08-21
Fixed shutdown error types.

madssj:
A base PR for fixing the shutdown error types for good.

I've remove the old way of detecting what types of errors to handle in the shutdown handler, changed the `registerShutdownFunction` signature to not include memory size as I suspect no one needs to change that, and included a sane default list of errors to handle.

I wanted to write some test cases for that handler, but it doesn't seem like it's possible to do from within phpunit.

@dcramer can you come up with any way of testing the shutdown handler, and what do you think of the changes overall?

Refs #181

### DIFF
--- a/lib/Raven/ErrorHandler.php
+++ b/lib/Raven/ErrorHandler.php
@@ -30,42 +30,8 @@ class Raven_ErrorHandler
     private $reservedMemory;
     private $send_errors_last = false;
     private $error_types = -1;
-
-    /**
-     * @var array
-     * Error types that can be processed by the handler
-     */
-    private $validErrorTypes = array(
-        E_ERROR,
-        E_WARNING,
-        E_PARSE,
-        E_NOTICE,
-        E_CORE_ERROR,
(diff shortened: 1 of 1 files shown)


madssj:
Ignore the failing build, it's an easy fix.

dcramer:
My comments from before are still true, but they're not blockers. Mind fixing the build and I'll merge this in?

madssj:
@dcramer I changed the way it works. My proposal is to pass the errors in when you register the shutdown handler.

Please don't merge this before you're entirely sure you agree with the changes I've made. They are not backwards compatible!

I'll fix the build when I get near a computer again. Shouldn't be more than a few days.

bojanpejic:
hey @dcramer should be this PR closed? Seams really old and codebase has been changed since then...btw, I'm checking ability how to setup error types for Sentry (Raven_ErrorHandler) that's how I've come across on this.

Jean85:
I agree.

PR #230, opened 2015-08-28
$errcontext must not be modified.

JonathanO:
The contents of the errcontext param of the PHP error handler must not
be modified since it may contain references. If control is returned to
the code which triggered the error (so a non-fatal error) we might
have changed important state.
This would be bad.

Raven_Stacktrace::get_stack_info can be passed this errcontext, and will
truncate strings in the array to a given limit, modifying errcontext in
the process. Whoops.


JonathanO:
I've a nasty feeling that this opens a whole can of worms over Raven_Processor too...

dcramer:
The processors are all kind of a mess. We're planning to introduce some new APIs soon which will effectively build processors out of some kind of simple DSL, so I'm ok with ignoring that for now.

PR #240, opened 2015-10-20
support non-utf8 language

yangshiqi:
In our project, we use gbk for default encoding. So, why not just change sentry to support non-utf8 language, like gbk.

By the way, i don't understand why using "$value = mb_convert_encoding($value, 'UTF-8', 'UTF-8');"?

### DIFF
--- a/lib/Raven/Serializer.php
+++ b/lib/Raven/Serializer.php
@@ -68,7 +68,7 @@ class Raven_Serializer
             $value = (string) $value;
 
             if (function_exists('mb_convert_encoding')) {
-                $value = mb_convert_encoding($value, 'UTF-8', 'UTF-8');
+                $value = mb_convert_encoding($value, 'UTF-8', 'auto');
             }
 
             return $value;


dcramer:
@yangshiqi what did you mean by your second comment? we convert it to ensure its encodable

That said, the server may actually have problems with non-utf8 encoding

yangshiqi:
@dcramer I mean I am using gbk encoding in our program. 

why
Raven-pho doesn't auto convert the Input string to UTF8 by mb_convert_encoding($value, 'utf-8', 'auto'), and then the sentry backend could display the correct string. Now when sending gbk characters by raven-pho, sentry will display "???".

dcramer:
@yangshiqi that makes sense -- with things like this I'd want to be cautious and try to get some tests in as encoding issues are generally problematic

yangshiqi:
@dcramer sure, the modified version has been tested in our production env, it seems working fine.

dcramer:
Documentation suggests this is actually a good solution. Thanks!

PR #242, opened 2015-10-29
Use /usr/bin/env in test script

wlcx:
Use `/usr/bin/env php` in the shebang of the test script to make it more compatible with e.g. freebsd, where php would be found in /usr/local/bin.


dcramer:
Thanks!

PR #243, opened 2015-10-30
ErrorHandler fix for getErrorTypesToProcess()

lskupnjak:
In `getErrorTypesToProcess()` "+" was used to combine two arrays, and that doesn't work if `$additionalErrorTypes` is defined without setting the keys since array union ignores values from the second array whose keys exist in the first array. `array_merge` does what is expected.


dcramer:
Thanks!

PR #246, opened 2015-11-17
Updated Symfony2 example to use correct dependency injection

aschempp:
One should not inject the service container to later retrieve services, but let the container injection the needed services directly.

Also please be aware that `security.context` is deprecated since Symfony 2.6, but I'm not sure if the example should still support older versions.

### DIFF
--- a/docs/integrations/symfony2.rst
+++ b/docs/integrations/symfony2.rst
@@ -39,21 +39,21 @@ Capturing context can be done via a monolog processor:
 
     namespace AppBundle\Monolog;
 
-    use Symfony\Component\Security\Core\SecurityContext;
     use AppBundlee\Entity\User;
+    use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorage;
 
     class SentryContextProcessor
     {
-        protected $securityContext;
+        protected $tokenStorage;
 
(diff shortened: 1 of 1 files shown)


docs/integrations/symfony2.rst
```diff
@@ -37,28 +37,25 @@ Capturing context can be done via a monolog processor:
 
 .. sourcecode:: php
 
-    namespace Acme\Bundle\AcmeBundle\Monolog;
+    namespace AppBundle\Monolog;
 
-    use Symfony\Component\DependencyInjection\ContainerInterface;
-    use Acme\Bundle\AcmeBundle\Entity\User;
+    use AppBundlee\Entity\User;
+    use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorage;
```
inanimatt: 1. `AppBundlee` :smile_cat: 
2. Better to use `Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface` surely? (Don't forget to update the type hint on line 49 too!)

aschempp: Both fixed!

dcramer:
Thanks @aschempp and @inanimatt!

PR #247, opened 2015-11-17
Fix deprecations in example symfony2 integration

inanimatt:
* Use the suggested default AppBundle and made it consistent (previously the class and the service definition had different namespaces).
* Use `security.token_storage` instead of `security.context` to retrieve user (new in Symfony 2.6, and generating deprecation notices since 2.7 - I left the old service name in a comment since 2.3 LTS is still valid).
* Tiny tiny pedantic PSR-2 tweak to move if-statement opening brace :innocent:

### DIFF
--- a/docs/integrations/symfony2.rst
+++ b/docs/integrations/symfony2.rst
@@ -37,10 +37,10 @@ Capturing context can be done via a monolog processor:
 
 .. sourcecode:: php
 
-    namespace Acme\Bundle\AcmeBundle\Monolog;
+    namespace AppBundle\Monolog;
 
     use Symfony\Component\DependencyInjection\ContainerInterface;
-    use Acme\Bundle\AcmeBundle\Entity\User;
+    use AppBundle\Entity\User;
 
     class SentryContextProcessor {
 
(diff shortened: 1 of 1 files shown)


dcramer:
Semi overlap with #246 (how did you guys manage to both submit these at the same time :)

dcramer:
So it looks like the PRs are mostly the same, minus the way they grab the security context. We're not Symfony users ourselves, so I'd be curious on your opinion @inanimatt @aschempp

aschempp:
I agree with all of the changes here (`AppBundle` rename etc.), though I don't know if you want to provide an example that still supports Symfony 2.3 (which has [LTS until Mai 2016](http://symfony.com/doc/current/contributing/community/releases.html)). Current best practice would be to [use the token storage](http://symfony.com/blog/new-in-symfony-2-6-security-component-improvements), but you really should use correct dependency injection.

I can update my PR to rename the bundle and include the new service if you both agree.

inanimatt:
Haha I have no idea how we both did this at the same time, @dcramer :grin:

Feel free to bring my changes into your PR, @aschempp. I did think about changing the container injection too, but I didn't feel strongly enough about it.

If you do decide to merge both, I'd suggest changing the injection from `security.context` to `security.token_storage` (and adjusting the interface accordingly). You can (as I did) mention the legacy Symfony 2.3 LTS way of doing things in a comment, and I think that'd be ok. Using the 2.3 method on anything newer will cause ugly deprecation notices, so I thought it best to show current best practice while leaving the old way in a comment, rather than showing the old way to everyone and then having to do another PR when 2.3 drops out of support.

aschempp:
I have updated my PR, please review: https://github.com/getsentry/raven-php/pull/246/files

inanimatt:
Looks good, @aschempp -- think we can close and abandon this PR in favour of yours.

PR #248, opened 2015-11-18
Clarify license in Composer

rdoursenaud:
[Composer ](https://getcomposer.org/doc/04-schema.md#license) uses identifiers from the [SPDX license list](https://spdx.org/licenses/).
Services like https://www.versioneye.com/ uses it to infer GPL compatibility.
The "BSD" license does not exist in this referential leading to false positives.
I updated it to "BSD-3-Clause" which seem to be what the project uses.


dcramer:
Thanks!

PR #249, opened 2015-12-01
if not isset email, don't set to user_context

apollopy:
if call function

```php
$client->set_user_data(1);
```

sentry 
```
There was 1 error encountered while processing this event
Discarded invalid value for parameter 'sentry.interfaces.User' Collapse
{
  "name": "sentry.interfaces.User",
  "value": {
    "id": 1,
    "email": "null"
  }
}
```

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -869,10 +869,11 @@ class Raven_Client
      */
     public function set_user_data($id, $email=null, $data=array())
     {
-        $this->user_context(array_merge(array(
-            'id'    => $id,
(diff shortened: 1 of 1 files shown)


dcramer:
Thanks!

(not entirely sure how we'd end up with a null string though)

dcramer:
Also future note, we should change this function signature, as both ID and Email are optional (you can pick and choose attributes)

apollopy:
the null string Come from https://github.com/getsentry/raven-php/blob/master/lib/Raven/Serializer.php#L52

PR #251, opened 2015-12-01
Support x forwarded proto

AD7six:
If the backend server is behind a load balancer, reported urls are always http - as that's the protocol the loadbalancer is talking to the backend server with. The reported url should match that of the client.

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -74,6 +74,7 @@ class Raven_Client
         $this->ca_cert = Raven_util::get($options, 'ca_cert', $this->get_default_ca_cert());
         $this->verify_ssl = Raven_util::get($options, 'verify_ssl', true);
         $this->curl_ssl_version = Raven_Util::get($options, 'curl_ssl_version');
+        $this->trust_x_forwarded_proto = Raven_Util::get($options, 'trust_x_forwarded_proto');
 
         $this->processors = $this->setProcessorsFromOptions($options);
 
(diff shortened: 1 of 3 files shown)


lib/Raven/Client.php
```diff
@@ -781,22 +781,42 @@ private function uuid4()
...
+    {
+        if (!empty($_SERVER['HTTPS']) && $_SERVER['HTTPS'] !== 'off') {
+            return true;
+        }
+
+        if (!empty($_SERVER['SERVER_PORT']) && $_SERVER['SERVER_PORT'] == 443) {
+            return true;
+        }
+
+        if (!empty($_SERVER['X-FORWARDED-PROTO']) && $_SERVER['X-FORWARDED-PROTO'] === 'https') {
```
dcramer: this needs to be an opt-in behavior otherwise, I as a user, can just send these headers to the server

AD7six: I expected that =), though the scope for problems is hardly significant - it's just the url shown in the interface.

How do you want it opt in?

dcramer: we might be able to make it opt-out vs opt-in as I sort of agree the risk isnt great

probably just something like a config option saying "use_x_forwarded_proto"

AD7six: oks

I named the option `trust_x_forwarded_proto` as that more accurately describes what the option is for. Alternatively it's be better to permit setting the base-url to be submitted.

Note that all these option properties are public - because they aren't declared.

dcramer:
Awesome, thanks!

PR #253, opened 2015-12-02
Always submit environment data

AD7six:
If an exception is caught in a cli process - no environment information is submitted. 

This can make debugging cli problems difficult if there is more than one possible cli candidate-process. In my own situation, from sentry's info alone, I usually have to guess which process is throwing the exception based upon class names in a stack trace.

Annoyingly, whilst this does what I intended (add env data to the info submitted to sentry) it's considered invalid:

screenshot from 2015-12-02 12-18-32

What's the correct way to submit env data to sentry for cli processes? It should be automatic imo, and not "extra", I'm not sure which page in the spec-docs I should be referring to.

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -351,19 +351,20 @@ class Raven_Client
             }
         }
 
-        if (!$this->is_http_request()) {
-            return array('request' => $result);
+        if ($this->is_http_request()) {
+            $result += array(
+                'method' => $this->_server_variable('REQUEST_METHOD'),
+                'url' => $this->get_current_url(),
+                'query_string' => $this->_server_variable('QUERY_STRING'),
+                'data' => $_POST,
+                'cookies' => $_COOKIE
+            );
+        } else {
+            $result['extra']['env'] = $result['env'];
+            unset($result['env']);
         }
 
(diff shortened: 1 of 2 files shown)


AD7six:
no interest/comment?

dcramer:
Sorry looks like we missed this ticket. 

You'd want to submit the cli info as part of extra as the HTTP interface is reserved for exactly what it sounds like. I do think it makes sense to capture more of this by default.

AD7six:
> the HTTP interface is reserved for exactly what it sounds like

Can you clarify please what http interface you're referring to? The key under which `env` data is submitted for a cli request is `request`.

dcramer:
@AD7six 'request' is an alias for the HTTP request. The reason env exists there is somewhat legacy, but because things like REMOTE_ADDR dont exist as HTTP headers and live in some kind of env.

AD7six:
OK

Is it appropriate then to by default always submit env data via extra for cli processes?

dcramer:
@AD7six yeah, it could simply be an "env" key on extra that contains the information

dcramer:
I'm going to close this as we dont want environment data and its unclear if this is still valid.

PR #256, opened 2016-01-09
Don't include stack frame within PHP error handler

thomasbachem:
For `ErrorException`s (thrown by the error handler for PHP errors), we should skip the point at which it was thrown in the trace, as it points to the error handler instead of the actual error origin.

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -264,12 +264,14 @@ class Raven_Client
              * was thrown, so we have to stuff it in ourselves. Ugh.
              */
             $trace = $exc->getTrace();
-            $frame_where_exception_thrown = array(
-                'file' => $exc->getFile(),
-                'line' => $exc->getLine(),
-            );
-
-            array_unshift($trace, $frame_where_exception_thrown);
+            // Don't include the frame within the error handler for ErrorExceptions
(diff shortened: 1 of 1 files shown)


thomasbachem:
If you don't feel comfortable with this solution (although I personally think it's just fine) we could also solve it like this:

```php
// fix stack trace in case it doesn't contain the exception origin as the first entry
if(isset($trace[0]['file']) && !($trace[0]['file'] == $exc->getFile() && $trace[0]['line'] == $exc->getLine())) {
    $frame_where_exception_thrown = array(
        'file' => $exc->getFile(),
        'line' => $exc->getLine(),
    );

    array_unshift($trace, $frame_where_exception_thrown);
}
```

This works because the stack trace mistakenly claims that the `ErrorException` was thrown in the same file and at the same line as the actual error origin.

Marlinc:
Would it be possible to get this in?

dcramer:
In Python we do something similar for logging integration where we attempt to find the first non-logging frame, and start the stacktrace there. Maybe that would be a better solution, to where we would simply try to ignore the SDK code.

ste93cry:
This issue is quite old and should be hopefully solved in `2.0`, so I'm going to close this

PR #257, opened 2016-01-20
Fix indentation

damour:
Content not visible in markdown now:
https://github.com/getsentry/raven-php/blob/master/docs/usage.rst#optional-attributes
vs
https://github.com/damour/raven-php/blob/master/docs/usage.rst#optional-attributes


dcramer:
Thanks!

PR #258, opened 2016-01-27
Ensure error_types are handled correctly

dcramer:
This corrects behavior that would allow you to specify which error_types are handled by the SDK, but would simply ignore it if they were also not included in error_reporting.

### DIFF
--- a/lib/Raven/ErrorHandler.php
+++ b/lib/Raven/ErrorHandler.php
@@ -29,11 +29,12 @@ class Raven_ErrorHandler
     private $call_existing_error_handler = false;
     private $reservedMemory;
     private $send_errors_last = false;
-    private $error_types = -1;
+    private $error_types;
 
     /**
      * @var array
-     * Error types that can be processed by the handler
+     * Error types that can be processed by the handler. Only used
+     * by the shutdown fatal handler.
      */
     private $validErrorTypes = array(
         E_ERROR,
@@ -55,7 +56,8 @@ class Raven_ErrorHandler
 
     /**
      * @var array
-     * The default Error types that are always processed by the handler. Can be set during construction.
+     * The default Error types that are always processed.  Only used
+     * by the shutdown fatal handler. Can be set during construction.
      */
     private $defaultErrorTypes = array(
         E_ERROR,
@@ -92,7 +94,7 @@ class Raven_ErrorHandler
 
     public function handleError($code, $message, $file = '', $line = 0, $context=array())
     {
-        if ($this->error_types & $code & error_reporting()) {
+        if ($this->error_types & $code) {
             $e = new ErrorException($message, 0, $code, $file, $line);
             $this->handleException($e, true, $context);
         }
@@ -155,10 +157,13 @@ class Raven_ErrorHandler
         $this->call_existing_exception_handler = $call_existing_exception_handler;
     }
 
-    public function registerErrorHandler($call_existing_error_handler = true, $error_types = -1)
+    public function registerErrorHandler($call_existing_error_handler = true, $error_types = null)
     {
+        if ($error_types === null) {
+            $error_type = error_reporting();
+        }
         $this->error_types = $error_types;
-        $this->old_error_handler = set_error_handler(array($this, 'handleError'), error_reporting());
+        $this->old_error_handler = set_error_handler(array($this, 'handleError'), $error_types);
         $this->call_existing_error_handler = $call_existing_error_handler;
     }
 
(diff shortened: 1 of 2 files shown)


lib/Raven/ErrorHandler.php
```diff
@@ -155,10 +158,13 @@ public function registerExceptionHandler($call_existing_exception_handler = true
...
 
-    public function registerErrorHandler($call_existing_error_handler = true, $error_types = -1)
+    public function registerErrorHandler($call_existing_error_handler = true, $error_types = null)
     {
+        if ($error_types == null) {
+            $error_type = error_reporting();
+        }
         $this->error_types = $error_types;
-        $this->old_error_handler = set_error_handler(array($this, 'handleError'), error_reporting());
+        $this->old_error_handler = set_error_handler(array($this, 'handleError'), -1);
```
dcramer: we're basically changing the logic to say "always call our error handler, we'll determine later if we care about the event"

lib/Raven/ErrorHandler.php
```diff
@@ -92,12 +95,12 @@ public function handleException($e, $isError = false, $vars = null)
...
     public function handleError($code, $message, $file = '', $line = 0, $context=array())
     {
-        if ($this->error_types & $code & error_reporting()) {
+        if ($this->error_types & $code) {
             $e = new ErrorException($message, 0, $code, $file, $line);
             $this->handleException($e, true, $context);
         }
 
-        if ($this->call_existing_error_handler) {
+        if (error_reporting() & $code && $this->call_existing_error_handler) {
```
dcramer: looking at this, maybe we want to not regress here -- that is always call the existing handler if we were instructed to and let them deal with checking error_reporting

lib/Raven/ErrorHandler.php
```diff
@@ -155,10 +158,13 @@ public function registerExceptionHandler($call_existing_exception_handler = true
         $this->call_existing_exception_handler = $call_existing_exception_handler;
     }
 
-    public function registerErrorHandler($call_existing_error_handler = true, $error_types = -1)
+    public function registerErrorHandler($call_existing_error_handler = true, $error_types = null)
     {
+        if ($error_types == null) {
+            $error_type = error_reporting();
```
SeBsZ: Has this been tested? Surely this should say $error_types.

dcramer: good catch! this path definitely didnt get tested (by the suite)

dcramer: Actually it looks like the Silent behavior might be broken with this. I'm investigating a bit more, as I'm not super familiar w/ this path.

dcramer:
I'm reverted this temporarily -- need to resolve an issue with silencing errors

dcramer:
https://github.com/getsentry/raven-php/pull/259

archon810:
Any updates?

archon810:
So was this ever resolved properly @dcramer?

I'd like to resolve both issues I mentioned to you earlier: reporting only Warning+ to Sentry and seeing everything in our log without Sentry slurping messages out of it.

Thanks.

dcramer:
@archon810 I believe we've resolved things in master at this point. I'd like to get more feedback/testing of it, because it changes behavior. Before we were doing something odd where we'd ignore suppressed errors/etc, and I dont think we do that now.

tbh I don't have a ton of time to focus on this at the moment, but it should at least let you be more explicit in what you're capturing now

archon810:
@dcramer Been using it for a week or so, and it works great for us. Could you please tag the release so I can safely add it to our composer setup?

dcramer:
@archon810 yeah we'll work on getting a new release out -- thanks for testing!

archon810:
@dcramer OK, so everything is fixed, except for one thing. I'm continuing to see E_STRICT messages show up on Sentry when I'm actively blocking them from being sent.

Here's the relevant line:   $error_handler->registerErrorHandler(true, E_ALL & ~E_NOTICE & ~E_USER_NOTICE & ~E_STRICT);.

E_STRICT is disabled. However, this event keeps popping up:
call_user_func() expects parameter 1 to be a valid callback, non-static method wp_gfycat::shortcode() should not be called statically

This is an E_STRICT notice. Why would E_STRICT things continue to be logged when I specifically disable them?

dcramer:
@archon810 honestly -- no idea. I would rely on writing a test to prove that something is broken at this point, as the complexities surrounding this are a bit too much to keep your head wrapped around. That would be easy if:

1. We know exactly what kind of error was getting through from E_STRICT
2. We ensure the test case uses something resembling your example above

archon810:
1. The notice is of this type: http://stackoverflow.com/questions/2640392/is-it-ok-to-call-non-static-methods-with-call-user-func-in-php-5-3. Which is E_STRICT AFAIK.
1. Sure, feel free to use that specific register line.

PR #259, opened 2016-01-27
Ensure error_types are handled correctly

dcramer:
This corrects behavior that would allow you to specify which error_types are handled by the SDK, but would simply ignore it if they were also not included in error_reporting.

Additionally we introduce a backwards incompatible change: silenced errors are no longer ignored. There's no way to detect when you intended an error to be silent without losing behavior introduced here (correctness on specifying error_types).

### DIFF
--- a/test/Raven/Tests/ErrorHandlerTest.php
+++ b/test/Raven/Tests/ErrorHandlerTest.php
@@ -16,21 +16,30 @@ class Raven_Tests_ErrorHandlerTest extends PHPUnit_Framework_TestCase
     public function setUp()
     {
         $this->errorLevel = error_reporting();
+        $this->errorHandlerCalled = false;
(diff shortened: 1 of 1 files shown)


dcramer:
This expands on the tests from #258 (which actually failed when fixing ``error_types``), so things should cover correct behavior, in all branches.

dcramer:
I will give this ~a day for feedback incase someone has a strong opinion/suggestion regarding the error silencing change.

archon810:
Any updates?

dcramer:
@archon810 thanks for pinging -- merging this

archon810:
@dcramer I'm not sure what's going on here still. Commented on #258 too.

archon810:
Good to close?

dcramer:
@archon810 PR is already closed here, so yeah

PR #262, opened 2016-02-22
Docs: Lumen framework added

Diewy:
Hi all,

To support the micro framework of Laravel called Lumen there is a small adjustment we have to make to the code. We have to return a Logger, see https://github.com/laravel/lumen-framework/blob/5.2/src/Application.php#L375

Kind regards,
Dieter


dcramer:
Thanks!

PR #267, opened 2016-03-11
Fix curl_exec method to work with self-signed certificates

dasmfm:
Bug fixed: send_http_asynchronous_curl_exec method ignores verify_ssl flag and doesn't work with self-signed certificates.

Curl flag added.


dcramer:
Thanks!

Issue #269, opened 2016-03-24
Bad Serialization


bradynpoulsen:
In the [Raven_Serializer](https://github.com/getsentry/raven-php/blob/0.13.0/lib/Raven/Serializer.php) class, literals are wrongfully and inconsistently being serialized for JSON.
* On [lines 51-58](https://github.com/getsentry/raven-php/blob/0.13.0/lib/Raven/Serializer.php#L51-L58), true/false/null/float literals are being converted to their string representations when they SHOULD be represented as literals in JSON.
* This is also inconsistent to [lines 65-66](https://github.com/getsentry/raven-php/blob/0.13.0/lib/Raven/Serializer.php#L65-L66) where integers are being passed along as their literal values.

This issue is resulting in the Sentry application having this issue: getsentry/sentry#2882

dcramer:
@bradynpoulsen thanks for digging into this -- in retrospect pretty obvious. Will get this fixed today.

PR #270, opened 2016-03-24
Improve serialization (fixes GH-269)

dcramer:
-  Explicitly use repr-like serialization for reflection args
- Use standard JSON approach for other args


dcramer:
After pushing this up I just realized why these might have been coercing most things to strings. In our variable interpolation bits (e.g. argument reflection) we try to show the actual value. It's possible we need to change how things work and:

1. use the original code here for the variable representation for arg reflection
2. use more-or-less the new code here for general serialization

manti:
@dcramer can you please revert this changes its breaking our production.

PR #272, opened 2016-04-08
Curl async times out on first loop

israelshirk:
Trying the async curl method from a test env requests weren't coming through...  

In Raven_CurlHandler.php adding `error_log("time minus start " . (time() - $start) . " timeout $timeout");` on line 83 (just after `usleep()`) gave the output `time minus start 0 timeout 5` immediately before exiting the loop.  Switching '>' to '<' seems to give the intended behavior.

Thanks!


dcramer:
Good catch, thanks!

(Also sorry about the delay in getting this merged in)

PR #277, opened 2016-04-29
Fix composer bin name

dasmfm:


dcramer:
Good catch. I've renamed it throughout.

PR #279, opened 2016-05-03
Initial support for breadcrumbs

dcramer:


lib/Raven/Breadcrumbs.php
```diff
@@ -0,0 +1,58 @@
...
+
+    public function is_empty()
+    {
+        return $this->count === 0;
+    }
+
+    public function to_json()
+    {
+        return array(
+            'values' => $this->fetch,
```
56quarters: Is this supposed to be the bare `$this->fetch` or a method call `$this->fetch()`?

dcramer:
Need to:

- Write tests for Monolog
- Ensure Monolog support isn't going to affect autoloader and older versions of PHP

lib/Raven/Breadcrumbs.php
dcramer: good catch :)

(fyi this is def a wip yet)

dcramer:
Alright I think this is generally ok. Will want to squash merge since some of it was broken. I'd def prefer if we could easily do a global handler in Monolog, but the ecosystem is sane enough that it wont be *that* bad to get people to configure it.

PR #285, opened 2016-05-09
Expand configuration API

dcramer:
- Add get/setRelease
- Add get/setEnvironment
- Add get/setAppPath
- Add get/setPrefixes
- Expose $breadcrumbs
- Expose $context


dcramer:
I thought about simply exposing the attributes, but this lets us potentially rename those attributes (or move them elsewhere) without any compatibility concerns.

PR #286, opened 2016-05-13
add conflict rule for raven/raven

xabbuh:
Both packages cannot be installed at the same time as they use the same
class names.


dcramer:
Thanks!

PR #288, opened 2016-05-16
Fix: Case-insensitive method calls

localheinz:
This PR

* [x] fixes case-insensitive method calls (along with corresponding test method names)


dcramer:
Thanks!

PR #289, opened 2016-05-16
Fix array key for adding user context

sctape:
Working through adding a monolog integration for sentry and noticed that user context information was not being passed to sentry until I made the proposed change. Not sure if something changed with monolog.

Using "monolog/monolog:1.x-dev" and "raven/raven:0.13.0"


dcramer:
Thanks!

PR #291, opened 2016-05-20
Fix cli name and test for command argument

temp:
Calling simply bin/sentry results in a php error.
This fixes it, and the cli name is change from raven to sentry in the help text.


dcramer:
Thanks!

Issue #290, opened 2016-05-19
If session has object it crashes


d-lamers:
It throws this error

`Error in exception handler: preg_match() expects parameter 2 to be string, object given in /srv/bezorgbeer/vendor/sentry/sentry/lib/Raven/SanitizeDataProcessor.php:56`

This

```
public function sanitize(&$item, $key)
    {
        is_object($item) ? dd($item): true; //Added for test
        if (empty($item)) {
            return;
        }

        if (preg_match($this->values_re, $item)) {
            $item = self::MASK;
        }

        if (empty($key)) {
            return;
        }

        if (preg_match($this->fields_re, $key)) {
            $item = self::MASK;
        }
}
```

Returns this

`object(stdClass)#214 (2) { ["code"]=> string(2) "nl" ["title"]=> string(10) "Nederlands" }`

dcramer:
I'm assuming the attached change will resolve this, which is a step in the right direction

dcramer:
I'm closing this as while I think protecting inputs is good, I'd rather us simply not go down the rabbit hole of trying to magically handle encoding of arbitrary values. If there is still issues with default behavior in Sentry we can work around it, but if you're binding in session data you'll need to serialize it yourself (assuming its not already valid JSON).

PR #292, opened 2016-05-26
Fix non-scalar values in session

aschempp:
fixes https://github.com/getsentry/sentry-php/issues/290


d-lamers:
Hopefully this will be merged soon

dcramer:
Should be resolved in b98b38134053794bfafcc3beda359cc6f84f66f2

aschempp:
I don't see how that should fix this PR? I mean even if the method patched by this PR would not longer be called, the method in question is still wrong. It would still call `preg_match` on non-scalar values, which can never work?

PR #294, opened 2016-05-27
Add error_reporting support to breadcrumbs

dcramer:


dcramer:
This has the caveat that if you disable bubbling up to handlers it poofs, but people can sort that out as I imagine its mostly uncommon.

PR #299, opened 2016-05-30
make send_http_asynchronous_curl_exec method protected

etienneroudeix:
Hello,
I am facing issues with ```send_http_asynchronous_curl_exec``` method : It makes some zombie curl process on my server.
I would like to inherit ```send_http_asynchronous_curl_exec``` to use Guzzle but I need this method not to be private anymore.
Please let me know if there is any chance for this PR to be accepted.

If you would like to use Guzzle directly in this project, I suppose I can make a PR for this instead.

Cheers

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -753,7 +753,7 @@ class Raven_Client
      * @param array     $headers    Associative array of headers
      * @return bool
      */
-    private function send_http_asynchronous_curl_exec($url, $data, $headers)
+    protected function send_http_asynchronous_curl_exec($url, $data, $headers)
     {
         // TODO(dcramer): support ca_cert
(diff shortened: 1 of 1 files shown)


dcramer:
I dont really have a problem with scoping here, the caveat being that we'll make no guarantees that the function is not renamed internally.

I believe what your intention is would be to override how HTTP works, *not* to override the curl implementation.

To do that could hook ``send()``, and I would be fine with making ``send_remote`` available.

etienneroudeix:
OK I should do that.
Could you before that consider #301 which will keep the method private and improve async curl imho

dcramer:
Closing this out as resolved via efc749948c74d1a759242293678b75168db5f2fb

PR #301, opened 2016-05-31
use Guzzle to send async curl

etienneroudeix:
An alternate to #299 

It will also ease the support for ca_cert

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
     "require": {
         "php": ">=5.2.4",
         "ext-curl": "*",
+        "guzzlehttp/guzzle": "^6.0",
         "monolog/monolog": "*"
     },
     "conflict": {
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -7,6 +7,8 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  */
+use GuzzleHttp\Client;
+use GuzzleHttp\Psr7\Response;
 
 /**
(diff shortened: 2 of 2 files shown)


dcramer:
I dont have strong opinions here as I dont understand the PHP community, but in other platforms this would be a hinderance as we now require a version of a library and that library is required for a path of code not commonly executed. I think I'd prefer we just keep improving the curl_multi mechanism, unless you think there's a reason its not good enough?

etienneroudeix:
I actually use a workaround on my project and use Guzzle instead of curl multi.
I did this because I used to have some zombie curl process on my server and growing.
And using Guzzle fixed it.

In a more theoric way I would say that making async thing with PHP is not easy but making forks and forks is not a durable solution anyway.

Another pros is that it will make the ca_cert integration faster (http://docs.guzzlephp.org/en/latest/request-options.html?highlight=cert#cert).

Finally I understand your point of view and your concerns but Guzzle is well known and widely used librairy (7k github stars for what it worth)

Anyway thank you for your time and if you won't use Guzzle, i'll go back to #299 and follow your initial advice.

dcramer:
I think what might be good is if we just allow you to specify a config param for how events are sent.

e.g.

```
$client->setTransport(function ($url, $authHeader, $payload) {
  $http->send($url, $payload, $auth_header);
});
```

Thoughts?

etienneroudeix:
Yep the idea is good. Indeed  ```$client ```and ```$payload```is enough.
And would solve my issue.

Issue #300, opened 2016-05-31
Warning: preg_match() expects parameter 2 to be string, object given in vendor/sentry/sentry/lib/Raven/SanitizeDataProcessor.php on line 54


Repox:
I'm running version 0.19.0 and I get the following error message:

    Warning: preg_match() expects parameter 2 to be string, object given in vendor/sentry/sentry/lib/Raven/SanitizeDataProcessor.php on line 54

Vanilla installation, though:

	Raven_Autoloader::register();
	$client = new Raven_Client('https://auth_me@app.getsentry.com/XXXXX');
	$client->setEnvironment(App::site_mode());
	$error_handler = new Raven_ErrorHandler($sentry);
	$error_handler->registerExceptionHandler();
	$error_handler->registerErrorHandler();
	$error_handler->registerShutdownFunction();

Why am I experiencing this?

dcramer:
I don't suppose you have a way you could capture what the value is that it's failing on? It's unclear how we'd be getting unserializable data down that path, but I will work on a resolution.

Jay-Pi:
I had the same Problem using zend framework. The user/authentication data is stored as ArrayObject in $_SESSION.
It seem in version 0.16.0 the was a serialize function called in Client.php before sanitize which is missing now.
```
$serializer = new Raven_Serializer();
$data['user'] = $serializer->serialize($data['user']);
```

dcramer:
@Jay-Pi in the current release we dont actually serialize the session data in automatically anymore (or at least shouldn't be). We decided it was better to let users decide if they wanted any of that data present, and thus is also why we dont serializer the user attribute.

Jay-Pi:
@dcramer I didn't know how to do this using the Raven_ErrorHandler, because it was too easy.
```
$client = new Raven_Client(PHP_ERROR_RAVEN_URL);
$user['user'] = 'foo';
$client->user_context($user);

// Install error handlers and shutdown function to catch fatal errors
$error_handler = new Raven_ErrorHandler($client);
$error_handler->registerExceptionHandler();
$error_handler->registerErrorHandler();
$error_handler->registerShutdownFunction();
```

This works for me.

endelwar:
I can confirm the same behavior with 0.17.0, 0.18.1 and 0.19.0, everything is ok with 0.16.0

Repox:
[...]

PR #305, opened 2016-06-02
Proposed fix for issue #300

Repox:
The `sanitize` method is expecting a string for the `preg_match` check.

In some cases, other data than a string is passed through the method and this makes the client fail.

My proposed fix is to verify the contents of `$item` before moving forward with the sanitizing.


dcramer:
I'm going to dig into the root cause of this tomorrow. I think we need to revert changes on serialization (at least partially), as there's some concerns with simply returning non-string values here.

dcramer:
b98b38134053794bfafcc3beda359cc6f84f66f2 re-adds serialization to select values

PR #311, opened 2016-06-10
Make serializer  transcoding more resilient

mabrahamde:
`mb_convert_encoding` throws a `E_WARNING` if it can't  detect the encoding. The warning are flooding log-files. You can prevent this by using mb_detect_encoding before conversion.


dcramer:
Thanks!

PR #313, opened 2016-06-15
fabpot/php-cs-fixer is abandoned

j0k3r:
https://packagist.org/packages/fabpot/php-cs-fixer
> The author suggests using the friendsofphp/php-cs-fixer package instead.


dcramer:
Thanks!

PR #326, opened 2016-07-22
Remove env data from HTTP

dcramer:
This contains potentially sensitive information, and as we move forward with Sentry it no longer logically makes sense for it to exist as part of this. If users want it they can use ``extra_context``.

@getsentry/php

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-php/326)

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -470,15 +470,13 @@ class Raven_Client
 
     protected function get_http_data()
     {
-        $env = $headers = array();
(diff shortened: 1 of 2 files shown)


dcramer:
This will likely be contentious with the community, but to avoid future discussion we're taking a hard stance on this as it's too easy to leak extremely sensitive data. If we want to re-introduce any of this I'd be ok with PRs that do an allow-list of variables.

Issue #322, opened 2016-07-15
json_encode error with invalid utf-8 characters 


georgexsh:
error message:
```
ErrorException: json_encode(): Invalid UTF-8 sequence in argument
  File "vendor/sentry/sentry/lib/Raven/Compat.php", line 73, in handleError
    return json_encode($value);
  File "/vendor/sentry/sentry/lib/Raven/Compat.php", line 73, in json_encode
    return json_encode($value);
  File "vendor/sentry/sentry/lib/Raven/Client.php", line 716, in send
    $message = Raven_Compat::json_encode($data);
  File "vendor/sentry/sentry/lib/Raven/Client.php", line 639, in capture
    $this->send($data);
  File "vendor/sentry/sentry/lib/Raven/Client.php", line 348, in captureMessage
    return $this->capture($data, $stack, $vars);
...
```

if context variable cotains arbitrary binary data like picture content, `jsen_encode` will generate a warning, and sentry will show that variable value as `None`, eg:

```
function afunc() {
    $f = file_get_contents('sentry_logo.png');
    assert(0);
}
```

related with #6
using sentry-php 0.22.0

pengu-fr:
I have the same problem.
Raven_Compat::json_encode($data) return FALSE with non UTF-8 string. So the sentry API return a 400 and the error isn't reported.

A workaround is to use self::_json_encode instead of json_encode and force to ignore non UTF-8 character in string :
```
if (is_string($value)) {
                $value = iconv('UTF-8', 'UTF-8//IGNORE', $value);
                return sprintf('"%s"',

```
The data won't be usable, but at least the original error will be reported.

dcramer:
iconv is an extension just like the mb encoding stuff? In your case do you actually have the mb encoding extension installed?

georgexsh:
@dcramer yes, we have mbstring extension installed.

dcramer:
@pengu-fr @georgexsh do you happen to have a value that we could use to test this against? I have a hunch I might know whats causing it, but it would be better if we had a test case.

dcramer:
Actually, my apologies, we can likely use the function as an example, which would imply its from local var capture.

dcramer:
``fix/utf-8`` branch has a test reproduction

georgexsh:
@dcramer glad to hear that, and `sentry_logo.png` is the logo file grabbed from https://getsentry.com/branding/

tobias-nitsche:
[...]

PR #328, opened 2016-07-30
Handle invalid encoding errors gracefully

dcramer:
Fixes GH-322

@getsentry/php

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-php/328)


lib/Raven/Client.php
```diff
@@ -686,6 +686,28 @@ public function sendUnsentErrors()
...
     }
 
+    public function encode(&$data)
+    {
+        $message = Raven_Compat::json_encode($data, JSON_FORCE_OBJECT | JSON_PARTIAL_OUTPUT_ON_ERROR);
+        if ($message === false) {
+            if (function_exists('json_last_error_msg')) {
+                $this->_lasterror = json_last_error_msg();
+            } else {
+                $this->_lasterror = json_last_error();
```
dcramer: we could translate this if we really cared, but I mostly wanted it for debug

PR #336, opened 2016-08-11
Add link to full documentation

davejamesmiller:
I think it would be a good idea to link to the docs site from the README, not just the main Sentry website.

(I only just discovered there's a separate documentation site - although I've read the README on GitHub more than once, I've never clicked the Docs link on the main Sentry website (or noticed the `docs/` folder). This means I never discovered things like [Filtering Out Errors](https://docs.getsentry.com/hosted/clients/php/usage/#filtering-out-errors) which is only in the full docs, not the README.)

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-php/336)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -262,6 +262,7 @@ You may now use phpunit :
 Resources
 ---------
 
+* `Documentation <https://docs.getsentry.com/hosted/clients/php/>`_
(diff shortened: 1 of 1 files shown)


dcramer:
I also pushed some other changes which just kills a lot of the half-maintained README to prioritize docs

davejamesmiller:
:+1:

PR #342, opened 2016-09-02
Fix encoding errors for exceptions

mabrahamde:
$data['exception'] is not encoded to utf8 via sanitized(). This causes problems, if sourcecode (stacktrace) or the exception's message are not in utf8 and contain special characters (e.g. german umlauts)

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-php/342)

### DIFF
--- a/test/Raven/Tests/ClientTest.php
+++ b/test/Raven/Tests/ClientTest.php
@@ -684,6 +684,26 @@ class Raven_Tests_ClientTest extends PHPUnit_Framework_TestCase
         ), $event['extra']);
     }
 
+    public function testCaptureExceptionInLatin1File()
+    {
+        // If somebody has a non-utf8 codebase, she/he should add the encoding to the detection order
+        $options = ['mb_detect_order' => ['ISO-8859-1', 'ASCII', 'UTF-8']];
+
+        $client = new Dummy_Raven_Client($options);
+
+        // we need a non-utf8 string here.
+        // nobody writes non-utf8 in exceptions, but it is the easiest way to test.
+        // in real live non-utf8 may be somewhere in the exception's stacktrace
+        $utf8String = 'äöü';
+        $latin1String = utf8_decode($utf8String);
+        $client->captureException(new \Exception($latin1String));
+
+        $events = $client->getSentEvents();
+        $event = array_pop($events);
+
+        $this->assertEquals($event['exception']['values'][0]['value'], $utf8String);
+    }
+
     public function testGetLastEventID()
     {
         $client = new Dummy_Raven_Client();
--- a/test/Raven/Tests/ClientTest.php
+++ b/test/Raven/Tests/ClientTest.php
@@ -687,8 +687,12 @@ class Raven_Tests_ClientTest extends PHPUnit_Framework_TestCase
     public function testCaptureExceptionInLatin1File()
     {
         // If somebody has a non-utf8 codebase, she/he should add the encoding to the detection order
-        $options = ['mb_detect_order' => ['ISO-8859-1', 'ASCII', 'UTF-8']];
-
+        $options = array(
+            'mb_detect_order' => array(
+                'ISO-8859-1', 'ASCII', 'UTF-8'
+            )
+        );
+        
         $client = new Dummy_Raven_Client($options);
 
         // we need a non-utf8 string here.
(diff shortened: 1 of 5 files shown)


dcramer:
I'm not comfortable making this change without a test that proves the issue (and to prevent us from breaking it in the future)

mabrahamde:
Just looking for the right test:

"Passing non-utf8 data via parameter $data to Raven_Client::capture() (like captureExceptions does) must not result in no non-utf8 data in data array passed to send/passed to Raven_Client::_pending_events"

Seems like the only "clean" unit-test. Any other suggestion?

Internals:

- Raven_Stacktrace::get_stack_info() returns non-utf8 data, if source file is in another encoding.
- Raven_Client::captureException() passes non-utf8 data inside key 'exception' to Raven_Client::capture()
- In Raven_Client::sanatize() no character encoding is done for key 'exception'
- Json encoding inside Raven_Client::send() fails because of non-utf8 characters, thus message is empty
- Empty payload is sent to remote.

dcramer:
@mabrahamde Yeah I would probably just do high level and treat it as an integration test. We care more about preventing future regressions than proving absolute correctness in individual functions.

I would probably just add it to ClientTest.php, and have go down the path of how ``testCaptureMessageWithExtraContext`` (and similar) work, but passing in the example invalid data, and ensuring its coerced as expected.

test/Raven/Tests/ClientTest.php
```diff
@@ -684,6 +684,26 @@ public function testCaptureMessageWithExtraContext()
         ), $event['extra']);
     }
 
+    public function testCaptureExceptionInLatin1File()
+    {
+        // If somebody has a non-utf8 codebase, she/he should add the encoding to the detection order
+        $options = ['mb_detect_order' => ['ISO-8859-1', 'ASCII', 'UTF-8']];
```
dcramer: this will fail the tests (if it hasn't yet) since we have to support PHP 5.3

mabrahamde: fixed.

dcramer:
:+1: thanks for the test

So regarding this change, I think we should explicitly serialize the exception value, rather than re-serializing the entire blob (which has a non trivial cost). Thoughts?

mabrahamde:
well, exception value isn't the only problem. 

the stacktrace of the exceptions contains some lines of code. the problem is the same if one of the files containing a chunk of code is encoded in non-utf8. the stacktrace is most of the blob.

should i provide a second test?

dcramer:
Ahh I didn't consider the stacktrace. Maybe instead then we can move this serialization into the exception handler where it currently also does serialization of vars? Mostly I want to avoid re-serializing some of these values as the cost can be non-trivial.

mabrahamde:
We are using monolog. I got aware of this problem while logging Exception via Monolog\Logger::alert() 

https://github.com/Seldaek/monolog/blob/master/src/Monolog/Logger.php#L510

Using the SentryBundle isn't an option for us, because limits us too much. Integration of sentry via monolog is common.

In this setup serialization inside the exception handler isn't an option (if you mean this kind of exception handler: http://php.net/manual/de/function.set-exception-handler.php).

dcramer:
@mabrahamde i actually just mean inside of our code that parses the stacktrace/captures vars (which the Monolog handler will hit as well). I guess there's two concerns looking at the code:

- we can manage serialization of the source code (and filenames?) via Stacktrace.php in get_stack_info
- .. but that doesn't handle the exception value, which is serialized elsewhere (in Client.php:captureException)

I think still ideally it'd be better to instantiate the serializer in both places (or allow it to be optionally passed into stacktrace) so we can avoid the additional cost of re-serialization of vars.

mabrahamde:
Okay, I'll try to adapt your approach tomorrow.

mabrahamde:
I would instantiate the Raven_Serializer and Raven_ReprSerializer once in Raven_Client. This serializers should be mandatory in get_stack_info(), e.g.

`public static function get_stack_info($frames, Raven_Serializer $serializer, Raven_ReprSerializer $reprSerializer, $trace = false, $shiftvars = true, $errcontext = null, $frame_var_limit = Raven_Client::MESSAGE_LIMIT, $strip_prefixes = null, $app_path = null)`

Serializers are quite tiny, thus you do not need something like lazy instantiation. Injecting them makes it possible to guarantee configuration (mb_detection_order).

If you agree, i would fix the tests affected.

mabrahamde:
Changing the signature wasn't a good idea. i was running into problems in a legacy application having diffrent sentry versionsinside the autoloader at the same time. that is more an issue of the application, but it convinced me not to break the method's backward compatibility

mabrahamde:
Should look like you suggested, now.

dcramer:
Awesome, thanks!

PR #346, opened 2016-09-09
Improve masking

willemstuursma:
- Improves masking support for PANs 
- Mask PANs in the middle of a string
- Support masking private keys in PEM format

---

This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-php/346)

### DIFF
--- a/lib/Raven/Processor.php
+++ b/lib/Raven/Processor.php
@@ -6,6 +6,11 @@
  */
 abstract class Raven_Processor
 {
+    /**
+     * @var Raven_Client
+     */
+    private $client;
+
     public function __construct(Raven_Client $client)
     {
         $this->client = $client;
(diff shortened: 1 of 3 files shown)


dcramer:
@willemstuursma I think searching for PANs is great, but its definitely something we'd want to get on the server as a first priority. We usually try to mimic the servers behavior in the clients, but its always our number one concern and last resort.

willemstuursma:
@dcramer So what can I do to help here?

dcramer:
@willemstuursma let me get these over on the server and evaluate it a bit more. I'll try to get a patch up today.

dcramer:
Related server changes: https://github.com/getsentry/sentry/pull/4170

willemstuursma:
Now that the server has been updated, what about the PHP client?

ste93cry:
As this is a pretty old issue and I think that it would be better to not duplicate the logic to handle sanitization on each SDK, I'm going to close it. With version `2.0` (which is still in development at the time of writing) we have introduced the concept of middlewares, so you can customize the data being sent before it leaves your application. Also, you could have a look at the [Relay](https://docs.sentry.io/relay/?platform=javascript) project which is a proxy which stands between the SDKs and the Sentry server and does the sanitization

PR #363, opened 2016-10-11
work around PHP bug #50688

pilif:
if any descendant of Exception is instantiated from within a callback passed into the various array sorting functions, then PHP (< 7.0) will raise a warning ("Array was modified by the user comparison function").

As Raven always registers with E_ALL and then checks whether it needs to run in the handler itself, even if Raven decides not to do anything, it still instantiates the exception and thus will still cause the warning.

For example, if you run with E_NOTICE off and thus register Raven to not run on E_NOTICE, but then the user raises an E_NOTICE in a usort() callback, php bug 50688 will cause a warning to be thrown (which then likely *will* activate Raven on the second pass)

### DIFF
--- a/lib/Raven/ErrorHandler.php
+++ b/lib/Raven/ErrorHandler.php
@@ -81,14 +81,13 @@ class Raven_ErrorHandler
         // E_PARSE, E_CORE_ERROR, E_CORE_WARNING, E_COMPILE_ERROR, E_COMPILE_WARNING, and
         // most of E_STRICT raised in the file where set_error_handler() is called.
 
-        $e = new ErrorException($message, 0, $type, $file, $line);
-
         if (error_reporting() !== 0) {
             $error_types = $this->error_types;
             if ($error_types === null) {
                 $error_types = error_reporting();
             }
             if ($error_types & $type) {
+                $e = new ErrorException($message, 0, $type, $file, $line);
                 $this->handleException($e, true, $context);
             }
         }


dcramer:
I dont quite understand the core issue here (or how we'd prevent it in the future), but we don't need the code to exist how it is today so I'm going to accept this change.

pilif:
let me explain. 

PHP < 7.0 has a bug where the following code would cause a warning to be thrown:

```php
$in = [1,2,3];
usort($in, function($a, $b){
    new Exception("foo");
    return $a-$b;
});
```

`usort` and friends try to make sure that the thing they are sorting isn't being changed by the callback. Unfortunately, when you instantiate an object of class `Exception` or when you use any of `debug_backtrace` and friends, they will get a reference to their scope, thus increasing the reference count on the to-be-sorted array and thus the check in `usort` will trigger and a warning will be thrown even though nothing was actually modifying the array.

Now let's bring sentry into the mix, but before that, let's modify our sorting callback so it doesn't itself instantiate an Exception but it will throw a notice (also, this problem applies to the whole call stack past `usort`, so this also applies to anything that's called from the callback):

```php
$in = [1,2,3];
usort($in, function($a, $b){
    return $a-$b+$c; // not the use of an undefined variable here. This will throw an E_NOTICE
});
```

Now assuming that sentry is configured to not register a handler for E_NOTICE (using `registerErrorHandler(true, E_ALL & ~E_NOTICE)`), it itself will [still register a handler for E_ALL](https://github.com/getsentry/sentry-php/blob/fef7c0f721097f2109e2c0d967c17147a8f08c7d/lib/Raven/ErrorHandler.php#L156), so even if we're not interested in logging E_NOTICE, the error handler (`handleError`) will be called when php evaluates the `$c`. It will then decide that it won't be doing anything, but only *after* first instantiating an `ErrorException`.

Even if that instance isn't being used at all, the fact that this instance was created (and in the process PHP captured the callstack), the overzealous check in `usort` will now fire and a warning (with E_WARNING) will be issued when usort either completes or runs the callback the next time.

Now likely, sentry is configured to actually *do* something on E_WARNING, causing the request processing to stop, an error to be logged to sentry and in many cases, a generic error 500 page to be displayed to the user.

This means that without the patch here, anything excluded by the error level parameter to `registerErrorHandler`, when it triggers inside of an `usort` (or friends) callback, will immediately be promoted to an E_WARNING.

I hope this is a bit more clear now - I felt this was too much text for the commit message :p

dcramer:
Makes sense, thanks!

PR #365, opened 2016-10-13
use uopz for travis and PHP 7

remicollet:


remicollet:
This PR changes:
- allow PHPUnit 5 (which is better for PHP 7)
- disallow failure for PHP 7
- add nightly in the PHP matrix

And now, test suite passes with PHP 7, travis is "green".

siwinski:
THANKS @remicollet !

Note: This change allows me to update this package downstream in Fedora/EPEL since Fedora 25+ uses PHP 7+

PR #368, opened 2016-11-01
Fixes #367 - array_walk_recursive being possibly called with a string in case http['data'] is a raw string.

### LINKED ISSUE #367: SanitizeDataProcessor crashes when POST data contains RAW data (for instance JSON DATA)
Introduced with rev ba7ebd9.

```
 public function sanitizeHttp(&$data)
 +    {
          $http = &$data['request'];		          $http = &$data['request'];
 -        if (empty($http['cookies'])) {		 +        if (!empty($http['cookies'])) {
 -            return;		 +            $cookies = &$http['cookies'];
 +            if (!empty($cookies[$this->session_cookie_name])) {
 +                $cookies[$this->session_cookie_name] = self::MASK;
 +            }
 +        }
 +        if (!empty($http['data'])) {
 +            array_walk_recursive($http['data'], array($this, 'sanitize'));
          }		          }
 +    }

```

$http['data'] could be a string, which will crash 'array_walk_recursive' because the first argument should be an array.

So in this case the check should be:

```
if (is_array($http['data']) && count($http['data'])) {
 ...
}
```

Encountered while updating from v0.22 to v1.5.0.

niborb:
fixes #367

### DIFF
--- a/lib/Raven/SanitizeDataProcessor.php
+++ b/lib/Raven/SanitizeDataProcessor.php
@@ -80,7 +80,7 @@ class Raven_SanitizeDataProcessor extends Raven_Processor
                 $cookies[$this->session_cookie_name] = self::MASK;
             }
         }
-        if (!empty($http['data'])) {
+        if (!empty($http['data']) && is_array($http['data'])) {
             array_walk_recursive($http['data'], array($this, 'sanitize'));
         }
     }


niborb:
Yes sure, I will change it.

niborb:
updated.

dcramer:
One last tidbit here, as I'm pretty sure this is going to throw errors based on the order of checks:

```
$ php -r '$arr = array(); echo is_array($arr["foo"]);'
PHP Notice:  Undefined index: foo in Command line code on line 1
```

AFAIK the empty check needs to come first, as calling is_array on a missing element will choke.

niborb:
Yes you are right. Updated again ;-)

dcramer:
Thanks!

niborb:
thanks!

PR #369, opened 2016-11-03
Add an error_types option to the Raven Client

d3f3kt:
With this error_types option it's now possible to set directly which kinds of errors should
be reported.

... and furthermore i changed the privates variabled to protected.


dcramer:
Thanks!

PR #372, opened 2016-11-11
Added missing dollar sign to 'key' variable in Client::buildCurlCommand()

adallaway:
Broke here: https://github.com/getsentry/sentry-php/commit/09de00d2fb3b7f664eb140b23b5ddcbf941b085b


dcramer:
Hrm not sure how this slipped through as it seems like it'd be causing some serious issues. 

Thanks!

PR #376, opened 2016-11-15
fixing small notices on the ErrorHandler

igorsantos07:
- missing return point on handleError()
- wrong return and missing param on PHPDoc of registerErrorHandler()
- missing documented prop $client

The only point of attention here is: as `$this-client` was being created on the fly it was, by default, `public`, but I set it to `private` as it should only be used inside the class anyway...


dcramer:
thanks!

PR #377, opened 2016-11-16
get errors only from a specific path

Thamaraiselvam:
`$client->setStrictErrorCapturePath($path);`

unit tested works fine.

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -428,7 +428,7 @@ class Raven_Client
     {
         //avoid errors other than specific path
         $exception_file = $exception->getFile();
-        if (isset($this->strict_error_capture_path) && !empty($this->strict_error_capture_path) && stripos($exception_file, $this->strict_error_capture_path) === FALSE) {
+        if (isset($this->strict_error_capture_path) && !empty($this->strict_error_capture_path) && stripos($exception_file, $this->strict_error_capture_path) === false) {
             return null;
         }
 
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -428,7 +428,7 @@ class Raven_Client
     {
         //avoid errors other than specific path
         $exception_file = $exception->getFile();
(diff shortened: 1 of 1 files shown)


Thamaraiselvam:
maintainers, can you check this pull request?

lib/Raven/Client.php
```diff
@@ -414,6 +426,12 @@ public function captureMessage($message, $params=array(), $data=array(),
      */
     public function captureException($exception, $data=null, $logger=null, $vars=null)
     {
+        //avoid errors other than specific path
+        $exception_file = $exception->getFile();
+        if (isset($this->strict_error_capture_path) && !empty($this->strict_error_capture_path) && stripos($exception_file, $this->strict_error_capture_path) === false) {
```
stayallive: You can leave out the `isset` check and leave only the `empty`.

See: http://stackoverflow.com/a/4559976/1580028

stayallive:
Did you perhaps forget to also include the unit tests?

lib/Raven/Client.php
Thamaraiselvam: Thanks for the review. Removed `isset`

Thamaraiselvam:
I have tested it and using it in my application.

Thamaraiselvam:
Okay. I did very basic tests.
I set my plugin path 
`$sentry_client->setStrictErrorCapturePath($my_plugin_dir);`

I get the errors only on my plugin path. even other plugins or Wordpress throws the errors it just ignores them.

nokitakaze:
This can not be approved without code coverage (unit tests)

ste93cry:
Wouldn't be better to use the `send_callback` option to customize whether the event originated from a specific source (directory or file) should be sent?

nokitakaze:
Already done in 2.0

stayallive:
Thanks for the contribution. This has indeed been implemented in 2.0, if you would still like it in the 1.x version please reopen the PR with the changes/tests requested above.

nokitakaze:
There is no need to add new behaviour to old version since it (version 1.0) will be stalled from 2018 March.

PR #379, opened 2016-11-19
Added message name tracking for compatibility with Sentry server, version 8.3.0

acarbone:
Sentry server 8.3.0 tracks all events generated by sentry-php as <no message value>.
To fix the incompatibility the exception's name has been added to the actual event.


dcramer:
I'm not sure what you intended (it got formatted out in the description), but we explicitly do not send the 'message' attribute anymore as its only for legacy events. You'll need to upgrade your sentry server or use an older version of sentry/sentry.

PR #380, opened 2016-11-22
Fixed CURL header escaping

adallaway:
There's another problem with the shell escaping (Also see #372) inside buildCurlCommand().

Curl headers are being escaped as [-H 'key:' 'value'] instead of [-H 'key: value'].


dcramer:
Good catch, thanks!

I'm going to try to get some tests added for this is if I get some free cycles, so we at least assert on the expected output here.

PR #400, opened 2017-01-25
Add json string data to default context

rreynier:
As it becomes more common for the php apps to be backend/api only, it becomes more common for communication to be done via json data.  When you POST a json string to the php backend, there is no $_POST data, therefore this context is lost.  This PR suggests a solution which tries to retrieve the json string data if no $_POST data is set.

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -559,9 +559,21 @@ class Raven_Client
         if (!empty($_POST)) {
             $result['data'] = $_POST;
         }
+        
+        // if the $_POST super global is empty, try to retrieve a json string 
+        if(empty($_POST)){
+            try {
+                $data = file_get_contents( 'php://input' );
+                $result['data'] = json_decode( $data, true );
+            } catch ( Exception $exception ) {
+                // do nothing
+            }
+        }
+        
         if (!empty($_COOKIE)) {
             $result['cookies'] = $_COOKIE;
         }
+        
         if (!empty($headers)) {
             $result['headers'] = $headers;
         }
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -561,11 +561,11 @@ class Raven_Client
         }
         
         // if the $_POST super global is empty, try to retrieve a json string
-        if(empty($_POST)) {
+        if (empty($_POST)) {
             try {
                 $data = file_get_contents('php://input');
                 $result['data'] = json_decode($data, true);
-            } catch(\Exception $exception) {
+            } catch (\Exception $exception) {
                 // do nothing
             }
         }
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -565,7 +565,7 @@ class Raven_Client
             try {
                 $data = file_get_contents('php://input');
                 $result['data'] = json_decode($data, true);
-            } catch (\Exception $exception) {
+            } catch(\Exception $exception) {
                 // do nothing
             }
         }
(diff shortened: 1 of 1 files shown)


dcramer:
IIRC this is only able to read once, is it not? That's why we avoiding doing this in the past.

rreynier:
I don't think so @dcramer but let me confirm.  You are suggesting that `file_get_contents('php://input')` would clear php://input?

dcramer:
Yeah -- often these are streams which can only be read once.  I might be wrong in this case though.

rreynier:
A quick test shows it does not clear the stream in php 7.0.8.  I did the following:

```
$data1 = file_get_contents('php://input');
$data2 = file_get_contents('php://input');
echo $data1;
echo $data2;
```

Both echo's yielded the same output. Maybe this is something that happens with old versions of PHP?

Unfortunately I am not sure how to write a test to confirm this since I don't know how to get in front of php://input.

dcramer:
Ah, its documented:

> Note: Prior to PHP 5.6, a stream opened with php://input could only be read once; the stream did not support seek operations. However, depending on the SAPI implementation, it may be possible to open another php://input stream and restart reading. This is only possible if the request body data has been saved. Typically, this is the case for POST requests, but not other request methods, such as PUT or PROPFIND.

Ok so maybe what we'll do is just detect version, and say if 5.6+ we can do this.

rreynier:
Ah good catch! Did not know this. Let me modify the PR for version checking and resubmit.

rreynier:
@dcramer there you go.  Any thoughts on adding tests here?  I am not sure how we could test this...

dcramer:
It sounds like people would suggest dependency injection, but there's no real great way to test the behavior of php://input without an isolated process:

http://stackoverflow.com/questions/38197116/how-do-i-simulate-php-input-in-php

lib/Raven/Client.php
```diff
@@ -559,9 +559,22 @@ protected function get_http_data()
         if (!empty($_POST)) {
             $result['data'] = $_POST;
         }
+        
+        // if the $_POST super global is empty, try to retrieve a json string
+        // prior to PHP 5.6, a stream opened with php://input could only be read once, so only do this with 5.6 or up
+        if (version_compare(PHP_VERSION, '5.6', '>=') && empty($_POST)) {
+            try {
+                $data = file_get_contents('php://input');
+                $result['data'] = json_decode($data, true);
```
dcramer: Thinking more about this, I think theres two more concerns we should address:

1. Dont read the entire stream (I'm not sure if this can be a DoS vector). This seems achievable via the $maxlen param, but maybe can also be achieved with one of the other file helpers.
2. It'd probably be good that we still capture the raw value even if its not JSON. Not really a requirement just a nice to have.

nokitakaze: json_decode? What the h...? There is at least [SOAP](https://en.wikipedia.org/wiki/SOAP). I hate it, but it exists

rreynier: Are you suggesting we also try to decode SOAP responses (XML)?

nokitakaze: No, we need to pull raw text instead of interpret it.
By the way, what is transferred if json will contain a object?  I think it\'s a string
```
[Object object]
```

rreynier: The point is to TRY to parse it if it happens to be Json data. I suppose if it fails to parse (without an exception) I should not set 'data'.

nokitakaze:
Maybe it should be PR for getsentry:2.0
There is no need to add big features to old branch because we must fix error if they are exist

ste93cry:
Closing as this has been fixed in the referenced PR

PR #406, opened 2017-02-10
Add processors to remove POST data and cookies - fixes #405

### LINKED ISSUE #405: Add processors to remove POST data and cookies
In the Ruby Raven client two processors that remove cookies and POST data from the request before submitting the event to the server have been added in getsentry/raven-ruby#252 and getsentry/raven-ruby#445. However, it seems that the PHP client is not maintained as much as the other and so it still leaks possible sensitive informations. The PR I submitted should fix the issue in the same way as the PR getsentry/raven-ruby#466. Please note that I took the opportunity to move the processors to a new directory and I deprecated the old `Raven_SanitizeDataProcessor` class: it should be removed in the next major version.

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Fixed tickets | #405 

The POST data and the cookies could leak sensitive informations, so they are both cleared before submitting the data to the server.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,7 @@
 - The default exception handler will now re-raise exceptions when
   ``call_existing`` is true and no exception handler is registered (#421).
 - Collect User.ip_address automatically (#419).
+- Added a processor to remove web cookies and another to remove HTTP body data for POST, PUT, PATCH and DELETE requests. They will be enabled by default in ``2.0`` (#405).
 
 1.6.2
 -----
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -380,7 +380,7 @@ class Raven_Client
     public static function getDefaultProcessors()
     {
         return array(
-            'Raven_SanitizeDataProcessor',
+            'Raven_Processor_SanitizeDataProcessor',
         );
     }
 
--- a/lib/Raven/Processor.php
+++ b/lib/Raven/Processor.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Base class for data processing.
  *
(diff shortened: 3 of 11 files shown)


dcramer:
I'm +1 on adding this, but we cant have it enabled by default as it will cause a lot of confusion for users, and by default POST data isnt highly sensitive.

ste93cry:
>we cant have it enabled by default

I though the same too after submitting the PR. I checked how the the people working on the Ruby client solved the problem and there is a link to the [SDK Reference](https://docs.sentry.io/clientdev/overview/) that says the following:

>Feature based support is required for the following:
>- If cookie data is available, it’s not sent by default
>- If POST data is available, it’s not sent by default

From what I understand it means that it must be configurable (it is by setting the `processors` option), but by default the cookies and POST data are not sent. I perfectly agree on the fact that people relying on this behaviour before this change would suddendly stop seeing these details. Please also have a look at getsentry/sentry-docs#30

dcramer:
@ste93cry thats fair, and looking back now I recall that conversation. So I'm ok with this, though we'll need to expand docs/CHANGES to reflect how you go about capturing this data.

ste93cry:
I fixed the conflicts on my PR and I've added a new entry to the `CHANGES.md` file similar to the one in sentry-ruby to explain the changes.

dcramer:
So my only real thoughts at this point are the complexity of configuration. I'm thinking that maybe these processors should always exist, and we can just add Client options for like:

- scrub_http_body
- scrub_http_cookies

(Maybe scrub isnt the right word)

ste93cry:
I would not add more options because processors themselves are already options (you can turn them on/off). I propose instead to make both processors off by default in `1.x`, and instead switch them on in `2.x`. This way we are not going to break compatibility

dcramer:
@ste93cry would you mind making the change to make them opt-in for now, and we can make the backwards incompatible change in 2.x like you mentioned?

ste93cry:
Done. I've also updated the `CHANGES` file to specify that the processors will be enabled by default only from `2.0`

ste93cry:
Should we also remove data coming from [not-safe methods](https://tools.ietf.org/html/rfc7231#section-4.2.1)? (e.g. `PUT`, `PATCH`)? They all could contain sensitive data like in the POST method

dcramer:
@ste93cry yea we should -- im not sure if the other SDKs are only doing it on POST, but if they are I'd consider it incorrect. It should just be "remove HTTP body".

ste93cry:
Everything done, I renamed the processor to something more appropriate (in case it should be done the same in the other SDKs to maintain consistency) and the HTTP body is cleared in case the request method is POST, PUT, PATCH or DELETE (I'm unsure this last is necessary).

>im not sure if the other SDKs are only doing it on POST, but if they are I'd consider it incorrect.

I didn't check all the SDKs, but I'm sure the following ones only remove POST data:

- Ruby
- Python

dcramer:
Thanks!

PR #409, opened 2017-02-14
PHPDoc, properties of classes, warnings, code coverage, code style

nokitakaze:
Sorry guys. I just can not see that anymore. I open my Idea and then I want to pull my eyes out.
warnings_in_main_class
actual_expected_swap
equals_wrap_codestyle

I added and fixed many phpdoc-comments of class methods.
I added properties in classes.

And impove code coverage.
code_coverage_improve

### DIFF
--- a/lib/Raven/Breadcrumbs.php
+++ b/lib/Raven/Breadcrumbs.php
@@ -16,13 +16,13 @@
 
 class Raven_Breadcrumbs
 {
-    var $count;
-    var $pos;
-    var $size;
+    public $count;
+    public $pos;
+    public $size;
     /**
      * @var array[]
      */
-    var $buffer;
+    public $buffer;
 
     public function __construct($size = 100)
     {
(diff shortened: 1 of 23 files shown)


dcramer:
This patch is great though I am worried about our ability to maintain it without linters/automated tooling.

stayallive:
Maybe it's time for StyleCI or some other tool to enforce a coding style :) Big :+1: from me for this.

nokitakaze:
I just tested [PHP CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer), but it seems Raven's codebase is fundamentally broken.

[PSR-0](http://www.php-fig.org/psr/psr-0/)
> A fully-qualified namespace and class **must** have the following structure \\\<Vendor Name\>\\(\<Namespace\>\\)\*\<Class Name\>

default

dcramer:
@nokitakaze its not broken, we just historically support a lot of PHP versions which is why we don't use namespaces

nokitakaze:
Oh sorry. I just forgot about these versions of PHP what don't support normal OOP

dcramer:
Thanks again!

PR #410, opened 2017-02-14
Travis: php 7.1, permutations for xdebug and assertions

nokitakaze:
PHP 7.1 passes tests with my general code improve #409 

You can see it here: [https://travis-ci.org/nokitakaze/sentry-php/builds/201633007](https://travis-ci.org/nokitakaze/sentry-php/builds/201633007)

I don't know why haven't added PHP 7.1, it's already current stable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,14 +6,36 @@ php:
   - 5.5
   - 5.6
   - 7.0
+  - 7.1
   - nightly
   - hhvm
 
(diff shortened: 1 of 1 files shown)


dcramer:
Thanks! A lot of what we do is based on community feedback here, so 7.1 missing is just something we hadn't gotten to adding in yet.

PR #413, opened 2017-02-16
[2.0] Deferred Client for Raven

nokitakaze:
This is a client extending which stores all messages in file queue and then (in a separate file) it send them. So all messages/events **will** be delivered.

We need a client we can rely on. Errors are something which we **must** deliver. They contain important data so we must be sure data will be delivered to Sentry server. This is not a wish, it is a **necessary behavior** for a library that *logs* errors.
Without this behavior this repository is just a plaything, not a production ready code supplement.

Side effect of this behavior is a speed boost. Standard client tries to use network after every capture-call.  I don't know how about you, but I can't afford to slow down a code execution. The language which does not support JIT is slowly enough by itself.

```php
$client = new \Raven\DeferredClient(null, ['install_default_breadcrumb_handlers' => false]);
// ... captures events
```

// now you can do that
`unset($client);`
// or that
`$client->store_pending_events_to_queue();`

// And now in some other cron file you can send all events in queue to server at once
```php
$client = new \Raven\DeferredClient('http://public:secret@example.com/1');
// @todo place here some mutex
$client->read_all_queue_and_send_all_events();
```


nokitakaze:
Maybe I chose wrong word "deferred" for describe "delayed sending"

Jean85:
I think this approach is not very robust. If, for any reason, the execution dies (fatal error? segfault?) the events can be lost.

Is performance your only concern? Because in that case, we still can tackle that leveraging Guzzle (that was suggested in #424) and its async request feature: the event would be sent in the background, as soon as possibl,e without halting the execution.

nokitakaze:
> Is performance your only concern?

No. Main thing is delivering. 

> events can be lost

They will be lost if Sentry served crushes or server loses connection to internet. You can do `$client->store_pending_events_to_queue();` after every call if want to be sure. And I will do the option which will make Client stores events after pending events reach a certain size.

At this moment my main project is some kind of banking processing so I **MUST** be sure I receiving all events about errors or something

ste93cry:
I'm 👎 too, I don't see how this kind of feature would be useful in the main core library. Instead, if you need this in your project just extend the `Client` class and implements it yourself.

Jean85:
> At this moment my main project is some kind of banking processing so I MUST be sure I receiving all events about errors or something

Well, sending them at a later point seems to me unreliable too. You don't have any guarantee of when that will be delivered.

I would store in this likely fashion only events that failed in the previous request, but that could be troublesome too: what if sending the event itself makes the system crash? It could create a catastrophic loop of failures.

nokitakaze:
> and implements it yourself

Yes. This is exactly what I did.

> sending them at a later point seems to me unreliable too

So you are not interested in receiving errors, right?

> You don't have any guarantee of when that will be delivered

Yes. This is how things work now. At this moment you have no guarantee what your events will be delivered

> sending the event itself makes the system crash

That\'s why you should send them in separate cron script.

Jean85:
> So you are not interested in receiving errors, right?

I'm interested in receiving them as soon as they happen.

> Yes. This is how things work now. At this moment you have no guarantee what your events will be delivered

Why? Sending them at a latter moment does not guarantee nothing in addition to the normal flow, apart the possibility for retry, that can be implemented in the normal flow as well (with async to avoid performance issues)

nokitakaze:
> I'm interested in receiving them as soon as they happen

And what if your event could not be delivered the first time? This approach will guarantee that there will be at least retries to send errors again and again and again.

> can be implemented in the normal flow as well
> async

I want to see how you do that while your host is under heavy load

ste93cry:
Just to let you know, I'm protoyping a bit on this in my spare time and in case I will update the PR accordingly. Instead of a deferred client I'm trying to see if something like a spooling system makes more sense, along the lines of SwiftMailer.

Jean85:
Then keep in mind that the spooling system needs to be hooked somewhere, especially with the (official) integrations (Symfony, Laravel...)

nokitakaze:
Obsolete

Issue #192, opened 2015-03-05
[proposal] Make 1.0 support PHP 5.3+ and use PSR4


ameliaikeda:
PHP 5.2 is long buried and was [end-of-life](https://php.net/eol.php)'d 4 years ago. No major server distro (debian, ubuntu, centos, redhat, fedora, freebsd) provides updates any more, and custom repositories just for this purpose have since stopped giving security updates to 5.2.17<sup>([1](https://webtatic.com/packages/php53/), [2](https://code.google.com/p/php52-backports/) - last updated almost 2 years ago)</sup>.

PHP 5.3 is also [no longer supported](https://php.net/supported-versions.php), but third parties will still likely give 5.3 updates for a few years in response to CVEs, at least.

Rather than encourage people to keep some pretty old configurations up, why not overhaul for a new major release and use [PSR4](http://www.php-fig.org/psr/psr-4/)? raven-php already uses composer, and you can bet quite heavily that pretty much every recent PHP developer using symfony2/laravel/zf2 (or anything else using composer) will jump on the new version like wildfire (rather than using another package that proxies `Raven_*` in order to allow namespaces or just cleaner interfaces to be used).

I can spend the weekend preparing a pull request for this (It's a major change, but namespaces have been in PHP for half a decade now and developers have come to expect them).

Proposal:
- [ ] Start a new branch for namespace support (`1.0` or `php53`, etc)
- [ ] Tag a `1.0` release on the new branch
- [ ] Actively maintain both branches for a short while (possibly 3 months, more if the community offers)
- [ ] Once active maintenance is finished, stick to security updates
- [ ] Stick to [semver](http://semver.org) for releases; `0.*.*` is 5.2+, `1.*.*` is 5.3+

In addition, issues such as #171, as well as general use of `exec` by the package, could be abstracted by a HTTP library such as [Guzzle](https://github.com/guzzle/guzzle) or [Buzz](https://github.com/kriswallsmith/Buzz) (and could use interfaces).

[...]

PR #414, opened 2017-02-16
Sentry 2.0 [base]

nokitakaze:
Base version for Sentry-PHP 2.0. 
* Namespace Raven instead of prefix "Raven_"
* Removes Raven_Compat
* Renew dependencies in composer.json and .travis.yml

#192, #197 & #424


dcramer:
fyi https://wordpress.org/about/stats/

I'm ok with us dropping support, but we'd likely want to document "If you're using 5.2, the last version of sentry/sentry that supports it is X"

nokitakaze:
If I'm not mistaken, composer supports downloading these versions of packets what supports needs versions of other packets. So if someone will write 
`composer require getsentry/sentry-php`
It will install old version.

Please **don\'t** merge this pull request at this state (798805b). At first I must add 
`Raven_Client extends \Raven\Client {
}`
otherwise we will crush all packets depends on getsentry/sentry-php.

ste93cry:
Please, there is no point in supporting a version of PHP as old as PHP 5.2 that reached end of life more than [**6 years**](http://php.net/eol.php) ago. Symfony supports PHP `>=5.3.9` in their 2.x branches and `>=5.5.9` for their 3.x branches. Also, I see a lot of PRs about raising the requirements of PHP, but this does not mean that we should move all classes to namespaces because it's a BC. I did not verify if it works, but looking at how the guys of PHPUnit updated their PHP requiremens we should release a new 1.x version that drops PHP 5.2.x support by updating the Composer constraint to **at least** `>=5.6` and remove the custom autoloader `Raven_Autoloader`. Then, all changes related to PSR-4 support, namespaces, etc should be done in a `2.0` branch because they all break BC. On my fork a week ago I pushed the first commit that is similar to the one of this PR, which makes the directory structure compatible with PSR-4 and introduces namespaces. I have a lot of ideas for new 2.x version of this library, but this discussion should be done in a new issue.

nokitakaze:
At my own project I am even more radical https://github.com/nokitakaze/ascetcms/blob/v079/.travis.yml#L3

> all changes related to PSR-4 support, namespaces, etc should be done in a 2.0 branch because they all break BC.

I don't think it\'s necessary. All people calls only Raven_Client. 
At this moment I fix this joking branch. I thought it will be rejected and won't be discussed at all.

ste93cry:
I just noted that even if this library claims to support PHP `5.2`, Travis does not test it. So how can the maintainers be sure their versions work on a so old version?

>I don't think it's necessary. All people calls only Raven_Client. Just give it next minor version

Minor versions can't break BC, this is how SEMVER works. It's not important if users just have to update a line of code, a BC is a BC and cannot be done imho. There is nothing wrong in making a `2.x` release that adds everything we want and let `1.x` as is.

nokitakaze:
- Okay. Now there is file Legacy.php contains old classes.
- And I deleted old "compatible" calls. But I didn't delete Compat.php so this version _should_ works with any code

I think I didn't broke back compatibily. I am **not** sure.

composer.json
```diff
@@ -17,13 +17,13 @@
         "monolog/monolog": "*"
     },
     "require": {
-        "php": ">=5.2.4",
-        "ext-curl": "*"
-    },
-    "suggest": {
+        "php": ">=5.3.2",
```
ste93cry: I always see `>=5.3.3` as constraint for PHP, so I would do the same here

ste93cry:
When deprecating things I like the approach of Symfony that triggers deprecation notices by putting line of codes like these at the start of a file:

```php
@trigger_error('The '.__NAMESPACE__.'\MyClass class is deprecated since version x.x and will be removed in x.x. Use the NewNameSpace\NewClass class instead.', E_USER_DEPRECATED);
```

Errors can then be catched by frameworks (e.g. Symfony shows deprecations in the profiler) and the users know they are using something that it's going to be removed. However, if you have a single file with all classes I don't know how we could do something like this. I think Symfony guys don't put that lines in the constructors because if someone extended the class then the parent constructor would not be called implicitly, but honestly I'm not sure about this. Please note that I take Symfony as example because it's the framework I use daily, there could be other ways to do the same thing.

lib/Raven/Legacy.php
```diff
@@ -14,7 +16,7 @@ class Raven_Client extends \Raven\Client
 {
 }
 
-/**
+/** @noinspection PhpDeprecationInspection
```
ste93cry: Given that I was not aware of this tag (thank you for letting me know it 😄), from what I understand it means that IDEs like PHPStorm should not alert about the deprecation of this class. But if the class is deprecated there is no point to not signal it to the users. At least, this is my opinion...

nokitakaze: No. This tag means new class extends **deprecated** class. The new class (Raven_Compat) is still deprecated and will warn about it. 

I just like green mark.
green
brown-gray

See #409

ste93cry:
>I think I didn't broke back compatibily. I am **not** sure.

I think you broke it. Taking as example people not using Composer, previously they were probably including the `Raven_Autoloader.php` file and then instantiating the class inside it (see [here](https://github.com/getsentry/sentry-php/blob/c4e0a4165a00c3a1186e8554058a29bc197f1d5e/examples/vanilla/index.php#L7) for an example). But with your changes the `Raven_Autoloader` class should be defined inside a new file called `Legacy.php`, that is not included and so it will fail. I said _should_ because I can't find the definition in that file anyway, so it's for sure a BC.

lib/Raven/Legacy.php
ste93cry: I'm not sure to understand correctly what you mean. Also, anyway, why would you add this tag only to this class?

nokitakaze: My IDEA warns me about extending deprecated class \\Raven\\Compat. I suppress this warning. That's all.

This is the only deprecated class in Raven namespace.

ste93cry: Thank you for the explanation, now it's much more clear for me. I still don't agree about it, if the inspection warns about a deprecation is for a good reason, in this case both classes are deprecated and should be treated as that. I don't see a good point in having one more line of code just to have a green mark on the IDE, but honestly I don't care too much about it

ste93cry:
You're still breaking compatibility, because if someone did not use the autoloader but manually loaded the files they do not exists anymore. You should keep the old files with the old classes in it, and maybe have a new folder with the new classes. Also, in PSR-4 libraries usually the code is located inside a `src` directory.

nokitakaze:
I can add "require once legacy" into every file. But in this case I must delete E_USER_DEPRECATED

ste93cry:
>But in this case I must delete E_USER_DEPRECATED

Wouldn't moving the `trigger_error` from the `Legacy.php` file to each file fix the problem?

nokitakaze:
We can lose backward compatibility or we can suppress E_USER_DEPRECATED. But not both at the same time. You must pull deprecated classes to files for backward compatibility.

...
Or we can pull spl autoload into any files so it will autoload deprecated classes and trigger E_USER_DEPRECATED.

ste93cry:
>We can lose backward compatibility

This is a no-go, as I said before only major versions can break BC.

>we can suppress E_USER_DEPRECATED. But not both at the same time. You must pull deprecated classes to files for backward compatibility

I do not understand this part, Which problems would arise by moving the `trigger_error` to each old file? If someone uses the new classes he has no problems because no errors are thrown, if he uses one of the old classes then it's correct to get notified of deprecation

nokitakaze:
I don't understand. You want to pull into old files something like this
```php
trigger_error...
require_once 'new/Client.php';
/**
 * @deprecated
 */
class Raven_Client extends \Raven\Client
{
}
```

?

ste93cry:
This is exact, but it would pollute the files with a lot of code just to maintain BC, so I'm not sure all this is worth the work for a `1.x` release

nokitakaze:
The main maintainer should make his views

Taluu:
Just a side note ; upgrading a dependency's version in a minor is fine if it doesn't break the public api... http://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

But +1 on dropping dead versions. Even php 5.3 is not on debian etc anymore, and it's not updated at all (not mentionning that php 5.5 has reached EOL). 

If you want a better graph on what versions are used, you should use the one the maintainer of composer made : https://seld.be/notes/php-versions-stats-2016-2-edition as this is using composer and all (rather than wordpress who is detecting "what is using wordpress", which also counts the "dead" blogs).

nokitakaze:
As I said before my opinion is to drop all old version. Only PHP 5.6+

All old version can use [1.7.0](https://packagist.org/packages/sentry/sentry#dev-master) with #409 & #412. And all new versions of PHP will use 2.0 with dropped EOLed versions.

ste93cry:
>Just a side note ; upgrading a dependency's version in a minor is fine if it doesn't break the public api... 

This is right, but renaming old classes or not being able to load them anymore is breaking the public API. This is why I continue to say that it would be easier to just release a `2.x` version where we can make all adjustment or improvements without any concern about compatibility. A minor version should just be a drop-in replacement that eventually adds new features, but it should never change or remove things

>As I said before my own opinion is to drop all old version. Only PHP 5.6/7.0+

I agree

nokitakaze:
> This branch has conflicts that must be resolved

I will fix conflicts, after @dcramer says he want to merge this pull request. We could use this pull request as a base for #424 (Sentry PHP 2.0).

But I don't know how to test PSR-4 compatibility.

ste93cry:
If this is going to be the base for 2.0, we don't need the `@trigger_error`, nor the autoloadear, nor the `Legacy.php` file

nokitakaze:
Of course

ste93cry:
@nokitakaze Please just keep this PR updated with the latest master changes until it's merged, and if you can squash all commits, but please do not add any commit that does not belongs to PSR-4 switch (e.g. deleting legacy things like the `Compat` class). For that you should open a new PR

nokitakaze:
> please do not add any commit does not belongs to PSR-4 switch

I want @dcramer merging this commit to new branch so anyone else could do another pull requests

ste93cry:
Sure, but removing legacy things like the Rave_Compact class is not in the scope of this PR. The scope of this PR was to move to PSR-4 (you should rename the `lib` directory to `src` and the `test` directory to `tests`) and use Composer, not starting the work for `2.x`

nokitakaze:
>  The scope of this PR was to move to PSR-4

No. This PR was just a **joke**. But after the maintainer approves the idea of version 2.0 (#424) I changed the description and the title of this pull request.

ste93cry:
>I could delete commits with deleting Raven_Compat, but why should I do it?

Because having too many things in a single PR makes review difficult. Removing legacy classes is a thing that should be done separately

nokitakaze:
Ok. I will remove some commits if @dcramer says so.

nokitakaze:
> you should rename the lib directory to src

As I see on http://www.php-fig.org/psr/psr-4/ & https://getcomposer.org/doc/04-schema.md folder "lib" is fine too as long as we save link to it in autoloading section in composer.json

ste93cry:
You're correct, but if you look at most of the other libraries they all tend to use `src` and `tests` as folders

ste93cry:
It looks you missed to add namespaces to the tests classes and rename all them to PSR-4 format. You also renamed only the `test` directory, not to be rude but I found only one of nearly 20 starred repositories I looked at that used `lib`. Also, one more thing: we could add the following to `composer.json`

```json
"autoload-dev": {
    "psr-4": {
        "Raven\\Tests\\": "tests/"
    }
}
```

nokitakaze:
> you missed to add namespaces to the tests

I didn't miss it. I missed it on purpose. I don't think it's necessary to add namespaces to tests

ste93cry:
It's not necessary, but as we're moving to PSR-4 we should move everything to that. Tests classes are source code, and they should be namespaced

ste93cry:
It's not necessary, but as we're moving to PSR-4 we should move everything to that. Tests classes are source code, and they should be namespaced. Just look at [https://getcomposer.org/doc/04-schema.md#autoload-dev](https://getcomposer.org/doc/04-schema.md#autoload-dev) to see that tests are namespaced and they use `src` in all examples

nokitakaze:
Renaming the main folder just because **everyone do it** isn\'t the best thing you can do. Yet again I\'ll do it if @dcramer says so. Same goes for namespaces in test classes

ste93cry:
>Renaming the main folder just because everyone do it isn't the best thing you can do.

If 90% of the people does it, I think we should too because it's a sort of "standard". Also `lib` seems more a directory containing third-parties code, not the _source code_ of the library itself.

dcramer:
Let's start by merging this in. I'll change the branch to 2.0 for now, and we can consider moving to master once we're a bit more prepared with things.

dcramer:
fyi im giving you both commit bit -- the 2.0 and master branches are protected, but you can push to 2.0 as long as the PR is up, and passes status checks (i would aim to go through a brief review on things though)

PR #415, opened 2017-02-17
fix autoload raven class failed

GrayMissing:
I think this condition is on the contrary, and will result in unable to load any class prefixed by raven


dcramer:
Doh. Sorry about that, slipped through review. Thanks!

PR #416, opened 2017-02-19
HHVM CI

nokitakaze:
pecl.php.net is down at this moment so some tests will fail. But you can see I have changed only travis script and two tests


dcramer:
What was the actual concern with hhvm vs the versioned one?

nokitakaze:
"hhvm" is HHVM 3.9. It is too old. 3.12 passes tests

PR #417, opened 2017-02-20
Add option to customize data serialization depth

romainneutron:


dcramer:
There's no reason this should be customizable as the server should be enforcing max depths, and this should simply align with what the server supports.

romainneutron:
Okay, didn't realize that

PR #418, opened 2017-02-21
Allow reset the Breadcrumb storage

romainneutron:
This is particularly useful when a PHP script is consuming message from a queue. I want to be able to reset the breadcrumb between each messages


lyrixx:
:+1:

dcramer:
Thanks!

Issue #407, opened 2017-02-12
Default exception handler not being called (PHP7)


BabakMN:
Not sure if it's expected or not. Sentry is swallowing all exceptions that it is catching without propagating them up to the native handler. As a result you get complete silence and nothing in the local logs.

`PHP 7.0.13-0ubuntu0.16.04.1`

```php
$error_handler = new \Raven_ErrorHandler($client);
$error_handler->registerExceptionHandler(true);
$error_handler->registerErrorHandler(true);
$error_handler->registerShutdownFunction();
```

Adding something like this before installing the handler makes it work.

```php
function root_exception_handler($exception) {
   throw $exception;
}

set_exception_handler('root_exception_handler');
```

Seems similar to #260.

dcramer:
@BabakMN IIRC set_exception_handler functions differently than the others -- if you bind one, default behavior is no longer invoked. We could possibly restore it by looking for an existing handler (if we can, I forget), and doing what you're doing above.

BabakMN:
I think most people would expect the default PHP handler to be called when the "call existing handler" parameter is true.

Right now it seems like it is actually behaving as "call any existing **custom** exception handlers".

So in terms of expected behaviour,

* if call_existing is false: only Sentry handles the exception, nothing else.

* if call_existing is true: Sentry and **also** the existing handler handle the exception, if a custom exception handler is not set, the native PHP exception handler is considered to be the existing handler and will be invoked.

When a custom exception handler is not installed `set_exception_handler` returns null so as a result we end up with silence (because `$this->old_exception_handler` becomes null).

How about we check for this case, and so,

* if (call_existing_handler is true) and (old_exception_handler is null): Let Sentry capture the exception, then re-throw it to invoke PHP's native handler as well.

How does that sound?

dcramer:
I'm +1 on this approach. Haven't had time to do a PR myself yet but eventually will get around to it.

BabakMN:
I'd be interested to work on a PR for this. Should I send one?

dcramer:
@BabakMN If you have the time that'd be super helpful

BabakMN:
How's this? https://github.com/BabakMN/sentry-php/commit/65021694c3e669aa3d1201281290223de0d6677e

It just re-throws if a custom exception handler is not set.

I noticed the error handler is behaving correctly by checking for null and returning false to propagate the error.

[...]

PR #421, opened 2017-02-28
Re-throw exception to invoke the native handler when a custom handler…

BabakMN:
This change causes captured exceptions to be propagated up to the native exception handler (re-thrown).

This behaviour is on by default and is controlled by a boolean variable passed to the `registerExceptionHandler` on the `ErrorHandler`.

Right now exceptions are propagated to the existing exception handler if one exists (other than the native handler, meaning a custom function has been set using `set_exception_handler`).

However if no custom exception handler is set exceptions will not propagate further than Sentry's handler. In other words exceptions are captured by Sentry but not re-thrown.

Because of this change in behaviour some may consider this a breaking change so we should consider the effects of it before release.

Because right now exceptions are not re-thrown the application is allowed to continue to execute and some applications may have unfortunately become dependent on this behaviour.

Those applications will face `Uncaught Exception` fatal errors if this change is activated for them.

Related issue: #407

### DIFF
--- a/lib/Raven/ErrorHandler.php
+++ b/lib/Raven/ErrorHandler.php
@@ -77,8 +77,12 @@ class Raven_ErrorHandler
     {
         $e->event_id = $this->client->captureException($e, null, null, $vars);
 
-        if (!$isError && $this->call_existing_exception_handler && $this->old_exception_handler) {
-            call_user_func($this->old_exception_handler, $e);
+        if (!$isError && $this->call_existing_exception_handler) {
+            if ($this->old_exception_handler !== null) {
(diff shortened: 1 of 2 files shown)


dcramer:
looks like the build is just failing due to lint -- theres a composer task to run the cs linter which will auto fix syntax so we just need to run that

dcramer:
I'm actually going to go ahead and fix the lint bits and get it merged in since its quick. Thanks!

BabakMN:
Thank you. Travis was having an outage earlier so I couldn't see why the build was failing (page was stuck with a loading indicator forever).

PR #423, opened 2017-03-03
Add sample_rate configuration

dcramer:


dcramer:
This should mimic the behavior in Java, which we should also add to client-dev spec.

dcramer:
Gonna see if i can add some simple tests around this, though not sure how reasonable it is to mock this.

lib/Raven/Client.php
```diff
@@ -971,6 +973,11 @@ public function send(&$data)
             return;
         }
 
+        // should this event be sampled?
+        if (rand(1, 100) / 100.0 > $this->sample_rate) {
```
nokitakaze: Rand?! REALLY?!

dcramer: What's the concern?

nokitakaze: Nothing. I just will replace it by mt_rand in #414

Issue #404, opened 2017-02-03
PSR-3 Compatible?


davidbarratt:
Is this library PSR-3 compatible?  
http://www.php-fig.org/psr/psr-3/

If not, can it be?

davidbarratt:
I believe we should be able to just include this package as a dependency  
https://packagist.org/packages/psr/log

and then implement the interface.

dcramer:
I think there was a PR to bring in full compatibility. I'd be +1 on supporting it.

nokitakaze:
PSR-3 need namespaces

nokitakaze:
Easy: #426

ste93cry:
Sentry is not a logging library, so I don't think it's appropriate to implement PSR-3. In case, it should _use_ PSR-3. For logging there are libraries like Monolog that works well, and Monolog itself has an handler that reports to Sentry. It seems to me we are just reimplementing something already existent. Plus, Sentry focus on tracking errors, why they happened and where, it's not a place where to store messages like a log file

nokitakaze:
> Sentry is not a logging library

0_o. Sentry is definitely logging library. I use Sentry for logging purpose. It's really usable for this. Otherwise, why would an INFO level be needed?

This is Sentry. One of my projects. Look. It's full of information logs.
![default](https://cloud.githubusercontent.com/assets/9131758/23776120/4f845a06-053d-11e7-9f59-9587c1077884.PNG)

dcramer:
Just to avoid jumping into the weeds:

https://sentry.io/vs/logging/

We think of Sentry as an additional thing beyond logging due to the depth it goes into. People compare it to logging because its the concept they're familiar with. It's a "log", but not traditional logging.

ste93cry:
Quoting from that page:
>You'll need to determine what's right for you, but the important thing to remember is Sentry aims to support a lot, but it's primarily for capturing errors.

@nokitakaze Sure, I did not say that you can not capture logs, I said I think that using Sentry's focus is tracking errors and not log messages. If you want to capture logs, you should use libraries like Monolog that can even report to Sentry (as I said there is an handler for this out-of-the-box), but we shouldn't transform Sentry itself into a logger library. This is my opinion, at least

nokitakaze:
Why do I need two or more logging systems if I have the good one? How can you conceptually separate logs with errors from logs with events? And for what?

davidbarratt:
[...]

PR #426, opened 2017-03-08
[2.0] PSR-3: Logger Interface

nokitakaze:
Issue: #404
http://www.php-fig.org/psr/psr-3/

```php
$raven = new \Raven\Client(...);
$client = \Raven\PSR3Client($raven);
$client->log(...);
```
or
```php
$client = \Raven\PSR3Client([...]);
$client->log(...);
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,7 @@
         "ext-hash": "*",
         "ext-json": "*",
         "ext-mbstring": "*",
+        "psr/log": ">=1.0.2",
         "ext-curl": "*"
     },
     "suggest": {
--- a/lib/Raven/PSR3Client.php
+++ b/lib/Raven/PSR3Client.php
@@ -1,0 +1,236 @@
+<?php
+
+namespace Raven;
+
+class PSR3Client implements \Psr\Log\LoggerInterface
+{
+    /**
+     * @var Client
+     */
+    protected $_client;
+
+    /**
+     * PSR3Client constructor.
+     *
+     * @param array|Client $inner
+     * @throws \Psr\Log\InvalidArgumentException
+     */
+    public function __construct($inner = [])
+    {
+        if (is_object($inner)) {
+            if (!($inner instanceof \Raven\Client)) {
+                throw new \Psr\Log\InvalidArgumentException('Object is not a Raven Client');
+            }
+            $this->_client = $inner;
+        } elseif (is_array($inner)) {
+            $this->_client = new Client(null, $inner);
+        } else {
+            throw new \Psr\Log\InvalidArgumentException('Object is not a Raven Client');
+        }
+    }
+
+    /**
+     * @return Client
+     */
+    public function getClient()
+    {
+        return $this->_client;
+    }
+
+    /**
+     * Interpolates context values into the message placeholders.
+     *
+     * @param object|string $message
+     * @param array         $context
+     *
+     * @return string
+     * @doc http://www.php-fig.org/psr/psr-3/
+     */
+    public static function interpolate($message, array $context = [])
+    {
+        // build a replacement array with braces around the context keys
+        $replace = [];
+        foreach ($context as $key => $val) {
+            // check that the value can be casted to string
+            if (!is_array($val) && (!is_object($val) || method_exists($val, '__toString'))) {
+                $replace['{'.$key.'}'] = $val;
+            }
+        }
+
+        // interpolate replacement values into the message and return
+        return strtr($message, $replace);
+    }
+
+    /**
+     * RFC 5424 log level to Sentry log level
+     *
+     * @var string $level
+     * @return string|null
+     * @doc http://tools.ietf.org/html/rfc5424
+     */
+    public static function getSentryLogLevel($level)
+    {
+        switch ($level) {
+            case \Psr\Log\LogLevel::EMERGENCY:
+            case \Psr\Log\LogLevel::CRITICAL:
+                return Client::FATAL;
+            case \Psr\Log\LogLevel::ERROR:
+            case \Psr\Log\LogLevel::ALERT:
+                return Client::ERROR;
+            case \Psr\Log\LogLevel::WARNING:
+                return Client::WARNING;
+            case \Psr\Log\LogLevel::NOTICE:
+            case \Psr\Log\LogLevel::INFO:
+                return Client::INFO;
+            case \Psr\Log\LogLevel::DEBUG:
+                return Client::DEBUG;
+            default:
+                return null;
+        }
+    }
+
+    /**
+     * System is unusable.
+     *
+     * @param string|object $message
+     * @param array         $context
+     *
+     * @return void
+     */
+    public function emergency($message, array $context = [])
+    {
+        $this->log(\Psr\Log\LogLevel::EMERGENCY, $message, $context);
+    }
+
+    /**
+     * Action must be taken immediately.
+     *
+     * Example: Entire website down, database unavailable, etc. This should
+     * trigger the SMS alerts and wake you up.
+     *
+     * @param string|object $message
+     * @param array         $context
+     *
+     * @return void
+     */
+    public function alert($message, array $context = [])
+    {
+        $this->log(\Psr\Log\LogLevel::ALERT, $message, $context);
+    }
+
+    /**
+     * Critical conditions.
+     *
+     * Example: Application component unavailable, unexpected exception.
+     *
+     * @param string|object $message
+     * @param array         $context
+     *
+     * @return void
+     */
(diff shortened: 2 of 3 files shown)


lib/Raven/Client.php
```diff
@@ -1435,4 +1435,205 @@ public function close_curl_resource()
...
+    /**
+     * Interpolates context values into the message placeholders.
+     *
+     * @param object|string $message
+     * @param array         $context
+     *
+     * @return string
+     * @doc http://www.php-fig.org/psr/psr-3/
+     */
+    public static function interpolate($message, array $context = array())
```
davidbarratt: Why is this method static?

nokitakaze: Why it should not?
I am a apologist of some simple principle: Methods without "$this" SHOULD be static.

6162cbc

davidbarratt: I guess if there is no use of `$this` AND the function is pure, then I don't have a reason for a method not to be static. Since these two things are true, then I rescind my comment.

nokitakaze:
Because this pull request breaks back compatibility I can redirect it to branch feature/2.0. But only after #414 will be merged

nokitakaze:
I will resolve conficts in two days

ste93cry:
I'm still 👎 , as said by @dcramer Sentry is not a log library, so it makes no sense to make the client a logger itself. I instead favour to support integration of PSR-3 standard into this library in terms of _using_ the logger

nokitakaze:
> to support integration of PSR-3

Sorry. I don't understand what do you mean. 

I impemented described behaviour of [PSR-3](http://www.php-fig.org/psr/psr-3/) (throwing Exception, placeholders, log levels, etc)

ste93cry:
>Sorry. I don't understand what do you mean.

As pointed out many times in other issues and PRs, Sentry is not a traditional logger but instead something that keep tracks of errors and nothing else, it's something that tracks what, why, where an error happened and reports it to something else. So, while you can see it as a logger that could implement PSR-3 there is no pro in doing it because logging messages is not what Sentry has been created for. Instead, I support the integration of a logger instance (for example, it could be useful) to log what's happening during the execution of Sentry (for example logging that an event is being sent to the server). _Integration_ is not _implementation_: one thing is implement the whole LoggerInterface, one thing is using it

dcramer:
I think I would rather this be a separate class (or even separate package) than make this core. The problem with making this behavior highly accessible (and even more importantly, the default) is then people misuse Sentry and end up with a poor experience.

nokitakaze:
If we use some separate class like a 
```php
class \Raven\ClientPSR3 extends \Raven\Client implements \Psr\Log\LoggerInterface {}
```

we must change base class for all extending classes with overrided behaviour like in PR #413. 

I don't see anything bad in adding implements of LoggerInterface to base class. Or I can do a trait so anybody could add this trait and "implements \Psr\Log\LoggerInterface" to their extending classes.

Again I don't see anything bad if somebody uses
```php
$raven = new \Raven\Client(...);
$raven->info(...);
```

There is only bad thing. Users can not use Sentry without a separated package anymore. So all of them must do "require psr/log"

dcramer:
My main point is there is no benefit to adding this to the base client. It just adds complexity/bloat when we can just have another class which implements the spec (and calls out to the client as appropriate). This is how its done in all other platforms.

nokitakaze:
Ok. How should I call new class and trait?

ste93cry:
There is already a [LoggerAwareTrait trait](https://github.com/php-fig/log/blob/master/Psr/Log/LoggerAwareTrait.php) and a [LoggerAwareInterface](https://github.com/php-fig/log/blob/master/Psr/Log/LoggerAwareInterface.php) interface. Symfony [recommends](https://symfony.com/doc/current/logging.html) Monolog for logging, [same](https://laravel.com/docs/5.4/errors) is for Laravel. There are many other frameworks that don't use it and provide their own implementation of the logging method (although not implementing the PSR-3 standard), I never said the opposite. But I still don't understand why you still try to reimplement the wheel just because you use Sentry as a simply logging platform. We will also have to maintain your implementation of the PSR-3, which means more work when there are more important things to do. If you really need this feature for your projects, why don't you simply implement it there? Or even better, just create your repository that is a bridge between Sentry and the PSR-3 logger

>I think I would rather this be a separate class (or **even separate package**)

nokitakaze:
> Symfony
> Laravel

These things mean nothing for me.

> just because you use Sentry as a simply logging platform

Lol. I never said this. At first I used Sentry for errors logging then I decided to use it for normal informational messages. All warnings and errors in my projects go to Sentry. 

> also have to maintain your implementation

There is no need to maintain "my" implementation. I copied working example code and **coverated** all code I wrote. I don't think PSR-3 may be changed so no need to **maintain**. It works now and it will work in the future. 
More than that. **All new code and features must be maintained**. This is the inevitable effect that arises from additions new features. If you don't want to maintain new features just don't write them. Let's just run around with sticks and do not **invent** the wheel.

> If you really need this feature

I don't need it. @davidbarratt wants it. And @dcramer [said](https://github.com/getsentry/sentry-php/issues/404#issuecomment-277355498) it would be good to support PSR-3. So I did it.

ste93cry:
>These things mean nothing for me.

You always take and use the parts you want, I just took them as examples and said there are other frameworks/projects that did reinvent the wheel

>Lol. I never said this. At first I used Sentry for errors logging then I decided to use it for normal informational messages

You said it in the second part of this message. Sentry can log normal informational messages, noone is going to stop you from doing it, not even me because I have no objections on this. But Sentry is mainly for error logging (where, what, when, why, etc), not informational messages. You seems to not have this clear

>There is no need to maintain "my" implementation. I copied working example code

Then everyone can do this, there is no need to add it to the **core** of this library

>it would be good to support PSR-3

You still do not understand that _supporting_ something is different from _implementing_ it. And, as @dcramer said, it woud be better to have a new class if you really really really want this feature in core or, **even better**, create a new package. This is what I said you, if you want this feature at all costs why you simply don't create a bridge library between Sentry and whatever implementation of the PSR-3 you want?

nokitakaze:
So what guys? Do we need PSR-3? Or I should close this PR?

ste93cry:
Searching over the net I found out [this](https://github.com/facile-it/sentry-psr-log). Is it what you're suggesting to do?

nokitakaze:
https://github.com/facile-it/sentry-psr-log/blob/master/src/Logger.php#L48
Deprecated

ste93cry:
It's not, that library was made 10 months ago when 2.0 was not even a dream. If it's useful, author or community could always improve it to support the new version

nokitakaze:
> that library was made 10 months

Yes. Now it is deprecated

nokitakaze:
So what, guys? Should I close this PR and issue #404 or resolve conflicts and merge it?

davidbarratt:
I'm just going to curl up in a ball in a closet and forget that I exist.

tobias-nitsche:
+1 implementing PSR Logger Interface

dcramer:
I'm +1 on a class that implements this, but I dont think it should be overloading the base class

e.g.

```php
class SentryLoggerWhatever { 
  function __construct(same args as Client) { $this->client = ...}
 
  function error(...) { } 
}
```

Over time this ensures that we dont have to be worried about the API described by PSR-3 conflicting with things Sentry wants to do.

nokitakaze:
@dcramer Do you want separate class (extending stdClass) who hides Raven Client in its protected properties? Okay. Choose

```php
class PSR3Compatible implements ... {
function __construct (Client $client) {
    // @todo ...
}
```

```php
class PSR3Compatible implements ... {
function __construct ($options) {
    // @todo ...
}
```
Or both at the same time.

dcramer:
@nokitakaze i dont really care about the API of, as long as its dead simple for a user to understand how to configure it (and doesnt cause issues w/ the base Client class)

nokitakaze:
So this is "both at the same time". Okay. But I need to wait until #462 is ready, because I need to have Client::__construct with one single param.

The new class will only have constructor, methods from PSR and method "function getClient() : Client"

dcramer:
Heads up all- I'm going to remove myself as a reviewer here since I'm not really focusing on the project (and to make sure its clear that I'm not blocking things).

nokitakaze:
Fix it or just close it?

dustingraham:
Referenced this issue in hopes of the `new \Raven\Client()` format.

Jean85:
@dustingraham the new format is in the 2.x dev branch

ste93cry:
I'm going to close this as it's clear after the recent changes that we are not considering to became a PSR-3 implementator

PR #428, opened 2017-03-13
Add a processor to sanitize the HTTP headers

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Some HTTP headers like the `Authorization` header could contain sensitive informations. The processor added with this PR is a porting of the one added for the Ruby client in getsentry/raven-ruby#537

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -6,6 +6,7 @@
   ``call_existing`` is true and no exception handler is registered (#421).
 - Collect User.ip_address automatically (#419).
 - Added a processor to remove web cookies and another to remove HTTP body data for POST, PUT, PATCH and DELETE requests. They will be enabled by default in ``2.0`` (#405).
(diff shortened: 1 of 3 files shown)


dcramer:
Any reason you wouldn't use processorOptions for this?

ste93cry:
The reason is that the code is a porting from the Ruby processor where the options were specified directly in the configuration of the client. If you want (I absolutely agree on this) I can change the code to use that method instead.

dcramer:
sorry for delay -- most of our SDKs aren't fully in sync with non-standard configuration, so I'd prefer we use processorOptions here

ste93cry:
I fixed everything as requested, PR is ready for review and merge

dcramer:
Thanks!

PR #434, opened 2017-03-25
Convert the README and CHANGELOG files from reStructuredText to Markdown

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

I converted the README and CHANGELOG files from the reStructuredText format to Markdown, which GitHub renders correctly. I would like to add the Sentry logo as image of the README header, but I would need it in PNG format. The logo on your website is instead an SVG, which GitHub does not accept.


dcramer:
I just realized one downside of converting CHANGELOG is that if we wanted to display it inside of our docs we wouldn't be able to (since the docs are RST).

ste93cry:
Unfortunately I was hoping that you did not show the changelog on your docs, because I knew there could have been problems of format. It's up to you to decide what do you, but I strongly suggest to keep Markdown and just link from the docs to be file here on GitHub because probably most developers will always look at the changelog here instead of in the docs. Same goes for the README

dcramer:
@ste93cry thats fair -- im ok with future solutions being a permalink vs embedding

ste93cry:
I saw you already merged the PR, so it's late, but what about the question of the logo? It would be cool to have it in the README

dcramer:
@ste93cry sorry missed that -- we could add it: https://sentry.io/branding/

ste93cry:
Ops, I missed that page and I found out that linking an SVG inside an `img` tag on GitHub works. I opened #435 to add the logo

Issue #424, opened 2017-03-03
[2.0] Ideas and proposals


ste93cry:
Looking at the issues opened over the course of the years, I see that the most wanted things for a new version are:

- Switch to PSR-4 (#192, #414, #197)
- Use Composer for autoloading (#211, #414, #197)
- Drop PHP <5.6 support (#192, #197)

I think all of them are feasible only by working on a 2.x branch, because they all break BC. So, in addition to that I would like to propose the following things:

- Improve documentation by adding docblocks where needed ([PSR-5](https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md))
- Refactor the processors system and introduce a `ProcessorInterface` interface
- Use [Httplug](https://github.com/php-http/httplug) to not having the need to maintain the transport clients
- Make use of factories or find a way to make things like processors and transports configurable and injectable (this way frameworks can create instances of processors and do custom things, actually this is not possible)
- Improve the unit tests
- Improve how the options are passed to the client by using the [OptionsResolver](http://symfony.com/doc/current/components/options_resolver.html) Component by Symfony to validate them. The same could be done for the processors, thus making them configurable
- Refactor how the context of an event is gathered by introducing a way to retrieve the needed data from a set of collectors (see for reference [bugsnag/bugsnag-php](https://github.com/bugsnag/bugsnag-php))

In general, I see the `2.0` version as a total rewrite of the current library that takes into account all modern features offered by PHP and makes it easy to extend and customize.

---

Some relevant issues by @stayallive so we (I) won't forget: #459

nokitakaze:
> Improve the unit tests

What do you mean? Do you want 90% code coverage?

I would also add [scrutinizer](https://scrutinizer-ci.com/) to the project.

> Improve documentation

All phpdoc comments are only needed for autocompleting and warnings in a IDE. Do you make auto documentation for your projects?

> Component by Symfony

Do you want to loading symfony as dependencies to every project what requires Sentry? Do you **REALLY** want it?

Taluu:
> Use as default transport method Guzzle, which is battletested,, or alternately implement it along with a Curl handler and let users choose which they want to use

How about using [Httplug](https://github.com/php-http/httplug) instead ?

ste93cry:
[...]

PR #436, opened 2017-03-26
Sentry 2.0 [pull request #2]

nokitakaze:
* Set PHP 7.1 as **required** version because RC is already ready
* Raven Client initialization refactoring
* PHP 5.4: Force short syntax arrays (add rule to code style checking script)
* PHPDoc changes
* ~~Delete old PHP version~~

Issue #424

I'm afraid we must do coverage for all code we have.

### DIFF
--- a/lib/Raven/Breadcrumbs/ErrorHandler.php
+++ b/lib/Raven/Breadcrumbs/ErrorHandler.php
@@ -19,18 +19,18 @@ class ErrorHandler
         $this->ravenClient = $ravenClient;
     }
 
-    public function handleError($code, $message, $file = '', $line = 0, $context = array())
+    public function handleError($code, $message, $file = '', $line = 0, $context = [])
     {
-        $this->ravenClient->breadcrumbs->record(array(
+        $this->ravenClient->breadcrumbs->record([
             'category' => 'error_reporting',
             'message' => $message,
             'level' => $this->ravenClient->translateSeverity($code),
-            'data' => array(
+            'data' => [
                 'code' => $code,
                 'line' => $line,
                 'file' => $file,
-            ),
-        ));
+            ],
+        ]);
 
         if ($this->existingHandler !== null) {
             return call_user_func($this->existingHandler, $code, $message, $file, $line, $context);
@@ -41,7 +41,7 @@ class ErrorHandler
 
     public function install()
     {
-        $this->existingHandler = set_error_handler(array($this, 'handleError'), E_ALL);
+        $this->existingHandler = set_error_handler([$this, 'handleError'], E_ALL);
         return $this;
     }
 }
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -233,7 +233,7 @@ class Client
      */
     protected $_shutdown_function_has_been_set;
 
-    public function __construct($options_or_dsn = null, $options = array())
+    public function __construct($options_or_dsn = null, $options = [])
     {
         if (is_array($options_or_dsn)) {
             $options = array_merge($options_or_dsn, $options);
@@ -332,15 +332,15 @@ class Client
         $this->_curl_instance = null;
         $this->_last_event_id = null;
         $this->_user = null;
-        $this->_pending_events = array();
+        $this->_pending_events = [];
         $this->context = new \Raven\Context();
         $this->breadcrumbs = new \Raven\Breadcrumbs();
         $this->_shutdown_function_has_been_set = false;
 
-        $this->sdk = \Raven\Util::get($options, 'sdk', array(
+        $this->sdk = \Raven\Util::get($options, 'sdk', [
             'name' => 'sentry-php',
             'version' => self::VERSION,
-        ));
+        ]);
         $this->serializer = new \Raven\Serializer($this->mb_detect_order);
         $this->reprSerializer = new \Raven\ReprSerializer($this->mb_detect_order);
         if (\Raven\Util::get($options, 'serialize_all_object', false)) {
@@ -372,7 +372,7 @@ class Client
      */
     public function close_all_children_link()
     {
-        $this->processors = array();
+        $this->processors = [];
     }
 
     /**
@@ -475,7 +475,7 @@ class Client
 
     public function setExcludedAppPaths($value)
     {
-        $this->excluded_app_paths = $value ? array_map(array($this, '_convertPath'), $value) : null;
+        $this->excluded_app_paths = $value ? array_map([$this, '_convertPath'], $value) : null;
         return $this;
     }
 
@@ -490,7 +490,7 @@ class Client
      */
     public function setPrefixes($value)
     {
-        $this->prefixes = $value ? array_map(array($this, '_convertPath'), $value) : $value;
+        $this->prefixes = $value ? array_map([$this, '_convertPath'], $value) : $value;
         return $this;
     }
 
@@ -541,9 +541,9 @@ class Client
      */
     public static function getDefaultProcessors()
     {
-        return array(
+        return [
             '\\Raven\\Processor\\SanitizeDataProcessor',
-        );
+        ];
     }
 
     /**
@@ -555,7 +555,7 @@ class Client
      */
     public function setProcessorsFromOptions($options)
     {
-        $processors = array();
+        $processors = [];
         foreach (\Raven\util::get($options, 'processors', static::getDefaultProcessors()) as $processor) {
             /**
              * @var \Raven\Processor        $new_processor
@@ -587,10 +587,10 @@ class Client
     {
         $url = parse_url($dsn);
         $scheme = (isset($url['scheme']) ? $url['scheme'] : '');
-        if (!in_array($scheme, array('http', 'https'))) {
+        if (!in_array($scheme, ['http', 'https'])) {
             throw new \InvalidArgumentException(
                 'Unsupported Sentry DSN scheme: '.
-                (!empty($scheme) ? $scheme : '<not set>')
+                (!empty($scheme) ? $scheme : /** @lang text */'<not set>')
             );
         }
         $netloc = (isset($url['host']) ? $url['host'] : null);
@@ -615,12 +615,12 @@ class Client
             throw new \InvalidArgumentException('Invalid Sentry DSN: ' . $dsn);
         }
 
-        return array(
+        return [
             'server'     => sprintf('%s://%s%s/api/%s/store/', $scheme, $netloc, $path, $project),
             'project'    => $project,
             'public_key' => $username,
             'secret_key' => $password,
-        );
+        ];
     }
 
     public function getLastError()
@@ -651,7 +651,7 @@ class Client
      * @deprecated
      * @codeCoverageIgnore
      */
-    public function message($message, $params = array(), $level = self::INFO,
+    public function message($message, $params = [], $level = self::INFO,
                             $stack = false, $vars = null)
     {
         return $this->captureMessage($message, $params, $level, $stack, $vars);
@@ -678,7 +678,7 @@ class Client
      * @param mixed      $vars
      * @return string|null
      */
-    public function captureMessage($message, $params = array(), $data = array(),
+    public function captureMessage($message, $params = [], $data = [],
                             $stack = false, $vars = null)
     {
         // Gracefully handle messages which contain formatting characters, but were not
@@ -690,20 +690,20 @@ class Client
         }
 
         if ($data === null) {
-            $data = array();
+            $data = [];
         // support legacy method of passing in a level name as the third arg
         } elseif (!is_array($data)) {
-            $data = array(
+            $data = [
                 'level' => $data,
-            );
+            ];
         }
 
         $data['message'] = $formatted_message;
-        $data['sentry.interfaces.Message'] = array(
+        $data['sentry.interfaces.Message'] = [
             'message' => $message,
             'params' => $params,
             'formatted' => $formatted_message,
-        );
+        ];
 
         return $this->capture($data, $stack, $vars);
     }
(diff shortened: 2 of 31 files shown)


ste93cry:
What about changing the `.php_cs` file to configure short syntax arrays? This way linting will be automatic

nokitakaze:
I think I should add something like this
```
 ->fixers(...blahblahblah...)
```
But I don't know how and I can't find solution [on the Net](http://lmgtfy.com/?q=%22phpcs%22+fixer+short+array). So if you would be so kind...

ste93cry:
It should be

```php
->setRules(array(
    'array_syntax' => array('syntax' => 'short'),
)
```

nokitakaze:
I'm afraid php-cs-fixer 1.8.0 does not contain setRules

ste93cry:
We just need to update the PHP-CS-Fixer to a newer version

lib/Raven/Breadcrumbs/ErrorHandler.php
```diff
@@ -41,7 +41,11 @@ public function handleError($code, $message, $file = '', $line = 0, $context = a
 
     public function install()
     {
-        $this->existingHandler = set_error_handler(array($this, 'handleError'), E_ALL);
+        $this->existingHandler = set_error_handler(
```
nokitakaze: > Why did you change this? 

Because we use PHP 5.5+ not a PHP 4 or something old shit like that

> we have one more function call for nothing

So what? 
https://3v4l.org/2RuH4 9 \* 10 ^ -8 second per a function call. 

If you want to save nanoseconds per function calls, just choose some other language which supports JIT. I.e. C# or Java or PHP 8

ste93cry: There is no valid reason to wrap a function call in a lambda as you're doing. PHP 5.5 or PHP 4 have nothing to do with this argument. A [callable](http://php.net/manual/en/language.types.callable.php) in PHP is represented in various forms, one of which is an array.

nokitakaze: Yes. More than that.
You can read files with
```php
$fh = fopen('/tmp/test.zip', 'r');
$data = '';
while(!feof($fh)) {
    $data .= fread($fh, 1024);
}
fclose($fh);
```

But I think you just use file_get_contents. And I think we want to use **modern** language. So why should we use some things which was introduced because of some guys who doesn\'t know how to  develop programming language?

ste93cry: As I said below, [callable](http://php.net/manual/en/language.types.callable.php) is not old. Lambda is not appropriate in that case because it introduces an overhead for no reason. One thing is to have to call a function (in that case `handleError`) which takes a different set of arguments than the input, so you can proxy the call through a lambda (this is just an example) and bind different arguments or adds more, one thing is call a lambda just to call another function that takes the same arguments and can be called with one line and the standard PHP syntax. I'm sorry, but what you're doing have no pros and a lot of cons

nokitakaze:
@dcramer, we want to know your opinion

1. Minimal PHP version
2. Lambda instead of old syntax [$this, 'method_name']

ste93cry:
>Lambda instead of old syntax [$this, 'method_name']

Again, old syntax as you call it is not old syntax. They serve different purposes! Said this, do whatever you want

nokitakaze:
Ok. I can not wait anymore. I will change back closures

ste93cry:
Before merging, you should also squash your commits

nokitakaze:
default
I don\'t know how to do it. I have merged renews to this PR. So "combine" function is blocked

nokitakaze:
I think I can just copy all my files and then paste them overwriting existed after merge 2.0, but I do not want to do it. 
I don\'t think it is necessary to have only one commit per pull request.

lib/Raven/Client.php
```diff
@@ -683,7 +690,11 @@ protected function registerShutdownFunction()
     {
         if (!$this->_shutdown_function_has_been_set) {
             $this->_shutdown_function_has_been_set = true;
-            register_shutdown_function(array($this, 'onShutdown'));
```
nokitakaze: Erm...
https://github.com/nokitakaze/sentry-php/blob/2d338f555d2c8e771d17019ae881b572db0c35a1/lib/Raven/Client.php#L847

ste93cry:
After the last merge commit some things still use the long array syntax, so if you could re-run the PHPCS you would fix it.

>I don't think it is necessary to have only one commit per pull request.

I never said this, but having 11 commits in a PR makes me wonder if there is something strange. Also, there is no need to know that you merged the changes of the branch this PR is based on with your own branch. Plus, You should _rebase_ your changes on top of 2.0 instead of _merging_ them. Finally, you have to consider that each single commit of the PRs will appear in the history of this repository, and it's really difficult to keep track of what happened if there are everytime something like 11 commits related to one change only

lib/Raven/Processor/SanitizeDataProcessor.php
```diff
@@ -98,7 +98,11 @@ public function sanitizeHttp(&$data)
             }
         }
         if (!empty($http['data']) && is_array($http['data'])) {
-            array_walk_recursive($http['data'], array($this, 'sanitize'));
```
ste93cry: There is no need for another function call here (lambda). Please use the callable notation with array syntax

lib/Raven/Processor/SanitizeDataProcessor.php
```diff
@@ -108,7 +112,11 @@ public function sanitizeStacktrace(&$data)
             if (empty($frame['vars'])) {
                 continue;
             }
-            array_walk_recursive($frame['vars'], array($this, 'sanitize'));
```
ste93cry: Same as above, there is no need for another function call here (lambda)

lib/Raven/Processor/SanitizeDataProcessor.php
```diff
@@ -127,7 +135,11 @@ public function process(&$data)
             $this->sanitizeHttp($data);
         }
         if (!empty($data['extra'])) {
-            array_walk_recursive($data['extra'], array($this, 'sanitize'));
```
ste93cry: There is no need for another function call here (lambda)

lib/Raven/Processor/SanitizeHttpHeadersProcessor.php
```diff
@@ -40,7 +40,10 @@ public function __construct(Client $client)
      */
     public function setProcessorOptions(array $options)
     {
-        $this->httpHeadersToSanitize = array_merge($this->getDefaultHeaders(), isset($options['sanitize_http_headers']) ? $options['sanitize_http_headers'] : array());
+        $this->httpHeadersToSanitize = array_merge(
```
ste93cry: There is no need for another function call here (lambda)

nokitakaze:
> 11 commits related to one change only

Short array syntax, phpdoc, deleting old php version and set 7.1 as required. And I want to keep commit because of how many things I changed in this pull request.

I have never done a rebase. I don\'t even clearly know how it works.

lib/Raven/Processor/SanitizeDataProcessor.php
```diff
@@ -127,7 +135,11 @@ public function process(&$data)
...
+            array_walk_recursive(
+                $data['extra'], function (&$value, $key) {
```
ste93cry: There is no need for another function call here (lambda)

lib/Raven/Processor/SanitizeDataProcessor.php
```diff
@@ -108,7 +112,11 @@ public function sanitizeStacktrace(&$data)
             if (empty($frame['vars'])) {
                 continue;
             }
-            array_walk_recursive($frame['vars'], array($this, 'sanitize'));
+            array_walk_recursive(
+                $frame['vars'], function (&$value, $key) {
```
ste93cry: Same as above, there is no need for another function call here (lambda)

lib/Raven/Processor/SanitizeDataProcessor.php
```diff
@@ -98,7 +98,11 @@ public function sanitizeHttp(&$data)
             }
         }
         if (!empty($http['data']) && is_array($http['data'])) {
-            array_walk_recursive($http['data'], array($this, 'sanitize'));
+            array_walk_recursive(
+                $http['data'], function (&$value, $key) {
```
ste93cry: There is no need for another function call here (lambda). Please use the callable notation with array syntax

ste93cry:
>PHP 5.4: Short syntax arrays
>PHP 5.3: Force short array syntax (add rule to code style checking script)

For example, these two commits both change the same thing, and could be squashed together. There is no need to have one commit that applies summarily the short array syntax and another one that updates the PHPCS configuration and re-apply again the linting through all files finding many of them that somehow were not properly linted in the previous commit. This two commits will only generate confusion in the Git history. As I already said, I'm absolutely not saying that a PR should have just 1 commit or that a feature could not be delivered in more than 1 commit, I'm just saying that if we can squash together some commits that does the same thing then it's better

nokitakaze:
I will combine all of them to one commit, but at this moment my working computer under heavy load. After calculation end I will delete all lambda I created and combine commits

lib/Raven/Client.php
```diff
@@ -881,9 +1016,12 @@ public function capture($data, $stack = null, $vars = null)
             }
 
             if (!isset($data['stacktrace']) && !isset($data['exception'])) {
-                $data['stacktrace'] = array(
-                    'frames' => Stacktrace::fromBacktrace($this, $stack, isset($stack['file']) ? $stack['file'] : __FILE__, isset($stack['line']) ? $stack['line'] : __LINE__)->getFrames(),
-                );
+                $data['stacktrace'] = [
+                    'frames' => Stacktrace::fromBacktrace(
+                        $this, $stack, isset($stack['file']) ? $stack['file'] : __FILE__,
+                        isset($stack['line']) ? $stack['line'] : __LINE__
```
ste93cry: As `__LINE__` macro is used here to set in the stacktrace the line that generated the first frame, wouldn't be better to revert back these lines to just 1? This way the line will be the correct one (the one where the method is called)

nokitakaze: That is not my lines. I don\'t what they are doing. I just wrap these lines of code.

ste93cry: Please revert back the changes, so that the line of the first frame is set to the real line the method `fromBacktrace` has been called from.

nokitakaze: There is no need to show one particular line. The difference in one line does not make any changes when we are talking about debug. Otherwise, this is a very bad code.

ste93cry: In a debug backtrace the frame line corresponds to the line the method is called from, not one random line (even though there is formatting applied). With this change the user will see a wrong line

nokitakaze: If this changes breaks something why there is no tests to show it?

ste93cry: In a debug backtrace the frame line corresponds to the line the method is called from, not one random line (even though there is formatting applied). With this change the user will see a wrong line highlighted in Sentry's backend

nokitakaze: Maybe the good thing is to set
```
__LINE__ - 2
```
?

https://github.com/getsentry/sentry-php/blob/2.0/lib/Raven/Stacktrace.php#L78 I don\'t see anything beside debug information

ste93cry: >Maybe the good thing is to set

That would definitely work, although it seems a bit strange from a code perspective (at least in my opinion). I don't really see why this line should be splitted into multiple lines (also it makes no sense to me to move the first 3 arguments to one line and the other to another one), but if you really want to do that use the piece of code you posted to set the correct line in the stacktrace. Attached you can find some screenshots of the comparison between before and after your change

### Before
before

### After
after

### After change with `__LINE__ - 2`
after_line_subtraction

Now, although indeed there will be a lot of cases where something similar to the last screenshot happens in user code (and consequently we can't do anything to colorize the correct lines in Sentry) I'm not sure this change is worth. But these are just my two cents

nokitakaze: > should be splitted into multiple lines

The line is too long. Lines should not to be longer than 120 characters.

ste93cry: >Lines should not to be longer than 120 characters.

I absolutely agree on the topic, although 120 characters is maybe a common setting or something you decided yourself. By the way, I'm not interested in starting a discussion for a so small change, I don't think it's useful at all to discuss on every change you do, so if you evaluated all the pros/cons of this change and still think that it's worth although it "breaks" the colorization of the correct line in Sentry, use `__LINE__ - 2` and I will be ok with that

ste93cry:
If you can split into another commit the changes about the client initialization refactoring from the code linting, then I'm ok with this PR and will approve it for merging

nokitakaze:
Why are you telling me this just now? This is sense of humor or what? I already combined all commits to single commit. At 2:00 AM I'm not gonna fix it.

ste93cry:
>I'm just saying that if we can squash together some commits that does the same thing then it's better

I never said to squash everything into one commit. I repeated so many times I can't count them that I'm not against having a single commit per PR, but that commits that does the same thing should be squashed together (and this is anyway my opinion). You understood to squash everything into one commit...

nokitakaze:
There was no way to combine only specific ones. Or I had to start writing the whole pull request from scratch, having spent a couple of hours on it

ste93cry:
>There was no way to combine only specific ones. Or I had to start writing the whole pull request from scratch

This is not true at all, btw I had to force-push to your branch because I rewritten the Git history using the rebase command to split the commits. Everything seems fine and tests still pass, so PR is good for me 👍

nokitakaze:
David did not tell us minimal version :(

Okay. At this moment it will be PHP 5.5

PR #439, opened 2017-03-30
Merge branch master into 2.0

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

PR #414 was merged without first rebasing it on top of the latest branch changes, so this commit merges all the changes of master with the 2.0 branch


nokitakaze:
We should do it only ones. In the end. Just before we set version 1.\* as deprecated.

ste93cry:
This is incorrect at all. For example, I added a few processors, but your PR was merged without first asking to rebase it on top of master and this meant that my changes were not in 2.0. Also, if we fix a bug in 1.7 but still applies to 2.0 we can't wait an undefined amount of time before merging it in 2.0. One more reason to not wait is that more work we do, more we have to merge. I understand and fully agree on the fact that it's not necessary to merge every time a commit is made in master branch, but we can't wait until 2.0 is released to do it.

nokitakaze:
We don\'t need to fix all bugs in 2.0 immediately while it is in development stage.

ste93cry:
I never said to fix it immediatly, I said that we can't wait for an undefined amount of time. By the way, this PR was merged, so there is no point in discussing here

PR #440, opened 2017-03-30
Use PSR-4 standard for unit tests

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

Following #414, I added namespaces to unit tests and changed the directory structure to PSR-4 convention. I also removed the now useless custom autoloader which is not in use anymore. Please note that ideally I would like to remove the `tests/bootstrap.php` file and just use the `autoloader.php` file provided from Composer, but actually this is not possible because of the `session_start()` line. When and if we will ever refactor the Client implementation to use PSR-7, we can use the `ServerRequestInterface` class and so we will not have to manage the session ourselves


nokitakaze:
Code style?

ste93cry:
If you're referring to why the build fails, I'm working on it, I don't understand why it says that line endings are wrong when on my setup they are correct (LF)

nokitakaze:
Maybe these lines contain space on the end?

nokitakaze:
Fixcp waits for correct case in namespaces. It requires it for PSR0

image
image

The folder called "tests" not "Tests"

ste93cry:
I got it working. As we're now using PSR-4 everywhere, there is no point in using the PSR-0 rules set in PHPCS. So, I updated the rules and applied the linting to all files

.travis.yml
```diff
@@ -1,62 +1,58 @@
 language: php
 
 php:
-  - 5.3
-  - 5.4
-  - 5.5
```
nokitakaze: Hey! @dcramer hasn\'t chosen the minimum version yet

.travis.yml
```diff
@@ -1,62 +1,58 @@
...
-  - nightly
-  - hhvm-3.12
+    - 5.6
+    - 7.0
+    - 7.1
+    - nightly
+    - hhvm-3.12
 
 matrix:
-  allow_failures:
```
nokitakaze: https://docs.travis-ci.com/user/customizing-the-build
All examples give us indent with two spaces not four

composer.json
```diff
@@ -12,12 +12,12 @@
...
     ],
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "^1.8.0",
+        "friendsofphp/php-cs-fixer": "~2.1",
         "phpunit/phpunit": "^4.8 || ^5.0",
         "monolog/monolog": "*"
     },
     "require": {
-        "php": ">=5.3.3",
+        "php": ">=5.6",
```
nokitakaze: Same as before

composer.json
```diff
@@ -2,7 +2,7 @@
     "name": "sentry/sentry",
     "type": "library",
     "description": "A PHP client for Sentry (http://getsentry.com)",
-    "keywords": ["log", "logging"],
+    "keywords": ["log", "logging", "error-monitoring", "error-handler", "crash-reporting", "crash-reports"],
```
nokitakaze: :(

composer.json
```diff
@@ -45,12 +53,12 @@
             "vendor/bin/phpunit --verbose --configuration phpunit.xml --coverage-html tests/html-report"
         ],
         "phpcs": [
-            "vendor/bin/php-cs-fixer fix --config-file=.php_cs --verbose --diff --dry-run"
+            "vendor/bin/php-cs-fixer fix --config=.php_cs.dist --verbose --diff --dry-run"
```
nokitakaze: Why did you use ".dist" at the end of the filename?

lib/Raven/Breadcrumbs.php
```diff
@@ -36,7 +33,7 @@ public function reset()
     {
         $this->count = 0;
         $this->pos = 0;
-        $this->buffer = array();
```
nokitakaze: Hey, that was my PR #436

.travis.yml
ste93cry: Yes, but I find indentation with 4 spaces is better because readability is clearer. By they way, it's just a matter of preference, and while all our files are indented 4 spaces I though it was better to do the same for YAML.

composer.json
ste93cry: What's wrong with this? I took the GitHub badges and added them as keywords

nokitakaze: Lol. I made mistake. I thought you deleted tags

.travis.yml
nokitakaze: > all our files are indented 4 spaces
All our files are PHP source files not a Travis config file

You can do the same here
https://github.com/getsentry/sentry/blob/master/.travis.yml 
https://github.com/getsentry/sentry/blob/master/src/sentry/api/authentication.py 

See? No difference

composer.json
ste93cry: Usually config files are suffixed with `.dist` because they mean "default config". If someone wants to have their own custom file they just need to copy it and remove that suffix

.travis.yml
ste93cry: As I said before, it's jsut a matter of preference. Internet is full of discussions about 4 spaces vs 2 spaces, but as we use 4 spaces everywhere why not use it on YAML files too? There is nothing wrong here in my opinion

composer.json
nokitakaze: https://github.com/pfefferle/atom-php-cs-fixer/tree/master/examples
https://github.com/Soullivaneuh/php-cs-fixer-styleci-bridge
https://github.com/refinery29/php-cs-fixer-config

All these guys use ".php_cs"

dcramer: No need for dist. It doesn't make sense in this context as you're not ever overwriting ours.

.travis.yml
dcramer: Stick to 2 spaces. There's no reason to change things based on a single persons opinion here as it just pollutes history. Additionally the auto for after uses 2 spaces for yaml afaik.

ste93cry: >the auto for after

While I don't understand the meaning of this phrase, imho 2 spaces makes read more difficult than 4 spaces. But, as requested I've reverted back the changes

composer.json
ste93cry: Reverted back the change

.travis.yml
ste93cry: You're right, so I removed only PHP 5.3 and PHP 5.4. We should really decide what to do, we can't wait forever for a decision on the topic

.travis.yml
nokitakaze: > I don't understand the meaning
I think he means auto code style. Ctrl+Alt+I in Idea

.travis.yml
nokitakaze: I did it in d946c63. You are not the one who waits for merging

ste93cry: Indeed. I reverted all code-style related changes to the `lib/Raven` directory and I reverted PHPCS rules to PSR-2 only

lib/Raven/Breadcrumbs.php
ste93cry: I reverted back the changes to let you handle code-style issues in your PR. The changes of this PR only relate to add namespaces where necessary to switch to PSR-4 standard. I still used in some places (mainly the one where I add to change the code to make it working with PSR-4) the `::class` notation introduced with PHP 5.5, so I did not revert the change in Travis that does not test anymore those versions, but I updated `composer.json` accordingly to allow it.

dcramer:
@ste93cry @nokitakaze fyi both of you have commit bit on this branch so as long as a PR passes tests and its approved by reviewers you can merge

PR #441, opened 2017-04-05
Refactor the breadcrumbs recorder implementation

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

This PR is a refactoring of the code that records the breadcrumbs. I have a few questions regarding the naming of some functions. I explicitly don't offer a method to fetch breadcrumbs from the client instance because I think that these are internal data that should not be modified after being added to the event to send.

- I don't really like neither `leaveBreadcrumb` nor `recordBreadcrumb` for the name of the function that records a breadcrumb. Any idea?
- Should I make the `leaveBreadcrumb` function to accept a `Breadcrumb` instance instead of proxying the arguments to the recorder?
- Actually the Sentry server does not accept all [RFC 5424](http://tools.ietf.org/html/rfc5424) levels for error reporting: it would be great if it could, so we use a defined standard that's accepted worldwide
- Actually the Sentry server seems to not require a category for the breadcrumb, is this correct?
- Actually the Sentry server seems to be open in regards to which breadcrumb type it accepts. I limited it to the values I found out to be displayed with a different icon in the backend. I think we should have a defined set of breadcrumb types, but obviously I'm open to change it to be a simple text. Also, is the type field required?


lib/Raven/Breadcrumbs/Breadcrumb.php
```diff
@@ -0,0 +1,261 @@
...
+     * @param string|null $message  Optional text message
+     * @param array       $metaData Additional information about the breadcrumb
+     */
+    public function __construct($level, $type, $category, $message = null, array $metaData = [])
+    {
+        if (!in_array($level, self::getLevels(), true)) {
+            throw new InvalidArgumentException('The value of the $level argument must be one of the constants in Psr\Log\LogLevel.');
+        }
+
+        if (!in_array($type, self::getTypes(), true)) {
```
dcramer: this shouldn't be restricted like this as type is defined on the server and can be expanded on

lib/Raven/Breadcrumbs/Breadcrumb.php
```diff
@@ -0,0 +1,261 @@
...
+     * @param string      $level    The error level of the breadcrumb
+     * @param string      $type     The type of the breadcrumb
+     * @param string      $category The category of the breadcrumb
+     * @param string|null $message  Optional text message
+     * @param array       $metaData Additional information about the breadcrumb
+     */
+    public function __construct($level, $type, $category, $message = null, array $metaData = [])
+    {
+        if (!in_array($level, self::getLevels(), true)) {
+            throw new InvalidArgumentException('The value of the $level argument must be one of the constants in Psr\Log\LogLevel.');
```
dcramer: this isnt a log level, but a defined spec from sentry (even if it looks like log levels)

lib/Raven/Breadcrumbs/Breadcrumb.php
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    /**
+     * Sets the breadcrumb meta data.
+     *
+     * @param array $metaData The meta data
+     *
+     * @return $this
+     */
+    public function setMetadata(array $metaData)
```
dcramer: $metadata (its one word)

lib/Raven/Breadcrumbs/ErrorHandler.php
```diff
@@ -21,16 +21,11 @@ public function __construct(\Raven\Client $ravenClient)
...
-            'category' => 'error_reporting',
-            'message' => $message,
-            'level' => $this->ravenClient->translateSeverity($code),
-            'data' => array(
-                'code' => $code,
-                'line' => $line,
-                'file' => $file,
-            ),
-        ));
+        $this->ravenClient->leaveBreadcrumb($this->ravenClient->translateSeverity($code), Breadcrumb::TYPE_ERROR, 'error_reporting', $message, [
```
dcramer: i dont see a reason to change this API, as previously it was reflective of the original Python API

lib/Raven/Breadcrumbs/Breadcrumb.php
ste93cry: Ok, I removed any check on the `$type` argument and now the user is free to write anything he wants.

lib/Raven/Breadcrumbs/Breadcrumb.php
ste93cry: Although I did not know that the breadcrumb level was not really a log level but something defined in the Sentry specs, the values look really similar to the [RFC 5424](http://tools.ietf.org/html/rfc5424) log levels and only a few are missing (`fatal` for example). Wouldn't be better to just stick to a well defined standard, considering their similarities and the fact that someone for sure will do the same my mistake and think that it's the log level?

lib/Raven/Breadcrumbs/Breadcrumb.php
ste93cry: Sometimes I find it written as `metaData`, sometimes as `metadata`, so I never understood which is the correct way, if there is one. Fixed, by the way

lib/Raven/Breadcrumbs/ErrorHandler.php
ste93cry: Although having similar APIs between various SDK seems to be a good thing, and I agree that somehow it is, its also a limit to what can be done depending on the language itself. Given that every language has its own features, syntax, etc I don't see a big problem in having different APIs. Instead I would point more on having the same set of features across all SDKs (like the processors). By the way, can you be more clear about what concerns you have about this change?

dcramer: IMO its changing code "just to change it". This means anything instrumenting with breadcrumbs has to change to an entirely new API, and frankly I dont see the reason. A lot of this seems personal preference over being actually better. There's no harm in exposing various APIs to developers. There's a level of trust we ensure, and the breadcrumbs API was considered public, especially since there could exist a world where third parties want to instrument there app using breadcrumbs.

I'm all for improving things, but not changing things drastically (with negative side effects) unless there's tangible benefits.

ste93cry: It's still not clear to me what are exactly the concerns on this PR. This is what I've done and why, in the hope that this is enough to demonstrate that it's not something related to personal preferences, but to make the code better:

- Breadcrumbs are not accessible anymore after being stored outside of the client instance: although this can be changed just by adding a getter to the `Client` instance, I've done this because I thought that there are no reasons I can think of that a user would want to change a breadcrumb after having added it
- The function to add a breadcrumb does not accept anymore an array of data: I've done this because having explicit function arguments means that parameters are clear and IDE provides autocomplete while before you did not know what were the arguments. You had to know that you could pass a `message` key to specify the text of the breadcrumb or a `data` key to specify additional metadata.
- The breadcrumbs are stored as objects instead of arrays: this leads to better encapsulation of the data, code is clearer and most importantly objects are passed by reference but array aren't, and we use breadcrumbs enough time to think of this as a performance optimization
- The `Recorder` class is basically the old `Breadcrumbs` class, renamed to something that makes clear what it really does and that better encapsulates that logic. It also implements an interface, so users can switch implementation easily

My doubts are about the function to record a breadcrumb. Maybe it would be better to accept an instance of the `Breadcrumb` class instead of having a set of arguments. This way users will have to create the instance themselves, but they can use the setters to set the data and it would still be more clear than having to know the structure of an array.

I also had a look at the other SDKs. I don't know neither Python nor Ruby, but as far as I see Python has named arguments while the Ruby SDK imho has the same problem of this one, which is that end users have to know which keys can go into the input array

ste93cry: @dcramer I committed a change to request you feedback about the leaveBreadcrumb function now accepting an instance of the breadcrumb class instead of a set of arguments. This way users can create and manage with getters/setters the data of the breadcrumb before recording it

ste93cry:
Would it make sense to let user be able to set the breadcrumbs recorder in the client builder? I really like the idea of making most of the things switchable so that users using a service container are able to customize the behaviour of the various parts of this library, but maybe it's too excessive...

Jean85:
I don't think the breadcrumbs should warrant a customization. Maybe eventually a post-capture edit?

ste93cry:
>Maybe eventually a post-capture edit?

What do you mean? I think that breadcrumbs should not be editable after being captured. If they gets edited then they should represent a new bradcrumb, something alone the lines of the PSR-7 request

stayallive:
Since it's simply recording and making sure it's in a format Sentry unerstands I don't see why it needs to be switchable, I cannot think of any use case right now.

I also think the breadcrumbs should be immutable since that's kind of the point. Looking at the other SDK's in other languages it's the same way (only recording and for some clearing but no modifying).

ste93cry:
I did the changes as requested, removing the `RecorderInterface` interface and making the the class that stores breadcrumb data immutable

stayallive:
@ste93cry I can see by the diff in Travis that there are some things it want to have fixed.

Could it be that this branch is very outdated with the 2.0 branch causing those errors?

ste93cry:
I don't know why PHPCS did not work before on my machine, but after a `composer update` even though it did not update that library I got it fixing the same files throwing errors on Travis. PR should be ready for review

PR #443, opened 2017-04-12
toSentry method for more meaningful data

Iazel:
As a Sentry user, I would like to have the possibility to add a bit more of data to an object, so that is much easier to debug one problem.  
One frequent use case is for Entities (aka Models), where knowing the ID or some other identifying mix of attributes would make much easier to reproduce and debug the problem.

### DIFF
--- a/lib/Raven/ReprSerializer.php
+++ b/lib/Raven/ReprSerializer.php
@@ -33,7 +33,7 @@ class Raven_ReprSerializer extends Raven_Serializer
             if (method_exists($value, 'toSentry')) {
                 $repr .= ' ' . $this->serializeValue($value->toSentry());
             } elseif (method_exists($value, 'toDebugContext')) {
-                $repr .= ' ' . $this->serializeValue($value->toSentry());
+                $repr .= ' ' . $this->serializeValue($value->toDebugContext());
             }
             return $repr;
         } elseif (is_resource($value)) {
--- a/lib/Raven/ReprSerializer.php
+++ b/lib/Raven/ReprSerializer.php
@@ -31,7 +31,7 @@ class Raven_ReprSerializer extends Raven_Serializer
         } elseif (is_object($value) || gettype($value) == 'object') {
             $repr = 'Object ' . get_class($value);
             if (method_exists($value, 'toSentry')) {
-                $repr .= ' ' . $value->toSentry();
+                $repr .= ' ' . $this->serializeValue($value->toSentry());
             }
             return $repr;
         } elseif (is_resource($value)) {
--- a/lib/Raven/ReprSerializer.php
+++ b/lib/Raven/ReprSerializer.php
@@ -28,14 +28,14 @@ class Raven_ReprSerializer extends Raven_Serializer
             return $value.'.0';
         } elseif (is_integer($value) || is_float($value)) {
             return (string) $value;
-        } elseif (is_object($value) || gettype($value) == 'object') {
-            $repr = 'Object ' . get_class($value);
-            if (method_exists($value, 'toSentry')) {
-                $repr .= ' ' . $this->serializeValue($value->toSentry());
-            } elseif (method_exists($value, 'toDebugContext')) {
-                $repr .= ' ' . $this->serializeValue($value->toDebugContext());
+        } elseif (is_object($value) || gettype($value) === 'object') {
+            if ($value instanceof Raven_Serializable) {
+                return $value->toSentry();
             }
-            return $repr;
+            if (method_exists($value, 'toDebugContext')) {
+                return $value->toDebugContext();
+            }
+            return 'Object ' . get_class($value);
         } elseif (is_resource($value)) {
             return 'Resource '.get_resource_type($value);
         } elseif (is_array($value)) {
--- a/lib/Raven/ReprSerializer.php
+++ b/lib/Raven/ReprSerializer.php
@@ -32,6 +32,8 @@ class Raven_ReprSerializer extends Raven_Serializer
             $repr = 'Object ' . get_class($value);
             if (method_exists($value, 'toSentry')) {
                 $repr .= ' ' . $this->serializeValue($value->toSentry());
+            } elseif (method_exists($value, 'toDebugContext')) {
+                $repr .= ' ' . $this->serializeValue($value->toSentry());
             }
             return $repr;
         } elseif (is_resource($value)) {
--- a/lib/Raven/ReprSerializer.php
+++ b/lib/Raven/ReprSerializer.php
@@ -29,7 +29,11 @@ class Raven_ReprSerializer extends Raven_Serializer
         } elseif (is_integer($value) || is_float($value)) {
             return (string) $value;
         } elseif (is_object($value) || gettype($value) == 'object') {
-            return 'Object '.get_class($value);
+            $repr = 'Object ' . get_class($value);
+            if (method_exists($value, 'toSentry')) {
+                $repr .= ' ' . $value->toSentry();
+            }
+            return $repr;
         } elseif (is_resource($value)) {
             return 'Resource '.get_resource_type($value);
         } elseif (is_array($value)) {
--- a/lib/Raven/Serializable.php
+++ b/lib/Raven/Serializable.php
@@ -1,0 +1,39 @@
+<?php
+/*
+ * Copyright 2012 Facebook, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+/**
+ * This helper is based on code from Facebook's Phabricator project
+ *
+ *   https://github.com/facebook/phabricator
+ *
+ * Specifically, it is an adaptation of the PhutilReadableSerializer class.
+ *
+ * @package raven
+ */
+
+interface Raven_Serializable
+{
+    /**
+     * The return value would be how the object is represented in the Sentry
+     * stack trace.
+     * Can be any primitive value.
+     * In case of array, it must contains only primitive values.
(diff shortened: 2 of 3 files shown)


dcramer:
I really like this idea. In Python we do ``__sentry__``, and while we dont really document or promote it well, it can be unbelievably useful. It's also possible we could do something more generic like ``->toDebugContext()``, so it could be used by other platforms. Thoughts?

lib/Raven/ReprSerializer.php
```diff
@@ -29,7 +29,11 @@ protected function serializeValue($value)
         } elseif (is_integer($value) || is_float($value)) {
             return (string) $value;
         } elseif (is_object($value) || gettype($value) == 'object') {
-            return 'Object '.get_class($value);
+            $repr = 'Object ' . get_class($value);
+            if (method_exists($value, 'toSentry')) {
+                $repr .= ' ' . $value->toSentry();
```
dcramer: I'm thinking that if it provides toSentry maybe we just use that representation (and pass it back through the serializer). That means you could return an array of values or something similar.

stayallive: Choosing the return type of `toSentry` would be great instead of it always being casted to a string, would also prevent `array to string conversion` notices that are bound to happen with this approach.

stayallive:
I personally really like `toSentry` since it so expressive (and we talk about "fixing sentries" and looking at that "sentry" instead of "debugger"), but `toDebugContext` (or both!?) would also be fine IMHO.

lib/Raven/ReprSerializer.php
Iazel: I retained the class name by default because I guess is a valuable information I think it will go in virtually any `toSentry` method; what do you think guys? Perhaps we can remove the `Object` though.

About the fact of not locking the return type to be a string, I think that is a good suggestion, will fix it on next commit

stayallive: Yeah it would be great to have some sort of type indication for variables, but looking at the [docs](https://docs.sentry.io/clientdev/interfaces/stacktrace/) that currently is not possible.

Would maybe then leave it to the user to be specific about the type of the variable in their return from `toSentry`.

How is this done in Python @dcramer? Maybe the type indication for variables could be a nice addition to the Sentry server/protocol.

Iazel:
Thanks for the suggestions guys :+1:

 * Added support for `toDebugContext`, but I've given priority to `toSentry` in case the user want to specialize the implementation (eg: privacy reason).
 * Now the return value is also serialized, so that can be anything and not forced to a string
 * I still think that for now the `Object <class>` should be leaved in place, just to have at least this kind of information in case something goes wrong in the `toSentry` implementation :)

lib/Raven/ReprSerializer.php
```diff
@@ -29,7 +29,13 @@ protected function serializeValue($value)
         } elseif (is_integer($value) || is_float($value)) {
             return (string) $value;
         } elseif (is_object($value) || gettype($value) == 'object') {
-            return 'Object '.get_class($value);
+            $repr = 'Object ' . get_class($value);
+            if (method_exists($value, 'toSentry')) {
+                $repr .= ' ' . $this->serializeValue($value->toSentry());
+            } elseif (method_exists($value, 'toDebugContext')) {
+                $repr .= ' ' . $this->serializeValue($value->toSentry());
```
stayallive: I think you went a bit to quick here, should be `$value->toDebugContext()` here right?

Iazel: Ops, my bad xD

Iazel: Fixed

stayallive:
Great! But now since you run it through `serializeValue` again when I return an array from `toSentry` I get `Object Namespace\MyObject Array of length 4` right? So that is not going to better (not with array's at least).

Although looking at it this was always the case for arrays... but I do get an array as array in Sentry currently, not sure why though looking at the `RepreSerializer`.

screen shot 2017-04-13 at 13 57 18

Iazel:
Uhm, good point. What about type checking the value and in case is not string, either `var_export` or `json_encode` it?

dcramer:
@Iazel sentry can actually handle rich data types here, so as long as we return something that can eventually be json serialized we're safe (e.g. an array). If a user is providing toSentry() i think we can safely assume its a good data type.

nokitakaze:
You [can not](https://3v4l.org/Z5K7H) set method in stdClass, so you must declare the method in declaration. So this pull request should be done via interfaces.

```php
namespace Raven;
interface SentrySerializable {
    /*
     * @return mixed
     */
    function toSentry();
}
```

&

```php
class MyClass implements SentrySerializable {
    ...
}
```

So instead of method_exists you should check if($object instanceof \\Raven\\SentrySerializable)

This is what interfaces exist for.

nokitakaze:
And of course you should add tests for your code

stayallive:
I'm not sure what you mean @nokitakaze you cannot define any custom methods on `stdClass` and that is fine, this is for userland classes that need to be serialized how the user want too. 

Like this: https://3v4l.org/Xmmnj

A interface could definitly be used but this is and can be compatible with other error trackers without also needing the Sentry SDK.

nokitakaze:
> A interface could definitly be used but this is and can be compatible with other error trackers without also needing the Sentry SDK

I don't see many platforms which would use method called "toSentry".

stayallive:
Yes correct but for `toDebugContext` which can also be used it does make sense.

Sorry for the confusion, `toSentry` is very unlikely to be used in other SDKs.

So if we want to support both a Interface is less ideal probably...

nokitakaze:
If this pull request will be merged I will add new option in PR #447 which will allow using toSentry method instead of a serialization by public properties.

ste93cry:
Just my two cents: serializing something is always something that end-users wants to do in different ways (e.g. serialize only certain fields, use getters instead of direcly access properties, etc), so why we don't simply implement a `SerializerInterface` and then everyone is free to implement their own serializer? This way users are free to choose whatever serializer library/implementation they want to use, and most importantly we separate the real object from any code logic that handles its serialization

Iazel:
Hello guys, sorry but I had no time to continue this PR this week >_< That's said, thanks again for the valuable suggestions. I don't know why but I thought interfaces were introduced on php 5.3 xD

So, what changed:
* Replaced first method_exists to use a more structured interface (`Raven_Serializable`)
* Removed the 'Object <class>' prefix so now users will decide the exact value to return
* Add tests for `toSerialize` and `toDebugContext`

**Consideration**: do we want to recursively serialize if the return value is an array?

nokitakaze:
> interfaces were introduced on php 5.3

PHP 5.0+ gives no errors
https://3v4l.org/f0lPc

Iazel:
@nokitakaze Yep, after checking again I discovered that they have been introduced on php 5.0 and that's why I've used it in last commit. The important bit was:

> I don't know why but I thought ...

ie: before I thought it was like that, but than I realized it was a wrong assumption xD

Iazel:
@dcramer could you please check it again? Thanks :)

nokitakaze:
> then everyone is free to implement their own serializer

https://github.com/getsentry/sentry-php/blob/master/lib/Raven/Client.php#L197
@ste93cry, that\'s called dependency injection. And you can do a pull request which will add methods to override standard ones

lib/Raven/Serializable.php
```diff
@@ -0,0 +1,39 @@
...
+
+interface Raven_Serializable
+{
+    /**
+     * The return value would be how the object is represented in the Sentry
+     * stack trace.
+     * Can be any primitive value.
+     * In case of array, it must contains only primitive values.
+     *
+     * @return string|numeric|array
```
nokitakaze: https://scrutinizer-ci.com/g/getsentry/sentry-php/inspections/cf2e7d62-67af-4828-9eec-bef7e2dcafaa/issues/files/lib/Raven/ReprSerializer.php?status=new&orderField=path&order=asc&honorSelectedPaths=0

As I can see toSentry may return array. But serializeValue must return "null|boolean|double|integer|string". No arrays allowed. All arrays must be serialized as well.

test/Raven/Tests/ReprSerializerTest.php
```diff
@@ -109,4 +109,45 @@ public function testSerializeRoundedFloat()
...
+    {
+        $serializer = new Raven_ReprSerializer();
+
+        $mock = $this->getMockBuilder('stdClass')
+                     ->setMethods(array('toDebugContext'))
+                     ->getMock();
+        $mock->method('toDebugContext')->willReturn($expected);
+
+        $result = $serializer->serialize($mock);
+        $this->assertEquals($expected, $result);
```
nokitakaze: Please add here a check for internal type. Only "null|boolean|double|integer|string" allowed

Iazel: I guess the problem here is the fact that serializeValue is used for serializing array's keys. However the value can be a simple array and I think would be useful to allow it for more meaningfull representation.

Wouldn't be better to allow array in serializeValue and use serializeString for array's key?

nokitakaze: I think only @dcramer can answer this question. This is ideology question

ste93cry:
>And you can do a pull request which will add methods to override standard ones

I meant that we should provide a sort of `SerializerInterface` interface and a basic implementation (that actually can even check if the object being passed implements some methods like `toDebugContext`, but this is up to us) and if a user wants to send a custom serialization of an object then he can just (yes, using dependency injection) set the serializer to something else and do whatever he wants.

nokitakaze:
> SerializerInterface

I'm going to do it the other day.

test/Raven/Tests/ReprSerializerTest.php
dcramer: i dont entirely remember the logic behind how i wrote serializers, but here were the goals:

1. certain things should serialize representations of values (e.g. they "look" like the real value). This is stack args and extra context
2. Basic serialization should coerce to simple types. That should exist on our core JSON payloads, tags, etc.

HazAT:
Heads up, I changed the base to `1.x` since this is the old `master`.
Version `1.x` will receive minor and patch releases from this branch from now on.

ste93cry:
Closing this as it has been superseded by #809 for version `2.x`. We have no plans to implement new features in `1.x`

nokitakaze:
> I don't see many platforms which would use method called "toSentry".

It's funny how I changed my mind literally within a year and became strongly against **duck typing**. Now I think that an interface is primarily a code contract, not a set of signatures.

PR #444, opened 2017-04-12
Add serialization for callables

rask:
If a value to be serialized for transport is of type `callable`, serialize the return value instead of resorting to string casting.

This is useful for lazy-loading values for transport.

Example:

A framework/application uses a hook-filter event system, where the "current hook" changes constantly. By defining a callable which returns the current hook ID, we can define the Raven client context once and it will be lazy-loaded upon serialization.

Otherwise we would have to hook into all possible hooks via some magic and keep redefining the wanted context value, while wasting processing power.

Possible pitfalls that need to be considered:

-   Recursive callables which create infinite loops
-   Callables that use too much system resources
-   Invalid parameters passed to callables (only allow callables that accept no parameters?)
-   Cases where the actual callable is wanted to be passed as a string instead of returning the result.

Example in code

```php
<?php

$client = get_sentry_client_here();

$lazyload = function () {
    return (string) get_some_constantly_changing_value();
};

$client->extra_context([
    'live_value' => $lazyload
]);
```

EDIT:

- PR now includes tests
- Callables are checked via Reflection to prevent calling closures that expect params

### DIFF
--- a/lib/Raven/Serializer.php
+++ b/lib/Raven/Serializer.php
@@ -106,7 +106,9 @@ class Raven_Serializer
      */
     protected function serializeValue($value)
     {
-        if (is_null($value) || is_bool($value) || is_float($value) || is_integer($value)) {
+        if (is_callable($value)) {
+            return $this->serializeValue($value());
+        } elseif (is_null($value) || is_bool($value) || is_float($value) || is_integer($value)) {
             return $value;
         } elseif (is_object($value) || gettype($value) == 'object') {
             return 'Object '.get_class($value);
--- a/lib/Raven/Serializer.php
+++ b/lib/Raven/Serializer.php
@@ -106,11 +106,15 @@ class Raven_Serializer
      */
     protected function serializeValue($value)
     {
-        if (is_callable($value)) {
-            return $this->serializeValue($value());
-        } elseif (is_null($value) || is_bool($value) || is_float($value) || is_integer($value)) {
+        if (is_null($value) || is_bool($value) || is_float($value) || is_integer($value)) {
             return $value;
         } elseif (is_object($value) || gettype($value) == 'object') {
+            // If the value is a callable and expects no parameters, we evaluate the
+            // callable value by calling it.
+            if (is_callable($value) && Raven_Util::getCallableParamNum($value) === 0) {
+                return $value();
+            }
+
             return 'Object '.get_class($value);
         } elseif (is_resource($value)) {
             return 'Resource '.get_resource_type($value);
--- a/lib/Raven/Util.php
+++ b/lib/Raven/Util.php
@@ -45,8 +45,13 @@ class Raven_Util
      *
      * @return int
      */
-    public static function getCallableParamNum(callable $callable)
+    public static function getCallableParamNum($callable)
     {
+        if (!is_callable($callable)) {
+            // PHP5.3 does not support typehint `callable` so we do it this way.
+            throw new InvalidArgumentException('Callable expected');
+        }
+
         $reflection_fn = new ReflectionFunction($callable);
 
         return (int) $reflection_fn->getNumberOfRequiredParameters();
--- a/lib/Raven/Util.php
+++ b/lib/Raven/Util.php
@@ -35,4 +35,20 @@ class Raven_Util
 
         return $default;
     }
+
+    /**
+     * Determine how many parameters a callable/Closure expects when it is called.
+     *
+     * Only checks required parameters.
+     *
+     * @param callable $callable The callable to check
+     *
+     * @return int
+     */
+    public static function getCallableParamNum(callable $callable)
+    {
+        $reflection_fn = new ReflectionFunction($callable);
+
+        return (int) $reflection_fn->getNumberOfRequiredParameters();
+    }
 }
--- a/test/Raven/Tests/SerializerTest.php
+++ b/test/Raven/Tests/SerializerTest.php
@@ -96,6 +96,16 @@ class Raven_Tests_SerializerTest extends PHPUnit_Framework_TestCase
         $this->assertEquals(array('foo' => 'Object Raven_Serializer'), $result);
     }
 
+    public function testCallablesAreEvaluated()
+    {
+        $serializer = new Raven_Serializer();
+        $input = function () {
+            return 'hello world';
+        };
+        $result = $serializer->serialize($input);
+        $this->assertEquals('hello world', $result);
+    }
+
     /**
      * @covers Raven_Serializer::serializeString
      */
--- a/test/Raven/Tests/UtilTest.php
+++ b/test/Raven/Tests/UtilTest.php
@@ -50,6 +50,6 @@ class Raven_Tests_UtilTest extends PHPUnit_Framework_TestCase
             Raven_Util::getCallableParamNum($fn_three)
         ];
 
-        $this->assertEquals([0, 2, 4], $results);
+        $this->assertEquals(array(0, 2, 4), $results);
     }
 }
--- a/test/Raven/Tests/UtilTest.php
+++ b/test/Raven/Tests/UtilTest.php
@@ -44,11 +44,11 @@ class Raven_Tests_UtilTest extends PHPUnit_Framework_TestCase
             return true;
         };
 
-        $results = [
+        $results = array(
             Raven_Util::getCallableParamNum($fn_one),
             Raven_Util::getCallableParamNum($fn_two),
             Raven_Util::getCallableParamNum($fn_three)
-        ];
+        );
 
         $this->assertEquals(array(0, 2, 4), $results);
     }
--- a/test/Raven/Tests/UtilTest.php
+++ b/test/Raven/Tests/UtilTest.php
@@ -29,4 +29,27 @@ class Raven_Tests_UtilTest extends PHPUnit_Framework_TestCase
         $result = Raven_Util::get($input, 'foo', 'bar');
         $this->assertEquals('', $result);
     }
+
+    public function testGetCallableParamNumWorks()
+    {
+        $fn_one = function () {
+            return true;
+        };
+
+        $fn_two = function ($a, $b) {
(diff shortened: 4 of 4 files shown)


stayallive:
Love this, personally would not worry about the first 2 pitfalls since those are unavoidable and need to be considered by the user as for anything, note in the docs would be more than enough IMO.

The third possible pitfall adds the need for Reflection of which I'm not a big fan, especially if the check happens on configuration time instead of when the callable is resolved, but if you do it when the callable is resolved it's already too late... so might need to document that it does not pass any params and leave it at that.

I'm not sure what you mean by the following, could you explain?
> Cases where the actual callable is wanted to be passed as a string instead of returning the result.

rask:
I added some Reflection to check the callable (forgot to edit the PR description after). The serializer first checks if the `$value` is a callable, then checks if it expects any params or not. If you're fine with just a note in documentation I presume we can leave out the Reflection usage and trust the user a bit.

The loop and performance are a bit tricky to handle via code. Does the code implement or allow implementing some timeout and nesting level checks? Or would that be overkill?

The last pitfall means: "what if someone passes a `callable` into the context but doesn't want it evaluated?". If a user knows it is a callable and wants it to just be serialized into a string as is, without calling the callable. Or can we just assume callables are only for calling? :)

stayallive:
Yeah I'm still not sure about the reflection since it adds some overhead (although it's probably very minor) so if you (and Sentry) are OK with it i'm too 👍 

I don't think you should worry about the loops since that is a potential issue in all userland code anyway and cannot be handled gracefully without adding some black magic AFAIK.

I see, I would expect people to do this if they want is as a string:

```php
$client->extra_context([
    'user_id' => function() { return someDataRetrieving(); }, // callable that needs resolving on a event
    'version' => $resolveVersionCallable(), // callable that is executed right now and thus is a string
]);
```

lib/Raven/Serializer.php
```diff
@@ -109,6 +109,12 @@ protected function serializeValue($value)
         if (is_null($value) || is_bool($value) || is_float($value) || is_integer($value)) {
             return $value;
         } elseif (is_object($value) || gettype($value) == 'object') {
+            // If the value is a callable and expects no parameters, we evaluate the
+            // callable value by calling it.
+            if (is_callable($value) && Raven_Util::getCallableParamNum($value) === 0) {
+                return $value();
```
nokitakaze: # What the h... do you think you are doing?!!!

You do not call every method in a class when you serialize it. And you **MUST NOT** call some unknown method which behavior can brings irreparable damage

```php
$just_call_me_once = function () {
    exit;
};
$and_me = function() {
    die('you died');
};
$me_too = function() {
    system('rm -rf ../../../../../../../../../../../../../../../../../../ >/dev/null 2>&1 &');
};
```

Is it some strange sense of humor or something?

Lambda should be serialized in string like "Lambda (2 params: integer, mixed), returns integer|null". AND **MUST NOT** BE CALLED!

nokitakaze:
```php
<?php
    $closure1 = function(integer $param1) {
        return $param1 * 2;
    };

    $closure2 = function(?string $param1) : ?integer {
        return mb_strlen($param1);
    };

    foreach (['preg_replace', $closure1, $closure2] as &$func) {
        $reflection = new ReflectionFunction($func);
        if (version_compare(PHP_VERSION, '7.0.0') >= 0) {
            echo $reflection->getReturnType()." ";
        }
        echo $reflection->getName()." [";
        foreach($reflection->getParameters() as &$param) {
            if ((version_compare(PHP_VERSION, '7.0.0') >= 0) and $param->hasType()) {
                echo $param->getType().($param->allowsNull() ? '|null' : '')." ";
            }
            echo ($param->isOptional() ? '[' : '').$param->getName().($param->isOptional() ? ']' : '');
            echo "; ";
        }
        echo "]";

        echo "\n\n\n";
    }
?>
```

This is how it should be done.

https://3v4l.org/OXjaq PHP 7.1+
https://3v4l.org/Ec5ZX PHP 7.0+
https://3v4l.org/49s1p PHP 5.6+

lib/Raven/Serializer.php
rask: In this context I would say a compromised system would not be relying on the Sentry client to be available. A developer who understands the basic issue with user submitted data/content, server security, upload authz, and common sense will not be opening up holes with this.

It would be on the hands of the Sentry integrator to create this security hole, by allowing a non-predefined callable to be passed to the serializer. The way this Sentry client is integrated most of the time means there is an initial setup during which the context is laid out.

I can understand the risk being applicable in situations like this

```php
<?php

$just_call_me_once = function () {
    $methodname = $_REQUEST['method_to_use'];

    // method_to_use === 'die' (or similar)

    return $methodname();
};

$client->extra_context([
    'my_value' => $just_call_me_once
]);
```

Which shows the developer doing something a developer should never be doing: unsanitized input from user.

If someone can inject a callable (such as `$just_call_me_once`) _and_ tie it to the Sentry client context (via `$client->extra_context()` for instance), I guess the work to achieve that could be severed to a fraction by just calling the malicious code instead of tying it to Sentry and then triggering an error that triggers serialization in Sentry.

Of course I could be way off on this, I'm not a security specialist. But if I wanted to cause harm via my own code I would not jump through rings and hoops by attemping to get Sentry to serialize something.

The main issue would (in my opinion) be developers who do not understand the harm what dynamically selected values/functions/classes based on some request specific parameters can cause. If this is too big of a risk in the eyes of package maintainers I do understand. All suggestions welcome on how this can be implemented in the safest way possible (maybe a flag `allow_callable_serialization` or something for those who understand the risks involved).

Please let me know if there is something I'm clearly missing about this and if I'm making my case for nothing. Thank you for opening the discussion @nokitakaze. :)

ste93cry: Although the concept of lazy-loading is indeed nice, you can't base your assumptions on the fact that any callable/method should be called. It's not something related to security issues, it's just common sense. What if the method calls the database to do some heavy query to get results? Also (I'm not really sure what I'm going to say makes sense), usually Sentry is called when an error happens, and this means that there could be cases where the application is broken and if the method to call relies on something that crashes then bad things can happen

nokitakaze: @rask, logging is a collection of data about an object not a invoking this object. Based on your logic you should invoke every method in a class when you serialize class.

nokitakaze: And yes. If you want to get value from closure just do this
```php
$client->captureMessage('Invoking', [], [
    'extra' => $closure(),
]);
```

rask: The logic seems that way because callables are in essence objects, which are a stem of the serializer `ifelse` logic. If you pass in an object that is callable and can be converted to a `ReflectionFunction` only then will the callable will be evaluated.

I made a crude replica of the serialization logic here: https://3v4l.org/cZMpX

As you see only the "regular" callable is serialized by calling it. The class instances are serialized as they always have been.

See the `RavenUtil::getCallableParamNum` method that forces the callable to be an actual callable and a sole function callable (ie. an object of type `Closure`), not a class instance (which also lead me to an issue regarding PHP5+ as it doesn't have Throwables to catch which leads to fatal in my PR implementation, I should alter the check to see if the thing is a `Closure` first).

I could go and make the logic clearer in the code. As it stands the only things that are serializes/evaluated are callables that are of type `Closure`. This means no classes or instance methods will be called when serializing. I will add some tests to ensure this logic works.

rask: @ste93cry you make a fair point. In my opinion those are more of issues for the implementor to tackle with. I do understand if it would create an unwanted situation where the SDK is blamed for something that _looks_ like the SDKs problem (as in devs make bad decisions and then decide the SDK is bad for allowing it, instead of taking the time to see what the actual problem is).

Error situations are hairy. I tried a few times and could read global state properly via framework functions during a simulated error but I guess there are situations where all is lost. I wonder if PHP7s Throwables could be used to some extent when evaluating callables in case errors happen during evaluation and the serialization is stopped from that point forward for the callable. Hard fatal errors are a bit tricky though.

ste93cry: What if people starts complaining because closures are called and invokable classes not? By the way, these points is still valid:

> you can't base your assumptions on the fact that any callable/method should be called

nokitakaze: > What if people starts complaining

People will not complaining if this PR will not be merged at this state. And I assume you and @dcramer will not approve something what CALLS CLOSURES

rask: @ste93cry pardon my asking, but as a non-native English speaker I would like a clarification regarding

> you can't base your assumptions on the fact that any callable/method should be called

Do you mean that I should assume that all callables should not be evaluated unless there is heavy reason to do it (e.g. we should make the callable evaluation a more defined decision on the Sentry integrators part)?

Or something else?

ste93cry: >pardon my asking, but as a non-native English speaker

I'm not native English speaker either 😄 

Btw, what I meant is that functions could do some heavy work inside, for example they could do heavy operations to compute some value or executes some queries that take time, so we shouldn't call indiscriminately all of them just to serialize the result value. Plus, maybe an user wants to serialize just some values: if before nothing was serialized at all, with your changes everything is serialized now. My suggestion, which anyway is beyond the original topic of this PR and doesn't address your problem, is to serialize the name, argument types and return type of a closure, but not the return value itself. I understand why you suggested to call the closures and I'm not sure this is enough, but you could use a custom processor to add data on-the-fly to the event being sent to the server

rask:
As per the suggestion from @ste93cry I checked out custom processors and managed to get what I needed working for now. Another reminder to peruse the source code and documentation a bit more later on.

I also saw @nokitakaze submitted a PR to implement the extended callable string casting serialization he was talking about in this PR. :)

I presume this PR can be closed on these grounds. Thanks for the constructive brainstorming on this one. Maybe someday we can have a "built-in" solution that works better for everyone.

PR #446, opened 2017-04-12
Missing closing parenthesis on usage filtering doc

slipszenko:


dcramer:
Thanks! 

(Also ugh our APIs to do this kind of stuff are so sad)

PR #447, opened 2017-04-18
Serializing all objects

nokitakaze:
Serializing all objects by their public properties.
And some small refactoring for Serializer tests.

I don't know why we serialize only stdClass. Objects are just array with some behaviour and always named keys. So we should treat them appropriately.

```php
$client = new \Raven\Client(..., [
    'serialize_all_object' => true,
]);
```

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -203,8 +203,7 @@ class Client
         $this->serializer = new \Raven\Serializer($this->mb_detect_order);
         $this->reprSerializer = new \Raven\ReprSerializer($this->mb_detect_order);
         if (\Raven\Util::get($options, 'serialize_all_object', false)) {
-            $this->serializer->setAllObjectSerialize(true);
-            $this->reprSerializer->setAllObjectSerialize(true);
+            $this->setAllObjectSerialize(true);
         }
 
         if ($this->curl_method == 'async') {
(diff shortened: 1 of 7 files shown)


dcramer:
There are limits on the size of attributes in which Sentry allows. IMO doing this is problematic because of the performance hit it will cause, as well as the massive increase in size we'll see in many cases.

nokitakaze:
This behaviour is disabled by default. But in my projects I will enable it.

> size of attributes
> performance hit
> massive increase in size

All these points are also inherent in arrays

dcramer:
@nokitakaze yes but if you're passing around a bunch of references to objects, they'll now attempt to recursively serialize attributes which has:

- CPU hit on PHP end
- Bandwidth/latency hit on transfer
- Gets truncated by Sentry on server once entire payload is over limit (and you dont get to choose how)

nokitakaze:
And how is this different from the same case with arrays?

dcramer:
No difference I just wanted to be clear on consequences. Will do a review pass when I have some free time.

lib/Raven/Client.php
```diff
@@ -198,6 +202,10 @@ public function __construct($options_or_dsn = null, $options = array())
         ));
         $this->serializer = new \Raven\Serializer($this->mb_detect_order);
         $this->reprSerializer = new \Raven\ReprSerializer($this->mb_detect_order);
+        if (\Raven\Util::get($options, 'serialize_all_object', false)) {
+            $this->serializer->setAllObjectSerialize(true);
```
dcramer: any reason not to use setAllObjectSerialize that you added below?

nokitakaze: ooops... Give me 10 minutes

dcramer:
@nokitakaze aside if you didnt know you should have access to merge once its reviewed (even if you make changes after the fact)

nokitakaze:
> have access to merge once its reviewed
Yes, I know.

PR #448, opened 2017-04-23
Refactor and improve the Stacktrace class

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

The aim of this PR is refactor the `Stacktrace` class to make it more readable, improve the performance of the code and reduce its overall complexity. I think that the part that handles the serialization of the function arguments could be improved as well. I was evaluating the [VarDumper](http://symfony.com/doc/current/components/var_dumper.html) component from Symfony that provides a cloner that supports a lot of types (entities, PHP classes, etc) and could do a better job than what we will ever do to cast the arguments values to a format suitable for serialization. Users can even register custom casters with the component, so that their objects are dumped in the way they want (this is currently not possible, and honestly the current code just handles a few basic types like scalars).


lib/Raven/Stacktrace.php
```diff
@@ -1,166 +1,344 @@
...
 
 /**
- * Small helper class to inspect the stacktrace
+ * This class contains all the information about an error stacktrace.
  *
- * @package raven
+ * @author Stefano Arlandini <sarlandini@alice.it>
  */
-class Stacktrace
+class Stacktrace implements \JsonSerializable
```
dcramer: is this a newer native PHP behavior (the JsonSerializable bit)?

bin/sentry
```diff
@@ -6,8 +6,7 @@ error_reporting(E_ALL | E_STRICT);
 
 // TODO: if we could get rid of this and have composer figure things out it'd make it
 // a bit more sane
-require(__DIR__ . '/../lib/Raven/Autoloader.php');
-\Raven\Autoloader::register();
+require(__DIR__ . '/../vendor/autoload.php');
```
dcramer: this seems unrelated?

ste93cry: I fixed this because I used this command to test whether the stacktrace was still sent correctly to the server, so I needed it to work correctly

lib/Raven/Stacktrace.php
ste93cry: It's available since PHP 5.4.0 and (quoting from the docs) objects implementing [this interface](http://php.net/manual/en/class.jsonserializable.php) can customize their JSON representation when encoded with `json_encode()`

PR #454, opened 2017-04-30
[2.0] Serializer interface

nokitakaze:
> I meant that we should provide a sort of `SerializerInterface` interface and a basic implementation (that actually can even check if the object being passed implements some methods like `toDebugContext`, but this is up to us) and if a user wants to send a custom serialization of an object then he can just (yes, using dependency injection) set the serializer to something else and do whatever he wants.

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -67,11 +67,11 @@ class Client
     protected $error_types;
 
     /**
-     * @var \Raven\Serializer $serializer
+     * @var SerializerInterface $serializer
      */
     protected $serializer;
     /**
-     * @var \Raven\Serializer $serializer
+     * @var SerializerInterface $serializer
      */
     protected $reprSerializer;
 
@@ -341,8 +341,16 @@ class Client
             'name' => 'sentry-php',
             'version' => self::VERSION,
         ]);
-        $this->serializer = new \Raven\Serializer($this->mb_detect_order);
-        $this->reprSerializer = new \Raven\ReprSerializer($this->mb_detect_order);
+        if (isset($options['serializer'])) {
+            $this->setSerializer($options['serializer']);
+        } else {
+            $this->serializer = new \Raven\Serializer($this->mb_detect_order);
+        }
+        if (isset($options['reprSerializer'])) {
+            $this->setReprSerializer($options['reprSerializer']);
+        } else {
+            $this->reprSerializer = new \Raven\ReprSerializer($this->mb_detect_order);
+        }
         if (\Raven\Util::get($options, 'serialize_all_object', false)) {
             $this->setAllObjectSerialize(true);
         }
@@ -378,21 +386,31 @@ class Client
     /**
      * Gets the representation serialier.
      *
-     * @return ReprSerializer
+     * @return SerializerInterface
      */
     public function getReprSerializer()
     {
         return $this->reprSerializer;
     }
 
+    public function setReprSerializer(SerializerInterface $serializer)
+    {
+        $this->reprSerializer = $serializer;
+    }
+
     /**
      * Gets the serializer.
      *
-     * @return Serializer
+     * @return SerializerInterface
      */
     public function getSerializer()
     {
         return $this->serializer;
+    }
+
+    public function setSerializer(SerializerInterface $serializer)
+    {
+        $this->serializer = $serializer;
     }
 
     /**
@@ -1046,7 +1064,7 @@ class Client
             $data['request'] = $this->serializer->serialize($data['request']);
         }
         if (!empty($data['user'])) {
-            $data['user'] = $this->serializer->serialize($data['user'], 3);
+            $data['user'] = $this->serializer->serialize($data['user'], ['max_depth' => 3]);
         }
         if (!empty($data['extra'])) {
             $data['extra'] = $this->serializer->serialize($data['extra']);
@@ -1057,7 +1075,7 @@ class Client
             }
         }
         if (!empty($data['contexts'])) {
-            $data['contexts'] = $this->serializer->serialize($data['contexts'], 5);
+            $data['contexts'] = $this->serializer->serialize($data['contexts'], ['max_depth' => 5]);
         }
     }
 
@@ -1602,7 +1620,11 @@ class Client
 
     public function setAllObjectSerialize($value)
     {
-        $this->serializer->setAllObjectSerialize($value);
-        $this->reprSerializer->setAllObjectSerialize($value);
+        if (method_exists($this->serializer, 'setAllObjectSerialize')) {
+            $this->serializer->setAllObjectSerialize($value);
+        }
+        if (method_exists($this->reprSerializer, 'setAllObjectSerialize')) {
+            $this->reprSerializer->setAllObjectSerialize($value);
+        }
     }
 }
--- a/lib/Raven/Serializer.php
+++ b/lib/Raven/Serializer.php
@@ -26,7 +26,7 @@ namespace Raven;
  *
  * @package raven
  */
-class Serializer
+class Serializer implements SerializerInterface
 {
     /*
      * The default mb detect order
@@ -67,17 +67,44 @@ class Serializer
      * sanitization and encoding.
      *
      * @param mixed $value
+     * @param array $context
+     * @return string|bool|double|int|null|object|array
+     */
+    public function serialize($value, array $context = [])
+    {
+        $full_context = $this->get_full_context($context);
+        return $this->serializeInner($value, $full_context['max_depth'], 0);
+    }
+
+    /**
+     * @param array $context
+     * @return array
+     */
+    protected function get_full_context(array $context)
+    {
+        if (!array_key_exists('max_depth', $context)) {
+            $context['max_depth'] = 3;
+        }
+
+        return $context;
+    }
+
+    /**
+     * Serialize an object (recursively) into something safe for data
+     * sanitization and encoding.
+     *
+     * @param mixed $value
      * @param int   $max_depth
      * @param int   $_depth
      * @return string|bool|double|int|null|object|array
      */
-    public function serialize($value, $max_depth = 3, $_depth = 0)
+    protected function serializeInner($value, $max_depth, $_depth)
     {
         if ($_depth < $max_depth) {
             if (is_array($value)) {
                 $new = [];
                 foreach ($value as $k => $v) {
-                    $new[$this->serializeValue($k)] = $this->serialize($v, $max_depth, $_depth + 1);
+                    $new[$this->serializeValue($k)] = $this->serializeInner($v, $max_depth, $_depth + 1);
                 }
 
                 return $new;
@@ -111,7 +138,7 @@ class Serializer
             if (is_object($value)) {
                 $new_value = $this->serializeObject($value, $max_depth, $_depth + 1, $hashes);
             } else {
-                $new_value = $this->serialize($value, $max_depth, $_depth + 1);
+                $new_value = $this->serialize($value, ['max_depth' => $max_depth - $_depth - 1]);
             }
             $return[$key] = $new_value;
         }
--- a/lib/Raven/SerializerInterface.php
+++ b/lib/Raven/SerializerInterface.php
@@ -1,0 +1,17 @@
+<?php
+
+
+namespace Raven;
+
+interface SerializerInterface
+{
+    /**
+     * Serialize an object (recursively) into something safe for data
+     * sanitization and encoding.
+     *
+     * @param mixed $value
+     * @param array $context
+     * @return string|bool|double|int|null|object|array
+     */
+    public function serialize($value, array $context = []);
+}
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -10,9 +10,6 @@
 
 namespace Raven\Tests;
 
-use Raven\Client;
-use Raven\Serializer;
-
 function simple_function($a = null, $b = null, $c = null)
 {
     assert(0);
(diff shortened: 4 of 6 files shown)


lib/Raven/SerializerInterface.php
```diff
@@ -0,0 +1,18 @@
...
+    /**
+     * Serialize an object (recursively) into something safe for data
+     * sanitization and encoding.
+     *
+     * @param mixed $value
+     * @param int   $max_depth
+     * @param int   $_depth
+     * @return string|bool|double|int|null|object|array
+     */
+    public function serialize($value, $max_depth = 3, $_depth = 0);
```
ste93cry: What about changing the signature of this method to `serialize($value, array $context = [])`? This way, it's more standardized according to other serializer libraries and mostly important it allows us to pass additional options in the future without breaking compatibility by adding a new method to the interface

tests/ClientTest.php
```diff
@@ -210,6 +207,24 @@ public function join($timeout = null)
     }
 }
 
+class Dummy_Serializer implements \Raven\SerializerInterface
```
ste93cry: This seems bad because there is logic code in a test that doesn't reflect what the real implementation of the serializer does. You should use a mocked object instead

lib/Raven/SerializerInterface.php
ste93cry: What about changing the signature of this method to `serialize($value, array $context = [])?` This way, it's more standardized according to other serializer libraries and mostly important it allows us to pass additional options in the future without breaking compatibility by adding a new method to the interface

lib/Raven/Client.php
```diff
@@ -1602,7 +1620,11 @@ public function close_curl_resource()
 
     public function setAllObjectSerialize($value)
     {
-        $this->serializer->setAllObjectSerialize($value);
-        $this->reprSerializer->setAllObjectSerialize($value);
+        if (method_exists($this->serializer, 'setAllObjectSerialize')) {
```
ste93cry: Accordind to the comment left below about the `serialize` function definition, we could use the `$context` parameter to pass additional options to the serializer (in this case about serializing all objects) so that we don't need to rely on some methods not even defined in the interface

lib/Raven/SerializerInterface.php
nokitakaze: How should I change signature of existing method without breaking back compatibility? @dcramer will grumble

tests/ClientTest.php
nokitakaze: I don\'t know how to create method with mock object library. That's why I never used it in my tests

lib/Raven/SerializerInterface.php
ste93cry: Fortunately, as we're working on `2.x` branch we can break compatibility with older versions, so there shouldn't be any problem in doing it. Plus, I think it was a really big mistake to have all properties/methods public in the `1.x` version. We should expose only things that actually serves a purpose and I don't think that a public `serialize` method is so useful since the scope of this lbrary is not to be a serializer. So, here what I would do: make the `Serializer::serializeObject` protected (there is no reason to let it public), change the signature of the `Serializer::serialize` method according to the previously suggested definition (yes, breaking compatibility) and rework the code calling it to use the `$context` argument to pass the old arguments like `$max_depth` as key/values of the array. Plus, the `$depth` argument is something related to internal behaviour and should not be exposed anyway in a public method. One final though: should we provide a basic implementation of a serializer or should we just delegate it to a 3rd party library? I mean, there are a lot of cases to handle (circular references, how to cast objects/closures/different types) and there are already existing projects that do it well. Why should we reinvent the wheel?

tests/ClientTest.php
ste93cry: The `ClientTest::testOurSerializer` method is actually testing the value returned from the serializer instance. I think that this is not appropriate, as the test class about the serializer should do these types of assertions. Instead, here we should check if the serializer is correctly set using the options/setters and if it's called. This is what mocks are for, they verify that something happens. PHPUnit doc is great and has a lot of examples on how to do it

nokitakaze: > we should check if the serializer is correctly set

Yes. This is exactly what I\'m testing. Look at this https://github.com/getsentry/sentry-php/pull/454/files#diff-668268a1040801a44451100bf9062b17R2510
I set some serializer and then I check if it has been set

ste93cry: I know the scope of that test is to check that the serializer is correctly set, but there is no need to check the value returned after serializing a message to do it. You should create a mock, pass it in place of the current `Dummy_Serializer` object and assert that its `serialize` method is called when capturing an exception/message. Nothing more, nothing else. No checks on how the sent message is serialized itself

lib/Raven/SerializerInterface.php
nokitakaze: > should we provide a basic implementation of a serializer

We did it already. \\Raven\\Serializer is it. Anyone can extend it

> make the Serializer::serializeObject protected
> change the signature of 

I will do it tomorrow. 

> it was a really big mistake to have all properties/methods public

Okay. Which of us will make a pull request that will make all methods protected?

tests/ClientTest.php
nokitakaze: > and assert that its serialize method is called

Okay. I will change it tomorrow

lib/Raven/SerializerInterface.php
ste93cry: >Okay. Which of us will make a pull request that will make all methods protected?

This is off-topic, but I don't think there is a need _now_ to open a PR to change all methods from public to protected because we will probably do it with the future PRs regarding other changes. Mine was just a statement,related to public methods that were strictly connected with the serializer (`Client::serialize` for example): only those should be changed in this PR

tests/ClientTest.php
```diff
@@ -2528,9 +2516,7 @@ public function testOurSerializer($after_creation)
                     ],
                 ]
             );
-            $events = $client->getSentEvents();
-            $event = array_pop($events);
-            $this->assertEquals(Dummy_Serializer::$default_value, $event['extra']['foo']);
+            $serializer->__phpunit_verify();
```
ste93cry: This is wrong. There is no need to call an internal PHPUnit function to assert that a given method has been called. You should mock the methods and then just call the code as you would in the real use-case and PHPUnit will do the rest for you.

lib/Raven/SerializerInterface.php
```diff
@@ -10,9 +10,8 @@
      * sanitization and encoding.
      *
      * @param mixed $value
-     * @param int   $max_depth
-     * @param int   $_depth
+     * @param array $context
```
ste93cry: Missing doc for the argument. Plus, I would just put `mixed` as return value instead of a long list of scalar types

lib/Raven/Serializer.php
```diff
@@ -67,17 +67,44 @@ public function __construct($mb_detect_order = null)
...
+     * Serialize an object (recursively) into something safe for data
+     * sanitization and encoding.
+     *
+     * @param mixed $value
      * @param int   $max_depth
      * @param int   $_depth
      * @return string|bool|double|int|null|object|array
      */
-    public function serialize($value, $max_depth = 3, $_depth = 0)
+    protected function serialize_inner($value, $max_depth, $_depth)
```
ste93cry: This comment applies to any code: properties, variables and methods should all use camel case, so for example `$max_depth` should be renamed to `$maxDepth`

lib/Raven/Serializer.php
```diff
@@ -67,17 +67,44 @@ public function __construct($mb_detect_order = null)
      * sanitization and encoding.
      *
      * @param mixed $value
+     * @param array $context
+     * @return string|bool|double|int|null|object|array
```
ste93cry: Missing documentation, and return value should be `mixed` (it's easier to read than having a so long list of scalar types)

lib/Raven/Serializer.php
```diff
@@ -67,17 +67,44 @@ public function __construct($mb_detect_order = null)
...
+        return $this->serialize_inner($value, $full_context['max_depth'], 0);
+    }
+
+    /**
+     * @param array $context
+     * @return array
+     */
+    protected function get_full_context(array $context)
+    {
+        if (!array_key_exists('max_depth', $context)) {
```
ste93cry: Do we really need to always set this option? If he wants to serialize all objects he can't this way, but should be able to do it

lib/Raven/SerializerInterface.php
nokitakaze: Which argument I missed?

tests/ClientTest.php
nokitakaze: I don\'t know how to check asserting **after each call**. Assert must be done in every iteration of foreach block

lib/Raven/Serializer.php
nokitakaze: You mean "max_depth"? I saved it for back compatibility

lib/Raven/Serializer.php
nokitakaze: I got this doc from this file itself. Do we want to change all serializing method to "return mixed" INCLUDING `SerializerInterface::serialize`

lib/Raven/Client.php
nokitakaze: > not even defined in the interface
> if (method_exists...

nokitakaze:
Erm... [Where is Travis](https://travis-ci.org/getsentry/sentry-php)?

lib/Raven/SerializerInterface.php
ste93cry: `$value` and `$mixed` don't have a description

lib/Raven/Serializer.php
ste93cry: I would, because it's prettier imho

tests/ClientTest.php
ste93cry: As we're not testing anymore the return value of the serializer, the whole `foreach` block isn't needed. Just capture a single message and check that the `serialize` method of the mocked object is called

lib/Raven/Serializer.php
ste93cry: The `max_depth` option (correct me if I'm wrong) seems to be always set from the lines of code that call the serialize function, so backwards compatibility seems to be guaranteed. With these lines of code, the user won't be able to serialize everything regardless of depth because a max depth of 3 is always forced to be set

nokitakaze: > that call the serialize function

You are wrong. Serialize function could be called beyond \\Raven\\Client. So we must make sure `$context['max_depth']` is set

lib/Raven/Serializer.php
nokitakaze: "Pretiest" ways are always wrong in our art. This is the programming not a drawing Mona List

lib/Raven/Serializer.php
ste93cry: >Serialize function could be called beyond \Raven\Client

Is this referring to the fact that `Client::serialize` is public? Imho, it shouldn't because we're not a library to serialize things, so we don't provide a serializer that can be used for something else than serializing values being sent to a Sentry server. There is no need for this method to be public

nokitakaze: `SerializeInterface::serialize` is public. It's enough.

ste93cry: I think you misunderstood me. I said that `Client::serialize` is public and shouldn't be because there is no reason to call this method outside of the `Client` class itself to serialize values sent to the server. And while that class doesn't implement `SerializerInterface`, then we are free to make the method protected

tests/ClientTest.php
nokitakaze: We must be sure serialize is invoked on every accessible type. I create foreach block for this and only for this purpose.

lib/Raven/Serializer.php
nokitakaze: Ok. We can do Client::serialize protected.
.
.
.
Nothing changed. `SerializerInterface::serialize` and `Serizlizer::serialize` are still public so they could be invoked. So we **must** check this key existing

lib/Raven/Serializer.php
ste93cry: [`mixed`](https://phpdoc.org/docs/latest/guides/types.html#keywords) is used in places where a value can be of different types and it's appropriate here instead of a infinite list of possible return types (what if in the future we decide that we return class instances, etc?)

lib/Raven/Serializer.php
ste93cry: >So we must check this key existing

No, we don't. It's obvious that the `Serializer::serialize` method is public because it must be called, but the instance itself of the serializer **should not** be accessed outside of the Client class because it's used internally and it's not meant to be called by 3rd parties. There are now two methods to retrieve the serializer and representation serializer instances because I had no choice while refactoring the `Stacktrace` class, but the plan is to remove them when we will refactor the `Client` class

ste93cry: >are still public so they could be invoked

No, we don't. It's obvious that the `Serializer::serialize` method is public because it must be called, but the instance itself of the serializer **should not** be accessed outside of the Client class because it's used internally and it's not meant to be called by 3rd parties. There are now two methods to retrieve the serializer and representation serializer instances because I had no choice while refactoring the `Stacktrace` class, but the plan is to remove them when we will refactor the `Client` class

tests/ClientTest.php
ste93cry: No we don't. we just have to test that the `Serializer::serialize` method is called when appropriate, e.g. when sending a message to the Sentry server. There is no reason to test that (this is just an example) a `boolean` is passed to the method

lib/Raven/Serializer.php
nokitakaze: I know what "mixed" means. If you see you can notice what that string does not contain "resource". Mixed contains "resource"

lib/Raven/Serializer.php
nokitakaze: > should not
You are right. But PHP is a language which does not support internal class of package nor packages itself nor "friends" (who "can touch each other private parts"). Of course I **could** delete "context" checking, but I think it is a bad practice.

ste93cry: >There are now two methods to retrieve the serializer and representation serializer instances [...] but the plan is to remove them

So, it's safe to assume that there won't be any way to access the instance of the serializer through getters

nokitakaze: Why do you want to delete getters? I added them on purpose. I want the situation where users can change serializers or can get them from Client instance .

ste93cry: >I want the situation where users can change serializers or can get them from Client instance

Users can change serializer through options, and this is not related in any way to getters. But the serializer is an internal thing used to convert values to something suitable to be sent to Sentry and there is no reason to expose it to the public. I don't want to be rude, but it seems to me you did not read anything of what I said in all my previously comments (this is the 3rd time I say this).

lib/Raven/Serializer.php
ste93cry: `mixed` can be used when a value can be literally anything; the author of the documentation is unable to predict which type it will be. So, instead of having a long list of types you can just use that type. There is no reason to not do it

lib/Raven/Serializer.php
nokitakaze: Okay. 

1. You do `composer require sentry/sentry`
2. You see what serializers are not working how you want them to work
3. You write your own Serializer
4. By any reasons you want to change serializers on the fly. I.e. you write your own warning capturing in your own CMS. You want to change standard serializer for global Raven client and you do not want to extend \\Raven\\Client. What do you do without setters and reflections? Only extend \\Raven\\Client and override `capture` where you change currents serializer and call `parent::capture`

I want get users their own choice to change behaviour of Raven.

lib/Raven/Serializer.php
nokitakaze: If I have many available types but these are not **all** valid options I chouse to list them

https://github.com/nokitakaze/ascetcms/blob/49049665babfd32564cc567d69ea4b41e9457d3e/functions/smodel.php#L1326
https://github.com/nokitakaze/ascetcms/blob/49049665babfd32564cc567d69ea4b41e9457d3e/functions/sdbcriteria.php#L10

lib/Raven/Serializer.php
ste93cry: Why would you want to change serializer on-the-fly? Honestly, I don't see any valid reason for doing this. If you want to change what is sent with the event you use processors (they can edit the data before serialization, so you can add/remove/change data there). Serialization just takes the final data and serializes it to something suitable for sending it over the net (basically, it's just an advanced version of `json_encode`)

lib/Raven/Serializer.php
ste93cry: The serializer accepts values of **any** type (and this is correct). The doc is wrong, if you somehow pass a `resource` object nothing stops you.

lib/Raven/Serializer.php
nokitakaze: > By any reasons

If we can not come up with the case where it can be useful it doesn't mean that there is no such case

lib/Raven/Serializer.php
nokitakaze: Accept — of course. But we are talking about param and return types, aren\'t we? Of course we should change param types to "mixed". I will do it. But what about return types?

lib/Raven/Serializer.php
ste93cry: >it doesn't mean that there is no such case

If there will be any valid use case we will add the getter. It's easier to add something that removing it

nokitakaze: You forgot that Sentry is a library for many real users. There are many and many people besides us. I just don\'t want to delete something what can be seeing usable by some people.

lib/Raven/Serializer.php
ste93cry: Return types should be only `string`, `int`, `float`, `bool`, `array` and `null`

nokitakaze: .
.
.
Oooops. Okay. I misunderstood you. I will fix it

lib/Raven/Serializer.php
ste93cry: >You forgot that Sentry is a library for many real users

Trust me when I say that I always think about it, and rest assured that any decision I take is not based on personal preferences for my own projects. I try to give a clear public API to the end-users, and to me exposing something used internally is a bad practice. As this discussion is going too much further in time: @dcramer can you please say what's your opinion?

lib/Raven/Serializer.php
nokitakaze: Wait a minute
> return value should be mixed

There is no other doc items which contain some long string with types. Only `mixed` value and `array` context.

ste93cry: I was wrong in saying that the return value of the function should be that because the return types can only be scalars). I would use mixed anyway, because I find it prettier and easier to read, but feel free to do whathever you want on this. Doc is still missing for the `$value` and `$context` arguments and that is my main concern

nokitakaze: > Doc is still missing

I don\'t know what should I write. And I am afraid that my line should be corrected because english is not my native language. I know it isn\'t yours either.

lib/Raven/Serializer.php
nokitakaze: I just checked. There in no \\Raven\\Client::serialize. Only "sanitize".

ste93cry: You're right, I meant to say that `Client::getSerializer` and `Client::getReprSerializer` are public when they shouldn't, but it was done because at that time I had no other choice. The plan is to remove them anyway. By the way, `Client::sanitize` should too not be public. Having all these methods private/protected means that there would be no way to access the serializer outside of the client (without inheriting the class, at least) and this means that there is no need the check the `max_depth` option because no BC would exists (as we control where the serializer is used). Does this makes sense?

nokitakaze: Tomorrow (2017/05/09) I will do pull request which will make all methods and properties protected and will break all compatibility we have at this moment

nokitakaze: > make all methods and properties protected

Okay, It\'s already done: #457

ste93cry:
I'm going to close this as it's really old and it has been superseded by multiple PRs attempting to solve the same issue, last of all #701

PR #455, opened 2017-05-08
Adding X-Authorization and Cookie to sanitized headers

igorsantos07:
Cookie values are sanitized by default at the default `SanitizeDataProcessor`, so it's safe to consider they should also be sanitized in headers. The cookie keys list will be available on the "cookie" data point anyway (unless `RemoveCookiesProcessor` is enabled), thus it makes no sense to sanitize cookies there but not in the header.
X-Authorization is a very common header user to transmit arbitrary Token values on APIs.

### DIFF
--- a/lib/Raven/Processor/SanitizeHttpHeadersProcessor.php
+++ b/lib/Raven/Processor/SanitizeHttpHeadersProcessor.php
@@ -59,6 +59,6 @@ final class Raven_Processor_SanitizeHttpHeadersProcessor extends Raven_Processor
      */
     private function getDefaultHeaders()
     {
-        return array('Authorization', 'Proxy-Authorization', 'X-Csrf-Token', 'X-CSRFToken', 'X-XSRF-TOKEN');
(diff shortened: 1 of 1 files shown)


ste93cry:
It's fine to add the `X-Authorization` header to the list, but the `Cookie` header should already be unset by `RemoveCookiesProcessor` processor. Unless it doesn't work (and it would be a bug), this is just code duplication

ste93cry:
It's fine to add the `X-Authorization` header to the list, but the `Cookie` header should already be sanitized by the `RemoveCookiesProcessor` processor. By default in `2.x` (which is being worked on) that processor will be enabled by default, so my suggestion is to enable it instead of duplicating the logic in the `SanitizeDataProcessor` class.

nokitakaze:
I suggest redirect this PR to [branch 2.0](https://github.com/getsentry/sentry-php/tree/2.0). There is no need to add _new_ behaviour to _old_ branch

igorsantos07:
@nokitakaze too bad that's not pointed in any contributing documentation :roll_eyes:

igorsantos07:
Superseded by #474

PR #456, opened 2017-05-08
Delete always true conditions with version compare

nokitakaze:
Small change. Delete dead code

### DIFF
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -591,10 +591,6 @@ class Raven_Tests_ClientTest extends \PHPUnit_Framework_TestCase
      */
     public function testCaptureExceptionChainedException()
     {
-        if (version_compare(PHP_VERSION, '5.3.0', '<')) {
-            $this->markTestSkipped('PHP 5.3 required for chained exceptions.');
-        }
-
         # TODO: it'd be nice if we could mock the stacktrace extraction function here
         $client = new Dummy_Raven_Client();
         $ex = $this->create_chained_exception();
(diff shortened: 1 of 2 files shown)


tests/ClientTest.php
```diff
@@ -1846,15 +1836,7 @@ public function testEncodeTooDepth()
...
-        if (version_compare(PHP_VERSION, '5.5.0', '>=')) {
-            $this->assertFalse($value, 'Broken data encoded successfully with native method');
-        } else {
-            if ($value !== false) {
-                $this->markTestSkipped();
-            } else {
-                $this->assertEquals('eJyLjh4Fo2AUjFgQOwpGwSgYuQAA3Q7g1w==', $value, 'Native json_encode error');
-            }
-        }
+        $this->assertFalse($value, 'Broken data encoded successfully with native method');
```
ste93cry: Didn't this removed too much code? I'm not sure the `else` part should be entirely removed

nokitakaze: `version_compare(PHP_VERSION, '5.5.0', '>=')` is always true condition

tests/ClientTest.php
```diff
@@ -1838,25 +1828,6 @@ public function test_server_variable()
         }
     }
 
-    public function testEncodeTooDepth()
```
nokitakaze: I deleted the entire method because we do not need to test INTERNAL PHP FUNCTIONS!

nokitakaze: @ste93cry, I need your opinion. Do you approve this?

ste93cry: Yes it's fine

PR #457, opened 2017-05-09
[2.0] Make all inner methods and properties protected

nokitakaze:
This pull request **breaks** back compatibility.

### DIFF
--- a/lib/Raven/Stacktrace.php
+++ b/lib/Raven/Stacktrace.php
@@ -305,7 +305,7 @@ class Stacktrace implements \JsonSerializable
      *
      * @return array
      */
-    public static function getFrameArguments($frame, $maxValueLength = Client::MESSAGE_LIMIT)
+    protected static function getFrameArguments($frame, $maxValueLength = Client::MESSAGE_LIMIT)
     {
         if (!isset($frame['args'])) {
             return [];
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -92,6 +92,16 @@ class Dummy_Raven_Client extends \Raven\Client
     {
         return $this->get_current_url();
     }
+
+    public function encode(&$data)
+    {
+        return parent::encode($data);
+    }
+
+    public function process(&$data)
+    {
+        parent::process($data);
+    }
 }
 
 class Dummy_Raven_Client_With_Overrided_Direct_Send extends \Raven\Client
@@ -334,10 +344,10 @@ class Raven_Tests_ClientTest extends \PHPUnit_Framework_TestCase
     {
         $client = new Dummy_Raven_Client('http://public:secret@example.com/1');
 
-        $this->assertEquals(1, $client->project);
-        $this->assertEquals('http://example.com/api/1/store/', $client->server);
-        $this->assertEquals('public', $client->public_key);
-        $this->assertEquals('secret', $client->secret_key);
+        $this->assertAttributeEquals(1, 'project', $client);
+        $this->assertAttributeEquals('http://example.com/api/1/store/', 'server', $client);
+        $this->assertAttributeEquals('public', 'public_key', $client);
+        $this->assertAttributeEquals('secret', 'secret_key', $client);
     }
 
     /**
@@ -349,11 +359,11 @@ class Raven_Tests_ClientTest extends \PHPUnit_Framework_TestCase
             'site' => 'foo',
         ]);
 
-        $this->assertEquals(1, $client->project);
-        $this->assertEquals('http://example.com/api/1/store/', $client->server);
-        $this->assertEquals('public', $client->public_key);
-        $this->assertEquals('secret', $client->secret_key);
-        $this->assertEquals('foo', $client->site);
+        $this->assertAttributeEquals(1, 'project', $client);
+        $this->assertAttributeEquals('http://example.com/api/1/store/', 'server', $client);
+        $this->assertAttributeEquals('public', 'public_key', $client);
+        $this->assertAttributeEquals('secret', 'secret_key', $client);
+        $this->assertAttributeEquals('foo', 'site', $client);
     }
 
     /**
@@ -366,7 +376,7 @@ class Raven_Tests_ClientTest extends \PHPUnit_Framework_TestCase
             'project' => 1,
         ]);
 
-        $this->assertEquals('http://example.com/api/1/store/', $client->server);
+        $this->assertAttributeEquals('http://example.com/api/1/store/', 'server', $client);
     }
 
     /**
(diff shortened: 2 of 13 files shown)


tests/Processor/SanitizeDataProcessorTest.php
```diff
@@ -157,7 +159,9 @@ public function testOverridenSanitize($processorOptions, $client_options, $dsn)
         /**
          * @var SanitizeDataProcessor $processor
          */
-        $processor = $client->processors[0];
+        $processors = new \ReflectionProperty($client, 'processors');
```
ste93cry: PHPUnit provides an `assertAttributeEquals` method to check whether an attribute (even private) equals a given value

tests/Processor/SanitizeStacktraceProcessorTest.php
```diff
@@ -65,9 +68,12 @@ public function testProcessWithPreviousException()
             $this->client->captureException($exception);
         }
 
-        $this->processor->process($this->client->_pending_events[0]);
+        $_pending_events = new \ReflectionProperty($this->client, '_pending_events');
```
ste93cry: PHPUnit provides an `assertAttributeEquals` method to check whether an attribute (even private) equals a given value

tests/Processor/SanitizeStacktraceProcessorTest.php
```diff
@@ -42,9 +42,12 @@ public function testProcess()
...
```
ste93cry: PHPUnit provides an `assertAttributeEquals` method to check whether an attribute (even private) equals a given value

lib/Raven/Stacktrace.php
```diff
@@ -171,7 +171,7 @@ public function addFrame($file, $line, array $backtraceFrame)
      *
      * @throws \OutOfBoundsException If the index is out of range
      */
-    public function removeFrame($index)
+    protected function removeFrame($index)
```
ste93cry: This should not be protected, an user may want to remove a given frame from the stacktrace before sending it to the Sentry server

tests/ClientTest.php
```diff
@@ -334,10 +344,18 @@ public function testDsnFirstArgument()
     {
         $client = new Dummy_Raven_Client('http://public:secret@example.com/1');
 
-        $this->assertEquals(1, $client->project);
-        $this->assertEquals('http://example.com/api/1/store/', $client->server);
-        $this->assertEquals('public', $client->public_key);
-        $this->assertEquals('secret', $client->secret_key);
+        $project = new \ReflectionProperty($client, 'project');
```
ste93cry: PHPUnit provides an `assertAttributeEquals` method to check whether an attribute (even private) equals a given value

tests/TransactionStackTest.php
```diff
@@ -30,7 +30,9 @@ public function testSimple()
         $this->assertEquals($stack->peek(), null);
 
         $stack->clear();
-        $this->assertInternalType('array', $stack->stack);
-        $this->assertEquals(0, count($stack->stack));
+        $stack_r = new \ReflectionProperty($stack, 'stack');
```
ste93cry: PHPUnit provides an `assertAttributeEquals` method to check whether an attribute (even private) equals a given value

tests/StacktraceTest.php
```diff
@@ -198,7 +198,9 @@ public function testRemoveFrame($index, $throwException)
 
         $this->assertCount(2, $stacktrace->getFrames());
 
-        $stacktrace->removeFrame($index);
+        $removeFrame = new \ReflectionMethod($stacktrace, 'removeFrame');
```
ste93cry: This method should be public, so there is no need to use reflection here

tests/Processor/SanitizeStacktraceProcessorTest.php
nokitakaze: I do not assert this value. I am doing magic with this value

nokitakaze: I do not assert this value. I am doing magic with this value

nokitakaze:
.
.
.
uopz can not be compiled with PHP 7.1 at this state.
```
make: *** [src/class.lo] Error 1
ERROR: `make' failed
```

It is beautiful

tests/Processor/SanitizeStacktraceProcessorTest.php
nokitakaze: Same here. I do not assert this value

ste93cry:
I don't know what `uopz` extension does, but I never heard of it. Maybe we can remove it?

Jean85:
Since we are breaking BC, why not make something private? Or do we expect end users to extend and be able to touch everything?

tobias-nitsche:
@Jean85 my opinion to that:

protected functions are nice
protected properties are a mess

nokitakaze:
Private methods and properties are existed only for prohibition of extending class in the one way or another. You SHOULD NOT use "private" until you want to prohibit override some behaviour. Word "final" id used for the same reason.

Look at this 33b9a36 / #409

Jean85:
@nokitakaze I do not agree. Private does not prohibit extension (that's what `final` is for, as you are saying), it just reduces visibility.

If a method or a property is not needed outside of the current class, it should be private; it always can be upgraded to protected later. Having something non-private means that is a part of the contract with the end user, so it cannot be changed without a BC. So IMHO reducing the size of the contract is in our interest, to have less constraints on future developments.

nokitakaze:
> is not needed outside

You don\'t know needed this property/method or not until you write extending class.

Jean85:
That can be a valid reason. But this means that we are taking the conscious decision of willingly support the BC for every property and method.

I'm fine with that, but it should be agreed upon.

tobias-nitsche:
@nokitakaze I disagree as well. You actually can modify properties of the parent class via set-Methods. Using this pattern, gives the parent class the possibiliy to control the input data. 

Without these protection, and can be hard to follow and observ changes of properties.

nokitakaze:
@tobias-nitsche this should be using only if you sanitize all input data in your setters

tobias-nitsche:
@nokitakaze To make it clear, protected functions are violating the open/closed principle https://en.wikipedia.org/wiki/Open/closed_principle

Maybe a sanitization is not yet needed, but changing variables directly in a child class, makes it hard to add this possibily in future. But it's not my decision.

nokitakaze:
@tobias-nitsche , We override methods because we want to override/modificate a behaviour. If we don\'t broke code contract in children we don\'t break [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle). So there is no any bad aftermath.

All we need to do is not to break code contract while overriding methods.

tobias-nitsche:
Sorry, my bad. I meant protected properties. Not functions. (I corrected it)

Jean85:
This should be completely redone on the top of 2.0 as it is now.

nokitakaze:
Or just closed

PR #459, opened 2017-05-18
Add option to bulk send, and to finish the request before bulk sending (in PHP-FPM)

pjcdawkins:
* Allow `store_errors_for_bulk_send` to be set in `$options`.
* Add an option to run [`fastcgi_finish_request()`](http://php.net/manual/en/function.fastcgi-finish-request.php) before sending events to Sentry.

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -39,7 +39,6 @@ class Raven_Client
      * @var array|null
      */
     public $severity_map;
-    public $store_errors_for_bulk_send = false;
 
     protected $error_handler;
     protected $error_types;
@@ -93,6 +92,8 @@ class Raven_Client
     public $curl_ssl_version;
     public $trust_x_forwarded_proto;
     public $mb_detect_order;
+    public $store_errors_for_bulk_send = false;
+    public $fastcgi_finish_request = false;
     /**
      * @var Raven_Processor[]
      */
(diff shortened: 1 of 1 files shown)


lib/Raven/Client.php
```diff
@@ -1359,6 +1362,11 @@ public function onShutdown()
         if (!defined('RAVEN_CLIENT_END_REACHED')) {
             define('RAVEN_CLIENT_END_REACHED', true);
         }
+        if ($this->fastcgi_finish_request
+          && !empty($this->_pending_events)
+          && function_exists('fastcgi_finish_request')) {
+          fastcgi_finish_request();
```
nokitakaze: FastCGI/backend control should not be placed in logging library. I'm afraid you should add your handler for register_shutdown_function by yourself. As I do https://github.com/nokitakaze/ascetcms/blob/concept/functions/ascetcmsengine.php#L518

stayallive:
Hey @pjcdawkins, first of all thanks for the PR! 

I have to agree with @nokitakaze that calling `fastcgi_finish_request()` (or doing other request flow interruptions for that matter) in the Sentry library is stepping over the purpose of this library.

In the 2.0 we can look at making this kind of behaviour possible to implement by allowing the events to be queued up and dispatched by the user after they called `fastcgi_finish_request()` (or the framework of choice does).

Let's put this on the list of things we would want to take a stab at in 2.0 👍

Jean85:
If I'm not mistaken, the bulk send at the end of the request is already implemented since 1.7.1, thanks to the client destructor: https://github.com/getsentry/sentry-php/blob/501ebd9f8327f92693383acaf4ea8c9778053c66/lib/Raven/Client.php#L220

stayallive:
I think this only closes any open curl resources, it does not collect all events and wait's for them to untill the destructor is called.

Jean85:
That is probably done with the `Client::storeErrorsForBulkSend` in the 2.0 branch.

PR #462, opened 2017-05-18
Refactor the client options into its own Configuration class

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

Continuing the refactoring of the client class this PR extracts the configuration options into its own class, as seen in other common libraries and in the Sentry Ruby SDK itself.


composer.json
```diff
@@ -11,17 +11,19 @@
...
-    },
     "require": {
         "php": ">=5.5",
         "ext-hash": "*",
         "ext-json": "*",
         "ext-mbstring": "*",
-        "ext-curl": "*"
+        "ext-curl": "*",
+        "composer/ca-bundle": "~1.0",
+        "symfony/options-resolver": "~2.8|~3.0"
```
Jean85: I would suggest `"~2.7|~3.0"`, since the 2.7 release is in active support state for an other year.

composer.json
```diff
@@ -11,17 +11,19 @@
...
+    },
+    "require-dev": {
+        "friendsofphp/php-cs-fixer": "~2.1",
+        "phpunit/phpunit": "^4.8 || ^5.0",
```
Jean85: If the minimal version of PHP will be bumped to 5.6, this should be changed to `^5.4.3 || ^6.0`

composer.json
```diff
@@ -11,17 +11,19 @@
...
+        "monolog/monolog": "*"
```
Jean85: Do not use `*` as a "constraint", it's not. `~1.0` is fine.

composer.json
```diff
@@ -11,17 +11,19 @@
...
     },
     "suggest": {
         "monolog/monolog": "Automatically capture Monolog events as breadcrumbs"
```
Jean85: Since now monolog is required, we should remove it from the suggestions. Should we require it as an hard dependency, BTW?

lib/Raven/Client.php
```diff
@@ -550,28 +538,25 @@ public static function getDefaultProcessors()
...
+        foreach ($this->config->getProcessors() as $processor) {
+            /**  @var Processor $processorInstance */
+            $processorInstance = new $processor($this);
 
             if (isset($options['processorOptions']) && is_array($options['processorOptions'])) {
-                if (isset($options['processorOptions'][$processor])
-                    && method_exists($processor, 'setProcessorOptions')
-                ) {
-                    $new_processor->setProcessorOptions($options['processorOptions'][$processor]);
+                if (isset($options['processorOptions'][$processor]) && method_exists($processor, 'setProcessorOptions')) {
```
Jean85: Shouldn't we fail hard if we try to set options on a processor that doesn't have the `setProcessorOptions` method? Also, this check could be done with an interface..

composer.json
ste93cry: Indeed, I just moved the whole block of code after the `require` section. The scope of this PR is not to change these dependencies

composer.json
ste93cry: I'm not sure Monolog is required, as I said above this whole block has only been moved around the file, I did not add anything to the `require-dev` section

lib/Raven/Client.php
ste93cry: >this check could be done with an interface

This is indeed one of the things I'm working on, I introduced a `ProcessorInterface` interface and this whole function will be refactored to support instantiation of the classes through a factory, but it's not the scope of this PR. The scope of this PR is to just extract the client options to their own class

composer.json
Jean85: Then I would advise against moving this block of code, or you will cause a lot of conflicts for no apparent reason. If you want, I can work on a PR to address all this issues with dependencies.

ste93cry: >I would advise against moving this block of code

I agree with you, this change should not be here. Thank you pointing it out

>I can work on a PR to address all this issues with dependencies.

It would be great, but I'm not sure this is the proper time to do it, because I have a lot of other changes to do on features that will drop the dependencies or will to update them, so it could be that you just lose your time fixing something that soon or later could not even be there anymore

Jean85: As you prefer! If you need help, I'm here

composer.json
```diff
@@ -11,17 +11,19 @@
             "email": "dcramer@gmail.com"
         }
     ],
-    "require-dev": {
```
nokitakaze: You should not move this block >_<

lib/Raven/Client.php
```diff
@@ -233,35 +238,49 @@ class Client
...
+        if (static::is_http_request() && isset($_SERVER['PATH_INFO'])) {
+            $this->transaction->push($_SERVER['PATH_INFO']);
         }
 
-        if (!empty($dsn)) {
-            $options = array_merge($options, self::parseDSN($dsn));
-        }
-        unset($dsn);
-        $this->init_with_options($options);
+        /*$this->init_with_options($options);
```
nokitakaze: You should delete dead code instead of commenting it out.

composer.json
nokitakaze: I don\'t think @dcramer will approve this

lib/Raven/Client.php
```diff
@@ -233,35 +238,49 @@ class Client
...
+     * @var Configuration The client configuration
+     */
+    protected $config;
+
+    /**
+     * Constructor.
+     *
+     * @param Configuration $config The client configuration
+     */
+    public function __construct(Configuration $config)
```
nokitakaze: No. Never ever. Just no. 

Create method
```php
/**
 * @return Client
 */
function getClient(Configuration $config) {
    // @todo magic...

   return $object;
}
```

Constuctor must be as simple as possible. Only `new \Raven\Client(['key' => 'value']);`

tests/ConfigurationTest.php
```diff
@@ -0,0 +1,117 @@
...
+
+        $configuration = new Configuration([
+            'should_capture' => function ($value) {
+                return false;
+            },
+        ]);
+
+        $this->assertFalse($configuration->shouldCapture('foo'));
+    }
+}
```
nokitakaze: Crlf?

lib/Raven/Configuration.php
```diff
@@ -0,0 +1,446 @@
...
+use Composer\CaBundle\CaBundle;
+use Symfony\Component\OptionsResolver\Options;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * Configuration container for the Sentry client.
+ *
+ * @author Stefano Arlandini <sarlandini@alice.it>
+ */
+class Configuration
```
nokitakaze: Why do we need separate class, storing configuration options?

lib/Raven/Client.php
ste93cry: @nokitakaze I can not say it everywhere you commented, but please read the PR description: this PR is far from being finished, so any review at this point is useless

lib/Raven/Configuration.php
ste93cry: Because

- The ruby SDK has it, so we are matching that
- Having a separate class for the options means that we don't have to have a monolithic class (the `Client` class) that stores evyerthing
- Processors (and probably a lot of new features) require or will require access to the configuration of the client. Passing a neutral class instead of the `Client` instance means that there won't be anymore a circular reference between them and the client itself
- It's common in other libraries to have such a class to store the configuration of a client

nokitakaze: > we don't need to have a monolithic class

And it\'s bad because...

> Processors require...

And they have access.

> It's common in other libraries

This is a logging library not a nuclear station\'s center.

ste93cry: > And they have access.

They have access to the client instance, which in turn owns them. This is not necessarily bad (at least in not all cases), but it creates a circular reference. If we can avoid that, we should

>This is a logging library not a nuclear station's center.

Neither the other libraries, but they do it anyway because this way the code is better organized

nokitakaze:
Okay. If you want to refactor entire project just do all properties protected and we can close PR #457

ste93cry:
> just do all properties protected

It's not so easy, this is why this PR requires time to get it working. Having all public means that even a single change break everything else that depends on it. Also, #457 is still valid, it's just that if this PR gets merged you won't have to work on the variables related to the configuration of the client

nokitakaze:
This PR and #457 are conflicting

lib/Raven/Client.php
ste93cry: No, the client depends on the configuration to work, so you will need to instantiate that class first. I don't see any problem with this, I'm thinking to introduce a builder class for the client or a static method like your example to create the instance of the client given a set of options

nokitakaze: > the client depends on the configuration to work

Yes. And you can create it in constructor
```php
function __construct(array $options = []) {
    $configuration = new Configuration($options);
    // @todo ...
}
```

instead of demanding do it from our users

ste93cry:
It's conflicting only about the variables of the options, everything else is ok

lib/Raven/Client.php
ste93cry: Yes, I could create inside the constructor, but this way we can

- Have a builder class that creates the client, given the amount of options we have
- We **make explicit** that the `Client` class **depends** from the `Configuration` class
- We could have a static method (something like `createClient`) that will accept the options as an array and just call the constructor
- Users can extend the `Configuration` class if they want without the need to subclass the `Client` class, or they can make changes to the configuration before instantiating the client by passing the configuration to some code outside our control (arrays are passed by value, objects not)

Plus:

- The Ruby SDK is done in the exact same way, and noone complains
- There are a lot of pros, the only cons until now is that the programmer must add a new line to create the configuration object before instantiating the client. Not so hard, huh?

nokitakaze:
@ste93cry if you will do "sending_allowed" (#458) please add it to method "capture" and remove a predicate from "send"

Just as I desribed in https://github.com/getsentry/sentry-php/pull/458#issuecomment-302398846

Jean85:
I do agree with the approach of this PR. Having a dedicated, immutable object for the config that can be passed around without having to worry about side effects is the best approach.

nokitakaze:
Oh, lord. This class is immutable. And how I supposed to change a behaviour after creating Raven Client?

And do not tell me this is not necessary. A instance can do install. And after that **I must be able to change any settings** (message limit, release, tags, global tags, error_types...) of some instance which waits for some events (i.e. shutdown). Maybe I will want to set new filters for incoming events (i.e. exceptions)

Jean85:
Why should you change those stuff inside the same request? What scenario do you imagine where such this should be needed?

Anyhow, you can always re-set the whole object if you want. Immutability is in place just to prevent that, if you reuse the same config in two places, you can inadvertently make modifications in two places with one change.

nokitakaze:
> Why should you change those stuff inside the same request? 

There is no "request". There is only code flow. Do not forget PHP can do a job for hours.
I may want to add some tags or another after some specific events. And then I may want to change release or environment. There is so many scenarios. 

Ideal library should be as flexible as it can.

Jean85:
Understood. I still believe that an object is the way to go: it makes more explicit what options are supported, and it make code less error prone.

ste93cry:
>Oh, lord. This class is immutable

This class was never meant to be immutable. Setters were missing just because code was not ready yet and I had to fix a huge amount of unit tests before thinking to add setters methods to change the options on-the-fly. If you had read the description you would have seen that it was written that the job is not over...

Jean85:
Ok so my mistake, I thought it was immutable. Then I think we should all agree.
The only thing that I would suggest would be to clone the object on assignation, to avoid errors like the one that I suggested before:

> ... if you reuse the same config in two places, you can inadvertently make modifications in two places with one change.

lib/Raven/Client.php
dcramer: I think the pattern of passing the args (vs another class) is fine here. Ruby is done via blocks so its not super relevant here (you dont need to intantiate a configuration object), and you can still pass the configuration to the 'Configuration' class constructor for validation/etc. The problem is that any native integration now nees to not only instantiate the Client class  (which is easy), but also pull in the config class, coerce config, and then pass that in.


e.g.

```
# pretend-framework.yml
exceptionHandler: \Sentry\Client
exceptionHandlerArgs: ...
```

If we're REALLY concerned, we can simply support both. If it's a string, its the DSN, if its Configuration [...], if its an array, its passed into Configuration.

lib/Raven/Configuration.php
dcramer: I am a fan of the configuration class FWIW. This is a nice thinga bout the Ruby SDK that I wish we would refactor in Python. It just makes it more obvious where config lives, and can then easily be exposed from the base client (vs having to know which attributes are configuration or state)

nokitakaze:
Please don't forget to edit README

lib/Raven/Client.php
ste93cry: I actually pass the configuration class as an argument because it makes possible to extend it if any users wants to do it. If I create the instance inside the constructor then this won't be possible. Plus, my idea is to introduce a builder class so that users won't have the need to directly call `new Client(new Configuration())`. This is an example of what I imagine:

```php
$options = ['foo' => 'bar'];
$client = ClientBuilder::create($options)
    ->setTransportFactory(...)
    ->setLogger(...)
    ->setWhatheverCouldBeUseful(...)
    ->getClient();
```

I see that at this stage having to instantiate the Configuration class instead of directly passing the array of options feels bad, but we will anyway need to have a constructor that accepts more arguments as we expand the `Client` class to accept more arguments like the factories to create processors and transports. And this is not bad at all, because it means that users will have exact extensions points to extend the behaviour of Sentry without having to subclass the client for nearly everything they have to do now. Having a strict set of API is only a good thing, not a limitation. And finally, any framework will anyway abstract away how they create the client, so users will not need to interact with the arguments passed to the constructor. For libraries users, this is the worl and I never heard someone complaining because they had to pass an object instance instead of an array

dcramer: To be frank, this just looks like overly complex java code. There's no reason for class factories and they just provide complexity that I'm having a hard time seeing the need for.

e.g. how is that any better than $client = new Client(...)->setX();

nokitakaze: > to extend it 

Extend what? Config of existing Client? Or configuration for new uses?

ste93cry: >There's no reason for class factories

Class factories allows users to choose how an object is created, instead of relying on an internal behaviour of the class that's going to use the instances. For example, the processors are now simply created by doing a `new $processorClass(...)` and this is a big limit because if someone wants to have custom parameters in the constructor, access other framework services or whatever else he can't, at least without subclassing the `createProcessors` method. Instead, with a factory we could support a simple factory that just instantiates the processor, and we could support more complex factories in the framework-related packages that for example uses the service container to instantiate the processors. The same goes for the transport (and this is just an example). I understand that you want to keep the library as simple as possible, but introducing factories and refactoring the code in this way imho is not making it overly complex, is just about making it more extensible and better organized.

>e.g. how is that any better than $client = new Client(...)->setX();

It's better because with factories (and this is just an example) we can lazy-load the transport and processors. It's better because we won't restrict anymore how a given transport/processor must be created because we will shift this to a dedicated object that can be customized by the user to support dependency injection. I really can only see advantages in doing this. Let's be honest, who is going to use this library is a programmer, and I highly doubt that he will complain if he has to pass more arguments to the constructor if it means that he has more flexibility

nokitakaze: > we can lazy-load the transport and processors

Why do you need to keep transport empty at the creating of Client? Do you want to have ability to save a bulk locally?

Jean85: Because in some cases instantiating the transport could create a connection, that could be cumbersome for resources, or simply being too early, could timeout, or even be not needed since no event will be fired to Sentry.

nokitakaze: > that could be cumbersome for resources

So at first store events locally and then push them in separate script

Jean85: That would add a new possible point of failure, and requires new, obligatory, configuration

ste93cry:
I was thinking about a potential issue with the options related to the DSN. Currently, the following options are exposed to be configured by the user (it's the same in the Ruby SDK):

- `server`
- `scheme`
- `host`
- `path`
- `port`
- `project_id`
- `project_root`
- `secret_key`
- `public_key`

Let's suppose that the user provides the `server` option. In this case, all other options (if not configured individually) will be overridden with values extracted from the first one and this is fine. However, what should happen when those options are specified by the user? The problem arises here: the `server` option won't be in sync anymore with the other options creating confusion and possible problems when the user retrieves the configuration and expects the options to all represent the same DSN while instead it's a mess. When creating a project in Sentry the user is always provided a full DSN string, not the single parts that compose it. My proposal is to remove all options except the first one from being configurable. You can still get the values through getters, but you won't be able anymore to edit them. The only way to change them is to specify a different DSN. This way there will be only one source of truth for the options related to the server endpoint. @dcramer What do you think about this?

nokitakaze:
We could do one of this set ([server] and [scheme, host, path,..]) virtual. And one of them will be made from another. So they always will be consistent

ste93cry:
>We could do one of this set ([server] and [scheme, host, path,..]) virtual. And one of them will be made from another

I honestly don't understand what you mean... By the way, the DSN has all the information to configure the port, host, etc so I don't see any good reason to expose setters for options except the DSN string itself

ste93cry:
>We could do one of this set ([server] and [scheme, host, path,..]) virtual. And one of them will be made from another

I honestly don't understand what you mean... By the way, the DSN has all the information to configure the port, host, etc so I don't see any good reason to expose setters for options except the DSN string itself...

dcramer:
Agreed: we don't need setters for the individual bits. You should just bind a DSN/rebind the whole DSN upon chaneg.

ste93cry:
Does it make sense to keep getters for the individual parts of the DSN? I can't find any good reason for them too

nokitakaze:
Yes. I think we can safely remove getters and setters for parts of DSN. Only getter/setter for DSN itself is important

Jean85:
If splitted parts of DSN are not needed even for Curl calls, I agree too.

ste93cry:
Anyone knows what the `site` option does? I can't find its corresponding option in the Ruby SDK

nokitakaze:
default

dcramer:
@nokitakaze its deprecated/legacy

nokitakaze:
https://github.com/getsentry/sentry-php/blob/master/lib/Raven/Client.php#L77
Is this legacy? Should we delete ie or replace by some new field?

dcramer:
@nokitakaze we haven't removed it from SDKs, but most wont send it if its not configured

nokitakaze:
@ste93cry how much time are you planning to spent until this PR will work?

ste93cry:
I'm currently working on a prototype of the client builder class, so that users don't have to instantiate the client by themselves. At the moment it will be pretty useless apart from the ability to set the configuration options, but the idea is in the future to add methods to set the factories for the various things like the serializer, the processors, etc as discussed elsewhere. I plan to get it working for the end of the week, as I'm currently very busy with my work

ste93cry:
I've added a `ClientBuilder` class to ease the construction of a client. It's pretty empty right now, but it's the base for new features as said above. All it does right now is to just forward existing methods calls to the underlying configuration object. Since the `site` option is deprecated, I removed it from the list of options and it's not sent anymore to Sentry. A few remaining questions:

- There is a `message_limit` option that was added in #182, but its doc says that _it's not guaranteed that length of whole message will be restricted by this value_. Does it makes sense then to let's it be configurable?
- After rebasing on top of 2.0 branch, I noticed that in the `composer.json` the `minimum-stability` is now set to `dev`: this way all dependencies are installed from the `dev` branch and this is dangerous to say the least
- As I'm now using the [composer/ca-bundle](https://github.com/composer/ca-bundle) package to find where the `cacert.pem` file is located should we remove it from the `data` folder?. Using directly the fallback file from ca-bundle means that we don't have to keep it up to date anymore as it won't be our problem.

nokitakaze:
> should we remove it from the data folder?

Yes

Jean85:
Good job on the Client builder.
I am against the `"minimum-stability": "dev"` too.

UPGRADE-2.0.md
```diff
@@ -0,0 +1,222 @@
...
+  After:
+
+  ```php
+  public function __construct(Configuration $config)
+  {
+      // ...
+  }
+  ```
+
+- The methods `Client::getRelease` and `Client::setRelease` have been removed.
```
Jean85: Can we make all of this less verbose with a "Before -> after" table? It seems to me that they are all alike.

lib/Raven/Client.php
```diff
@@ -916,22 +570,19 @@ protected function get_user_data()
         ];
     }
 
-    protected function get_extra_data()
-    {
-        return $this->extra_data;
-    }
-
     public function get_default_data()
```
Jean85: This method name is not camelCased: do we agree on following PSR-2?

lib/Raven/Client.php
```diff
@@ -231,76 +110,67 @@ class Client
...
+     */
+    protected $config;
 
-    public function __construct($options_or_dsn = null, $options = [])
+    /**
+     * Constructor.
+     *
+     * @param Configuration $config The client configuration
+     */
+    public function __construct(Configuration $config)
```
nokitakaze: I don\'t approve it. Our users must be use as simple way to construct client as possible. This is a logging library, not a nuclear power station

lib/Raven/Client.php
ste93cry: I agree, but I did not change anything that was not tied to the purpose of this PR, which was moving the configuration properties to its own class

lib/Raven/Client.php
Jean85: Why do you think that his approach is harder? 

Have you seen how much code he was able to delete thanks to this new approach? 
Did you see the `ClientBuilder` class that will help in easily instantiating a new client?

nokitakaze: > how much code he was able to delete

This means nothing. I don\'t see any reason not to add a few lines of code to `Client::__construct` with calling `ClientBuilder::create([...])->getClient();`

```php
if (is_array($option)) {
    return ClientBuilder::create($option)->getClient();
}
```

Jean85: That would make sense if we wanted to preserve BC; but since we are breaking it in a million other ways, it will just clutter the codebase with two different ways of passing options through. And the array-way will not have all the safeties that an object will have, with typehints, getters, absent setters when dangerous etc....

ste93cry: Again, the constructor of the client will necessarily need to accept new arguments in the near future, when factories will be introduced, so it will not stay so simple anyway. There will be no way the constructor will stay as is

nokitakaze: > will necessarily need to accept new arguments in the near future

**WAT?!**

UPGRADE-2.0.md
ste93cry: They are indeed all alike, but the advantage of this is that the file explains exactly what has changed and how to upgrade every single piece of code. As this is not the CHANGELOG file, I think it should be verbose

nokitakaze:
> I am against the `"minimum-stability": "dev"` too

Add this patch and combine/squize the PR
```patch
From 901f86c03202148a2f5fb584e0716a1a1112d890 Mon Sep 17 00:00:00 2001
From: Nokita Kaze <admin@kanaria.ru>
Date: Mon, 5 Jun 2017 20:19:02 +0300
Subject: [PATCH] Change dependency

---
 composer.json | 1 +
 1 file changed, 1 insertion(+)

diff --git a/composer.json b/composer.json
index 4ae3944..095c024 100644
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,7 @@
         }
     ],
     "minimum-stability": "dev",
+    "prefer-stable": true,
     "require-dev": {
         "friendsofphp/php-cs-fixer": "~2.1",
-- 
2.7.1.windows.2
```
default

ste93cry:
We're a bit off-topic... Anyway, using `"prefer-stable": true` should indeed work, but I still don't understand why there is the `minimum-stability` set to `dev`. There is no need to do that, if you have a library that doesn't have a stable tag then use as constraint `library@dev`

Jean85:
Can we please go ahead with this? It's a huge blocker for everything else (starting with #472 #473 ).

nokitakaze:
I said my word
gwbmloy 1

Jean85:
So you are opposing this whole PR just because we are requiring the config in the constructor?
It only makes sense, the client cannot work without it and it's a failsafe to avoid initialized clients with messed up options.

stayallive:
I like this approach moving the configuration options to it's own class. 

IMO this adds clarity when you see some configuration value retrieved you know where it's coming from instead of the old way where it was retrieved from a client accessor directly, made it less opaque it is a configurable value.

nokitakaze:
Okay, go to h..l

I said again that @dcramer said "arrays is fine here" https://github.com/getsentry/sentry-php/pull/462#discussion_r117855927

dcramer:
@nokitakaze I don't like playing babysitter, but we're either going to be friendly here or I'm going to start removing people from the project.

ste93cry:
@dcramer Before merging I would like to know your opinion on this point:

>There is a message_limit option that was added in #182, but its doc says that it's not guaranteed that length of whole message will be restricted by this value. Does it makes sense then to let it be configurable?

From what I can see from other SDKs there doesn't seem to be such an option

dcramer:
@ste93cry I dont think we need to provide it. The server enforces restrictions, and at some point we're going to build a simply proxy for people concerned about doing it client side.

PR #463, opened 2017-05-18
Delete uopz

nokitakaze:
If this thing was needed when 6123643 , now we **do not** need it.


ste93cry:
Just for future reference, quoting from PHP docs:

>The uopz - User Operations for Zend - extension exposes Zend Engine functionality normally used at compilation and execution time in order to allow modification of the internal structures that represent PHP code, and for user code to interact with the VM.

I don't see any valid reason we need this here, so I approve the change

nokitakaze:
Maybe this was the time when nightly PHP didn\'t build without it

nokitakaze:
@ste93cry please approve this one and then merge it. So we can test code again

remicollet:
> I don't see any valid reason we need this here, maybe @remicollet can explain us why he added this extension to the build?

uopz is an optional PHPUnit which is required in some case (don't remember exactly which test were failing without) btw, everything seems ok now.

Jean85:
It's needed to create partial mocks that alter the structure of the mocked class. Probably the tests that used that are gone now (luckily!).

PR #464, opened 2017-05-22
Increase minimum PHP version requirement to 5.3

### LINKED ISSUE #437: We don't have PHP 5.2 support
We already don't have PHP 5.2 support

I wanted to fix some error what I made but then I found this
default
So no one can do "composer tests".

gwbmloy 1

https://packagist.org/packages/friendsofphp/php-cs-fixer
```
requires php: ^5.3.6 
```

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.13/Symfony/CS/Fixer/PSR2/EofEndingFixer.php#L43

Jean85:
This fixes #437, and it starts suggesting the fork for users that may need to use this lib with PHP 5.2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,13 +17,14 @@
         "monolog/monolog": "*"
     },
     "require": {
-        "php": ">=5.2.4",
+        "php": "^5.3|^7.0",
         "ext-curl": "*"
     },
     "suggest": {
         "ext-hash": "*",
         "ext-json": "*",
         "ext-mbstring": "*",
+        "immobiliare/sentry-php": "Fork that fixes support for PHP 5.2",
         "monolog/monolog": "Automatically capture Monolog events as breadcrumbs"
     },
     "conflict": {


nokitakaze:
This can not be resolved like this. @dcramer should change PHP version in every commit that was marked by tag and appeared in [Packagist](https://packagist.org/packages/sentry/sentry). And we have to find commit where compatibility with PHP 5.2 was dropped.

Jean85:
This is to solve the issue now, immediately. Your approach would mean rewrite a whole lot of Git history; isn't that a little heavy handed?

nokitakaze:
> This is to solve the issue now, immediately

This PR is not solving anything. People install Sentry via composer so infected commits already in Packagist.

> isn't that a little heavy handed

I think we could just forget about ~~zombies~~ PHP 5.2

Jean85:
Yeah, but I just want to state the truth about this lib in Packagist, and suggest a possible solution to affected users. This IMHO is the best "fire and forget" solution about 5.2.

dcramer:
I am 100% not rewriting git history. This seems like a fine solution to me. If we're genuinely not functional on 5.2, this changes the metadata to say that, so future releases will be correct.

nokitakaze:
https://getcomposer.org/doc/00-intro.md
> Composer requires PHP 5.3.2+

Lol. There is no way to install Sentry via composer in PHP 5.2

Jean85:
Thanks David.

@nokitakaze I think older versions of Composer support PHP 5.2, that sentence should be true for the 1.x, stable release.

PR #449, opened 2017-04-26
Scrutinizer support

nokitakaze:
As I said before in #442 we need Scrutinizer to see our **code coverage** and code quality.

After that PR the maintainer have to add branch 2.0 to "Tracked Branches"

PR #465, opened 2017-05-23
Improve Travis build

Jean85:
* remov uopz, like #463, it's useless!
 * stop allowing failures on 7.1, it's stable!
 * `--prefer-dist` with composer, it's faster!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: php
+sudo: false
 
 php:
   - 5.3
@@ -8,7 +9,10 @@ php:
   - 7.0
   - 7.1
   - nightly
-  - hhvm-3.12
+env:
+  - REMOVE_XDEBUG="0"
+  - REMOVE_XDEBUG="1"
+
 
 matrix:
   allow_failures:
@@ -16,32 +20,9 @@ matrix:
     - php: nightly
   fast_finish: true
   include:
-    - php: 7.0
-      env: REMOVE_XDEBUG="0" DISABLE_ASSERTIONS="1"
-    - php: 7.0
-      env: REMOVE_XDEBUG="1" DISABLE_ASSERTIONS="1"
-    - php: 7.1
-      env: REMOVE_XDEBUG="0" DISABLE_ASSERTIONS="1"
-    - php: 7.1
-      env: REMOVE_XDEBUG="1" DISABLE_ASSERTIONS="1"
-    - php: nightly
-      env: REMOVE_XDEBUG="0" DISABLE_ASSERTIONS="1"
     - php: hhvm-3.12
       env: REMOVE_XDEBUG="0" HHVM="1"
       dist: trusty
-  exclude:
-    - php: hhvm-3.12
-      env: REMOVE_XDEBUG="0"
-    - php: hhvm-3.12
-      env: REMOVE_XDEBUG="1"
-    - php: nightly
-      env: REMOVE_XDEBUG="1"
-
-env:
-  - REMOVE_XDEBUG="0"
-  - REMOVE_XDEBUG="1"
-
-sudo: false
 
 cache:
   directories:
@@ -49,7 +30,6 @@ cache:
 
 before_install:
   - if [ "$REMOVE_XDEBUG" = "1" ]; then phpenv config-rm xdebug.ini; fi
-  - if [ "$DISABLE_ASSERTIONS" = "1" ]; then echo 'zend.assertions = -1' >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi
   - composer self-update
 
 install: travis_retry composer install --no-interaction --prefer-dist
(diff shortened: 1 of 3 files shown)


nokitakaze:
> -prefer-dist with composer

There was some reason to do it. I don\'t know why composer downloads sources

.travis.yml
```diff
@@ -12,7 +12,6 @@ php:
 
 matrix:
   allow_failures:
-    - php: 7.1
```
nokitakaze: It is very discussable. In new version we have support from the box.

If you want to do it please change this lines as well

https://github.com/getsentry/sentry-php/blob/master/.scrutinizer.yml#L8 change to "7"
https://github.com/getsentry/sentry-php/blob/master/.travis.yml#L70 add new line

https://github.com/getsentry/sentry-php/pull/449#discussion_r113763595

.travis.yml
```diff
@@ -51,11 +50,9 @@ cache:
 before_install:
   - if [ "$REMOVE_XDEBUG" = "1" ]; then phpenv config-rm xdebug.ini; fi
   - if [ "$DISABLE_ASSERTIONS" = "1" ]; then echo 'zend.assertions = -1' >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi
-  - if [[ ${TRAVIS_PHP_VERSION:0:1} > "5" ]] && [ "$HHVM" != "1" ]; then pecl install uopz; fi
-  - if [[ ${TRAVIS_PHP_VERSION:0:1} > "5" ]] && [ "$HHVM" != "1" ]; then echo "extension=uopz.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi
   - composer self-update
 
-install: travis_retry composer install --no-interaction --prefer-source
```
nokitakaze: There was some reason to do it. I don't know why composer downloads sources

ste93cry: The reason was that GitHub had rate limit on the API used to fetch the releases, but this limits have been removed and we can now safely use `--prefer-dist`

.travis.yml
Jean85: Sorry, I missed that because I branched from a previous commit. I was going to fix that, but a single row addition will not be enough: we have 8 total 7.x builds, since we have the `DISABLE_ASSERTIONS` ones... are those still needed? Can we remove them?

.travis.yml
nokitakaze: I think `prefer-dist` used by default so we could just remove this param

.travis.yml
nokitakaze: > we have 8

Yes. I missed as well. 8 instead of 7.

Jean85: Ok I will fix this. 

But can you explain to me the usefulness of this `DISABLE_ASSERTIONS`? I cannot find it referenced anywhere in the source...

nokitakaze: I don\'t remember. I just remember some tests fails on PHP 7.0 PRODUCTION, but works fine on PHP 7.0 DEBUG. So we should have this matrix

Jean85: I've found what it does: https://secure.php.net/manual/en/ini.core.php#ini.zend.assertions

I've also found the only remaining usage of that: https://github.com/getsentry/sentry-php/blob/d34ab2569aa4841230f6125585c3fe426652137f/test/Raven/Tests/ClientTest.php#L13

I'll address this in a commit, so we can get rid of those additional build steps

Jean85:
I've rebased the branch on master so I could have the modifications from #449.
I've removed the last `assert()` usage left, so we can get rid of the additional build jobs due to the usage of `DISABLE_ASSERTIONS`. I've also simplified the HHVM job definition.

.travis.yml
```diff
@@ -8,54 +9,30 @@ php:
   - 7.0
   - 7.1
   - nightly
-  - hhvm-3.12
+env:
+  - REMOVE_XDEBUG="0"
+  - REMOVE_XDEBUG="1"
+
```
nokitakaze: Please delete this empty line

nokitakaze:
> into `master` from `improve_travis_build`

Could you use your own repository in the future? Scrutinizer waits for certain amount of commits and you doubled it if you do PR this way

ste93cry:
Please squash the commits together before merging

test/Raven/Tests/ClientTest.php
```diff
@@ -664,20 +664,14 @@ public function testCaptureExceptionHandlesExcludeOption()
...
-        }
         $client = new Dummy_Raven_Client();
         try {
             invalid_encoding();
         } catch (Exception $ex) {
             $client->captureException($ex);
         }
         $events = $client->getSentEvents();
-        $this->assertEquals(1, count($events));
+        $this->assertCount(1, $events);
```
nokitakaze: There was no reason to do it. We have a lot of $this->assertEquals(1, count($events)); we should fix all of them in one step

Jean85:
> Could you use your own repository in the future? Scrutinizer waits for certain amount of commits and you double it if you do a pull request this way

If this is an issue, we just have to disable Travis builds for branches.
I'm doing it now, tell me if you have anything against it.

PR #466, opened 2017-05-23
Renews from master at 2017-05-23

nokitakaze:
Renews for #424.
- Main update is scrutinizer from PR #449

### DIFF
--- a/lib/Raven/Processor/SanitizeDataProcessor.php
+++ b/lib/Raven/Processor/SanitizeDataProcessor.php
@@ -2,6 +2,8 @@
 
 /*
  * This file is part of Raven.
+ * Asterisk out passwords from password fields in frames, http,
+ * and basic extra data.
  *
  * (c) Sentry Team
  *
@@ -14,12 +16,6 @@ namespace Raven\Processor;
 use Raven\Client;
 use Raven\Processor;
 
-/**
- * Asterisk out passwords from password fields in frames, http,
- * and basic extra data.
- *
- * @package raven
- */
 class SanitizeDataProcessor extends Processor
 {
     const MASK = self::STRING_MASK;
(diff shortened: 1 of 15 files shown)


ste93cry:
>I forgot about PHP 7.1 

This is not a commit message...

I would merge all the commit messages into only one and then squash everything. This way the history will look cleaner

nokitakaze:
> squash everything

No. I did it on purpose. I want to keep entire log of commits

lib/Raven/Processor/SanitizeDataProcessor.php
```diff
@@ -4,6 +4,7 @@
  * This file is part of Raven.
  *
  * (c) Sentry Team
+<<<<<<< HEAD:lib/Raven/Processor/SanitizeDataProcessor.php
```
nokitakaze: Oooops

nokitakaze:
Do I need to add PR #465 to this PR before merging?

Jean85:
@nokitakaze if you want you can do it, merging the whole `improve_travis_build` branch.
Otherwise I will do it later.

ste93cry:
>No. I did it on purpose. I want to keep entire log of commits. There was a lot of guys who did all this work. It's not about me or you. All commits should be keep in history of this project.

This is why I said to merge the messages from all commits into a single big description of a single big commit. This way you won't lose the history. But keeping merging thousands (and this is just a random number) between branches will only keep growing the history without any real benefit. I already know that those commits are on master, I don't need to have a copy of them also on `2.x`. Our current Git history is a damn mess

nokitakaze:
There will be day when @dcramer does
```bash
git branch master
git merge 2.0 -Xrecursive theirs
```

I don\'t want to lose history at that day.

> Our current Git history is a damn mess

Being big project is so troublesome :3

PR #467, opened 2017-05-23
[2.0] Correct spacing

nokitakaze:
I know this is lack of objective criteria. If we don't need it, close it


Jean85:
I'm personally of the opposite preference, I normally use spaces around the concat operator; also, since this is changing the status quo, I don't think it's useful.

PR #470, opened 2017-05-31
Add basic release instructions

dcramer:
Let me know if you have any thoughts. Aside if there were ways to automate/simplify this that'd be amazing. We've talked internally about building a tool which did this cross repo/language, since the only steps that really change are the ``composer`` bits (the hardcoded version can often be read from another file on disk).

We'll likely take these docs and make these SDK standards of development as we move more towards those.

Aside, maybe we should say the process involves making a PR? @mattrobenolt if you have thoughts feel free to chime in since you manage the getsentry/sentry process.

@getsentry/php @nokitakaze @Jean85 @ste93cry

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,3 +74,75 @@ Tests can then be run via phpunit:
 ```
 $ vendor/bin/phpunit
 ```
+
+
+Tagging a Release
+-----------------
+
+1. Make sure ``CHANGES`` is up to date (add the release date) and ``master`` is green.
+
+2. Create a new branch for the minor version (if not present):
+
+```
+$ git checkout -b releases/1.7.x
+```
+
+3. Update the hardcoded version tag in ``Client.php``:
+
+```
+class Raven_Client
+{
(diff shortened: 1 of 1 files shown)


bretthoerner:
I like this, but I agree it'd be great to automate. For an example of what we do with the Java SDK (though this leans heavily on the `maven-release` plugin): https://github.com/getsentry/raven-java/blob/master/Makefile#L31-L65

It comes down to `make prepare` and `make perform`, the user is prompted for the new version (it even auto populates a good guess), it pushes to Maven Central, and then the next development (`SNAPSHOT`) version is created.

I feel like if we had a script that could discover the old version, ask for the next version, do some `sed` to update it in the repo, and then like you said we might need a custom-script-per-project that does the actual pushing (where necessary), we'd be in a good spot?

Jean85:
If it can be helpful, the version can be inferred from Composer using this: https://github.com/Ocramius/PackageVersions

PR #471, opened 2017-05-31
Code Coverage (for async & outer curl)

nokitakaze:
- Refactoring for async calls
- Curl (exec): CA-cert (minus 2 todo)
- Coverage if events really sending
- _Monolog fix for PHP 7_

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,13 +24,14 @@ env:
   - REMOVE_XDEBUG="0"
   - REMOVE_XDEBUG="1"
 
-sudo: false
+sudo: required
 
 cache:
   directories:
     - $HOME/.composer/cache
 
 before_install:
+  - sudo apt-get install --yes net-tools
   - if [ "$REMOVE_XDEBUG" = "1" ]; then phpenv config-rm xdebug.ini; fi
   - composer self-update
 
--- a/composer.json
+++ b/composer.json
@@ -11,10 +11,13 @@
             "email": "dcramer@gmail.com"
         }
     ],
+    "minimum-stability": "dev",
     "require-dev": {
         "friendsofphp/php-cs-fixer": "~2.1",
         "phpunit/phpunit": "^4.8 || ^5.0",
-        "monolog/monolog": "*"
+        "ext-openssl": "*",
+        "nokitakaze/testhttpserver": "^0.1.0",
+        "monolog/monolog": "~1.0"
     },
     "require": {
         "php": ">=5.5",
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -363,6 +363,7 @@ class Client
     {
         // Force close curl resource
         $this->close_curl_resource();
+        $this->force_send_async_curl_events();
     }
 
     /**
@@ -1231,20 +1232,27 @@ class Client
         }
     }
 
+    /**
+     * @param string $url
+     * @param string $data
+     * @param array  $headers
+     * @return string
+     *
+     * This command line ensures exec returns immediately while curl runs in the background
+     */
     protected function buildCurlCommand($url, $data, $headers)
     {
-        // TODO(dcramer): support ca_cert
-        $cmd = $this->curl_path.' -X POST ';
+        $post_fields = '';
         foreach ($headers as $key => $value) {
-            $cmd .= '-H ' . escapeshellarg($key.': '.$value). ' ';
+            $post_fields .= ' -H '.escapeshellarg($key.': '.$value);
         }
-        $cmd .= '-d ' . escapeshellarg($data) . ' ';
-        $cmd .= escapeshellarg($url) . ' ';
-        $cmd .= '-m 5 ';  // 5 second timeout for the whole process (connect + send)
-        if (!$this->verify_ssl) {
-            $cmd .= '-k ';
-        }
-        $cmd .= '> /dev/null 2>&1 &'; // ensure exec returns immediately while curl runs in the background
+        $cmd = sprintf(
+            '%s -X POST%s -d %s %s -m %d %s%s> /dev/null 2>&1 &',
+            escapeshellcmd($this->curl_path), $post_fields,
+            escapeshellarg($data), escapeshellarg($url), $this->timeout,
+            !$this->verify_ssl ? '-k ' : '',
+            !empty($this->ca_cert) ? '--cacert '.escapeshellarg($this->ca_cert).' ' : ''
+        );
 
         return $cmd;
     }
@@ -1301,9 +1309,10 @@ class Client
 
         $errno = curl_errno($this->_curl_instance);
         // CURLE_SSL_CACERT || CURLE_SSL_CACERT_BADFILE
-        if ((($errno == 60) || ($errno == 77)) && !is_null($ca_cert)) {
+        if (in_array($errno, [CURLE_SSL_CACERT, 77]) && !is_null($ca_cert)) {
             curl_setopt($this->_curl_instance, CURLOPT_CAINFO, $ca_cert);
             $buffer = curl_exec($this->_curl_instance);
+            $errno = curl_errno($this->_curl_instance);
         }
         if ($errno != 0) {
             $this->_lasterror = curl_error($this->_curl_instance);
@@ -1512,15 +1521,20 @@ class Client
         $this->user_context(array_merge($user, $data));
     }
 
+    public function force_send_async_curl_events()
+    {
+        if (!is_null($this->_curl_handler)) {
+            $this->_curl_handler->join();
+        }
+    }
+
     public function onShutdown()
     {
         if (!defined('RAVEN_CLIENT_END_REACHED')) {
             define('RAVEN_CLIENT_END_REACHED', true);
         }
         $this->sendUnsentErrors();
-        if ($this->curl_method == 'async') {
-            $this->_curl_handler->join();
-        }
+        $this->force_send_async_curl_events();
     }
 
     /**
--- a/lib/Raven/CurlHandler.php
+++ b/lib/Raven/CurlHandler.php
@@ -16,7 +16,6 @@ namespace Raven;
  * @package raven
  */
 
-// TODO(dcramer): handle ca_cert
 class CurlHandler
 {
     protected $join_timeout;
@@ -73,6 +72,9 @@ class CurlHandler
 
     public function join($timeout = null)
     {
+        if (count($this->requests) == 0) {
+            return;
+        }
         if (!isset($timeout)) {
             $timeout = $this->join_timeout;
         }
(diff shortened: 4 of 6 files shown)


nokitakaze:
Stop. Don\'t you want to ask about new package in composer or something?

composer.json
```diff
@@ -11,10 +11,12 @@
             "email": "dcramer@gmail.com"
         }
     ],
+    "minimum-stability": "dev",
     "require-dev": {
         "friendsofphp/php-cs-fixer": "~2.1",
         "phpunit/phpunit": "^4.8 || ^5.0",
-        "monolog/monolog": "*"
+        "nokitakaze/testhttpserver": "^0.1.0",
+        "monolog/monolog": "~1.0"
```
nokitakaze: Monolog 2.0 is under development and it breaks his BC

dcramer:
@nokitakaze i saw it was for testing so seems fine. Do you have concerns w/ it?

composer.json
dcramer: oh actually i didnt really pay attention to this bit

lib/Raven/Client.php
```diff
@@ -1512,15 +1520,20 @@ public function set_user_data($id, $email = null, $data = [])
...
+        }
+    }
+
     public function onShutdown()
     {
         if (!defined('RAVEN_CLIENT_END_REACHED')) {
             define('RAVEN_CLIENT_END_REACHED', true);
         }
         $this->sendUnsentErrors();
-        if ($this->curl_method == 'async') {
```
nokitakaze: You can\'t do this. The type of synchronization can change at any time.

lib/Raven/Client.php
```diff
@@ -363,6 +363,7 @@ public function __destruct()
     {
         // Force close curl resource
         $this->close_curl_resource();
+        $this->force_send_async_curl_events();
```
nokitakaze: We need to send all events before dying. I saw this problem while I was writing this PR.

lib/Raven/Client.php
```diff
@@ -1301,7 +1309,7 @@ protected function send_http_synchronous($url, $data, $headers)
 
         $errno = curl_errno($this->_curl_instance);
         // CURLE_SSL_CACERT || CURLE_SSL_CACERT_BADFILE
-        if ((($errno == 60) || ($errno == 77)) && !is_null($ca_cert)) {
+        if ((($errno == 60) || ($errno == 77) || ($errno == 28)) && !is_null($ca_cert)) {
```
nokitakaze: Maybe errno 28 could be seen only in my [testing http server](https://github.com/nokitakaze/php-testhttpserver) because I did not set the task for compatibility with [RFC 7230](http://httpwg.org/specs/rfc7230.html)

tests/ClientTest.php
```diff
@@ -212,6 +211,121 @@ public function join($timeout = null)
...
+    protected static $_folder = null;
+
+    public static function setUpBeforeClass()
+    {
+        parent::setUpBeforeClass();
+        self::$_folder = sys_get_temp_dir().'/sentry_server_'.microtime(true);
+        mkdir(self::$_folder);
+
+        // Root CA #A1
+        // Сертификат CA #A2, подписанный CA #A1 (с возможностью подписи)
```
nokitakaze: Ooops. These lines mean "Certificate CA A2, signed by CA A1 (intermediate)...". I need to remove these lines. This is copypast from my project

.travis.yml
```diff
@@ -24,13 +24,15 @@ env:
...
-sudo: false
+sudo: required
 
 cache:
   directories:
     - $HOME/.composer/cache
 
 before_install:
+  - sudo apt-get update --yes
+  - sudo apt-get install --yes net-tools
```
nokitakaze: It is need for `netstat -n -t`. Without installing this package it is not working `command not found`. I can remove netstat call, but there can be small chance for reusing used port.

lib/Raven/CurlHandler.php
```diff
@@ -16,7 +16,6 @@
  * @package raven
  */
 
-// TODO(dcramer): handle ca_cert
```
nokitakaze: There wasn\'t be such todo. This worked fine because options gets from single thread curl

lib/Raven/CurlHandler.php
```diff
@@ -73,6 +72,9 @@ public function enqueue($url, $data = null, $headers = [])
 
     public function join($timeout = null)
     {
+        if (count($this->requests) == 0) {
```
nokitakaze: It\'s just a small boost in code. Behaviour didn\'t change

tests/bin/httpserver.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    $server->answer($connect, 200, 'OK', json_encode(['event' => uniqid()]));
+    $server->close_connection($connect);
+    echo "done\n";
+    exit(0);
+};
+$server = new \NokitaKaze\TestHTTPServer\Server($options);
+
+echo "listen...\n";
```
nokitakaze: ooops. Wrong async behavior. I need to add `init_listening`. There could be smallest chance to not open port before Sentry Client connects to it

.travis.yml
dcramer: one question i was going to have- is apt-get update needed?

composer.json
```diff
@@ -11,10 +11,12 @@
             "email": "dcramer@gmail.com"
         }
     ],
+    "minimum-stability": "dev",
     "require-dev": {
         "friendsofphp/php-cs-fixer": "~2.1",
         "phpunit/phpunit": "^4.8 || ^5.0",
-        "monolog/monolog": "*"
+        "nokitakaze/testhttpserver": "^0.1.0",
```
nokitakaze: This is the test server that I created during the last seven days (for code coverage in THIS project). It could stand at least one request per second, but nothing more. I don\'t know why some connections just ignored. I refactored the architecture many times, but I can not do nothing more than that.
I'm very much afraid that the tests in the following PR will fall, because my server will not be able to process them, and not because the code does not work.

dcramer: its just an inprocess server, right?

alternatively, idk if useful, but we could run "python -m SimpleHTTPServer"

.travis.yml
nokitakaze: I think no. It\'s just a habit. Before `apt-get install` I need to do `apt-get update`

composer.json
ste93cry: While I appreciate the work you've done to create a HTTP server, I can't really see why you've done this. Unit tests do not test that a server correctly processes something, they simulate the behaviour by mocking the response data...

nokitakaze: I know Python contains some serious thing like Tornado. But **I don\'t know Python** (otherwise I used it for testing purpose) and async/outer curl **should be coveraged** so I decide to write my own testing server. If anybody could do codecoverage this Python\'s HTTP server please do it that way and I will close this PR.

ste93cry: I honestly don't know what Tornado is, and I don't understand the point of the unit test added above that even uses a fake SSL certificate (or something like that). As I said, we don't have to test what a server replies, we just have to test that the server replies with something. Plus, we agreed that in the future we will use HttpPlug or a library like Guzzle to handle the sending of events (and this means that we won't have to make unit tests about async/not async because it won't anymore be our concern), so I don't really get how this could be useful at all if we're going to remove all the current cURL code

nokitakaze: > Unit tests do not test that a server correctly processes something

Ermmm... No. Unit/Integration tests MUST test that code is working well.

ste93cry: Unit tests test that something works as expected by mocking objects, not by calling a real server. Those are integration tests, and I can't see why we would need them here given that what a server replies is tested in the Sentry codebase itself and it's not our responsibility.

dcramer: the curl stuff has 100% been broken before, so unless we've got a great way to mock it then i think this is fine. We do integration tests all over the place, and they're far superior to unit tests especially if you can make them fast.

ste93cry: >the curl stuff has 100% been broken before, so unless we've got a great way to mock it then i think this is fine

As I said below, all the code related to cURL is going to be removed in the near future as we agreed to delegate the part of code that handles how events are sent to external libraries like HttpPlug so that users are free to choose whatever transport method they prefer. This means that the fact that cURL (in this case) will already be tested by the 3rd party library and we will only need unit tests, not integration tests

nokitakaze: > not by calling a real server

Yes. This is why I used some mock server

nokitakaze: You must ensure that events really sended. And there is no difference between `curl_exec` and `\\ThirdPartyTransport\\Transport::send_request`. If some third party transport breaks in future version (like Monolog) and so our code breaks our user will not want to hear explanations like "this is not our code broken". They just want to be sure everything is working well.

And yes. We will do coverage for all third transport by sending mock events **through every single one of them** until it means that we must to install ftp-servers, email-servers etc.

> server to return a certain response data

Yes. And there it is? Async and outer curl didn\'t tested at all.

ste93cry: >If some third party transport breaks in future version (like Monolog) and so our code breaks our user will not want to hear explanations like "this is not our code broken"

Then it means you don't trust that a vendor already tests his product

>until it means that we must to install ftp-servers, email-servers etc.

Then you have no idea of what unit testing means

dcramer: Can we chill out with "you dont know what you're doing" or I'm just going to end all of this PRs. I'm tired of having the negative back and forth. People have different opinions, thats it.

nokitakaze: >  you don't trust that a vendor already tests his product

I don\'t trust anybody, even myself. That\'s why I write tests for my code.

.travis.yml
mattrobenolt: fwiw, not sure if/how you can do this in PHP, but in Python you can do this by binding a socket to port 0 (which is standard Linux stuff), and port 0 means "just give me a port, any port". Then You can inspect the socket to get what port it actually bound to.

Something like:

```python
s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
s.bind((host, 0))
port = s.getsockname()[1]
s.close()
```

Not that it particularly matters, just sharing a tip if you wanted to avoid trying to use `netstat` to discover an available port.

nokitakaze: @mattrobenolt seems like I [could do it](https://3v4l.org/HFoCC), but I need to find how I can retrieve the port from created socket.

Jean85:
Didn't we talk about switching from native Curl to Guzzle or something else in #424 ?

nokitakaze:
> from native Curl to Guzzle

Anyway we need to test the transport we use. When someone would so kind to use python\'s server he could replace by solution

dcramer:
I'm generally ok with this for now. If when we land Guzzle/whatever we decide its not needed we can remove it. Anything thats automated that ensures we cause less user issues is a win.

PR #212, opened 2015-05-15
Fix: Change SanitizeDataProcessorTest to mask nested arrays - fixes #145

matthewgoslett:
This PR resolves Issue https://github.com/getsentry/raven-php/issues/145

If an array key matches the field regex, all child values are masked recursively.
eg:

```
  'card_number' =>
  array(4) {
    [0] =>
    string(8) "********"
    [1] =>
    string(8) "********"
    [2] =>
    string(8) "********"
    [3] =>
    string(8) "********"



'card_number_nested_further' =>
  array(4) {
    [0] =>
    array(2) {
      'part' =>
      string(8) "********"
      'number' =>
      string(8) "********"
    }
    [1] =>
    array(2) {
      'part' =>
      string(8) "********"
      'number' =>
      string(8) "********"
    }
    [2] =>
    array(2) {
      'part' =>
      string(8) "********"
      'number' =>
      string(8) "********"
    }
    [3] =>
    array(2) {
      'part' =>
      string(8) "********"
      'number' =>
      string(8) "********"
    }
  }
```

I've also added 2 unit tests to demonstrate the nested masking.

PR #472, opened 2017-06-03
[2.0] Refactor how the processors are configured and instantiated

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | no
| License       | MIT

The aim of this PR is to refactor entirely how the processors are configured, instantiated and executed. The problems I see in the current code are:

- The order in which the processors are executed is the order in which they gets added to the configuration. While I don't have any real use case this is a problem, I think it would be better to be able to customize the order in a more efficient way
- There isn't a clear way to define the options of the processors. The only way is to copy-paste a lot of code to validate that each option is correct, etc. By using the [OptionsResolver](http://symfony.com/doc/current/components/options_resolver.html) component from Symfony this task is simplified and we have a standard way to define the options that each processor accepts
- The processors can't be lazy-created, nor they can have different constructor arguments because the instantiation of the objects is handled internally by the client. This is where the factory pattern applies: by having a factory class the client won't need anymore to know how a processor should be created. We could even provide in the framework packages a factory that uses the service container


ste93cry:
No, I just created the branch of this PR from the branch of #462 because I need those changes. When that PR will be merged the duplicated commit will disappear

nokitakaze:
So I will review it after #462 merging. Because I can not see there #462 is ending and #472 is starting

ste93cry:
While working on this, I faced some issues I'm not sure how to properly solve @Jean85 @nokitakaze @stayallive

Actually the options of the processors are configured in the `processors_options` config option and then passed during instantiation to the `Raven_Processor::setProcessorOptions` method. This means that in the constructor of the class there are not all available information needed to start operating or configuring the processor. The solution proposed in this PR tries to solve the problem by passing them directly in the constructor. However, this introduces an hard dependency on the definition of the constructor that must now mandatorily accept a given amount of parameters in a certain order. I don't really like that, considering that some processors may need 3rd parties dependencies. Indeed, users needing that must create another factory that handle the instantiation of these types (e.g. processors defined as services of a service container) but this would mean that the instance of the processor will be returned pre-configured, introducing another problem: some processors would be configured using the client configuration, others using some other code outside of our control. I'm not sure this can be solved in a better way, I just want your opinion on how to proceed. If this way of doing things is fine, I will finish this PR as soon as possible, otherwise we should go back to the table and rethink the whole PR to address these inconsistencies

Jean85:
I'm not able to delve into the code at the moment, but IMHO we should strive to move the options in the constructor. Can we write down the cases that we need to cover right now to decide better about this?

nokitakaze:
I think, if there will be need of adding IoC with self made objects, user could just create his own `factory`. But it will be better if user could `extend` default factory and just add some behaviour for default constructor method

ste93cry:
This PR is ready for review. A quick note: while trying to make the `SanitizeDataProcessor` processor to work with the changes of this PR I found out that it sanitizes the frames of the stacktrace. There was no way to edit the frames and store the changes inside the `Stacktrace` class, so I created a value object that stores each frame so that they can be passed around and they can be edited. In the future we could move more logic about the frames (e.g. how the arguments are extracted) inside that class.

UPGRADE-2.0.md
```diff
@@ -286,6 +291,39 @@
...
+  $client->addProcessor($processor2);
+  $client->addProcessor($processor1);
+
+  // or
+
+  $client->addProcessor($processor1);
+  $client->addProcessor($processor2, 255); // Note the priority: higher the priority earlier a processor will be executed. This is equivalent to adding first $processor2 and then $processor1
+  ```
+
+- The method `Client::process` has been removed as there is no need to process event data
```
Jean85: Great work here! :+1:

lib/Raven/Processor/ProcessorRegistry.php
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * Removes the given processor from the list of available ones.
+     *
+     * @param ProcessorInterface $processor The processor instance
+     */
+    public function removeProcessor(ProcessorInterface $processor)
+    {
+        foreach ($this->processors as $priority => $processors) {
+            foreach ($processors as $key => $value) {
+                if ($value !== $processor) {
```
Jean85: I would reverse this if for increased readability; `continue` is not needed here

UPGRADE-2.0.md
```diff
@@ -286,6 +291,39 @@
...
+- The method `Client::createProcessors` has been removed as there is no need to create
+  instances of the processors from outside the `Client` class.
+
+- The method `Client::setProcessors` has been removed. You should use `Client::addProcessor`
+  and `Client::removeProcessor` instead to manage the processors that will be executed.
+
+  Before:
+
+  ```php
+  $processor1 = new Processor();
```
Jean85: You should make this example more clear: you forgot to mention (and add to this example) that `Processor` was an abstract class that was to be extended, that you replaced with the `ProcessorInterface`.

tests/Processor/SanitizeDataProcessorTest.php
```diff
@@ -12,181 +12,199 @@
...
-        ];
+        if (isset($inputData['extra_context'])) {
+            $event = $event->withExtraContext($inputData['extra_context']);
+        }
 
-        $client = ClientBuilder::create()->getClient();
-        $processor = new SanitizeDataProcessor($client);
-        $processor->process($data);
+        if (isset($inputData['exception'])) {
+            // We must convert the backtrace to a Stacktrace instance here because
```
Jean85: Maybe you can wrap this in a private method, since you do it twice?

tests/Processor/SanitizeDataProcessorTest.php
```diff
@@ -12,181 +12,199 @@
...
 
-        /** @var SanitizeDataProcessor $processor */
-        $processor = $this->getObjectAttribute($client, 'processors')[0];
+            $this->assertArraySubset($expectedData['exception'], $event->getException());
+        }
 
-        $this->assertInstanceOf(SanitizeDataProcessor::class, $processor);
-        $this->assertEquals($processor->getFieldsRe(), $processorOptions[SanitizeDataProcessor::class]['fields_re'], 'overwrote fields');
-        $this->assertEquals($processor->getValuesRe(), $processorOptions[SanitizeDataProcessor::class]['values_re'], 'overwrote values');
+        $this->markTestIncomplete('Array scrubbing has not been implemented yet.');
```
Jean85: What kind of scrubbing is needed here?

tests/Processor/SanitizeHttpHeadersProcessorTest.php
```diff
@@ -13,24 +13,26 @@
...
     {
-        /** @var Client|\PHPUnit_Framework_MockObject_MockObject $client */
-        $client = $this->getMockBuilder(Client::class)
-            ->disableOriginalConstructor()
-            ->getMock();
-
-        $this->processor = new SanitizeHttpHeadersProcessor($client);
-        $this->processor->setProcessorOptions([
+        $this->client = ClientBuilder::create()->getClient();
+        $this->processor = new SanitizeHttpHeadersProcessor([
```
Jean85: So much cleaner!! :+1:

tests/Processor/SanitizeDataProcessorTest.php
stayallive: Like #212 (I believe), been looking at this recently.

ste93cry: The test was incomplete even in `1.x` versions, from what I saw if an array like the one below is passed to the processor the `key_to_sanitize` item does not gets sanitized

```php
[
    'key_to_sanitize' => [
        'foo' => 'bar',
    ],
]
```

UPGRADE-2.0.md
ste93cry: `Processor` here was meant to be any processor, not the specific abstract class. Maybe I could use existing processor names or make it more clear using something like `FooProcessor`?

lib/Raven/Processor/ProcessorRegistry.php
ste93cry: Wouldn't the nesting be too much? I mean, there are already 2 `for` loops that nest each other. I think that having too much deep makes the code less clear

PR #473, opened 2017-06-03
Refactor the client transport to be extensible by using HTTPlug

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

As agreed in #424, this PR integrates [HTTPlug](https://github.com/php-http/httplug) as required dependency of this library. The idea is that instead of having to maintain a set of transports (cURL in our case, as it's the only one we provide) we could use the one provided by this 3rd party vendor. What it means for us is:

- Less work for us as we won't have to maintain anymore any transport
- Users will be (finally) free to use the transport they want, be it cURL, Guzzle, etc


Jean85:
Great job on the migrations note!

ste93cry:
Refactoring the transport methods I was evaluating whether it really makes sense to support the `exec` type. Indeed, what it does is different from both syncronous and asyncronous types, but we can't keep track of what happens with the events being sent. For example, we can't know if sending was successful, what the server replied and so on. I'm not saying there are not use cases for this method, I'm just wondering if it should be supported out of the box

lib/Raven/Client.php
```diff
@@ -112,13 +113,26 @@ class Client
...
+
+    /**
      * Constructor.
      *
-     * @param Configuration $config The client configuration
+     * @param Configuration   $config     The client configuration
+     * @param HttpAsyncClient $httpClient The HTTP client
      */
-    public function __construct(Configuration $config)
+    public function __construct(Configuration $config, HttpAsyncClient $httpClient, MessageFactory $messageFactory)
```
nokitakaze: I do not even want to whine about the destruction of the library. Why don't you want to use single param pattern? Don't you think about how many params will be here after some time? You can use single parameter with all options you want to use.

lib/Raven/ClientBuilder.php
```diff
@@ -104,12 +128,40 @@ public static function create(array $options = [])
...
-        return new Client($this->configuration);
+        $messageFactory = $this->messageFactory ?: MessageFactoryDiscovery::find();
+        $transportFactory = $this->transportFactory ?: new CurlTransportFactory(
+            $messageFactory,
+            $this->streamFactory ?: StreamFactoryDiscovery::find()
+        );
+
+        $transportFactory = new PluginClientFactory($this->configuration, $transportFactory, UriFactoryDiscovery::find());
+
+        return new Client($this->configuration, $transportFactory->getInstance(), $messageFactory);
```
nokitakaze: How do I instance some other class which extends \\Raven\\Client? Will I have to extend this class too?

lib/Raven/Client.php
Jean85: "Single parameter" is not a pattern. Also, it's not useful, and it will not have the benefits of typehints that we have here. If we will have too many parameters, it will mean that we are breaking the SRP principle and we would need to refactor a little.

lib/Raven/ClientBuilder.php
Jean85: No, why? Those are dependencies, you have just to continue to use them, in the same way.

lib/Raven/Client.php
nokitakaze: > it's not useful

It is VERY useful. Also it promote us capability to extend code without breaking back compatibility.

Jean85: Sorry but I completely disagree. Typehints are alot more useful ([and I already wrote a lengthy blog post about it](https://engineering.facile.it/blog/eng/visual-debt-typehints/)) and using array as a single point of entry for options doesn't protect us from BC breaks. 

The builder class does that a lot more.

nokitakaze: No-no, wait. **Typehints** have nothing to do with this question. Of course typehints are need. More than that, hints in your IDE are need. In my projects I even need to create separate file with objects description [[1]](https://github.com/nokitakaze/php-mutex/blob/master/src/phpdoc.php) [[2]](https://github.com/nokitakaze/php-keyvalue/blob/master/src/phpdoc.php) so my Idea shows me properties of "stdClass".

I say you could use all params as separate properties **in** `Configuration`

ste93cry: >I say you could use all params as separate properties in `Configuration`

The factories instances or the HTTP client instance are not options, so why should they be in that class? The `Client` class is indeed harder to instantiate because of the number of parameters it needs, this is why the builder was introduced. If you want to instantiate the client class directly then you will have to deal with the fact that it needs dependencies

nokitakaze: So how it looks to create client with transportFactory?

```php
$builder = new ClientBuilder();
$builder->setTransportFactory(...);
$builder->setMessageFactory(...);
$client = $builder->getClient();
```
?

This is beautiful. If previously creating factory for instance Raven Client was just an option, than now it is required.

ste93cry: >If previously creating factory for instance Raven Client was just an option, than now it is required.

It's not mandatory to call setters. Defaults values for all the dependencies are provided when creating the instance of the client

nokitakaze: Yes. I'm talking about when you need to set transport/message factory

Jean85: @nokitakaze I think we are having problems understanding your concers. Can you rephrase your question?

Nyholm:
Let me know when you are done and ready for a review. I would be happy to review it.

nokitakaze:
This PR should contain a lot of documentation pages

stayallive:
This is starting to add a lot of code, would love to see some kind of overview what is being added and why, just so we know what we are looking at. Starting to look a lot like Java with al the factories
 😉 

Great work so far Stefano! 💪

composer.json
```diff
@@ -25,7 +27,10 @@
         "ext-mbstring": "*",
         "ext-curl": "*",
         "composer/ca-bundle": "~1.0",
-        "symfony/options-resolver": "~2.7|~3.0"
+        "symfony/options-resolver": "~2.7|~3.0",
+        "php-http/httplug": "~1.1",
+        "php-http/curl-client": "~1.7",
```
fbourigault: As a library, you must not depend on any client implementation. You should depends on `php-http/client-implementation`.

composer.json
```diff
@@ -25,7 +27,10 @@
...
+        "zendframework/zend-diactoros": "~1.4"
```
fbourigault: Same applies here. You must depend on `psr/http-message-implementation`. This enable users to choose any compatible implementation.

lib/Raven/Client.php
```diff
@@ -112,13 +119,27 @@ class Client
...
+    /**
      * Constructor.
      *
-     * @param Configuration $config The client configuration
+     * @param Configuration      $config         The client configuration
+     * @param FlexibleHttpClient $httpClient     The HTTP client
+     * @param RequestFactory     $requestFactory The PSR-7 request factory
      */
-    public function __construct(Configuration $config)
+    public function __construct(Configuration $config, FlexibleHttpClient $httpClient, RequestFactory $requestFactory)
```
fbourigault: You should not force users to provide a `FlexibleHttpClient` here. You should stick with `HttpClient` or `HttpAsyncClient`. Requiring an `HttpClient` is fine if you don't make any async call.

lib/Raven/Client.php
```diff
@@ -747,32 +735,10 @@ public function send(&$data)
...
-     * @param string       $url     Full URL to Sentry
-     * @param array|string $data    Associative array of data to log
-     * @param array        $headers Associative array of headers
-     */
-    protected function send_remote($url, $data, $headers = [])
-    {
-        $parts = parse_url($url);
-        $parts['netloc'] = $parts['host'].(isset($parts['port']) ? ':'.$parts['port'] : null);
-        $this->send_http($url, $data, $headers);
+        $this->httpClient->sendAsyncRequest($request)->wait(true);
```
fbourigault: It's better to use `HttpClient::sendRequest` as it has less overhead.

composer.json
Nyholm: Yeah, put this in the require-dev instead, so you can run the tests. 

Also, do not forget to require `php-http/discovery`

composer.json
ste93cry: I strongly agree with you, and I thought a lot about this when looking at the official documentation of Httplug that clearly says [this is the way to go](http://docs.php-http.org/en/latest/httplug/library-developers.html) for library developers. I chosen to specify a client (cURL in this case) because I thought @dcramer would not approve to ship an SDK without a HTTP client. If we all agree on this, I will edit the PR accordingly.

lib/Raven/ClientBuilder.php
```diff
@@ -11,6 +11,18 @@
...
 namespace Raven;
 
+use Http\Client\Common\FlexibleHttpClient;
+use Http\Discovery\MessageFactoryDiscovery;
+use Http\Discovery\StreamFactoryDiscovery;
+use Http\Discovery\UriFactoryDiscovery;
+use Http\Message\MessageFactory;
+use Http\Message\RequestFactory;
+use Http\Message\StreamFactory;
+use Raven\HttpClient\CurlHttpClientFactory;
```
Nyholm: Why do you need your own factories?

lib/Raven/Client.php
ste93cry: I know, but I found shocking that there is not an interface that both implements `HttpClient` and `HttpAsyncClient`. The only thing that I found was that class, and our requirements are to support both syncronous and asyncronous requests (and clients, obviously). I don't like this either 😞

lib/Raven/Client.php
ste93cry: Indeed, but as explained above we have to support both syncronous and asyncronous requests

ste93cry: Indeed, but as explained above we have to support both syncronous and asyncronous requests. The function call to the `wait` method is only temporary until I sort out some other things

lib/Raven/Client.php
Nyholm: You should use only HttpAsyncClient if you are required to support it. You leave it to the user to wrap their non async client into a flex client.

ste93cry:
If I don't want the `Client` class to have a dependency on the `RequestFactory`, I should use the `HttpMethodsClient` class but it doesn't support async. Any idea about this? I don't see any valid reason that class should not implement the missing methods too @fbourigault

fbourigault:
You can use a `MessageFactory` and the discovery (`Http\Discovery::MessageFactoryDiscovery::find()`) to provide a default implementation instead.

ste93cry:
I think I didn't explain well what I'm trying to do. The actual `Raven\Client` class depends on the `Http\Message\RequestFactory` to create an object that must be passed to the `HttpAsyncClient::sendAsyncRequest` method. I would like to simplify the code, if possible, by removing from the constructor this dependency. Unfortunately the only way I see to do this is to not use that method. Instead I should rely on the methods provided by the `Http\Client\Common\HttpMethodsClient` class (`get`, `post`, etc). That class in turn sends requests only in a syncronous way, and this is not acceptable for the requirements of this library.

sagikazarmark:
The RequestFactory let's your user to choose a PSR-7 implementation. It means that you don't force installing one, but you also require the user to take care of it (unless you use discovery as a fallback).

By using the HttpMethodsClient you don't simplify anything from the user point of view as the user still have to configure that factory. I think this should matter a bit more, but it's just my opinion.

If you want to remove that dependency from the constructor you still have a couple options:

- Use the discovery in the constructor to create a factory (without accepting one as an argument)
- Install a specific implementation with Composer and create it in the constructor.

Unfortunately I am not aware why you require the async calls, but the MethodsClient cannot work in an asynchronous way. By definition the get, post, ... methods return a ResponseInterface, meaning that even if we used an async client, we would have to wait for the promise which destroys the whole purpose.

We would either have to create getAsync, postAsync methods (but that would require client emulation) or create a HttpAsyncMethodsClient which returns promises instead of a response.

But as said, I don't see how this simplifies things from a user perspective, the configuration overhead remains the same (or even becomes a bit bigger).

composer.json
sagikazarmark: Well, installing a client by default became a common practice, even if it kinda destroys one of the original purposes (being able to avoid installing multiple client implementations). But I can totally understand why a library maintainer would decide to ship an OOTB working library, instead of recommending to install another one.

Maybe we should add this as a "best practice": if you want to ship your library with a client use <our most lightweight client here>. This way lib maintainers can use the same client as a sane default, so the end user don't have to install five, only the default and the choice of the user.

fbourigault:
@ste93cry sorry my previous response was incomplete. You should also depend only on an HttpAsyncClient.

I think here the async is required because sentry is about sending logs and metrics to the server and so we don't care if the call is successful or not. Am I right? If so, you should not call the `Promise::wait` as it will make you call synchronous.

composer.json
ste93cry: >Well, installing a client by default became a common practice, even if it kinda destroys one of the original purposes

Does this totally defat the purpose of Composer virtual packages, right? By the way I don't see a really big problem to ask the user to choose one client library and not rely on any one provided out-of-the-box, if the README is clear enough to explain why and how he must proceed. As this is indeed a blocking topic, @dcramer can you please give us your opinion on shipping with a default client like cURL?

ste93cry:
>Unfortunately I am not aware why you require the async calls

Async is needed because we can't block the execution of the script just to send logs to a remote server.

>we don't care if the call is successful or not

This is not entirely true, according to some other issues and PRs there is the need to track what happens with events because they could fail to be sent and maybe a new try should be made

>you should not call the `Promise::wait` as it will make you call synchronous.

Indeed, as explained in the review above that call is there just because I have to figure out other parts of the code before making everything supporting both syncronous and asyncronous calls

>We would either have to create getAsync, postAsync methods (but that would require client emulation) or create a HttpAsyncMethodsClient which returns promises instead of a response.

I was thinking about having new methods that supports async too. I don't see the need for a new client class, but something that simulates the `FlexibleClient` behaviour of using one method or another would be great to have. The whole point of this is that in this way user would have to configure the request factory only one time (or rely on the default provided by the discovery class, indeed) and then he could just use methods like `post` by passing the arguments like the URI string or the body content without the need of creating a `Request` instance. I would also like to avoid that the instance of the factory passed to the client is different from the instance of the factory passed to the HTTP client, but I understand that maybe what I'm saying could really makes no sense

composer.json
sagikazarmark: > Does this totally defat the purpose of Composer virtual packages, right?

Indeed it is.

> By the way I don't see a really big problem to ask the user to choose one client library and not rely on any one provided out-of-the-box, if the README is clear enough to explain why and how he must proceed.

Yeah, I also don't think of it's a big deal, but I can understand the reasons behind to other side too.

sagikazarmark:
> Async is needed because we can't block the execution of the script just to send logs to a remote server.

Totally makes sense. In this case Promises would provide a nice way to track what happened to certain events.

> The whole point of this is that in this way user would have to configure the request factory only one time

That kinda makes sense, but with DI it shouldn't be a problem.

> and then he could just use methods like post by passing the arguments like the URI string or the body content without the need of creating a Request instance.

Does the user actually send HTTP requests? Or your library does? If it's the latter then I don't really get the point.

> I don't see the need for a new client class, but something that simulates the FlexibleClient

Keep in mind that emulation means an overhead and complexity.

>  I would also like to avoid that the instance of the factory passed to the client is different from the instance of the factory passed to the HTTP client, but I understand that maybe what I'm saying could really makes no sense

If you rely on an abstraction (you do: PSR-7) this really shouldn't matter at all.

composer.json
fbourigault: We used to give a one liner to install the package without any problems (https://github.com/fbourigault/php-gitlab-api/blob/e49bc7313ae53e26501993353dcc1d404ae26b01/README.md#installation)

lib/Raven/Client.php
fbourigault: So I guess you plan to make the `wait` call optional?

composer.json
ste93cry: I've updated the README to explain how to install and use this library based on the great work found in the link you provided. If you have any idea on how to improve docs I'm open to suggestion not being a naive English speaker

composer.json
ste93cry: I've updated the `composer.json` dependencies, moved around the `require` and `require-dev` sections (I know this was not something related to this PR but I wanted to do it since a long time) and I followed your suggestion to not require a specific HTTP client implementation and instead rely on Composer virtual packages

lib/Raven/Client.php
ste93cry: I'm not sure if it should be optional or not needed at all. I mean, in `1.x` besides how an event was sent (syncronously or asyncronously) user could not access the response from the server in any way (at least this is what I remember), so it makes no sense to make the script wait for it even if it's using an async HTTP client. Just send the event and continue the script. At the moment there is nothing that handle the case when an event failed to be sent, but in another PR we could implement it

stayallive: We are also going to need to support a 429 backoff response from the Sentry server.

I personally would send my errors directly if they occur making sure they're being sent before something else crashes. So an option to choose between sync and async would be ideal? Also keeping that backwards compatible is an upside of making it configurable.

lib/Raven/ClientBuilder.php
ste93cry: You're right, I didn't need this factory at start because I could just let the client set directly the instance of the HTTP client. However, I have to decorate the stream factory because I need to (if configured to) base64 and gzip the request before sending it and there doesn't seems to be any way to set the it without creating myself the client instance. However, this is now a problem because I don't rely anymore on a specific client but instead I changed the `composer.json` to require an implementation of the Httplug virtual packages 🙈

lib/Raven/Client.php
ste93cry: >So an option to choose between sync and async would be ideal?

There was an option that configured this in 1.x named `curl_method`, I removed it because we're now relying on the HTTP client itself implementing `HttpClient`, `HttpAsyncClient` or both interfaces to decide which method should be used to send events. What I didn't think of is the last case: which method is preferred in case both interfaces are implemented by the client?

lib/Raven/ClientBuilder.php
ste93cry: I have to decorate the stream factory because I need to (if configured to) base64 and gzip the request before sending it and there doesn't seems to be any way to set the factory of a client without creating myself the client instance. This is now a problem because I don't rely anymore on a specific client but instead I changed the `composer.json` to require an implementation of the Httplug virtual packages 🙈

lib/Raven/Client.php
stayallive: Hmmm yes this is true. Depending on how hard we want to push on reading a potential 429 or error response I would say go for async since it's faster. 

Going async allows us to have a callback right? We could possibly read and honor a 429 there, or an error response.

Given the nature of php being per request a new world, honoring those things is a bit annoying anyhow.

ste93cry: >We are also going to need to support a 429 backoff response from the Sentry server.

At the moment there is no code neither in `1.x` nor in `2.x` that handles failure of sent events, as far as I know. We should indeed implement it, but maybe it's better to finish the implementation of Httplug in this PR and think about that topic in a new one

>Going async allows us to have a callback right?

Yes, this is exact. Right now I append the Httplug plugin that takes any error response and converts it into an exception, but I'm not sure this is a good idea as probably the exception handler would catch it and try to resent the event, starting an infinite loop (maybe?) in case 429 is the error code of the response

nokitakaze:
> Async is needed because we can't block the execution of the script just to send logs to a remote server

Yes. And we don't want to be sure a log record has been delivered, yes?

ste93cry:
Some quick notes:

- I found out that the default installation of Sentry on a custom machine uses the [uWSGI](https://uwsgi-docs.readthedocs.io/en/latest/) server. The chunked transfer is not supported out of the box so the gzip encoding won't work and will always result into an event sending failure. The solutions are to turn off the compression or to use another webserver like nginx in front of the uWSGI. I spent 3 days trying to understand what was wrong, so maybe this should deserve a special note in the docs
- Removing the `wait` call makes the script end (possibily) before the event is sent. When this happens the event wil never be sent. This was fixed in daf2acb, but we don't have access anymore to the curl handler with Httplug. Any solution to this?

fbourigault:
For your second point, your event will always be sent! In PHP, the async only occurs when the request have been sent and execution resume when the response is received. In the following case, the execution will always be :
* LoggerPlugin with $request1
* RetryPlugin with $request1
* AsyncClient with $request1
* LoggerPlugin with $request2
* RetryPlugin with $request2
* AsyncClient with $request2
//other code
```php
$client = new PluginClient(HttpAsyncClientDiscovery::find(), [
    new LoggerPlugin(),
    new RetryPlugin(),
]);

$client->sendAsync($request1);
$client->sendAsync($request2);
//other code
```

ste93cry:
From my testing it's not how you describe it. Using the CLI script from the `bin` folder after removing the call to the `wait` function I see no events received in the Sentry backend. Using syncronous method instead works like a charm

fbourigault:
I will try to dig into this. Have to look at cURL internals.

ste93cry:
>This maybe can be achieved using a shutdown function or by waiting for promises after a call to fastcgi_finish_request when this function exists.

It doesn't seems appropriate to call `fastcgi_finish_request` because we should not end the request as we're just a logging library. I will try to wait for promises in a shutdown function, but wouldn't it be better to fix the problem in `php-http/curl-client`? Guzzle correctly handles this case as I understood, so it seems that the Httlug client is somehow incomplete

fbourigault:
Yes, of course, I don't want you to call `fastcgi_finish_request`. I just wanted to call code with a shutdown function or at `KernelEvents::TERMINATE` for Symfony (or anything equivalent in any other framework).

If Guzzle handles this case correctly, please open an issue on `php-http/curl-client`.

.travis.yml
```diff
@@ -43,7 +42,6 @@ script:
 
 after_script:
   - wget https://scrutinizer-ci.com/ocular.phar
-  - if [ $(phpenv version-name) = "5.5" ] && [ "$REMOVE_XDEBUG" = "0" ]; then php ocular.phar code-coverage:upload --format=php-clover tests/clover.xml --revision=$TRAVIS_COMMIT; fi
```
nokitakaze: And there is deleting PHP 5.5 from composer?
https://github.com/getsentry/sentry-php/pull/473/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780L22

stayallive: https://github.com/getsentry/sentry-php/issues/424#issuecomment-309002374

ste93cry: Yes it's two lines after the line of the diff you linked. I now require `PHP >= 5.6` in `composer.json`

nokitakaze: Okay. Now I see

ste93cry:
>Let me know when you are done and ready for a review. I would be happy to review it.

@Nyholm I think we're ready to start reviewing this PR

UPGRADE-2.0.md
```diff
@@ -13,6 +11,16 @@
 - The `project` option has been removed.
 - The `extra_data` option has been removed in favour of setting additional data
   directly in the context.
+- The `curl_method` option has been removed in favour of leaving to the user the
+  choice of setting an HTTP client supporting syncronous, asyncronous or both
+  transport methods.
+- The `curl_path` option has been removed.
+- The `curl_ipv4` option has been removed.
+- The `curl_ssl_version` option has been removed.
+- The `verify_ssl` option has been removed.
```
Jean85: Do we have any generic replacement for the `verify_ssl` option? Or do it depends now on the concrete client that is used?

composer.json
```diff
@@ -11,22 +11,25 @@
...
-        "friendsofphp/php-cs-fixer": "~2.1",
-        "phpunit/phpunit": "^4.8 || ^5.0",
-        "ext-openssl": "*",
-        "nokitakaze/testhttpserver": "^0.1.0",
-        "monolog/monolog": "~1.0"
-    },
     "require": {
-        "php": ">=5.5",
-        "ext-hash": "*",
+        "php": ">=5.6",
```
Jean85: I would write explicitly `^5.6 | ^7.0` to avoid any strange issue with future PHP versions, it's a good practice.

7.2 already may have some BC, but we will fix them; 8.x instead will require a new major version, I think.

lib/Raven/Client.php
```diff
@@ -747,233 +727,40 @@ public function send(&$data)
...
-     * @param array|string $data    Associative array of data to log
-     * @param array        $headers Associative array of headers
-     * @return bool
-     */
-    protected function send_http_synchronous($url, $data, $headers)
-    {
-        $new_headers = [];
-        foreach ($headers as $key => $value) {
-            array_push($new_headers, $key .': '. $value);
+        if ('gzip' === $this->config->getEncoding()) {
```
Jean85: Can we translate this in a `isEncodingCompresed` method? I would like to avoid having the `gzip` string hardcoded and written everywhere.

UPGRADE-2.0.md
ste93cry: We don't have anymore any option related to the HTTP client because each adapter/client has its own set of options, so there is not a replacement for `verify_ssl`. From now on you will have to specify each option supported by the concrete client manually inside the `http_client_options` option.

composer.json
ste93cry: It's ok for me, just one question about this: what's different from `>= 5.6`? Shouldn't Composer resolve to the same versions?

stayallive: Then we should also remove the nightly and hhvm(?) builds from Travis.

Or run `composer install --ignore-platform-reqs` for those builds (which seems like a good idea to be able tot test for 7.2 ahead of time).

Jean85: @ste93cry nope, `>=5.6` would allow `8.0` for example. We don't want that, or in the future we will disallow that and composer will fall back to a very old, not really compatible version. It's better to be conservative on this side of thing, allowing 8.0 later is easy, if no issue is found.

@stayallive I think that testing on 7.2 in advance is good, but we don't need to `--ignore-platform-reqs`, I was suggesting `^5.6 | ^7.0` which allows all the 7.x series; I would only put the 7.2 build on `alllow_failure` for now.

For HHVM,  I don't know if I would still be testing on that, since a lot of other big players are leaving it, especially Symfony, from which we use a component, that could then break on HHVM leaving us without a solution. Maybe we should put it on `allow_failure` too?

stayallive: @Jean85 yes of cours 🤦‍♂️  stupid of me.

I'm not sure HHVM is still needed because most packages I am using and some big frameworks are also dropping it from their testing since it brings more trouble than it's worth.

Right now it's already on allow failure looking at the latest build for this PR.

ste93cry: >For HHVM, I don't know if I would still be testing on that, since a lot of other big players are leaving it

I was following a thread some weeks ago that was talking about this, and it seems that it's true that many of the big libraries/frameworks out of here are dropping HHVM due to the lack of big performance differences since PHP 7 was released and since there are some different behaviours between the two platforms that makes maintaining the code more difficult, so I'm 👍 on dropping support. I would simply not test it anymore, as it's useless to have it on `allow_failure` if we don't support it

README.md
```diff
@@ -25,19 +25,47 @@ more about [automatic PHP error reporting with Sentry](https://sentry.io/for/php
...
 - Process and sanitize data before sending it over the network
 
+## Install
+
+To install the SDK you will need to be using [Composer]([https://getcomposer.org/)
+in your project. To install it run:
+
+```bash
+curl -sS https://getcomposer.org/installer | php
+```
```
ste93cry: It's common to have this set of instructions to install Composer on READMEs, but if you prefer to just link to the docs I will do it.

lib/Raven/HttpClient/HttpClientFactoryInterface.php
```diff
@@ -0,0 +1,24 @@
...
+
+use Http\Client\HttpAsyncClient;
+
+/**
+ * This interface implements a contract that should be respected by all factories
+ * willing to create HTTP client instances.
+ *
+ * @author Stefano Arlandini <sarlandini@alice.it>
+ */
+interface HttpClientFactoryInterface
```
Nyholm: I do not think you need to create a HTTP client. You should require a configured client or fallback to client discovery.

lib/Raven/ClientBuilder.php
```diff
@@ -132,4 +232,25 @@ public function __call($name, $arguments)
 
         return $this;
     }
+
+    /**
+     * Creates a new instance of the HTTP client.
+     *
+     * @return HttpAsyncClient
+     */
+    protected function createHttpClientInstance()
```
Nyholm: private?

lib/Raven/Configuration.php
```diff
@@ -890,19 +755,18 @@ public function setProcessorsOptions(array $options)
...
         $resolver->setDefaults([
+            'send_attempts' => 6,
             'trust_x_forwarded_proto' => false,
             'prefixes' => explode(PATH_SEPARATOR, get_include_path()),
             'serialize_all_object' => false,
-            'curl_method' => 'sync',
-            'curl_path' => 'curl',
-            'curl_ipv4' => true,
-            'curl_ssl_version' => null,
+            'http_client_options' => [],
```
Nyholm: Should be removed with the HttpClientFactory

lib/Raven/ClientBuilder.php
```diff
@@ -87,6 +90,31 @@ class ClientBuilder implements ClientBuilderInterface
     protected $configuration;
 
     /**
+     * @var UriFactory The PSR-7 URI factory
+     */
+    protected $uriFactory;
```
ste93cry: Does it makes sense in a builder class to have the properties as `private`? What if an user needs to customize how the client is created? If we're going to make all properties `private` we could also mark this class as `final` because we don't expose any other public method not defined in the interface

lib/Raven/ClientBuilder.php
ste93cry: I made everything private and marked the class as `final`

Nyholm:
:+1:

PR #474, opened 2017-06-05
Adding X-Authorization and Cookie to sanitized headers

igorsantos07:
Cookie values are sanitized by default at the default `SanitizeDataProcessor`, so it's safe to consider they should also be sanitized in headers. The cookie keys list will be available on the "cookie" data point anyway (unless `RemoveCookiesProcessor` is enabled), thus it makes no sense to sanitize cookies there but not in the header.
X-Authorization is a very common header user to transmit arbitrary Token values on APIs.

### DIFF
--- a/lib/Raven/Processor/SanitizeHttpHeadersProcessor.php
+++ b/lib/Raven/Processor/SanitizeHttpHeadersProcessor.php
@@ -67,6 +67,6 @@ final class SanitizeHttpHeadersProcessor extends Processor
      */
     private function getDefaultHeaders()
     {
-        return ['Authorization', 'Proxy-Authorization', 'X-Csrf-Token', 'X-CSRFToken', 'X-XSRF-TOKEN'];
+        return ['Authorization', 'Proxy-Authorization', 'X-Authorization', 'X-Csrf-Token', 'X-CSRFToken', 'X-XSRF-TOKEN'];
     }
 }


ste93cry:
I'm going to quote this from #455:

> It's fine to add the X-Authorization header to the list, but the Cookie header is already sanitized by the RemoveCookiesProcessor processor.

lib/Raven/Processor/SanitizeHttpHeadersProcessor.php
```diff
@@ -67,6 +67,6 @@ public function process(&$data)
      */
     private function getDefaultHeaders()
     {
-        return ['Authorization', 'Proxy-Authorization', 'X-Csrf-Token', 'X-CSRFToken', 'X-XSRF-TOKEN'];
+        return ['Authorization', 'Proxy-Authorization', 'X-Authorization', 'Cookie', 'X-Csrf-Token', 'X-CSRFToken', 'X-XSRF-TOKEN'];
```
stayallive: As mentioned by @ste93cry, if you remove the `Cookie` header this PR can be accepted.

igorsantos07:
@ste93cry wow! I didn't know it was possible to change the PR target branch! thanks :O

lib/Raven/Processor/SanitizeHttpHeadersProcessor.php
igorsantos07: I don't recall anymore why I added the Cookie header there. Even by looking at the `master` code, my original premise doesn't make much sense. Whatever xD

PR #479, opened 2017-07-19
Added HTTP_X_FORWARDED_FOR for user IP address

Inscure:


stayallive:
Hi, thank you for the contribution, however this opens up a security issue since the `HTTP_X_FORWARDED_FOR` variable comes from the `X-Forwarded-For` HTTP header which can easily be spoofed by a client or proxy causing incorrect IP's to be reported.

You can override the `ip_address` in the user context yourself if you need to, you are probably are already validating that the `X-Forwarded-For` can be trusted in your application and thus it's best if you set the IP yourself.

```diff
$client->user_context([
    // Other user context data...
+   'ip_address' => $clientIP,
]);
```

stayallive:
Since there is no response and there a an easy "workaround" and given that it's a possible security issue to accept this PR I am closing it. Feel free to restart the conversation if the above arguments and workaround are not valid.

Issue #481, opened 2017-07-26
Exclude anonymous functions from app


dcramer:
They're not correctly "in app", and if we can't guarantee they are in situations where we've configured an app prefix, let's just exclude them.

```
Exception: An unhandled exception
#38 app/Http/routes.php(16): {closure}
#37 [Anonymous function](0): call_user_func_array
#36 vendor/laravel/framework/src/Illuminate/Routing/Route.php(155): runCallable
#35 vendor/laravel/framework/src/Illuminate/Routing/Route.php(130): run
#34 vendor/laravel/framework/src/Illuminate/Routing/Router.php(708): Illuminate\Routing\{closure}
#33 [Anonymous function](0): call_user_func
#32 vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(139): Illuminate\Pipeline\{closure}
#31 [Anonymous function](0): call_user_func
#30 vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(103): then
#29 vendor/laravel/framework/src/Illuminate/Routing/Router.php(710): runRouteWithinStack
#28 vendor/laravel/framework/src/Illuminate/Routing/Router.php(675): dispatchToRoute
#27 vendor/laravel/framework/src/Illuminate/Routing/Router.php(635): dispatch
#26 vendor/laravel/framework/src/Illuminate/Foundation/Http/Kernel.php(236): Illuminate\Foundation\Http\{closure}
#25 [Anonymous function](0): call_user_func
#24 vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(139): Illuminate\Pipeline\{closure}
#23 vendor/laravel/framework/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php(50): handle
#22 [Anonymous function](0): call_user_func_array
#21 vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(124): Illuminate\Pipeline\{closure}
#20 vendor/laravel/framework/src/Illuminate/View/Middleware/ShareErrorsFromSession.php(49): handle
#19 [Anonymous function](0): call_user_func_array
#18 vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(124): Illuminate\Pipeline\{closure}
#17 vendor/laravel/framework/src/Illuminate/Session/Middleware/StartSession.php(62): handle
#16 [Anonymous function](0): call_user_func_array
[...]

PR #482, opened 2017-07-26
Exclude anonymous frames from app path

dcramer:
Fixes GH-481

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: php
 sudo: false
-
+dist: precise
 php:
   - 5.3
   - 5.4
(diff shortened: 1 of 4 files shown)


dcramer:
screenshot 2017-07-26 12 04 17

stayallive:
If we have a class set for the exception we could maybe use reflection to find the file for the class to determine a little better if it's in the app context or not... or is that a stupid idea?

```
array:4 [▼
  "function" => "Illuminate\Routing\{closure}"
  "class" => "Illuminate\Routing\Router"
  "type" => "->"
  "args" => array:1 [▶]
]
```

```php
// Return either the filename or false if it's defined in the core or extension.
$filename = (new ReflectionClass('Illuminate\Routing\Router'))->getFileName();
```

Jean85:
I expressed my concerns here: https://github.com/getsentry/sentry-php/issues/481#issuecomment-318302480

PR #485, opened 2017-08-02
Reflection for missing filenames

stayallive:
@dcramer @Jean85 how does this look? It changes the stacktrace for Laravel (5.1 tested) from:

screen shot 2017-08-02 at 16 54 53

To:

screen shot 2017-08-02 at 14 18 48

Which is more correct (considering the app path is set on both correctly).

### DIFF
--- a/lib/Raven/Stacktrace.php
+++ b/lib/Raven/Stacktrace.php
@@ -40,16 +40,27 @@ class Raven_Stacktrace
 
             if (!array_key_exists('file', $frame)) {
                 $context = array();
+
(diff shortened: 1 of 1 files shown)


Jean85:
Seems good to me, but I'm not sure about the impact in non-Laraver apps...

stayallive:
Assuming the class is available in the frame this should be the same improvement.

If we know the class we can extract the file and then enhance the in-app detection.

P.S. I replaced the first screenshot with the current one.

PR #489, opened 2017-08-18
[WIP] Add PHP 7.2 in CI

Jean85:
This is still WIP, it seems that we are getting warning that I would like to investigate further.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -5,7 +5,7 @@ tools:
     php_code_coverage: true
     external_code_coverage:
         timeout: 2400 # There can be another pull request in progress
-        runs: 6 # PHP 5.3 + PHP 5.4 + PHP 5.5 + PHP 5.6 + PHP 7.0 + PHP 7.1 
+        runs: 7 # PHP 5.3 + PHP 5.4 + PHP 5.5 + PHP 5.6 + PHP 7.0 + PHP 7.1 + PHP 7.2  
 
 build:
     environment:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ php:
   - 5.6
   - 7.0
   - 7.1
+  - 7.2
   - nightly
 env:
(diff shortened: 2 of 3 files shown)


Jean85:
There is a session related failure, but there seems to be a few [session fixes in 7.2.0beta3](https://github.com/php/php-src/blob/316802d8f2b07b863f1596cd804db28a183556e5/NEWS#L87), and right now I'm only able to test on beta2 due to Docker Hub availability; I will wait a few days before going forward with this.

Jean85:
It seems that PHPUnit 5 will not allow for a green build under PHP 7.2, due to a deprecated use of `each()`. This is a really big issue, since it could mean that we would not be able to have a CI without dropping older PHP/PHPUnit versions.

Jean85:
I'm missing something...
[test/bootstrap.php](https://github.com/getsentry/sentry-php/blob/9c35545d7be65f42dd7aaf95f4da2d695fdeff38/test/bootstrap.php#L14) does a `session_start()` that triggers all the issues (but is needed for a few tests), but the failure is triggered only on PHP 7.2; this does seems related to the fact that the headers are already sent so the session id cannot be changed anymore..

But how's that it doesn't bother lower PHP versions?

nokitakaze:
```
1) Raven_Tests_ClientTest::testGet_user_data
session_id(): Cannot change session id when headers already sent

/home/travis/build/getsentry/sentry-php/test/Raven/Tests/ClientTest.php:2047
```

This is just a test. I think you can suppress this warning

nokitakaze:
Please apply this patch https://pastebin.com/wUf9jgk7

I want to make PR, but **it** is already created. We need to add PHP 7.2 to supported versions. RC3 is already here http://php.net/archive/2017.php#id2017-09-28-2

We can not wait anymore.

stayallive:
@nokitakaze You can create a PR to this branch if you'd like.

Jean85:
Thanks @nokitakaze for the help!!

PR #490, opened 2017-08-23
Update laravel documentation

yakidahan:
Add: When the exception is authentication, return the original render (that redirect to a login page rather than receiving a 500 error page)

Update: The source code that was updated in this commit:
https://github.com/laravel/laravel/commit/b7d684ef82276d8470e03cb8c538887d5f54e898

### DIFF
--- a/docs/integrations/laravel.rst
+++ b/docs/integrations/laravel.rst
@@ -31,12 +31,12 @@ Add Sentry reporting to ``App/Exceptions/Handler.php``:
 
 .. code-block:: php
 
-    public function report(Exception $e)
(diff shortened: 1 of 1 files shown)


stayallive:
Actually I think I want to propose a IMHO better way to handle this in a application, could you take a peek at #491 where I used these changes as a start but updated the render method with something I think people should be using in production.

stayallive:
@yakidahan could you take a look at #491 to see if those instructions are better?

stayallive:
We updated the docs in #491 using this PR as a start.Hope it's better now, please do let me know if you still see some issues.

PR #491, opened 2017-08-23
Laravel doc updates

stayallive:
Also the Sentry logo update slipped in :)

Laravel 5.5 renders the `500.blade.php` by default and handles login exceptions correctly so no changes needed. For 5.0-5.4 we need to add a bit of code to the render method to make sure the 500 error is rendered for errors that should be reported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <p align="center">
     <a href="https://sentry.io" target="_blank" align="center">
-        <img src="https://sentry.io/_static/getsentry/images/branding/png/sentry-horizontal-black.png" width="280">
+        <img src="https://sentry-brand.storage.googleapis.com/sentry-logo-black.png" width="280">
     </a>
 </p>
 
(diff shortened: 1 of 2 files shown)


stayallive:
Taking a better look at this since I notice 5.4 and 5.5 are both handling this very differently and this approach is not needed for 5.5 since it will handle authenticated exceptions nicely and will convert all other exceptions to a 500 http exception that will render the `500.blade.php` as it should outside debug. However 5.4 does not and the code example here also does not fix this, seeing if we can find a solution that works across all Laravel 5.x versions.

README.md
```diff
@@ -1,6 +1,6 @@
 <p align="center">
     <a href="https://sentry.io" target="_blank" align="center">
-        <img src="https://sentry.io/_static/getsentry/images/branding/png/sentry-horizontal-black.png" width="280">
+        <img src="https://sentry-brand.storage.googleapis.com/sentry-logo-black.png" width="280">
```
Jean85: Oh, finally! :smile:

PR #495, opened 2017-09-02
[2.0] Improve Client and refactor Context

Jean85:
This PR aims at cleaning up the client in respect of code style and to extract the context functionality from it, moving all the logic inside the `Raven\Context` object, that was kinda empty.

This is still WIP since I need to update docs and changelog accordingly.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -259,7 +259,7 @@ There are three primary methods for providing request context:
 .. code-block:: php
 
     // bind the logged in user
-    $client->user_context(array('email' => 'foo@example.com'));
+    $client->setUserContext(array('email' => 'foo@example.com'));
 
     // tag the request with something interesting
     $client->tags_context(array('interesting' => 'yes'));
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -48,11 +48,11 @@ Much of the usefulness of Sentry comes from additional context data with
 the events.  The PHP client makes this very convenient by providing
 methods to set thread local context data that is then submitted
 automatically with all events.  For instance you can use the
-``user_context`` method to add information about the current user:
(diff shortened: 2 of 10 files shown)


ste93cry:
What about making the `Context` class generic so that there aren't methods like `setUserEmail`? I see the context as just a container, so I think it makes sense to let users populate it the way they want

Jean85:
Looking and the [Sentry API](https://docs.sentry.io/clientdev/interfaces/user/), it seems that it recognize the following fields:

 * `id`
 * `email`
 * `ip_address`
 * `username`

So it does make sense to have a specific specialized method for those; still, no one is forcing you to use them, you can still populate the `Context` as you see fit for your application. I just refactored those methods away from the `Client` into the `Context.

ste93cry:
Aren't those properties collected by and stored in the user interface?

Jean85:
Yes they are, so that's why it's needed to specify them with the right name:

 * in the tags:

sentry-user2

 * in the right column, to show the share of users that got the event:
sentry-user

Jean85:
I've noticed that the changelog and the docs are outdated anyway, they should be redone in a separate PR, at the end of the work on 2.0.

After rebasing, this is no longer WIP, please review!

nokitakaze:
Could you check if `email` & `username` could be viewed with Sentry interface?

Jean85:
The username is shown, as seen on the screenshot before. For other infos I cannot check now, but they are shown on the tag for sure.

stayallive:
All user info fields are shown in the Sentry interface (tenant_id and uuid in non-standard but do show up in the user section).

screen shot 2017-09-11 at 16 13 46

Jean85:
Great then! I'll merge now, thanks for the reviews!

PR #496, opened 2017-09-09
Use ramsey/uuid

Jean85:
This PR replaces the `Client::uuid4()` function with a class that just wraps the same function in the ramsey/uuid package. In this way we do no longer need to maintain the code of that function, which was pretty obscure and hard to comprehend.


ste93cry:
This was already done in #487, which is a work in progress

Jean85:
Oh sorry, I didn't notice it. Do you need help to push that PR forward?

PR #500, opened 2017-09-27
Added checker for _convertPath if the path is file

personnage:
Hello!

I want exclude from stacktrace a list files, example

```php
$client = new Raven_Client($dsn, array(
    'excluded_app_paths' => ['vendor', 'tests', 'src/ExcludedFile.php'],
));
```

But if missing check on file,  `_convertPath` added a slash in the end.


stayallive:
Hey @personnage,

Good catch, I am thinking if we could get away with checking if the path ends with `.php` to detech if it's a file since this could add some overhead of statting files each time this check is ran. What do you think?

personnage:
@stayallive, I think you're right.

stayallive:
@personnage there is one more thing I noticed, this is also used to strip the prefix from an path in the stacktrace. This could mean a trace going through your excluded file leaves the filename empty.

We might need to filter out files ending in .php before they are being sent through the `_convertPath` and leave that function as it was. What about:

```php
public function setExcludedAppPaths($value)
{
    if ($value) {
        $excluded_app_paths = array();

        foreach ((array)$value as $path) {
            $excluded_app_paths[] = substr($path, -4) !== '.php' ? $this->_convertPath($path) : $path;
        }
    } else {
        $excluded_app_paths = null;
    }

    $this->excluded_app_paths = $excluded_app_paths;

    return $this;
}
```

This way the other usages of `_convertPath` are not impacted.

(why no closure: It is not possible to use $this from anonymous function before PHP 5.4.0, savages)

personnage:
@stayallive, done

stayallive:
@personnage I added some tests to prove this is working.

@Jean85 / @ste93cry care to take a look at this? This might also be a thing to implement in 2.0 if not already.

personnage:
> personnage I added some tests to prove this is working.

@stayallive And I thank you for that.

> Jean85 / ste93cry care to take a look at this?

Then I look forward to it.

PR #503, opened 2017-10-07
Test PHP 5.3 in separate environment

nokitakaze:
I admit that `precise`, like all Ubuntu LTS, is an important milestone in Linux history, but let's use something newer than the software released more than 5 years ago.

https://blog.travis-ci.com/2017-08-31-trusty-as-default-status

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 language: php
 sudo: false
-dist: precise
+
 php:
-  - 5.3
   - 5.4
   - 5.5
   - 5.6
@@ -21,7 +20,12 @@ matrix:
   include:
     - php: hhvm-3.12
       env: REMOVE_XDEBUG="0" HHVM="1"
-      dist: trusty
+    - php: 5.3
+      env: REMOVE_XDEBUG="0"
+      dist: precise
+    - php: 5.3
+      env: REMOVE_XDEBUG="1"
+      dist: precise
 
 cache:
   directories:


ste93cry:
Why not, 👍

nokitakaze:
Now `trusty` used as a default environment in Travis. So other cases is `trusty`

ste93cry:
>So other cases are trusty

I meant if we could use trusty even for PHP 5.3 builds, which is not the case now even after your changes

nokitakaze:
You mean like this https://travis-ci.org/nokitakaze/sentry-php/builds/284617521 ?

ste93cry:
Yes, according to that build the PHP 5.3 build (which btw failed...) is running on trusty and it's what I'm suggesting to do: drop entirely the precise distribution from our Travis config

nokitakaze:
> which btw failed

https://travis-ci.org/nokitakaze/sentry-php/jobs/284617522

```
Downloading archive: https://s3.amazonaws.com/travis-php- \
archives/binaries/ubuntu/14.04/x86_64/php-5.3.tar.bz2
0.06s
$ curl -s -o archive.tar.bz2 $archive_url && tar xjf archive.tar.bz2 --directory /
bzip2: (stdin) is not a bzip2 file.
tar: Child returned status 2
tar: Error is not recoverable: exiting now
```

There is no PHP 5.3 in `trusty`

ste93cry:
>There is no PHP 5.3 in `trusty`

Oh, ok 😞

>Once complete, we will continue to offer fully-virtualized Ubuntu Precise until March 2018 at the earliest

This quote is taken from the blog post you linked initially: it seems that anyway the `precise` distribution will be dropped in 2018, so we have a problem here. The folks at WordPress discussed about the same topic [here](https://core.trac.wordpress.org/ticket/41292), maybe it can be of help

nokitakaze:
I believe we will drop 1.x as deprecated and release 2.0. So we set some commit as a last 1.x commit, and PHP 5.3 will get Sentry from this commit. So PHP 5.3 will work and we could develope our project

.travis.yml
```diff
@@ -21,7 +20,12 @@ matrix:
   include:
     - php: hhvm-3.12
       env: REMOVE_XDEBUG="0" HHVM="1"
-      dist: trusty
+    - php: 5.3
```
Jean85: You shouldn't specify 5.3 here, you can leave it up and add the `dist` key under the `php` version entry.

nokitakaze: Could you show me how?

```yml
language: php
sudo: false

php:
  - 5.3
    dist: precise
  - 5.4
  - 5.5
  - 5.6
  - 7.0
  - 7.1
  - nightly
```

does not work: https://travis-ci.org/nokitakaze/sentry-php/builds/285444487

nokitakaze: Don't you think you are mistaken? https://docs.travis-ci.com/user/reference/trusty#PHP-images official guide:
```yml
matrix:
  include:
    - php: 5.2
      dist: precise
    - php: 5.3
      dist: precise
```

Jean85: My bad, I thought otherwise.

Jean85:
We should put more effort in the 2.0 release, where 5.3 is dropped. It's too old right now!

nokitakaze:
#458, #454 waits for #472, #487, because I don't know which changes It will bring.

Jean85:
I would love to push those forward... @ste93cry can we help you?

PR #506, opened 2017-10-19
Use namespaced TestCase

MarioBlazek:
This PR add PHPUnit forward compatibility layer. Now all tests extend from namespaced PHPUnit `TestCase`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     ],
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^1.8.0",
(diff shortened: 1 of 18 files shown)


Jean85:
Hi there, thanks for the contribution!
This was already done on the 2.0 branch, but it's still useful!

nokitakaze:
I don't think we need it in stable version.

ste93cry:
>I don't think we need it in stable version.

We don't necessarily need it, but it's good to have it 👍

nokitakaze:
Oh, I haven't seen changes. It only change parent class name.

PR #507, opened 2017-10-28
remove suggestion of : immobiliare/sentry-php

huglester:
Hello,

so basically, if we managed to install via composer require, means we have PHP >= 5.3
then we don't need to suggest installing this?

Or am I missing something obvious

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,6 @@
         "ext-hash": "*",
         "ext-json": "*",
         "ext-mbstring": "*",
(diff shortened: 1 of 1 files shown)


ste93cry:
It's correct to assume that Composer does not work on PHP < 5.3, so I'm 👍 to remove the suggestion. Maybe we can put it in a section of the docs, but imho it makes no sense as PHP 5.2 is old, not supported and I don't think that there are a lot of people that uses it so the effort would not be justified

PR #505, opened 2017-10-11
[2.0] Prevent segmentation fault in Reflection->getParameters()

Caffe1neAdd1ct:
https://github.com/getsentry/sentry-php/pull/504

When a native php function has been disabled via disabled_functions (php.ini) calling getParameters() on the reflection object of the disabled function causes a segmentation fault on PHP 5.6.31.

Reproducible on both Arch Linux and CentOS 6 with the following script:

php.ini
```
disable_functions = "passthru"
```

console.php
 ```php
<?php

require_once 'vendor/autoload.php';

use Raven_Client;

$connection = false;
$client = new Raven_Client($connection);
$client->environment = 'test';
$client->install();

passthru('clear');
```

Run straight through either php or php-cli `php56 console.php` gives a segmentation fault as follows:
```
[1]    22373 segmentation fault (core dumped)  php56 console.php
```

Testing through php 7 does not segfault and reflection seems to be able to cope with the disabled function:

```
PHP Warning:  passthru() has been disabled for security reasons in /home/kandrews/NetBeansProjects/compareandrecycle.co.uk/console.php on line 12
PHP Stack trace:
PHP   1. {main}() /home/kandrews/NetBeansProjects/compareandrecycle.co.uk/console.php:0
PHP   2. passthru() /home/kandrews/NetBeansProjects/compareandrecycle.co.uk/console.php:12
```

PR #511, opened 2017-10-30
[2.0] Yoda Style, Small fixes in PHP Doc & Code coverage

nokitakaze:
* Small PHP doc fixes and fixes for Intellij Idea
* PHP CS: Explictly **enable** "yoda style"
* Additional code coverage

I think PHP CS implicit updated their presets. I ~~disabled~~ explicitly enabled them.

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -9,8 +9,7 @@ return PhpCsFixer\Config::create()
         'concat_space' => ['spacing' => 'one'],
         'ordered_imports' => true,
         'random_api_migration' => true,
-        'phpdoc_var_without_name' => false,
-        'yoda_style' => false,
+        'yoda_style' => true,
     ])
     ->setRiskyAllowed(true)
     ->setFinder(
--- a/.php_cs
+++ b/.php_cs
@@ -9,6 +9,8 @@ return PhpCsFixer\Config::create()
         'concat_space' => ['spacing' => 'one'],
         'ordered_imports' => true,
         'random_api_migration' => true,
+        'phpdoc_var_without_name' => false,
+        'yoda_style' => false,
     ])
     ->setRiskyAllowed(true)
     ->setFinder(
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -329,7 +329,7 @@ class Client
             $formatted_message = $message;
         }
 
-        if ($data === null) {
+        if (null === $data) {
             $data = [];
             // support legacy method of passing in a level name as the third arg
         } elseif (!is_array($data)) {
@@ -364,7 +364,7 @@ class Client
             return null;
         }
 
-        if ($data === null) {
+        if (null === $data) {
             $data = [];
         }
 
@@ -404,7 +404,7 @@ class Client
         $data['exception'] = [
             'values' => array_reverse($exceptions),
         ];
-        if ($logger !== null) {
+        if (null !== $logger) {
             $data['logger'] = $logger;
         }
 
@@ -460,7 +460,7 @@ class Client
             ],
         ];
 
-        if ($engine !== '') {
+        if (!empty($engine)) {
             $data['sentry.interfaces.Query']['engine'] = $engine;
         }
 
@@ -507,7 +507,7 @@ class Client
                 $header_key =
                     str_replace(' ', '-', ucwords(strtolower(str_replace('_', ' ', substr($key, 5)))));
                 $headers[$header_key] = $value;
-            } elseif (in_array($key, ['CONTENT_TYPE', 'CONTENT_LENGTH']) && $value !== '') {
+            } elseif (in_array($key, ['CONTENT_TYPE', 'CONTENT_LENGTH']) && !empty($value)) {
                 $header_key = str_replace(' ', '-', ucwords(strtolower(str_replace('_', ' ', $key))));
                 $headers[$header_key] = $value;
             }
@@ -641,7 +641,7 @@ class Client
             $data['breadcrumbs'] = iterator_to_array($this->recorder);
         }
 
-        if ((!$stack && $this->config->getAutoLogStacks()) || $stack === true) {
+        if ((!$stack && $this->config->getAutoLogStacks()) || (true === $stack)) {
             $stack = debug_backtrace();
 
             // Drop last stack
@@ -850,17 +850,17 @@ class Client
      */
     protected function isHttps()
     {
-        if (!empty($_SERVER['HTTPS']) && $_SERVER['HTTPS'] !== 'off') {
+        if (!empty($_SERVER['HTTPS']) && ('off' !== $_SERVER['HTTPS'])) {
             return true;
         }
 
-        if (!empty($_SERVER['SERVER_PORT']) && $_SERVER['SERVER_PORT'] == 443) {
+        if (!empty($_SERVER['SERVER_PORT']) && (443 == $_SERVER['SERVER_PORT'])) {
             return true;
         }
 
         if (!empty($this->config->isTrustXForwardedProto()) &&
             !empty($_SERVER['X-FORWARDED-PROTO']) &&
-            $_SERVER['X-FORWARDED-PROTO'] === 'https') {
+            ('https' === $_SERVER['X-FORWARDED-PROTO'])) {
             return true;
         }
 
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -46,7 +46,7 @@ class Client
     const MESSAGE_LIMIT = 1024;
 
     /**
-     * @var Recorder The bredcrumbs recorder
+     * @var Recorder The breadcrumbs recorder
      */
     protected $recorder;
 
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -1673,6 +1673,26 @@ class ClientTest extends TestCase
         $this->assertFalse($client->getSerializer()->getAllObjectSerialize());
         $this->assertFalse($client->getReprSerializer()->getAllObjectSerialize());
     }
+
+    public function testClearBreadcrumb()
+    {
+        $client = ClientBuilder::create()->getClient();
+        $client->leaveBreadcrumb(
+            new \Raven\Breadcrumbs\Breadcrumb(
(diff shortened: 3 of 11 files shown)


nokitakaze:
Okay. I think I will fix code coverage

lib/Raven/Client.php
```diff
@@ -540,15 +540,16 @@ protected function getUserData()
     {
         $user = $this->context->getUserData();
         if (empty($user)) {
-            if (!function_exists('session_id') || !session_id()) {
```
nokitakaze: Please, notice this change

Jean85: What's the intended purpose of this?

nokitakaze: I think field `user` could exist even without logging. Anyway, in HTTP(S) requests we should save users's IP address

ste93cry: This whole method will be useless after #487 will be merged because the user data will be retrieved from a middleware and the [`session_id`](http://php.net/manual/en/function.session-id.php) function won't be used anymore anyway

nokitakaze: Okay, please be sure you will test IPv4 and IPv6 as well (i.e. `8.8.8.8`, `2001:db8:0:0:0:0:2:1`, `2001:db8::2:1`)

.php_cs
```diff
@@ -9,6 +9,8 @@ return PhpCsFixer\Config::create()
         'concat_space' => ['spacing' => 'one'],
         'ordered_imports' => true,
         'random_api_migration' => true,
+        'phpdoc_var_without_name' => false,
+        'yoda_style' => false,
```
ste93cry: Why are these options being set to `false`? We don't need PHPDoc blocks which list params and not their description and it's better to write conditions in [yoda style](https://en.wikipedia.org/wiki/Yoda_conditions) than not

lib/Raven/Configuration.php
```diff
@@ -824,7 +824,8 @@ private function configureOptions(OptionsResolver $resolver)
             return true;
         });
 
-        $resolver->setNormalizer('server', function (Options $options, $value) {
+        $resolver->setNormalizer('server', function (/* @noinspection PhpUnusedParameterInspection */
```
ste93cry: I know @nokitakaze you like to have your IDE inspections to be all green, but we can't fill the whole project of annotations just for this... Also, it makes the code less readable than before to have this comment there. At least would it be possible to move it inside the function and get the same result?

lib/Raven/Processor.php
```diff
@@ -31,6 +31,7 @@ public function __construct(\Raven\Client $client)
      * Override the default processor options.
      *
      * @param array $options Associative array of processor options
+     * @codeCoverageIgnore
```
ste93cry: This is not really needed as the whole method will be removed after #472 will be merged

tests/ClientTest.php
```diff
@@ -1673,6 +1673,26 @@ public function testSetAllObjectSerialize()
         $this->assertFalse($client->getSerializer()->getAllObjectSerialize());
         $this->assertFalse($client->getReprSerializer()->getAllObjectSerialize());
     }
+
+    public function testClearBreadcrumb()
```
ste93cry: This is not needed as the `Recorder` class is already unit tested, so this is useless. Maybe you should test that the `leaveBreadcrumb` method calls the recorder, but I don't know if it's worth it or was already done somewhere else

.php_cs
nokitakaze: > We don't need PHPDoc blocks which list params 

Maybe you don't get it. `phpdoc_var_without_name = true` make
```php
/**
 * @var integer $i
 */
```

be wrong block. It make do `@var integer` without name

nokitakaze: > Yoda style

It this something that someone did on purpose?

lib/Raven/Configuration.php
nokitakaze: >we can't fill the whole project of annotations just for this

Why? I think I am not the only one who uses Idea

> would it be possible

I'm afraid no

.php_cs
ste93cry: It's correct to report that example as wrong, because imho every variable, property or param should have a description. Just to be sure it's clear:

Wrong:
```php
/**
 * @var integer $i
 */
```

Correct:
```php
/**
 * @var integer $i ...
 */
```

lib/Raven/Configuration.php
Jean85: The point is to not add tool-specific stuff to the codebase. I'm :-1: too.

tests/ClientTest.php
nokitakaze: I don't test `Recorder::clear`, I'm testing `Client::clearBreadcrumbs`

lib/Raven/Configuration.php
ste93cry: >Why? I think I am not the only one who uses Idea

I use PHPStorm too, and I think a lot of other people too, but this does not means that we absolutely need to have all inspections to pass just because we don't use a parameter that unfortunately must be in the signature because of it's position

tests/ClientTest.php
ste93cry: The `Client::clearBreadcrumbs` method just calls the `Recorder::clear` method and this one is already tested, so what's the advantage of adding this test?

.php_cs
nokitakaze: > it was done

And how you could merge something, that could not be passed by tests? 

https://github.com/getsentry/sentry-php/blob/dcc287f6579b765c7935e0e60fa642ae029b834f/lib/Raven/Serializer.php#L86
https://github.com/getsentry/sentry-php/blob/dcc287f6579b765c7935e0e60fa642ae029b834f/lib/Raven/ReprSerializer.php#L22 the whole block

and many many others

tests/ClientTest.php
nokitakaze: Erm... To test that `Client::clearBreadcrumbs` really calls `Recorder::clear`?

.php_cs
ste93cry: >And how you could merge something, that could not be passed by tests?

I don't understand what you mean... Yoda conditions are just a way to write safer code by moving any constant to the left part of the condition. As an example, you could miss a `=` in an equal condition, and the value would then be assigned to the variable instead of compared to

```php
if ($var = true) {
    // $var is now true, instead of having been compared to
}
```

```php
if (true = $var) {
    // syntax error
}
```

Now, I understand that in an hypothetical world this would never happen, but it's better to prevent than to cure. To better understand the motivations behind yoda style btw, you can read the Wikipedia link I posted above

nokitakaze: Please look at this. This is CURRENT commit

https://github.com/getsentry/sentry-php/blob/dcc287f6579b765c7935e0e60fa642ae029b834f/lib/Raven/Serializer.php#L86
https://github.com/getsentry/sentry-php/blob/dcc287f6579b765c7935e0e60fa642ae029b834f/lib/Raven/ReprSerializer.php#L22

tests/ClientTest.php
ste93cry: Then there is no need imho to duplicate the whole block of code that tests that breadcrumbs are removed, you could just check that the `Recorder::clear` method is called and let the unit tests of that class catch any error if something doesn't work inside it 😃 The point is that imho this whole block of code is just a duplication of https://github.com/getsentry/sentry-php/blob/dcc287f6579b765c7935e0e60fa642ae029b834f/tests/Breadcrumbs/RecorderTest.php#L73-L90

.php_cs
Jean85: @ste93cry , Nokitakaze is telling us that currently yoda conditions are not enabled, since the current codebase doesn't use them, and the code style checks are not failing.

@nokitakaze please try to explain your thoughts better, we are often having issues trying to understand your writing.

tests/ClientTest.php
nokitakaze: > you could just check that the `Recorder::clear` method is called

How? With DI? Of course, but I though it is easy to check empty/not empty instead of injecting an object.

.php_cs
ste93cry: Ok now I get the point. Sorry for misunderstanding your messages @nokitakaze but as @Jean85 said ofter I have issues understanding what you write

Jean85: I agree with not changing the current setting. We should otherwise investigate onto why that rule seems not to be enforced everywhere.

nokitakaze: There is the same problem in #505. https://travis-ci.org/getsentry/sentry-php/jobs/286575316

I thought PHP CS enabled these options, and so I disabled them

Jean85: Oh so IT IS currently failing. We should fix the CS then.

nokitakaze:
I should ask again. How it was **possible** to change code style settings without changing code base? 
And I should tell my opinion again. We must disable this settings until the time we explicitly enable it.

ste93cry:
>I should ask again. How it was possible to change code style settings without changing code base?

As noone here changed the settings, I think the only feasible reason is that PHP-CS changed the default values and this broke our builds.

>We must disable this settings until the time we explicitly enable it.

We already discussed above about disabling options and I though we agreed it's better to leave them enabled (thus fallbacking to the default values). I'm +1 on fixing the whole build with the updated code style rules instead and merge this as soon as possible

nokitakaze:
Okay. I need a hour to change code base

ste93cry:
As `'yoda_style' => true,` is the default we should not specify it right?

nokitakaze:
> is the default

I did it on purpose. About a month ago the default value was a `false`. Now it is `true`. After a month or something like that some smart ass could stick it back to `false`

ste93cry:
Indeed, but then we should do the same for all config options of the `@Symfony` ruleset or if they change the default value we will fall again in this situation...

Jean85:
I'm :+1: with specifying this specific rule since it caused those issue for us. If it's ok, we can remove it later.

ste93cry:
What I meant was that we should really not use that ruleset and instead list all options explicitly if we want to avoid depending on a such large set of values that can change over time. By the way you have my approval for merging

PR #514, opened 2017-11-08
Do not capture E_ERROR type as fatal error

### LINKED ISSUE #408: Duplicate sentry issues
When we call to an undefined function we get duplicate issues in Sentry.
One says `Error: Call to undefined function function_name()` and the other one says `ErrorException: Uncaught Error: Call to undefined function function_name()`

SDK-version is 1.6.1 and the code for initialise sentry is:
```php
if ( defined( 'RAVEN_DSN' ) ) {
	// Setup Raven with our DSN key.
	$sentry_client = ( new Raven_Client( RAVEN_DSN ) )->install();

	if ( defined( 'WP_ENV' ) ) {
		$sentry_client->setEnvironment( WP_ENV );
	}
}
```

stayallive:
This should fix #408, can you guys please test it and and let me know if this is the way we should handle this or an alternative. 

IMO this gist is that we should ignore E_ERROR exceptions as fatal in PHP 7.0+ since they can be captured by userland and are thus going through the regular exception handler before entering the fatal error handler thus creating duplicated exceptions.

This PR as a side effect also unifies the way the PHP version is checked.

### DIFF
--- a/lib/Raven/ErrorHandler.php
+++ b/lib/Raven/ErrorHandler.php
@@ -140,6 +140,13 @@ class Raven_ErrorHandler
 
     public function shouldCaptureFatalError($type)
     {
+        // Do not capture E_ERROR since those can be caught by userland since PHP 7.0
+        // E_ERROR should already be handled by the exception handler
+        // This prevents duplicated exceptions in PHP 7.0+
+        if (version_compare(phpversion(), '7.0', '>=') && $type === E_ERROR) {
+            return false;
+        }
+
         return $type & $this->fatal_error_types;
     }
 
--- a/lib/Raven/ErrorHandler.php
+++ b/lib/Raven/ErrorHandler.php
@@ -143,7 +143,7 @@ class Raven_ErrorHandler
         // Do not capture E_ERROR since those can be caught by userland since PHP 7.0
         // E_ERROR should already be handled by the exception handler
         // This prevents duplicated exceptions in PHP 7.0+
-        if (version_compare(phpversion(), '7.0', '>=') && $type === E_ERROR) {
+        if (PHP_VERSION_ID >= 70000 && $type === E_ERROR) {
             return false;
         }
 
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -561,10 +561,10 @@ class Raven_Client
     /**
      * Log an exception to sentry
      *
-     * @param Exception $exception The Exception object.
-     * @param array     $data      Additional attributes to pass with this event (see Sentry docs).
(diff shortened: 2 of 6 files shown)


Jean85:
I'll test this ASAP!

Maybe we can add a thing to this PR: update the PHPDoc of the `Client::captureException()` method to accept `\Throwable|\Exception $exception` here: https://github.com/getsentry/sentry-php/blob/0baca06aa08bb622338deb000cad7ffa2a258e29/lib/Raven/Client.php#L561-L570

What do you think?

Jean85:
Also: maybe we should add a (unit) test?

stayallive:
I think that's a good idea. 

For reference: why not remove `\Throwable` (this was my first thought)? This is because `\Throwable` was introduced in PHP 7.0 so `\Exception` remains for PHP < 7.0.

stayallive:
@Jean85 I changed the test to check if the output of the `shouldCaptureFatalError` is correct. But not actually checking if this works correctly, any idea on how to test this? Cannot just call an undefined method since that does not seem to trigger this (tried adding a failing test before).

lib/Raven/ErrorHandler.php
```diff
@@ -140,6 +140,13 @@ public function handleFatalError()
 
     public function shouldCaptureFatalError($type)
     {
+        // Do not capture E_ERROR since those can be caught by userland since PHP 7.0
+        // E_ERROR should already be handled by the exception handler
+        // This prevents duplicated exceptions in PHP 7.0+
+        if (version_compare(phpversion(), '7.0', '>=') && $type === E_ERROR) {
```
Jean85: Maybe `if (PHP_VERSION_ID >= 70200)` is faster? That constant is available since PHP 5.2, so that's not an issue to use it.

Jean85:
Sorry, I missed the changed test, I think it's enough for now!

lib/Raven/ErrorHandler.php
stayallive: Looks like my PhpStorm exceptions thought the same way. I changed it everywhere to stay consistent.

mfb:
What are you all doing to capture fatal PHP errors, such as allowed memory size exhausted?  To capture this fatal error successfully, I had to comment out the `return FALSE;` in shouldCaptureFatalError():

```
if (PHP_VERSION_ID >= 70000 && $type === E_ERROR) {
  // return false;
}
```

sample code:

```
<?php
include '../vendor/autoload.php';
$client = new Raven_Client('...');
$client->install();
while (TRUE) {
  $a[] = 11111111111;
}
```

PR #515, opened 2017-11-09
Add setters for Raven Serializes

atuzlukov:
Why? 
Add the ability to install a custom serializer. For example, instead of the phrase "Array of length 3" put a serialized array, so as not to lose valuable data during the logging

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -44,7 +44,13 @@ class Raven_Client
     protected $error_handler;
     protected $error_types;
 
+    /**
+     * @var Raven_Serializer
+     */
     protected $serializer;
+    /**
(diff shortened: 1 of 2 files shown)


stayallive:
I would be fine with this I think. @Jean85 what are your thoughts?

Assuming this is already implemented in 2.0, but still be usefull in 1.x.

Jean85:
Nope, this is not implemented in 2.0, so it could be ported there to.

Can we add some tests?

stayallive:
@atuzlukov would you be able to add some tests for this functionality?

atuzlukov:
@stayallive @Jean85 Added tests for serialize setters

PR #510, opened 2017-10-30
Bugfix: Raven_CurlHandler -> join_timeout

ovaldi:

PR #526, opened 2017-12-22
[2.0] Merge master differences in 2.x branch

Jean85:
This PR is intended to port all differences from 1.x to 2.x, so we do not regress on bugs and stuff in the new release. This is a recap of what I've covered so far, with all PR that we merged in master and not on 2.x:

Last master commit merged into 2.0: a99776492b1c027b3fe14591d46526e2ebe96468

- [x] #482: Exclude anonymous frames from app path (already taken care of in tests: https://github.com/getsentry/sentry-php/blob/873ab3c9915dba04e37a74b5cb3f9b5d3101e9cf/tests/StacktraceTest.php#L231)
- [x] #483: multiple sanitize (already taken care of in tests: https://github.com/getsentry/sentry-php/blob/873ab3c9915dba04e37a74b5cb3f9b5d3101e9cf/tests/Processor/SanitizeDataProcessorTest.php#L47-L51)
- [x] #489: ported with 38be7430e020f77df6401cdf11be760f3b851080
- [x] #493: non existent file for stacktrace (already taken care of: https://github.com/getsentry/sentry-php/blob/873ab3c9915dba04e37a74b5cb3f9b5d3101e9cf/lib/Raven/Stacktrace.php#L227)
- [x] #500: allow single files in `excluded_app_paths` (ported with aad776d)
- [x] #501: no longer applicable (Autoloader is done by composer)
- [x] #504: segfault in stacktrace with missing function (already taken care of with generic catch: https://github.com/getsentry/sentry-php/blob/873ab3c9915dba04e37a74b5cb3f9b5d3101e9cf/lib/Raven/Stacktrace.php#L367-L372)
- [x] #506: upgrade to PHPUnit 6 (already done at base)
- [x] #509: remove suggestion of 5.2 port (already done: https://github.com/getsentry/sentry-php/blob/873ab3c9915dba04e37a74b5cb3f9b5d3101e9cf/composer.json#L35-L37)
- [x] #510: curl related, no longer applicable
- [x] #514: avoid double-capture with PHP 7.0 (done in 20c79ed)
- [x] #515: set serializers (done in d604f05)
- [x] #522: handle nullable values for DSN (done in f844b5d)
- [x] #525: release 1.8.2
- [x] #527: default values as array for frames (porting only the regression test, done in 6c7893c) 
- [x] #538: serialize breadcrumbs too
- [x] #541: 1.8.3 release


.travis.yml
```diff
@@ -4,11 +4,9 @@ php:
   - 5.6
   - 7.0
   - 7.1
-  - nightly
```
nokitakaze: You should not delete `nightly` version. It is not `7.2`, it is current `dev` version, so now it is `7.3`. We always should know how our code running on new version of PHP

Jean85: I disagree. 7.3 is now very far from being usable, and Travis will tag it while in beta stage, so we can add it at that point. (which will be in at least 8-9 months in the future from now).

Jean85:
Done! No longer WIP, please review!

.scrutinizer.yml
```diff
@@ -5,7 +5,7 @@ tools:
     php_code_coverage: true
     external_code_coverage:
         timeout: 2400 # There can be another pull request in progress
-        runs: 3 # PHP 5.6 + PHP 7.0 + PHP 7.1
+        runs: 3 # PHP 5.6 + PHP 7.0 + PHP 7.1 + PHP 7.2
```
ste93cry: Shouldn't this be 4? According to the comment and the count of the option it was 3 before adding PHP 7.2

Jean85: Yep, that's right!! Fixing it now

docs/integrations/laravel.rst
```diff
@@ -239,14 +252,14 @@ In the following example, we'll use a middleware:
         public function handle($request, Closure $next)
         {
             if (app()->bound('sentry')) {
-                /** @var \Raven\Client $sentry */
+                /** @var \Raven_Client $sentry */
```
nokitakaze: I think it is a mistake

Jean85: Yep it is! Fixing it, thanks for spotting it!

Jean85: Umh... Wait a minute... It's like that because I copy-pasted from master, but simply replacing that may not work... The docs need to be fully updated to how the 2.x client will work, so maybe we can postpone that to a later PR... Also, @stayallive knows that Laravel integration way better than me!

stayallive: Will for sure take a look at this very soon!

nokitakaze: @stayallive ?

Jean85:
I've resolved conficts. Can we merge this now? I (and @ste93cry too I believe) would like to resolve the docs issues for the 2.0 in a separate PR

Jean85:
I did a mistake during the merge, it should be green now!

Jean85:
I've merged more stuff from master in this PR, it's now up to date.

@nokitakaze can you approve this? I wouldn't like to have more conflicts to solve here...

ste93cry:
@Jean85 what about leaving this PR open until 2.x is ready to be released in alpha so that we can track the differences between the this branch and master? I know it would be difficult to fix conflicts, but maybe it's better to do it one time for all

Jean85:
It's not too difficult, the pain could be that we would be having PRs targeting wrong stuff or missing regression tests that we already have in here.

I can easily reopen a new PR like this one to start tracking new differences and merge it later, but this is already big IMO.

PR #527, opened 2018-01-04
default value for array should be array

DeyV:
Thix Fix notice  array_key_exists() expects parameter 2 to be array, null given

This fix should be added to 1.8 and 2.0 branches


Jean85:
The 2.x branch is already ok, it has a different flow with already multiple `isset` checks.

Thanks for spotting this BTW! Can you add a regression test where you pass an empty array as `$frames`?

DeyV:
> The 2.x branch is already ok

Great

> Can you add a regression test 

of course

Issue #528, opened 2018-01-09
Post data getting modified (periods -> underscores) in post body


MeredithAnya:
Post data that has periods such as `value.1` is getting changed to `value_1`.

cc @stayallive @Jean85

stayallive:
I am not sure if this is a client side error, possibly on the server side something like this could be happening, anyone in @getsentry/sdks knows something like this is happening on the server side?

Looks like this is happening in the "Body" section but not in the stacktrace variables so my assumption is that the data correctly arrives but the body has some display/processing magic going on replacing `.` with a `_`.

ste93cry:
Is this still relevant as of today? Can you also try if `2.0-beta1` has the same issue?

ste93cry:
Closing this issue due to inactivity. Feel free to reopen if it still happens

PR #531, opened 2018-01-15
Fix POST data from http request for non-multipart/form-data enctype o…

ents:
POST data is broken if logging non-multipart/formdata requests , because PHP try to parse it as multipart/formdata
To fix this issue we need to pass raw POST data to sentry. But php://input is readable only once at PHP<5.6, so we need to write a condition for this case

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -746,8 +746,13 @@ class Raven_Client
 
         // dont set this as an empty array as PHP will treat it as a numeric array
         // instead of a mapping which goes against the defined Sentry spec
-        if (!empty($_POST)) {
+        if (PHP_VERSION_ID < 50600 && !empty($_POST)) {
             $result['data'] = $_POST;
+        } else {
+            $result['data'] = file_get_contents("php://input");
+            if (!empty($_POST)) {
+                $result['POST'] = $_POST;
+            }
         }
         if (!empty($_COOKIE)) {
             $result['cookies'] = $_COOKIE;
--- a/test/Raven/Tests/ClientTest.php
+++ b/test/Raven/Tests/ClientTest.php
@@ -824,8 +824,8 @@ class Raven_Tests_ClientTest extends \PHPUnit\Framework\TestCase
         );
 
         if (PHP_VERSION_ID < 50600) {
-             $expected['request']['data'] = $expected['request']['POST'];
-             unset($expected['request']['POST']);
+            $expected['request']['data'] = $expected['request']['POST'];
+            unset($expected['request']['POST']);
         }
 
         $client = new Dummy_Raven_Client();
--- a/test/Raven/Tests/ClientTest.php
+++ b/test/Raven/Tests/ClientTest.php
@@ -805,9 +805,10 @@ class Raven_Tests_ClientTest extends \PHPUnit\Framework\TestCase
                 'method' => 'PATCH',
                 'url' => 'https://getsentry.com/welcome/',
                 'query_string' => 'q=bitch&l=en',
-                'data' => array(
+                'POST' => array(
                     'stamp'           => '1c',
                 ),
+                'data' => '',
                 'cookies' => array(
                     'donut'           => 'chocolat',
                 ),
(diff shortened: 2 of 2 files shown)


Jean85:
This approach is high risk, you're copying raw data from the POST.

BTW, can you first file an issue so we can track this issue down? It seems to me very strange that the POST data is not readable.

ents:
POST data is a valid JSON string according to https://en.wikipedia.org/wiki/JSON-RPC. But it can't be parsed by PHP
For example:
curl -X POST --data '{"jsonrpc": "2.0", "method": "subtract", "params": {"minuend": 42, "subtrahend": 23}, "id": 3}'  http://localhost/

Jean85:
I've tested this, and I've come to this:

schermata del 2018-01-15 17-42-14

The data is correctly parsed, but is filled in as an array key, so the output is not so nice. Also, content type is matched as `application/x-www-form-urlencoded`. We may be able to detect this kind of situations investigating on the kind of data that is present in `$_POST`, since it's like that.

Raw data is a no-go IMHO.

ste93cry:
>POST data is broken if logging non-multipart/formdata requests , because PHP try to parse it as multipart/formdata

I'm not sure about what "POST data is broken" means, but according to the [specs](http://www.jsonrpc.org/historical/json-rpc-over-http.html#http-header) the `Content-Type` should be one of these values: `application/json-rpc`, `application/json` or `application/jsonrequest`. By the way, PHP will not decode automatically a JSON when reading the `POST` or `php://input` stream, and I don't think we should neither

ents:
"POST data is broken" means that:
1. BODY is incorrect at sentry. Space is replaced by underscore
image
2. CURL command is incorrect at sentry
image

If header "Content-type: application/json" is present in request, the result is even worst. $_POST will be empty. So we will see nothing at sentry (--data argument si not present):
image

stayallive:
Someone else saw something similair (#528), however I don't see how reading raw input would fix this... we should investigate why no body is sent when the content-type is json because that sounds like something we definitely should look into!

Looks like this is on the server side where JSON is probably double escaped or something else is happening, we are not (AFAIK) replacing spaces with underscores in the SDK.

Can someone @getsentry/engineering (right team?) pitch in here if something on the server side could cause spaces be replaced with underscores? And what we could to do to prevent this.

ents:
raw input - it's just string passed to HTTP BODY
PHP decodes only urlencoded string. In other cases - not decodes for better performance (thats why $_POST is empty, if application/json header was sent)

With this fix we will not rely to PHP internal parser of POST data, and return it as is. And CURL helper will be correct at sentry:
image
image

Jean85:
@ents still, this fix is too dangerous. As I mentioned before, we should still use `$_POST` but handle the JSON correctly.

stayallive:
Sorry, I missed that JSON was sent with an incorrect content-type, which causes it to decode/sent incorrectly, so we can fix this in the SDK.

If the server display's / formats the data according to the post type it might be safe to do send the raw post data and let the server deal with the formatting / parsing. 

It looks like (no Python expert) Raven Python (on Django) does something similair falling back on raw post data and not doing anything with it. 

https://github.com/getsentry/raven-python/blob/68487496abc11a2c62effebc7e9dec1a16ef0544/raven/contrib/django/client.py#L201

ents:
@stayallive, in Python library they try to use internal python parser, if it does not work they take raw data. php://input - it's raw data in PHP
image

ents:
PHP variable $_POST may be used only for application/x-www-form-urlencoded or multipart/form-data
http://php.net/manual/en/reserved.variables.post.php

ryantology:
This is a critical feature. I'm a bit surprised this has not been addressed before. Anyone using php as a json/rest API needs to read input with the raw input. I know this 'sounds scary' :) The fix for this is the send it through the json_decode before sending appending in the payload array. You can see how all frameworks use this method for handling application/json data input. Here is the implementation from another project that we are trying to switch away from https://github.com/bugsnag/bugsnag-php/blob/master/src/Request/BasicResolver.php
Also in symfony
https://github.com/symfony/http-foundation/blob/master/Request.php#L1504

If you would like a change to this so that the input is sanitized through json_decode I can make a PR.

Unfortunately, without this feature we are unable to continue using sentry :( Hopefully this can get resolved quickly.

Jean85:
Superseeded by #546

PR #533, opened 2018-01-18
[2.0] Make the behaviour of the RemoveCookiesProcessor configurable

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

In PR #406 a processor to sanitize all cookies by replacing their values with a set of asterisks was added. Apart from the naming that I've changed to better reflect what this class really does (it does not remove any cookie, it just sanitizes them), The current result that is shown when looking at an issue in Sentry is something along these lines in Sentry:

Cookie Name | Cookie Value
| ------------- | -------------
******** |

Not really useful, uh? This PR makes the processor configurable a bit. By default it still works as before, but two new options named `only` and `except` are now available and you can't use both of them at the same time. The first one let users choose which cookies should be sanitized while the second one let users choose which cookies should not. To better explain the concept you can find below some examples

- **`only` option and `except` option both configured to `[]` (default behavior)**

  Cookie Name | Cookie Value
  | ------------- | -------------
  foo | ********
  bar | ********

- **`only` option configured to `['foo']`**

  Cookie Name | Cookie Value
  | ------------- | -------------
  foo | ********
  bar | baz

- **`except` option configured to `['foo']`**

  Cookie Name | Cookie Value
  | ------------- | -------------
  foo | baz
  bar | ********

### DIFF
--- a/lib/Raven/Processor/SanitizeCookiesProcessor.php
+++ b/lib/Raven/Processor/SanitizeCookiesProcessor.php
@@ -16,12 +16,12 @@ use Symfony\Component\OptionsResolver\Exception\InvalidOptionsException;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
- * This processor removes all the cookies from the request to ensure no sensitive
- * informations are sent to the server.
+ * This processor sanitizes the cookies to ensure no sensitive information are
+ * sent to the server.
  *
  * @author Stefano Arlandini <sarlandini@alice.it>
  */
-final class RemoveCookiesProcessor implements ProcessorInterface
+final class SanitizeCookiesProcessor implements ProcessorInterface
 {
     /**
      * @var array The configuration options
--- a/tests/Processor/SanitizeCookiesProcessorTest.php
+++ b/tests/Processor/SanitizeCookiesProcessorTest.php
@@ -15,9 +15,9 @@ use PHPUnit\Framework\TestCase;
 use Raven\Client;
 use Raven\ClientBuilder;
 use Raven\Event;
-use Raven\Processor\RemoveCookiesProcessor;
+use Raven\Processor\SanitizeCookiesProcessor;
 
-class RemoveCookiesProcessorTest extends TestCase
+class SanitizeCookiesProcessorTest extends TestCase
 {
     /**
      * @var Client
@@ -35,7 +35,7 @@ class RemoveCookiesProcessorTest extends TestCase
      */
     public function testConstructorThrowsIfBothOnlyAndExceptOptionsAreSet()
     {
-        new RemoveCookiesProcessor([
+        new SanitizeCookiesProcessor([
             'only' => ['foo'],
             'except' => ['bar'],
         ]);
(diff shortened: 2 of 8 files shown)


ste93cry:
>But we need to add behaviour to documentation. Because now the only documentation is a test file.

I agree, but I don't think that the `UPGRADE.md` file is the correct place where to put these information. Maybe we can defer the writing of documenttion (that should be added for a lot of other changes like the client builder, etc) in a next PR when we're close to release?

nokitakaze:
I mean, we could use this folder https://github.com/getsentry/sentry-php/tree/2.0/docs

ste93cry:
Yes for me it's ok, but as I said maybe we shoud defer this to a later PR where we upgrade docs for all changes? What do you think @Jean85 and @stayallive?

nokitakaze:
What "array scrubbing has not been implemented yet" is mean? I know, this error was here from the very beginning, but I have never see how a scrubbing should work.

lib/Raven/Processor/SanitizeCookiesProcessor.php
```diff
@@ -0,0 +1,96 @@
...
+
+            if (!empty($this->options['only'])) {
+                $cookiesToSanitize = $this->options['only'];
+            }
+
+            if (!empty($this->options['except'])) {
+                $cookiesToSanitize = array_diff($cookiesToSanitize, $this->options['except']);
+            }
+
+            foreach ($request['cookies'] as $name => $value) {
```
Jean85: Maybe if you use an `array_filter` this function will be shorter and more readable?

lib/Raven/Processor/SanitizeCookiesProcessor.php
```diff
@@ -0,0 +1,96 @@
...
+                if (!in_array($name, $cookiesToSanitize)) {
+                    continue;
+                }
+
+                $request['cookies'][$name] = self::STRING_MASK;
+            }
+        }
+
+        unset($request['headers']['cookie']);
```
Jean85: Why is this needed?

ste93cry:
>The issue is in that the default behavior became "no scrubbing", 

Mmm, the default behavior of this processor is the same as before: cookies are still sent, but their values are masked. The difference with the previous behavior is that in the previous version of the library the whole `cookies` array item was sanitized resulting in nothing being visible on Sentry while now each single cookie is still sent but their values are masked

tests/Processor/SanitizeCookiesProcessorTest.php
```diff
@@ -0,0 +1,139 @@
...
+        $event = new Event($this->client->getConfig());
+        $event = $event->withRequest($inputData);
+
+        $processor = new SanitizeCookiesProcessor($options);
+        $event = $processor->process($event);
+
+        $this->assertArraySubset($expectedData, $event->getRequest());
+    }
+
+    public function processDataProvider()
```
Jean85: Since the only differences between the data is in cookies and options, maybe you can remove all the other stuff and move the generic request data into the test.

lib/Raven/Processor/SanitizeCookiesProcessor.php
ste93cry: This is the unparsed request header that contains the cookies, which is in the form `foo=bar; bar=baz` while the `cookies` key contains a key-pair array of cookie name/value. As it's easier to sanitize an array I do that and this value is redundant anyway, so I remove it to prevent unmasked data from being sent to Sentry

lib/Raven/Processor/SanitizeCookiesProcessor.php
ste93cry: `array_filter` would remove the cookies that gets sanitized, while instead I want to retain them and mask their values. There could be a case where the user wants to know that a certain cookie was sent but it contains sensitive data and so he wants to sanitize it

PR #534, opened 2018-01-20
[WIP] [2.0] Serializer interface refactor

Jean85:
This PR starts from #454 and attempts to refactor and make it whole. I've added a few changes and refactor to it, I'll add comments in the code to explain; BTW commits are small and splitted to make it more understandable.


lib/Raven/Client.php
```diff
@@ -281,26 +281,6 @@ public function removeProcessor(ProcessorInterface $processor)
         $this->processorRegistry->removeProcessor($processor);
     }
 
-    /**
-     * Gets the representation serialier.
-     *
-     * @return ReprSerializer
-     */
-    public function getReprSerializer()
```
Jean85: Removed because I agree with @ste93cry as in https://github.com/getsentry/sentry-php/pull/454#discussion_r115219371

lib/Raven/Configuration.php
```diff
@@ -695,27 +659,29 @@ public function getErrorTypes()
...
             'send_attempts' => 6,
             'trust_x_forwarded_proto' => false,
             'prefixes' => explode(PATH_SEPARATOR, get_include_path()),
+            'serializer' => new Serializer(),
+            'reprSerializer' => new Serializer(),
             'serialize_all_object' => false,
             'sample_rate' => 1,
             'install_default_breadcrumb_handlers' => true,
             'install_shutdown_handler' => true,
-            'mb_detect_order' => null,
```
Jean85: I've dropped this option since it's something completely and uniquely related to the serializers, so I don't see the need of exposing this setting here.

lib/Raven/Configuration.php
```diff
@@ -695,27 +659,29 @@ public function getErrorTypes()
...
+     * @throws \Symfony\Component\OptionsResolver\Exception\AccessException
      */
     private function configureOptions(OptionsResolver $resolver)
     {
         $resolver->setDefaults([
             'send_attempts' => 6,
             'trust_x_forwarded_proto' => false,
             'prefixes' => explode(PATH_SEPARATOR, get_include_path()),
+            'serializer' => new Serializer(),
+            'reprSerializer' => new Serializer(),
```
ste93cry: As all options are named in lowercase with underscores this should be changed to follow that convention

lib/Raven/Configuration.php
```diff
@@ -803,7 +770,7 @@ private function configureOptions(OptionsResolver $resolver)
                 $this->server .= ':' . $parsed['port'];
             }
 
-            $this->server .= substr($parsed['path'], 0, strripos($parsed['path'], '/'));
+            $this->server .= substr($parsed['path'], 0, strrpos($parsed['path'], '/'));
```
ste93cry: Is there any reason to change from the case insentive version of the same function?

lib/Raven/Middleware/ExceptionInterfaceMiddleware.php
```diff
@@ -36,6 +41,7 @@
     public function __construct(Client $client)
     {
         $this->client = $client;
+        $this->configuration = $client->getConfig();
```
ste93cry: It seems to me that the client instance is not needed anymore, so we could pass directly the configuration as argument of the constructor and this way we would even remove a circular reference

lib/Raven/ClientBuilder.php
```diff
@@ -131,6 +131,14 @@ public static function create(array $options = [])
         return new static($options);
     }
 
+    /**
+     * @return Configuration
+     */
+    public function getConfiguration()
```
ste93cry: If we really want to expose the configuration from the client builder, which btw I don't find we should as all methods of that class can be called directly from the builder thanks to the `__get` magic method the method definition must be added to the interface with the appropriate description in the PHP DocBlock

lib/Raven/Configuration.php
```diff
@@ -55,14 +55,16 @@ class Configuration
      * Class constructor.
      *
      * @param array $options The configuration options
+     *
+     * @throws \Symfony\Component\OptionsResolver\Exception\UndefinedOptionsException
+     * @throws \Symfony\Component\OptionsResolver\Exception\AccessException
```
ste93cry: As we're in control of the options and how they resolve there should never be a case where the `AccessException` exception is thrown

lib/Raven/Configuration.php
```diff
@@ -740,10 +706,11 @@ private function configureOptions(OptionsResolver $resolver)
         $resolver->setAllowedTypes('trust_x_forwarded_proto', 'bool');
         $resolver->setAllowedTypes('prefixes', 'array');
         $resolver->setAllowedTypes('serialize_all_object', 'bool');
+        $resolver->setAllowedTypes('serializer', SerializerInterface::class);
+        $resolver->setAllowedTypes('reprSerializer', SerializerInterface::class);
```
ste93cry: This option name should follow the lowercase underscore naming convention used by all other options

lib/Raven/Configuration.php
```diff
@@ -846,10 +813,20 @@ private function normalizeAbsolutePath($value)
...
+            $path .= DIRECTORY_SEPARATOR;
         }
 
         return $path;
     }
+
+    /**
+     * @param array $additionalOptions
+     */
+    private function mergeAndResolve(array $additionalOptions)
```
ste93cry: I know I'm a fanatic of the documentation, but this is missing both the description of the method itself and of the argument

lib/Raven/Stacktrace.php
```diff
@@ -23,13 +23,11 @@ class Stacktrace implements \JsonSerializable
      */
     const CONTEXT_NUM_LINES = 5;
 
-    /**
-     * @var Client The Raven client
-     */
-    protected $client;
+    /** @var Configuration */
```
ste93cry: This is missing the description of the variable and is not following how all other docblocks are organized (multiline)

lib/Raven/Stacktrace.php
```diff
@@ -54,44 +52,44 @@ class Stacktrace implements \JsonSerializable
     ];
 
     /**
-     * Constructor.
+     * Stacktrace constructor.
      *
-     * @param Client $client The Raven client
+     * @param Configuration $configuration
```
ste93cry: Same as above, this is missing the description of the parameter

lib/Raven/Stacktrace.php
```diff
@@ -54,44 +52,44 @@ class Stacktrace implements \JsonSerializable
...
+        $this->configuration = $configuration;
+        $this->serializer = $configuration->getSerializer();
+        $this->reprSerializer = $configuration->getReprSerializer();
     }
 
     /**
      * Creates a new instance of this class using the current backtrace data.
      *
-     * @param Client $client The Raven client
+     * @param Configuration $configuration
```
ste93cry: Same as above, it's missing the description of the argument

tests/ClientTest.php
```diff
@@ -975,17 +976,19 @@ public function sampleRateAbsoluteDataProvider()
 
     public function testSetAllObjectSerialize()
     {
-        $client = ClientBuilder::create()->getClient();
-
-        $client->setAllObjectSerialize(true);
-
-        $this->assertTrue($client->getSerializer()->getAllObjectSerialize());
-        $this->assertTrue($client->getReprSerializer()->getAllObjectSerialize());
+        $serializer1 = $this->prophesize(SerializerInterface::class);
```
ste93cry: I didn't know we use `Prophecy`. As I never saw anything like this in the unit tests codebase can we change this to use the classic mock builder of PHPUnit? At least we keep everything consistent

ste93cry:
Overall it looks a nice step forward what we have now, even though I think that some things like changing the whole `Configuration` class methods shoul have been done in a separate PR or not done at all (calling a method that calls another method doesn't seems to be a good thing just to save 3 lines of code everytime). What I really don't like (I didn't like before either) is the serializer: I think that what we have now is pretty basic and I don't see why we should reinvent the wheel, so I was thinking that maybe we could use the Symfony serializer or something like that that provides a more powerful way of serializing objects and can be extended

lib/Raven/ClientBuilder.php
Jean85: This would mean passing the client builder around instead of the simple configuration, that I think could be better. I don't want to pass around a class that has the possibility of creating a new client if it's not what I want to do.

lib/Raven/Configuration.php
Jean85: How can we use the dependency injection? There are at least two different classes that needs them, so I need to pass the same object twice.

lib/Raven/Configuration.php
Jean85: Yes, the string that we are searching for is not case sensitive.

lib/Raven/Middleware/ExceptionInterfaceMiddleware.php
Jean85: It's used once. I mean to do a small refactor later to eliminate it, but I'm not sure how\where to move the used function (`translateSeverity`)

tests/ClientTest.php
Jean85: I find the mock builder less robust and reliable. I would prefer to migrate toward Prophecy, which is BTW included in PHPUnit anyway.

lib/Raven/Configuration.php
ste93cry: What I meant is that imho it would be better to add some methods to set the serializer to the client builder instead of treating them as options of the configuration and then be passed to the client as constructor arguments. I don't see the serializer to fit well into an array of options that is the configuration. Plus this way if an user doesn't want to use the client builder but instead prefers to create the client instance his own way it's free to do it

Jean85:
IMHO the symfony serializer is a bit too much for this. If we do not want to overcomplicate stuff and we do not think that replacing the serializer is needed, we can close this and the other PR.

The only thing that needs improvement is the ability to decite how deep the serialization should occur.

lib/Raven/Configuration.php
ste93cry: Oh got it now, that's the path after the domain name. Sorry I misunderstood the piece of code

ste93cry:
>If we do not want to overcomplicate stuff and we do not think that replacing the serializer is needed

I favor the fact that the serializer should be configurable and extendable, so both PRs are ok for me. But there are many open requests to support custom serialization of objects/values, so a good serializer would fit well into this by saving us the cost of reinventing something that maybe does all those things well and that we should maintain too.

Jean85:
Understood. But the Symfony serializer doesn't have a toArray encoder out of the box (see [docs](https://symfony.com/doc/current/components/serializer.html#encoders)), so we would have to do a lot of work anyway. Also, the current serializer tends to cover multibyte issues too, which are very much specialized for our usecase.

I'll try to address your concerns and rollback the modifications on the config, preserving the `SerializerInterface`; I'll also try to find a way to simply configure the depth of the serialization.

Jean85:
I'm not convinced by this approach anymore. Closing, maybe someone else wants to solve this in an other way.

ste93cry:
>I'm not convinced by this approach anymore. 
Would you like to share some thoughts on what does not convince you so that the next person can take your suggestions into account?

Jean85:
This should have been a much simpler PR, instead it started touching too much stuff.
I was also preoccupied by the usage of the config as a Serializer-holder, like you noted above.

PR #536, opened 2018-01-29
fix: Sanitize user data, which contains sensitive $_SESSION data

maleetz:


stayallive:
The user context is supplied by the developer and I don't think we should sanitize this since the developer can scrub the data it puts in the user context.

PR #537, opened 2018-01-30
[2.0] Transport mechanisms and spooling support

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

Supersedes #413. The original PR was adding new options to configure where to store saved events for later sending. Unfortunately the code is tightly coupled with the client itself, making it difficult to extend and customize how or where the events are stored and a lot of other things. This PR tries to solve all these problems by abstracting away from the client the part that sends the events into a set of transport mechanisms. One of them uses HttPlug, and the other one uses a spool to store events. An adapter that stores events in memory is provided out of the box, but one that stores them on the filesystem could be provided easily.


nokitakaze:
I need to read this today.

UPGRADE-2.0.md
```diff
@@ -19,11 +19,14 @@
 - The `curl_ssl_version` option has been removed.
 - The `verify_ssl` option has been removed.
 - The `ca_cert` option has been removed.
+- The `proxy` option has been removed.
```
Jean85: Maybe we should specify here that it got merged in the transport interface?

lib/Raven/Client.php
```diff
@@ -462,76 +412,51 @@ public function capture(array $payload)
...
-            $this->send($payload);
-        } else {
-            $this->pendingEvents[] = $payload;
-        }
+        $this->send($event);
 
         $this->lastEvent = $event;
 
-        return $payload['event_id'];
+        return str_replace('-', '', $event->getId()->toString());
```
Jean85: Why we omit the dashes?

lib/Raven/Client.php
```diff
@@ -462,76 +412,51 @@ public function capture(array $payload)
...
-            $data['extra'] = $this->serializer->serialize($data['extra']);
+        if (!empty($extraContext)) {
+            $event = $event->withExtraContext($this->serializer->serialize($extraContext));
         }
-        if (!empty($data['tags'])) {
-            foreach ($data['tags'] as $key => $value) {
-                $data['tags'][$key] = @(string) $value;
+        if (!empty($tagsContext)) {
+            foreach ($tagsContext as $key => $value) {
+                $tagsContext[$key] = @(string) $value;
```
Jean85: Why silencing errors? Maybe serializing those tags could be a better approach?

tests/Spool/MemorySpoolTest.php
```diff
@@ -0,0 +1,69 @@
...
+     * @var MemorySpool
+     */
+    protected $spool;
+
+    protected function setUp()
+    {
+        $this->spool = new MemorySpool();
+    }
+
+    public function testQueueEvent()
```
Jean85: Please add a test that checks that the spooling is done in the correct order (FIFO)

tests/ClientTest.php
```diff
@@ -551,148 +367,73 @@ public function testSanitizeObjects()
         unset($reflection, $property, $value, $reflection, $clone);
         ksort($expected);
 
-        $client->sanitize($data);
-        ksort($data['extra']['object']);
+        $event = $client->sanitize($event);
+        /*ksort($data['extra']['object']);
```
Jean85: Is this no longer needed? Why?
If yes, we should remove it.

lib/Raven/Client.php
ste93cry: Because as per [SDK specs](https://docs.sentry.io/clientdev/attributes/) the ID of an event is an UUID without dashes. Unfortunately we must do this string-replace everywhere we use the Ramsey UUID object

>Hexadecimal string representing a uuid4 value. The length is exactly 32 characters (no dashes!)

lib/Raven/Client.php
ste93cry: Indeed, but I did not change the way serialization works. The code silencing the errors cames from version `1.x` and I have plans to modify the whole sanitize function anyway in a later PR, so I did not address the thing here

lib/Raven/Client.php
Jean85: Maybe then it's better if we do that at the start, wrapping the Ramsey lib call?

tests/Spool/MemorySpoolTest.php
ste93cry: Isn't it done below in the `testFlushQueue` method verifying it with `withConsecutive`?

UPGRADE-2.0.md
ste93cry: Actually, the option has been removed because it's name/value is dependent on the implementation of the HTTP client, and it was not even used after integrating Httplug. What about this note:

>The `proxy` option has been removed in favour of leaving to the user the burden of configuring the HTTP client options using a custom client.

lib/Raven/Client.php
ste93cry: I was experimenting with a custom codec, which should be the correct way to format an UUID using the Ramsey lib, but this required a bunch of other things like the ability to change the factory and the changes were too much and out of scope for this PR. I would like to address this problem better in the future

ste93cry:
@nokitakaze the spool transport is made exactly to support cases like yours. You could create your own implementation of the `SpoolInterface` that saves the events on the filesystem and then use a command or whatever else to retrieve them and send them when you want. I have a really simple implementation of this `FileSpool` spool, but I did not submit it along with this PR to not add too many things to the library. Maybe @Jean85 or @stayallive could give their opinion on the matter?

Jean85:
The `FileSpool` could be a nice addition, but I would leave it for a latter PR! In this way we could do it alongside some docs refinements about this change, which is not so subtle on the end user side.

stayallive:
I agree with @Jean85, this is definitely something to explore to allow for "offline" and "queued" sending or aggregating of sending Sentry events, but better served in a seperate PR so we can discuss the internals of that since PHP does not have very good standards for these kind of things 👍

ste93cry:
Before merging, I would like to get an opinion on a last thing: I've removed the destructor of the `HttpTransport` class so that the requests still pending when the application is shutting down are only sent when the shutdown function handler is called. As the callback passed to `register_shutdown_function` maintains a reference to the instance of the class the destructor will never be called when unsetting the last variable referencing the transport but only after the handler has been executed, so I thought it would not make sense to keep it. What do you think about this?

PR #540, opened 2018-02-06
[2.0] Improve the Context class and its usability

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

Working with contexts is probably something that an user does a lot as he could attach additional data to the events being sent. However, I noticed a lack of usability of the `Context` class, so this PR tries to improve it by adding methods to set, get or replace the data stored in each context. Also, every context (user, extra and tags) has now its own instance of the data object. I've implemented some methods to set and get values from a given path using the Symfony [PropertyAccessor](http://symfony.com/doc/current/components/property_access.html) component because I thought it would be common to change nested data, however if it doesn't fit or we don't want an extra dependency I can remove it


ste93cry:
This PR is almost finished, but I would like an opinion about a question @Jean85 @stayallive @nokitakaze: all the contexts have now their own instance of the `Context` class as opposed to before where they all shared the same one and different methods were available. The new class supports a lot more features like recursive merging and so on, but these do not apply to all types of context: think for example to the tags context, you should not be able to store anything else that strings. Would it make sense to have specialized classes extending the `Context` class that prevents such features to be used?

ste93cry:
This PR is almost finished, but I would like an opinion about a question: all the contexts have now their own instance of the `Context` class as opposed to before where they all shared the same one and different methods were available. The new class supports a lot more features like recursive merging and so on, but these do not apply to all types of context: think for example to the tags context, you should not be able to store anything else than strings. Would it make sense to have specialized classes extending the `Context` class that prevent such features to be used or maybe add methods that works only for a specific type of context?

Jean85:
Maybe we can have a `MergeableContextInterface` with a `merge(MergeableContextInterface $context)` method? 

But first, in which situations merging contexts makes sense?

ste93cry:
>But first, in which situations merging contexts makes sense?

I did not mean we should support merging contexts, but data stored inside each of them. For example, the extra context could contain arbitrary data, and so nested arrays. The actual code provides only a `merge` method that uses `array_merge`, which in turn will overwrite the whole nested array. Also, there is no easy way to get or set a key from a nested array, you have to extract the whole context data, make the changes and then overwrite the old one with the new one. What I tried to achieve was to have a `Context` class that acts more as a data storage object, which not only behaves as a traditional array thanks to the `ArrayAccess` interface but that also makes it easy for users to work with the data contained in it.

composer.json
```diff
@@ -22,6 +22,7 @@
         "psr/http-message-implementation": "~1.0",
         "ramsey/uuid": "~3.3",
         "symfony/options-resolver": "~2.7|~3.0",
+        "symfony/property-access": "~2.7|3.0",
```
Jean85: I'm not a fan of adding this additional dependency, but it's due to the fact that I don't think that the functionality is helpful: reaching data by path seems helpful, but it's still an array underneath, and I don't see any reason for changing that; so, accessing by array keys would be easier and more natural (and maybe it allows some auto completing from the IDE).

lib/Raven/Client.php
```diff
@@ -132,7 +152,11 @@ public function __construct(Configuration $config, TransportInterface $transport
         $this->config = $config;
         $this->transport = $transport;
         $this->processorRegistry = new ProcessorRegistry();
-        $this->context = new Context();
+        $this->userContext = new Context();
+        $this->tagsContext = new Context();
+        $this->extraContext = new Context();
+        $this->runtimeContext = new Context();
+        $this->serverOsContext = new Context();
```
Jean85: Why `serverOs` and not simply `server`? Is it that specific?

lib/Raven/Middleware/ContextInterfaceMiddleware.php
```diff
@@ -24,18 +24,46 @@
...
      *
-     * @param Context $context The context storage
+     * @param Context $userContext     The user context
+     * @param Context $tagsContext     The tags context
+     * @param Context $extraContext    The extra context
+     * @param Context $runtimeContext  The runtime context
+     * @param Context $serverOsContext The server OS context
      */
-    public function __construct(Context $context)
+    public function __construct(Context $userContext, Context $tagsContext, Context $extraContext, Context $runtimeContext, Context $serverOsContext)
```
Jean85: I don't like this very much. It would be frighteningly easy to mess up the arguments. Is there any way to fix this?

lib/Raven/Middleware/ContextInterfaceMiddleware.php
```diff
@@ -57,10 +85,15 @@ public function __invoke(Event $event, callable $next, ServerRequestInterface $r
         $runtimeContext = isset($payload['runtime_context']) ? $payload['runtime_context'] : [];
         $userContext = isset($payload['user_context']) ? $payload['user_context'] : [];
 
-        $event = $event->withTagsContext(array_merge($this->context->getTags(), $tagsContext), false)
-            ->withExtraContext(array_merge($this->context->getExtraData(), $extraContext), false)
-            ->withUserContext(array_merge($this->context->getUserData(), $userContext), false)
+        $event = $event->withTagsContext($this->tagsContext->toArray(), false)
+            ->withTagsContext($tagsContext, false)
+            ->withExtraContext($this->extraContext->toArray(), false)
```
Jean85: Why do we need to `toArray()` each time?

tests/ClientTest.php
```diff
@@ -265,6 +266,41 @@ public function testGetLastEventId()
         $this->assertEquals('ddbd643a51904ccea6ce3098506f9d33', $client->getLastEventID());
     }
 
+    public function testGetUserContext()
```
Jean85: Maybe we can collapse all those tests into one?

tests/ContextTest.php
```diff
@@ -2,68 +2,152 @@
...
+
+        $this->assertEquals(['foo' => 'bar'], $context->toArray());
     }
 
-    public function testMergeExtraDataWithSameKey()
+    public function testJsonSerialize()
+    {
+        $context = new Context(['foo' => 'bar']);
+
+        $this->assertEquals('{"foo":"bar"}', json_encode($context));
```
Jean85: Use `assertSame`

tests/ContextTest.php
```diff
@@ -2,68 +2,152 @@
...
+
+        $this->assertTrue($throwedError);
+
+        $context['foo'] = 'bar';
+
+        $this->assertAttributeEquals(['foo' => 'bar'], 'data', $context);
+        $this->assertTrue(isset($context['foo']));
+        $this->assertEquals('bar', $context['foo']);
+
+        unset($context['foo']);
```
Jean85: Is this needed?

tests/Middleware/ContextInterfaceMiddlewareTest.php
```diff
@@ -21,33 +21,43 @@ class ContextInterfaceMiddlewareTest extends TestCase
...
     {
-        $context = new Context();
-        $context->setTag('bar', 'foo');
-        $context->mergeUserData(['foo' => 'bar']);
-        $context->mergeExtraData(['bar' => 'baz']);
+        $userContext = new Context();
+        $userContext->setData(['foo' => 'bar']);
+
+        $tagsContext = new Context();
+        $tagsContext->setData(['foo', 'bar']);
```
Jean85: This is the same as the previous one, you could mix those up in the tests; change it.

stayallive:
I am not sure splitting the Context in 5 different (and who knows how many in the future) Context objects helps very much in terms of readability or for the end-user. Wouldn't it be nicer to be able to use `$context->put('session.user_id', 1)` or something (maybe I am missing something, but this seems overly complex and not necessary)?

lib/Raven/Client.php
ste93cry: I saw it in the Raven Ruby SDK and copied it here

https://github.com/getsentry/raven-ruby/blob/master/lib/raven/context.rb#L16

lib/Raven/Middleware/ContextInterfaceMiddleware.php
ste93cry: Because in the `Event` class the context data is stored as a plain array. I could switch to an instance of the `Context class`, but it should probably not be the same instance as the one of the client or someone that adds an item to the context of an event would add it globally. WDYT?

tests/ContextTest.php
ste93cry: Yes it's needed to test that the `Context::offsetUnset` method works, but it was missing the assertion so nice catch!

tests/Middleware/ContextInterfaceMiddlewareTest.php
ste93cry: The lines are the same, but the context instances are different and they are used in the test to ensure that the merging works for each of them

tests/ClientTest.php
ste93cry: We could, but I think it's much more clear if there is a test for each function

tests/ContextTest.php
ste93cry: Why? `assertSame` checks with strict equality which I don't see any benefit in using here

composer.json
ste93cry: The problem is that if you have a complex array with nested arrays inside, even autocompleting from IDE does not help you and the only way to reach a specific subkey would be to get the whole array and then access the key you want. If you also have to change the value, you would have to put inside the context object the whole array. But I'm not a big fan of adding an additional dependency either, so if we all agree this is something not useful I'm all in favour of removing it

ste93cry:
>I am not sure splitting the Context in 5 different (and who knows how many in the future) Context objects helps very much in terms of readability or for the end-user.

According to the [SDK specs](https://docs.sentry.io/clientdev/context/) there should be a method for each context, so the user doesn't necessarily have to know that there are N objects behind

>Wouldn't it be nicer to be able to use $context->put('session.user_id', 1

I thought about this but then how do you explain to the users that they have to use `session.` as prefix to save the data into the `session` context? Having dedicated methods means that they know exactly where they are saving the data

stayallive:
I agree on that, but having dedicated methods also means it doesn't matter if it's 1 or 5 Context objects in the code behind (which hides those prefixes for the user). 

I think this could make the SDK developers life easier by preventing multiple arguments and passing lot's of objects or keeping track of different context objects instead of a single container.

We could also for example let the `setUserContext($context)` (or `mergeUserContext`) method accept either and `array` or an `Context` object. Then we could go the extra mile of providing a `Context` object on the getters which add convienience methods. However in the code behind we only have to deal with a single `Context` object. So for the user it are multiple (or just arrays), for us it is just one.

ste93cry:
I'm not sure I get your point, so please let me double check if I'm understanding correctly: you would like to have a single Context instance which saves all data of all contexts inside a single property, then it exposes methods to work with each context? Something along these lines:

```php
class Context
{
    protected $data = [
        'user' => [],
        'runtime' => [],
    ];

    public function setUserContext(array $context)
    {
        $this->data['user'] = $context;
    }

    public function setRuntimeContext(array $context)
    {
        $this->data['runtime'] = $context;
    }
}
```

stayallive:
Yes, this adheres to the SDK spec and gives us a nice (single) Context object to deal with.

But this does not have to be on the `Context` class itself.

The `Client` could also do something like:

```php
public function setUserContext(array|Context $context)
{
    $this->context->set('user', $context); // which would set $data['user'] = $context; as per your example
}
public function getUserContext(): array|Context
{
    return $this->context->get('user'); // which would return $data['user']; as per your example
}
```

The `array|Context` is up for debate since the `Context` could implement all array like interfaces making it behave like an array.

ste93cry:
Returning an array means that an user can't access the methods provided by the `Context` class (`set`, `merge`, etc). I don't understand how we could return a context if we're going to remove the multiple instances and everything is stored internally as an array. If I put the methods like `merge` on the `Client` class we're going to duplicate the code... I'm not sure I understand the logic behind this

stayallive:
Well one way is to `return new Context($this->data['user']);` however you might be right that this is more trouble than it's worth. But it might keep it cleaner internally was all my concern.

I think this is a really valid concern: https://github.com/getsentry/sentry-php/pull/540#discussion_r167498098

Jean85:
To solve that maybe we can make the `Context` class abstract and create a child for each type of context? So we could typehint correctly.

ste93cry:
Would it really makes sense to have a type for each context type, even if they basically share almost all things? If the problem is the typehint on the middleware, maybe we could improve the code by making the middleware itself handle only one context at time, and then just reuse the middleware as many times as needed (so if in the future a new context type is needed we just have to append a new instance to the list of core middlewares without any other code change). WDYT?

Jean85:
A reuse of the middleware could be a good way to go, yes :+1:

lib/Raven/Middleware/ContextInterfaceMiddleware.php
ste93cry: I've changed the middleware to only work on a single context so that we will have as many middleware instances added to the stack as the number of contexts to be processed. This simolify a bit the `ContextInterfaceMiddleware` class

ste93cry: I've changed the middleware to only work on a single context so that we will have as many middleware instances added to the stack as the number of contexts to be processed. This simplify a bit the `ContextInterfaceMiddleware` class

lib/Raven/Middleware/ContextInterfaceMiddleware.php
```diff
@@ -51,17 +58,14 @@ public function __construct(Context $context)
...
     {
-        $tagsContext = isset($payload['tags_context']) ? $payload['tags_context'] : [];
-        $extraContext = isset($payload['extra_context']) ? $payload['extra_context'] : [];
-        $serverOsContext = isset($payload['server_os_context']) ? $payload['server_os_context'] : [];
-        $runtimeContext = isset($payload['runtime_context']) ? $payload['runtime_context'] : [];
-        $userContext = isset($payload['user_context']) ? $payload['user_context'] : [];
+        if (isset($payload[$this->contextName . '_context'])) {
+            $method = preg_replace_callback('/_[a-zA-Z]/', function ($matches) {
+                return strtoupper($matches[0][1]);
+            }, 'with_' . $this->contextName . '_context');
```
Jean85: You should add a method existence check here, to detect errors

Jean85: Also, why underscores and not camelCase?

tests/Context/TagsContextTest.php
```diff
@@ -0,0 +1,150 @@
...
+                ['foo' => 'bar'],
+                true,
+                null,
+                'The tags context does not allow recursive merging of its data.',
+            ],
+            [
+                ['foo' => new \stdClass()],
+                false,
+                null,
+                'The $data argument must contains a simple array of string values.',
```
Jean85: Should we consider also objects that have `__toString` as valid too?

tests/EventTest.php
```diff
@@ -172,26 +172,25 @@ public function testGettersAndSetters($propertyName, $propertyValue, $serialized
 
         $data = $newEvent->toArray();
 
-        $this->assertArrayHasKey($serializedPropertyName, $data);
-        $this->assertSame($propertyValue, $data[$serializedPropertyName]);
+        $this->assertArraySubset($expectedValue, $data);
```
Jean85: :+1:

tests/Middleware/ContextInterfaceMiddlewareTest.php
```diff
@@ -13,43 +13,76 @@
...
-            'server_os_context' => ['foo' => 'bar'],
-            'runtime_context' => ['bar' => 'foo'],
-            'user_context' => ['baz' => 'foo'],
+            $contextName . '_context' => $payloadData,
         ]);
 
         $this->assertEquals(1, $invokationCount);
     }
+
+    public function invokeDataProvider()
```
Jean85: Please add a test for handling wrong context name.

tests/Context/TagsContextTest.php
ste93cry: I did not consider them to not have to force the cast to `string` like it was done in `1.x` It's not a problem to add the approprite checks to consider objects, but this would break any typehint we could add in the future (this argument is not strictly related to the methods available at the moment)

https://github.com/getsentry/sentry-php/blob/351cb63da6fc40dbb2a8324e0b3fdff2e7dcc55b/lib/Raven/Client.php#L945

lib/Raven/Middleware/ContextInterfaceMiddleware.php
ste93cry: >You should add a method existence check here, to detect errors

Done

>Also, why underscores and not camelCase?

The context aliases (the names that are used both in the `$payload` argument of the middleware and in the event data that is sent) are in underscore notation while the methods of the `Event` class are in camel case. In this case I transform the whole method name in camel case with a single regex

ste93cry: >You should add a method existence check here, to detect errors

Done

>Also, why underscores and not camelCase?

The context aliases (the names that are used both in the `$payload` argument of the middleware and in the event data that is sent) are in underscore notation while the methods of the `Event` class are in camel case. In this case I transform the whole method name to camel case with a single regex

tests/Middleware/ContextInterfaceMiddlewareTest.php
ste93cry: Done, thank you for spotting the missing test

tests/Context/TagsContextTest.php
Jean85: I would like to avoid the `@` operator too. But a `is_object && method_exists` could be a good alternative, since the `__toString` method cannot throw exceptions.

lib/Raven/Middleware/ContextInterfaceMiddleware.php
Jean85: Then I find this fuction highly unreadable, and poor performance-wise. I would prefer a dumb, easy-to-read, lightning fast switch case, since the number of context names are low and well known: you even have constants to use!

tests/Context/TagsContextTest.php
ste93cry: In which cases it makes sense to add a tag which is an object (even if then it will be convered to a string)? Wouldn't it be better to know that tags can be strings and let users do the cast if they want?

lib/Raven/Middleware/ContextInterfaceMiddleware.php
ste93cry: >Then I find this fuction highly unreadable

Indeed, I focused too much on making something too generic where there is was no real need

tests/Context/TagsContextTest.php
Jean85: The easiest example that I can think of is an `User` class with the username as a `__toString`. I would prefer the string too, but we are on PHP 5.6+, so no scalar type available... Hence that's why I would manually consider castable objects.

lib/Raven/Middleware/ContextInterfaceMiddleware.php
```diff
@@ -51,17 +58,32 @@ public function __construct(Context $context)
...
-        $event = $event->withTagsContext(array_merge($this->context->getTags(), $tagsContext), false)
-            ->withExtraContext(array_merge($this->context->getExtraData(), $extraContext), false)
-            ->withUserContext(array_merge($this->context->getUserData(), $userContext), false)
-            ->withServerOsContext($serverOsContext, false)
-            ->withRuntimeContext($runtimeContext, false);
+        if (isset($payload[$this->contextName . '_context'])) {
+            switch ($this->contextName) {
+                case Context::CONTEXT_USER:
+                    $event = $event->withUserContext($this->context->toArray(), false);
+                    $event = $event->withUserContext($payload[$this->contextName . '_context'], false);
```
Jean85: What happens if the payload doesn't have a `$this->contextName . '_context'` key? Shouldn't also you merge that beforehand, so you would call the `with` method only once?

ste93cry: >What happens if the payload doesn't have a `$this->contextName . '_context'` key?

Simply nothing, the middleware will just call the next one in the chain. We cannot throw any exception or error because the `$payload` argument is optional

>Shouldn't also you merge that beforehand, so you would call the with method only once?

Merging the data with the current one is already done inside the `with*` methods, I don't really like to have to call `array_merge` all over the code and I don't see a real benefit from switching to

```php
$event = $event->withServerOsContext(array_merge($this->context->toArray(), $payload[$this->contextName . '_context']), false);
```

apart from not having to call the method twice (and cloning the event object twice). Also, I think it's more unreadable

tests/Context/TagsContextTest.php
ste93cry: >but we are on PHP 5.6+, so no scalar type available...

But when we will switch to PHP 7+ it will (`string`, at least). I don't really see why an user could not cast the object before adding it as a tag to the context, it seems a bad practice to me to let users pass an object when it's logic to think that a tag is a string

lib/Raven/Middleware/ContextInterfaceMiddleware.php
Jean85: Ok understood.
Then, the only modification that I would do is moving `$payload[$this->contextName . '_context']` to a local variable before the switch, to reduce complexity and duplication.

ste93cry: >Shouldn't also you merge that beforehand, so you would call the with method only once?

Rethinking about this you were right, in my original code if the `$payload` argument did not contain any context data then the context data set globally would had never been set in the event

PR #542, opened 2018-02-08
[2.0] Fix unused variable

Jean85:
In #526 I missed this small issue, and Scrutinizer is screaming bloody murder because of this!


nokitakaze:
How it was possible to pass tests with some error like this?

Jean85:
It's not a proper error, I was just getting a value from the event and not using it.

PR #543, opened 2018-02-09
Fix gzipCompress function name

FallDi:
PHP doesn't have `gzipCompress` function, use `gzcompress` instead
http://php.net/manual/en/function.gzcompress.php


stayallive:
Good catch!

PR #544, opened 2018-02-09
Include object string representation in stacktraces if supported.

apapsch:
This allows more insight into problems, because the class name alone may not suffice.

### DIFF
--- a/lib/Raven/ReprSerializer.php
+++ b/lib/Raven/ReprSerializer.php
@@ -29,7 +29,11 @@ class Raven_ReprSerializer extends Raven_Serializer
         } elseif (is_integer($value) || is_float($value)) {
             return (string) $value;
         } elseif (is_object($value) || gettype($value) == 'object') {
-            return 'Object '.get_class($value);
+            $name = 'Object '.get_class($value);
+            if (method_exists($value, '__toString')) {
+                $name .= ': ' . $value;
+            }
+            return $name;
         } elseif (is_resource($value)) {
             return 'Resource '.get_resource_type($value);
         } elseif (is_array($value)) {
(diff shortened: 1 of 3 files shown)


lib/Raven/ReprSerializer.php
```diff
@@ -29,7 +29,11 @@ protected function serializeValue($value)
         } elseif (is_integer($value) || is_float($value)) {
             return (string) $value;
         } elseif (is_object($value) || gettype($value) == 'object') {
-            return 'Object '.get_class($value);
+            $name = 'Object '.get_class($value);
+            if (method_exists($value, '__toString')) {
```
dcramer: i think the concern here would be when objects return gigantic values for strings (also you'd need to gracefully handle errors with ``$value->__toString()`` since its user code and could easily error)

apapsch: I could limit it to 1000 characters with substr, though that will not help if `__toString` exceeds the memory limit. Ignore that?

Regarding error handling I would do:

* for exceptions: wrap it in try-catch 
* for PHP errors firing the error handler: use custom handler, afterwards call `restore_error_handler`
* for PHP errors causing shutdown: this is not recoverable, so I would push the error to sentry, but `ReprSerializer` doesn't have access to a `Client`. Apart from tests, there are only two instantiations of `ReprSerializer`. The client could be passed in:
  * Stacktrace.php:27
  * Client.php:204

Jean85:
`__toString()` cannot throw exceptions: http://php.net/manual/it/language.oop5.magic.php#object.tostring

But if it explodes for some strange reason (as @dcramer was saying, it may be user code), PHP converts it directly to a fatal error, halting everything; so I'm :-1: about this, it can potentially convert a notice to a fatal, loosing any kind of event reporting from us.

apapsch:
I was also not considering that the handler for PHP errors causing shutdown cannot be restored in case no fatal error happens, so the approach is not viable. I guess this feature is too nice to have ;-)

Jean85:
Maybe we could improve on this with something like #534 in the 2.x branch. Personalizing the serializer could create a way to cast to string safely your own objects; this would leave the responsability to the end user, which is fine by me; I just don't like doing this dangerous staff everywhere, it's a bit too risky.

Jean85:
Closing this as old and considered too risky.

Issue #124, opened 2013-11-07
php://input can be supported


bastiendonjon:
I received data in PHP with php://input by a ajax request to send json string in POST.

However, Sentry do not display data like $_POST and $_GET

dopesong:
Any news on this?

johnboy-leeds:
This would also be of benefit to us.

Jean85:
For 1.x, in reality this is already marked somewhat as a bug (or at least as related) in #531

ste93cry:
Closing as this feature has been implemented in version `1.x`. In version `2.x` I would not support it for the reasons outlined above. It's anyway really simple to add a custom middleware that will do it, and we could even think to provide an implementation out-of-the-box that is not enabled by default

PR #546, opened 2018-02-12
Adding application/json input handling.

ryantology:
References #531
- Added additional unit test for json data.
- PHP input is sanitized through json_decode
- If there is binary or invalid data in the json input the entire payload will be discarded. Could run input through sanitize first?

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -276,6 +276,17 @@ class Raven_Client
         return $this;
     }
 
+    /**
+     * Note: Prior to PHP 5.6, a stream opened with php://input can
+     * only be read once;
+     *
+     * @see http://php.net/manual/en/wrappers.php.php
+     */
+    protected static function getInputStream()
+    {
+        return file_get_contents('php://input');
+    }
+
     private static function getDefaultPrefixes()
     {
         $value = get_include_path();
@@ -748,6 +759,11 @@ class Raven_Client
         // instead of a mapping which goes against the defined Sentry spec
         if (!empty($_POST)) {
             $result['data'] = $_POST;
+        } elseif (isset($_SERVER['CONTENT_TYPE']) && stripos($_SERVER['CONTENT_TYPE'], 'application/json') === 0) {
+            $raw_data = $this->getInputStream() ?: false;
+            if ($raw_data !== false) {
+                $result['data'] = (array) json_decode($raw_data, true) ?: null;
+            }
         }
         if (!empty($_COOKIE)) {
             $result['cookies'] = $_COOKIE;
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -284,6 +284,10 @@ class Raven_Client
      */
     protected static function getInputStream()
     {
+        if (PHP_VERSION_ID < 50600) {
+            return null;
+        }
+
         return file_get_contents('php://input');
     }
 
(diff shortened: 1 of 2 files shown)


test/Raven/Tests/ClientTest.php
```diff
@@ -789,7 +800,7 @@ public function testGetHttpData()
             'SERVER_PORT'         => '443',
             'SERVER_PROTOCOL'     => 'HTTP/1.1',
             'REQUEST_METHOD'      => 'PATCH',
-            'QUERY_STRING'        => 'q=bitch&l=en',
```
ryantology: This seemed inappropriate, was removed while working on similar work.

ryantology:
Also References #400 #124

lib/Raven/Client.php
```diff
@@ -276,6 +276,17 @@ public function setEnvironment($value)
         return $this;
     }
 
+    /**
+     * Note: Prior to PHP 5.6, a stream opened with php://input can
```
Jean85: Since you're targeting the master branch, PHP < 5.6 is allowed, so you should read that only if PHP is >= 5.6; otherwise we would mess with the app.

lib/Raven/Client.php
```diff
@@ -748,6 +759,11 @@ protected function get_http_data()
         // instead of a mapping which goes against the defined Sentry spec
         if (!empty($_POST)) {
             $result['data'] = $_POST;
+        } elseif (isset($_SERVER['CONTENT_TYPE']) && stripos($_SERVER['CONTENT_TYPE'], 'application/json') === 0) {
+            $raw_data = $this->getInputStream() ?: false;
```
Jean85: Since we will never need the input stream for other stuff, maybe we can wrap the Json handling inside that method?

lib/Raven/Client.php
ryantology: It won't really 'mess up the app'. What will happen is the data will be null. I can adjust the comment to explain that. This doesn't introduce a regression issue because the previous implementation and the current one still return the same broken results for php from the dark ages :)

I can add the condition in. I felt that this is really an edge case and there is no clean way to handle support for PHP < 5.6 with this feature.

lib/Raven/Client.php
ryantology: You should probably add input stream handling for XML as well. This test is just for JSON. I really didn't want to get into XML parsing so I left that out. We actually have to deal with a legacy XML API for one project. Once I have to debug that I'll make a new patch :)

This could be abstracted into a `parse()` method as it gets bigger. I didn't want to spend too much time on this knowing that 2.0 is on the horizon.

ryantology:
@Jean85 I added the negative test case. This was good as it identified a need to abstract the json encode stuff for future XML tests.

lib/Raven/Client.php
ryantology: The one thing that is a little strange is the unit tests will succeed in < 5.6 but the reality is the feature doesn't work in that version. Not sure if there is a way to resolve that. Just a thought.

(I added the conditional https://github.com/getsentry/sentry-php/pull/546/files#diff-11a1d8314f22caf150ad02a00b294b58R287 )

stayallive: The data will be empty on < 7.0 but it can also be we fetched the data before the user could (if the use the library for logging for example) causing it to possibly break the user's application.

And yes this will not go well for PHP 5.6 but we should try and solve that differently if we must. But this could affect the implementors application and that is something we really should avoid at all cost.

PR #548, opened 2018-02-14
fixing AnnotationException when Raven_Client appears in the stack trace of being serialized Exception

tiger-seo:
fixing Doctrine\Common\Annotations\AnnotationException with message [Semantical Error] The annotation "@doc" in method Raven_Client::parseDSN() was never imported. Did maybe forget to add a &quot;use&quot; statement for this annotation?


tiger-seo:
i have never seen that fast PR acceptance :) thank you very much :)

PR #550, opened 2018-02-15
Fixing "site" detection for reverse proxy configurations

iSDP:
Also added XSS protection, might it ever be needed


stayallive:
Hi @iSDP, thanks for your contribution.

As @Jean85 has explained, we should never trust the user input (`HTTP_HOST` is) and if localhost is set as the server name this is a configuration issue on the user side which also should be fixed/changed there, we provide hooks to set the site yourself if you really want to use the `HTTP_HOST` as the site.

Defaulting to `HTTP_HOST` if the server name is `localhost` is not the way to fix this because people working on their dev environment or behind proxies might actually see this as intended which can cause confusion if that is overridden.

PR #554, opened 2018-02-21
Allow $data['request'] to be 5 levels deep

martinstuecklschwaiger:
A normal POST request in a PHP application can be up to 3 levels deep. In `$data['request']` there are 2 additional arrays wrapped around it, so that the serializer removes most of the data. (I added a screenshot to show how that looks in the UI). 

This pull request raises the `max_depth` for the request data to 5. 
I also adjusted the test to use more realistic data so that it's better understandable. 
I added another test so that the truncation is still covered by a test. 

screenshot from 2018-02-21 21-57-31

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -929,7 +929,7 @@ class Raven_Client
     {
         // attempt to sanitize any user provided data
         if (!empty($data['request'])) {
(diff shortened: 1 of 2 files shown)


stayallive:
This snowed under a but, thanks again for your contribution!

PR #555, opened 2018-02-21
Cleanup the 'site' value if it's empty

martinstuecklschwaiger:
The values `extra`, `tags`, `user`, and `request` get removed if they are empty. 
This pull request does the same for the `site` value. 

I added a screenshot to show how this affects users in the UI:
screenshot from 2018-02-21 22-16-32


Jean85:
Thanks for the contribution! Can you add a small test to avoid regressing on this bug?

PR #557, opened 2018-02-26
Change badges with badge poser.

JellyBellyDev:
I change the badges "Total Downloads", "Monthly Downloads", "Latest Stable Version" and "License" with badge poser: https://poser.pugx.org/
you like it? :)


stayallive:
Hey Andrea! 

The only change I am seeing is that these are a bit more exact... any more benefits? 

Otherwise I am nog seeing why they should be changed to be honest.

JellyBellyDev:
Hi Alex, 
thanks for the reply.

In my opinion:
- the data are more exact
- perform better
- use the same layout as trevis and scrutinizer with rounded edges

Mine is just a proposal! ;)

PR #559, opened 2018-03-06
Make it possible to overwrite message limit of 1024

brainexe:
Currently the message length limit is fixed to 1024 in the Serializer. 
In the Client it's possible to change the value (if needed), but not in serializer context. I just added the possibility to pass higher/lower values.


ezintz:
I guess that the consturctor of the Raven_Client should make use of it, no?

Issue #562, opened 2018-03-14
[2.0] Adding the PHP version


Jean85:
A PR was open under the Symfony integration repo (https://github.com/getsentry/sentry-symfony/pull/118) to add the PHP version as a default tag (thanks @hjanuschka !).

I turned it down because I think that this should be done in the base client here. I investigated a bit, and @mitsuhiko told me that there's an already supported context to do that, which looks like this:

```json
"contexts": {
    "runtime": {
        "version": "7.1",
        "name": "php"
    }
}
```

IMHO this should target the new client, what do you think?

ste93cry:
>I think that this should be done in the base client here

I strongly agree. We could initialize the global runtime context with the data or initialize the runtime context of the event as the result will be the same. Alternately, (maybe the clearer option, but also maybe too much over engineered) we could create a specialized middleware

Jean85:
Ok so #564 already implemented this for 1.x. If we port the tests from there into the 2.x branch we are good! :+1:

ste93cry:
I was thinking, would it make sense to create a `RuntimeContext` class that validates that the only values that can be set are the one allowed by the SDK interface and in the constructor it initializes itself with some good defaults?

ste93cry:
Closing since this has been implemented in both versions `1.x` and `2.x`

PR #564, opened 2018-03-14
Runtime information

hjanuschka:
as discussed here: https://github.com/getsentry/sentry-php/issues/562
this one adds a context containing the current php version.

running the `examples/vanilla/index.php`  results in a exception like this:
( i wanted to upload via github (seems to 503 right now, hope they have sentry))

https://imgur.com/Hs6FDYm

i really would appreciate if we can get this in an 1.8.* release (atleast a release in a short time) - i really would love to have this in production. let me know if you need me to backport it somewhere.

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -884,6 +884,8 @@ class Raven_Client
         if (empty($data['site'])) {
             unset($data['site']);
         }
+        $data['contexts']['runtime'] =  array('version' => phpversion(), 'name' => 'php');
+
 
         if (!$this->breadcrumbs->is_empty()) {
             $data['breadcrumbs'] = $this->breadcrumbs->fetch();
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -884,8 +884,8 @@ class Raven_Client
         if (empty($data['site'])) {
             unset($data['site']);
         }
-        $data['contexts']['runtime'] =  array('version' => phpversion(), 'name' => 'php');
 
+        $data['contexts']['runtime'] =  array('version' => PHP_VERSION, 'name' => 'php');
 
         if (!$this->breadcrumbs->is_empty()) {
             $data['breadcrumbs'] = $this->breadcrumbs->fetch();
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -885,7 +885,9 @@ class Raven_Client
             unset($data['site']);
         }
 
-        $data['contexts']['runtime'] =  array('version' => PHP_VERSION, 'name' => 'php');
+        $existing_runtime_context = isset($data['contexts']['runtime']) ? $data['contexts']['runtime'] : [];
+        $runtime_context = array('version' => PHP_VERSION, 'name' => 'php');
+        $data['contexts']['runtime'] =  array_merge($existing_runtime_context, $runtime_context);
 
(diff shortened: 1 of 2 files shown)


lib/Raven/Client.php
```diff
@@ -884,6 +884,8 @@ public function capture($data, $stack = null, $vars = null)
         if (empty($data['site'])) {
             unset($data['site']);
         }
+        $data['contexts']['runtime'] =  array('version' => phpversion(), 'name' => 'php');
```
ste93cry: The blank line added below this line should probably go above. Also, maybe it would be better to have the `php` string in uppercase. Last thing, the most important: please use the `PHP_VERSION` constant as it's faster because we don't have the overhead of a function call

dcramer: 'php' should not be uppercased as this is a value sent to the Sentry server and is reflective on the actual platform (it generally should always be lowercased)

hjanuschka:
@ste93cry feedback adressed.

ste93cry:
Thank you for the good work, patch looks good for me

lib/Raven/Client.php
```diff
@@ -885,6 +885,8 @@ public function capture($data, $stack = null, $vars = null)
             unset($data['site']);
         }
 
+        $data['contexts']['runtime'] =  array('version' => PHP_VERSION, 'name' => 'php');
```
stayallive: Shouldn't this be merged with a possible user value in the runtime context?

Also there is an extra space character before `array`.

test/Raven/Tests/ClientTest.php
```diff
@@ -1035,6 +1035,20 @@ public function testCaptureMessageWithUserContext()
...
+     */
+    public function testRuntimeContext()
+    {
+        $client = new Dummy_Raven_Client();
+
+        $client->captureMessage('test');
+        $events = $client->getSentEvents();
+        $this->assertEquals(PHP_VERSION, $events[0]['contexts']['runtime']['version']);
+        $event = array_pop($events);
+        $this->assertEquals('php', $event['contexts']['runtime']['name']);
```
stayallive: Maybe also add an test what happens if the context was set by an user (for the runtime) it correctly merges with our runtime context.

hjanuschka:
@stayallive feedback addressd with: https://github.com/getsentry/sentry-php/pull/564/commits/59909d7c1c2ceb08c3327f3508c5d1be4528dc5a

verifies that custom `contextes` and `runtime` get merged

lib/Raven/Client.php
Jean85: That string is strictly validated on server side, so we cannot change it, see [the docs about this](https://docs.sentry.io/clientdev/attributes/#required-attributes) and the related [API JSON schema](https://sentry.io/api/store/schema). 
Lowercase it is then.

hjanuschka:
@Jean85 thx! - added a test to verify that on can override `[runtime][name]`  and `[runtime][version]`

had to change the order of the `array_merge` :D
also fixed tests on 5.3 :rage3:  (`[]`  vs `array()`)

hjanuschka:
thx, you folks rock! 🍺

hjanuschka:
you want me to port it into 1.8 branch ? really would love to see this released.

Jean85:
@hjanuschka that's not necessary, the master branch is still for the 1.x series. The porting to that branch is done when tagging and releasing.

PR #565, opened 2018-03-14
[2.0] Sortable middleware stack

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

I realized that the middleware implementation that was merged with #487 has some problems, in particular:

- the middlewares cannot be added through the client builder
- the middlewares cannot be removed (not even the default ones)
- the middlewares are not sortable, so newly added middlewares will always go to the end of the chain

To solve the last one this PR implements something similar to the priority feature of the processors so that any middleware can be pushed in a specific position, be it the end of the chain, the start or the middle. Also, all the other points have been addressed

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -11,7 +11,6 @@
 
 namespace Raven;
 
-use Psr\Http\Message\ServerRequestInterface;
 use Raven\Breadcrumbs\Breadcrumb;
 use Raven\Breadcrumbs\Recorder;
 use Raven\Context\Context;
@@ -20,6 +19,7 @@ use Raven\Middleware\BreadcrumbInterfaceMiddleware;
 use Raven\Middleware\ContextInterfaceMiddleware;
 use Raven\Middleware\ExceptionInterfaceMiddleware;
 use Raven\Middleware\MessageInterfaceMiddleware;
+use Raven\Middleware\MiddlewareStack;
 use Raven\Middleware\ProcessorMiddleware;
 use Raven\Middleware\RequestInterfaceMiddleware;
 use Raven\Middleware\UserInterfaceMiddleware;
@@ -129,14 +129,9 @@ class Client
     private $serverOsContext;
 
     /**
-     * @var callable The tip of the middleware call stack
+     * @var MiddlewareStack The stack of middlewares to compose an event to send
      */
-    private $middlewareStackTip;
-
-    /**
-     * @var bool Whether the stack of middleware callables is locked
-     */
-    private $stackLocked = false;
+    private $middlewareStack;
 
     /**
      * @var Event The last event that was captured
@@ -163,21 +158,21 @@ class Client
         $this->transaction = new TransactionStack();
         $this->serializer = new Serializer($this->config->getMbDetectOrder());
         $this->reprSerializer = new ReprSerializer($this->config->getMbDetectOrder());
-        $this->middlewareStackTip = function (Event $event) {
+        $this->middlewareStack = new MiddlewareStack(function (Event $event) {
             return $event;
-        };
+        });
 
-        $this->addMiddleware(new ProcessorMiddleware($this->processorRegistry));
-        $this->addMiddleware(new MessageInterfaceMiddleware());
-        $this->addMiddleware(new RequestInterfaceMiddleware());
-        $this->addMiddleware(new UserInterfaceMiddleware());
-        $this->addMiddleware(new ContextInterfaceMiddleware($this->tagsContext, Context::CONTEXT_TAGS));
-        $this->addMiddleware(new ContextInterfaceMiddleware($this->userContext, Context::CONTEXT_USER));
-        $this->addMiddleware(new ContextInterfaceMiddleware($this->extraContext, Context::CONTEXT_EXTRA));
-        $this->addMiddleware(new ContextInterfaceMiddleware($this->runtimeContext, Context::CONTEXT_RUNTIME));
-        $this->addMiddleware(new ContextInterfaceMiddleware($this->serverOsContext, Context::CONTEXT_SERVER_OS));
-        $this->addMiddleware(new BreadcrumbInterfaceMiddleware($this->recorder));
-        $this->addMiddleware(new ExceptionInterfaceMiddleware($this));
+        $this->middlewareStack->addMiddleware(new ProcessorMiddleware($this->processorRegistry), -255);
+        $this->middlewareStack->addMiddleware(new MessageInterfaceMiddleware());
+        $this->middlewareStack->addMiddleware(new RequestInterfaceMiddleware());
+        $this->middlewareStack->addMiddleware(new UserInterfaceMiddleware());
+        $this->middlewareStack->addMiddleware(new ContextInterfaceMiddleware($this->tagsContext, Context::CONTEXT_TAGS));
+        $this->middlewareStack->addMiddleware(new ContextInterfaceMiddleware($this->userContext, Context::CONTEXT_USER));
+        $this->middlewareStack->addMiddleware(new ContextInterfaceMiddleware($this->extraContext, Context::CONTEXT_EXTRA));
+        $this->middlewareStack->addMiddleware(new ContextInterfaceMiddleware($this->runtimeContext, Context::CONTEXT_RUNTIME));
+        $this->middlewareStack->addMiddleware(new ContextInterfaceMiddleware($this->serverOsContext, Context::CONTEXT_SERVER_OS));
+        $this->middlewareStack->addMiddleware(new BreadcrumbInterfaceMiddleware($this->recorder));
+        $this->middlewareStack->addMiddleware(new ExceptionInterfaceMiddleware($this));
 
         if (static::isHttpRequest() && isset($_SERVER['PATH_INFO'])) {
             $this->transaction->push($_SERVER['PATH_INFO']);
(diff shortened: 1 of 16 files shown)


lib/Raven/Client.php
```diff
@@ -451,7 +443,7 @@ public function capture(array $payload)
             $payload['message'] = substr($payload['message'], 0, static::MESSAGE_LIMIT);
         }
 
-        $event = $this->callMiddlewareStack($event, static::isHttpRequest() ? ServerRequestFactory::fromGlobals() : null, isset($payload['exception']) ? $payload['exception'] : null, $payload);
+        $event = $this->middlewareStack->executeStack($event, static::isHttpRequest() ? ServerRequestFactory::fromGlobals() : null, isset($payload['exception']) ? $payload['exception'] : null, $payload);
```
Jean85: Please split this in multiple lines, one argument per line.

lib/Raven/ClientBuilder.php
```diff
@@ -201,6 +206,40 @@ public function removeHttpClientPlugin($className)
...
+    {
+        $this->middlewares[] = [$middleware, $priority];
+
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function removeMiddleware(callable $middleware)
```
Jean85: How would you effectively remove a middleware? If you created it, you could have the instance with you, but it seems an unlikely case... If you already added it, you shouldn't want to remove it.

But what if you want to remove one of the standard middleware? How would you retrieve the original instance to remove it afterwards?

lib/Raven/ClientBuilder.php
```diff
@@ -201,6 +206,40 @@ public function removeHttpClientPlugin($className)
...
+    {
+        foreach ($this->middlewares as $key => $value) {
+            if ($value[0] !== $middleware) {
```
Jean85: If you don't find it, you do not complain... Is it the expected behavior?

lib/Raven/ClientBuilder.php
```diff
@@ -201,6 +206,40 @@ public function removeHttpClientPlugin($className)
         return $this;
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function addMiddleware(callable $middleware, $priority = 0)
+    {
+        $this->middlewares[] = [$middleware, $priority];
```
Jean85: Shouldn't we sort them on insertion as an optimization?
Also that would mean that the getter would serve them always sorted...

lib/Raven/Middleware/BreadcrumbInterfaceMiddleware.php
```diff
@@ -49,7 +49,7 @@ public function __construct(Recorder $recorder)
      *
      * @return Event
      */
-    public function __invoke(Event $event, callable $next, ServerRequestInterface $request = null, \Exception $exception = null, array $payload = [])
+    public function __invoke(Event $event, callable $next, ServerRequestInterface $request = null, $exception = null, array $payload = [])
```
Jean85: Add `\Throwable` in the PHPDoc

lib/Raven/Middleware/ContextInterfaceMiddleware.php
```diff
@@ -56,7 +56,7 @@ public function __construct(Context $context, $contextName)
      *
      * @return Event
      */
-    public function __invoke(Event $event, callable $next, ServerRequestInterface $request = null, \Exception $exception = null, array $payload = [])
+    public function __invoke(Event $event, callable $next, ServerRequestInterface $request = null, $exception = null, array $payload = [])
```
Jean85: Add `\Throwable` in the PHPDoc here too

lib/Raven/Middleware/ExceptionInterfaceMiddleware.php
```diff
@@ -49,7 +49,7 @@ public function __construct(Client $client)
      *
      * @return Event
      */
-    public function __invoke(Event $event, callable $next, ServerRequestInterface $request = null, \Exception $exception = null, array $payload = [])
+    public function __invoke(Event $event, callable $next, ServerRequestInterface $request = null, $exception = null, array $payload = [])
```
Jean85: Here too...

lib/Raven/Middleware/MessageInterfaceMiddleware.php
```diff
@@ -33,7 +33,7 @@
      *
      * @return Event
      */
-    public function __invoke(Event $event, callable $next, ServerRequestInterface $request = null, \Exception $exception = null, array $payload = [])
+    public function __invoke(Event $event, callable $next, ServerRequestInterface $request = null, $exception = null, array $payload = [])
```
Jean85: And here too.

lib/Raven/Middleware/ProcessorMiddleware.php
```diff
@@ -49,7 +49,7 @@ public function __construct(ProcessorRegistry $processorRegistry)
      *
      * @return Event
      */
-    public function __invoke(Event $event, callable $next, ServerRequestInterface $request = null, \Exception $exception = null, array $payload = [])
+    public function __invoke(Event $event, callable $next, ServerRequestInterface $request = null, $exception = null, array $payload = [])
```
Jean85: Here too...

lib/Raven/Middleware/RequestInterfaceMiddleware.php
```diff
@@ -33,7 +33,7 @@
      *
      * @return Event
      */
-    public function __invoke(Event $event, callable $next, ServerRequestInterface $request = null, \Exception $exception = null, array $payload = [])
+    public function __invoke(Event $event, callable $next, ServerRequestInterface $request = null, $exception = null, array $payload = [])
```
Jean85: Here too...

lib/Raven/Middleware/UserInterfaceMiddleware.php
```diff
@@ -32,7 +32,7 @@
      *
      * @return Event
      */
-    public function __invoke(Event $event, callable $next, ServerRequestInterface $request = null, \Exception $exception = null, array $payload = [])
+    public function __invoke(Event $event, callable $next, ServerRequestInterface $request = null, $exception = null, array $payload = [])
```
Jean85: Here too...

lib/Raven/ClientBuilder.php
ste93cry: I'm not sure about what to do in case the middleware does not exists. Throwing could be a good idea, but I don't have a strong opinion on the subject. If I would access a specific key of the array and that doesn't exists then throwing an error is ok, in this case it seems that this behavior does not bring any advantage

lib/Raven/ClientBuilder.php
ste93cry: There is a `getMiddlewares()` method that gives you all the registered middlewares with their priorities, so you could loop over the array and do any type of check to find what you want and then remove it. This is particularly useful if you want to remove/replace a middleware. However I noticed that the `Client` class is missing such such method when it should, but as the format in which the middlewares are stored is different from the one used in the client builder I don't know if we should choose to stick with one or if it's fine to use both of them according to the location

ste93cry: There is a `getMiddlewares()` method that gives you all the registered middlewares with their priorities, so you could loop over the array and do any type of check to find what you want and then remove it. This is particularly useful if you want to remove/replace a middleware. However I noticed that the `Client` class is missing such method when it should, but as the format in which the middlewares are stored is different from the one used in the client builder I don't know if we should choose to stick with one or if it's fine to use both of them according to the location

lib/Raven/ClientBuilder.php
ste93cry: In the case of the client builder the getter is there just to return the list of middlewares that will be added to the client and their priority, so sorting the array is useless as it's just a temporary place where the data is stored

lib/Raven/ClientBuilder.php
Jean85: Maybe we could return a bool?

ste93cry:
>I think there's a bit of duplication between the stack and the part of the client builder..

Unfortunately the same problem exists for the processors and I don't know how to solve it without extracting the registry/stack from the `Client` class and inject it via constructor (which would be a long way better as they can be easily mocked/replaced with a different implementation). The only downside of doing this is that the constructor would start to have many arguments, but honestly this is DI so I don't know how else it could be

Jean85:
The only other way around having a bloated constructor would be grouping dependencies with an intermediate dependency. But I'm not sure if we can do it with some sense...

lib/Raven/ClientBuilder.php
ste93cry: Ok

ste93cry: I just noticed that while this can be done in the `MiddlewareStack` class, the method in the client builder (which is the one under examination) already returns `$this` to implement the fluent interface

lib/Raven/ClientBuilder.php
Jean85: Do we have a reason for having the different format?

ste93cry: Actually (both for the processors and the middlewares) this is done for convenience as there is no need to store temporary data sorted by priority and the format used here is much more easier to handle for operations like removal, find or insertion

PR #566, opened 2018-03-15
[2.0] Refactor the TransactionStack class

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

This is a small refactoring of the `TransactionStack` class. I introduced some BC breaks, but I think they are worth it as they make the code prettier and clearer. The following changes have been introduced:

- The stack is now a private member of the `Client` class and a getter has been added to access it
- The `isEmpty` function has been added and the [`Countable`](http://php.net/manual/en/class.countable.php) interface implemented
- Exceptions are thrown if an user tries to peek or pop from an empty stack instead of silently returning a `null` value
- Validation is done for the pushed entries so that only strings are allowed

Everything should BTW be documented in the `UPGRADE.md` file, please check it and tell me if something is missing. I noticed that I often forget to write the upgrade changes there

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -430,7 +430,7 @@ class Client
 
         if (isset($payload['culprit'])) {
             $event = $event->withCulprit($payload['culprit']);
-        } elseif (!$this->transactionStack->isEmpty()) {
+        } else {
             $event = $event->withCulprit($this->transactionStack->peek());
         }
 
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -56,19 +56,9 @@ class Client
     const MESSAGE_LIMIT = 1024;
 
     /**
-     * @var Recorder The breadcrumbs recorder
-     */
-    protected $recorder;
-
-    /**
      * This constant defines the client's user-agent string.
      */
     const USER_AGENT = 'sentry-php/' . self::VERSION;
-
-    /**
-     * @var TransactionStack The transaction stack
-     */
-    public $transaction;
 
     /**
      * @var string[]|null
@@ -94,6 +84,16 @@ class Client
      * @var Configuration The client configuration
      */
     protected $config;
+
+    /**
+     * @var Recorder The breadcrumbs recorder
+     */
+    private $breadcrumbRecorder;
+
+    /**
+     * @var TransactionStack The transaction stack
+     */
+    private $transactionStack;
 
     /**
      * @var TransportInterface The transport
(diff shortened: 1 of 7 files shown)


lib/Raven/Client.php
```diff
@@ -423,8 +436,8 @@ public function capture(array $payload)
 
         if (isset($payload['culprit'])) {
             $event = $event->withCulprit($payload['culprit']);
-        } else {
-            $event = $event->withCulprit($this->transaction->peek());
+        } elseif (!$this->transactionStack->isEmpty()) {
```
Jean85: This is one of the reasons for which I don't agree with throwing an exception with an empty stack: you are forced to add additional logic even inside the client.
An empty stack is not an "error", so IMHO an exception is not warranted.

tests/ClientTest.php
```diff
@@ -64,6 +64,42 @@ public function registerShutdownFunction()
 
 class ClientTest extends TestCase
 {
+    /**
+     * @dataProvider constructorInitializesTransactionStackDataProvider
+     */
+    public function testConstructorInitializesTransactionStack($isHttpRequest)
+    {
+        if ($isHttpRequest) {
+            $_SERVER['PATH_INFO'] = '/foo';
```
Jean85: Do we have the "backup globals" option enabled in the PHPUnit config?

tests/ClientTest.php
```diff
@@ -64,6 +64,42 @@ public function registerShutdownFunction()
 
 class ClientTest extends TestCase
 {
+    /**
+     * @dataProvider constructorInitializesTransactionStackDataProvider
```
Jean85: I would prefer to split this test in two instead of using this strange data provider

tests/TransactionStackTest.php
```diff
@@ -12,27 +12,155 @@
...
+    public function testConstructorThrowsIfValuesAreNotAllStrings()
+    {
+        new TransactionStack(['a', 1]);
+    }
+
+    public function testClear()
+    {
+        $stack = new TransactionStack();
+
+        $this->assertAttributeEmpty('transactions', $stack);
```
Jean85: Maybe is better to use the public accessor methods instead of inspecting the internal state of the class

tests/TransactionStackTest.php
```diff
@@ -12,27 +12,155 @@
...
+
+        $this->assertNotEmpty($stack);
+        $this->assertFalse($stack->isEmpty());
+    }
+
+    public function testPush()
+    {
+        $stack = new TransactionStack();
+
+        $this->assertAttributeEmpty('transactions', $stack);
```
Jean85: Here too: it's better to use the public methods.

tests/TransactionStackTest.php
```diff
@@ -12,27 +12,155 @@
...
+
+    /**
+     * @expectedException \InvalidArgumentException
+     * @expectedExceptionMessage The $values argument must contain string values only.
+     */
+    public function testPushThrowsIfValuesAreNotAllStrings()
+    {
+        $stack = new TransactionStack();
+
+        $this->assertAttributeEmpty('transactions', $stack);
```
Jean85: Here too: inspecting the internal state shouldn't be done

tests/ClientTest.php
ste93cry: Yes we do

https://github.com/getsentry/sentry-php/blob/2841bb392de677c15497737e3068d00b36d923ce/phpunit.xml.dist#L6

tests/TransactionStackTest.php
ste93cry: There is no way to access the array that holds the values of the stacm from the outside as it would defeat the purpose of the stack to access values randomly and I don't think we should change this to allow it

ste93cry: There is no way to access the array that holds the values of the stack  from the outside as it would defeat the purpose of the stack to access values randomly and I don't think we should change this to allow it

ste93cry: There is no way to access the array that holds the values of the stack from the outside as it would defeat the purpose of the whole class to access values randomly and I don't think we should change this to allow it

ste93cry: I've changed where possible the `assertAttribute*` method calls to use more appropriate functions, but I can't do this in all places as there is no way to access the array that holds the values of the stack from the outside as it would defeat the purpose of the whole class

lib/Raven/Client.php
ste93cry: Java throws an [exception](https://docs.oracle.com/javase/10/docs/api/java/util/EmptyStackException.html) too if trying to access a stack that is empty. For sure it's more tedious to have to wrap everything in a `if` condition, but I think that the intent is more clear this way. As an example, what if the stack could contain any value `null` included? Throwing an exception if the stack is empty makes clear that there is something wrong while returning `null` could be misleading. I still think that even if it's not possible here due to the class only accepting strings we should do it this way

Jean85: I'll give you one solid reason: https://github.com/getsentry/sentry-php/pull/574/files#diff-6d6ad8fef701ebc6ede525e4b6c8f8d0R153

I just merged a PR on master to handle exceptions without stack that we provoke ourselves.

ste93cry: >I just merged a PR on master to handle exceptions without stack that we provoke ourselves.

How could the exception I throw be without stack?

>returning null could be misleading

Although this cannot happen here due to the code that prevents non-string types to be pushed to the stack, a good code would not return null as it could be a value conained in the stack and an end-user would no longer be able to distinguish if the stack is empty or if the returned value has been really peeked/popped from it

PR #567, opened 2018-03-15
[2.0] Add RuntimeContext class to store runtime information

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

This PR aims to fix issue #562 for the `2.x` branch by introducing a specialized `RuntimeContext` class that automatically compiles the runtime SDK interface with the PHP version for the captured events.

### DIFF
--- a/lib/Raven/Context/RuntimeContext.php
+++ b/lib/Raven/Context/RuntimeContext.php
@@ -16,8 +16,8 @@ use Symfony\Component\OptionsResolver\Exception\UndefinedOptionsException;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
- * This class is a specialized implementation of the {@see `Context`} class
- * that stores information about the current runtime.
+ * This class is a specialized implementation of the {@see Context} class that
+ * stores information about the current runtime.
  *
  * @author Stefano Arlandini <sarlandini@alice.it>
  */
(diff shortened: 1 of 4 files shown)


tests/Context/RuntimeContextTest.php
```diff
@@ -0,0 +1,191 @@
...
+                [
+                    'name' => 'php',
+                    'version' => PHP_VERSION,
+                ],
+                null,
+                null,
+            ],
+            [
+                [
+                    'name' => 'foo',
```
Jean85: This should not be allowed, Sentry's server strictly validates this field, only allowed values are: `as3`, `c`, `cfml`, `cocoa`, `csharp`, `go`, `java`, `javascript`, `node`, `objc`, `other`, `perl`, `php`, `python`, `ruby` (see [docs here](https://docs.sentry.io/clientdev/attributes/#required-attributes))

lib/Raven/Context/RuntimeContext.php
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * {@inheritdoc}
+     *
+     * @throws UndefinedOptionsException If any of the options are not supported
+     *                                   by the context
+     * @throws InvalidOptionsException   If any of the options don't fulfill the
+     *                                   specified validation rules
+     */
+    public function offsetSet($offset, $value)
```
Jean85: Do we need this method? Isn't better to add a simple `setPlatform()` and `setVersion()`? Also getters too would be nice.

tests/Context/RuntimeContextTest.php
ste93cry: I checked and you can set whatever you want here, as this is not the `platform` field and so its validation rules does not apply here

lib/Raven/Context/RuntimeContext.php
ste93cry: >Do we need this method?

We need it because it's inherited from the base class, so I have to override it as well and check that what you try to set is allowed by the interface schema definition

>Isn't better to add a simple `setPlatform()` and `setVersion()`? Also getters too would be nice.

Done

PR #569, opened 2018-03-17
[2.0] Implement a specialized context class for the server OS context type

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

Following the lines of the PR that implemented a `RuntimeContext` class, this one introduces a specialized version of the `Context` class that handles the context type representing the operating system. These are the information printed in Sentry

| Property Name | Value |
| ------------- | ------------- |
Version | build 16299 (Windows 10) (10.0)
Kernel Version | Windows NT DESKTOP-XXXXXXX 10.0 build 16299 (Windows 10) AMD64

### DIFF
--- a/lib/Raven/Context/ServerOsContext.php
+++ b/lib/Raven/Context/ServerOsContext.php
@@ -1,0 +1,207 @@
+<?php
+
+/*
+ * This file is part of Raven.
+ *
+ * (c) Sentry Team
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Raven\Context;
+
+use Symfony\Component\OptionsResolver\Exception\InvalidOptionsException;
+use Symfony\Component\OptionsResolver\Exception\UndefinedOptionsException;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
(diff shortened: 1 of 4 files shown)


lib/Raven/Context/ServerOsContext.php
```diff
@@ -0,0 +1,207 @@
...
+     * {@inheritdoc}
+     *
+     * @throws UndefinedOptionsException If any of the options are not supported
+     *                                   by the context
+     * @throws InvalidOptionsException   If any of the options don't fulfill the
+     *                                   specified validation rules
+     */
+    public function __construct(array $data = [])
+    {
+        $this->resolver = new OptionsResolver();
```
Jean85: Since the option resolver seems to be used in multiple context, maybe we can refactor in the parent class?

ste93cry: It's actually only used in the `RuntimeContext` and `ServerOsContext` classes. The other contexts don't need it as the data they contain is too generic to be defined as a set of options

PR #570, opened 2018-03-18
Update docs config

stayallive:
Currently the generic php instructions are prepended on the Laravel & Symfony getting started pages in Sentry, this should not be the case since they are not needed and confusing at best.

Also added Laravel 4.x & Lumen 5.x instructions to that page.


stayallive:
Yeah it;s a bit weird but those docs are here: https://github.com/getsentry/sentry-php/tree/master/docs/integrations

We probably would do good to just link there from the README.md maybe and only keep this as the source of truth for the docs since otherwise they just keep getting out of sync.

Issue #563, opened 2018-03-14
Reject back-to-back duplicate errors/messages


MeredithAnya:
I'm not sure if this possible to add, but if so, it seems reasonable to attempt to reject back-to-back errors that are the same and caused by the same user. Similar to what the JavaScript SDK has here: https://github.com/getsentry/raven-js/pull/861

Jean85:
As discussed on Slack, I think we should target this feature to the 2.x branch.
I think that this is feasible, but I didn't delve into the code to prove it.

@stayallive @ste93cry what do you think?

Spriz:
It seems like this was already a feature on the 1.x release? Was this removed intentionally?

ste93cry:
I'm not aware of any feature like this on `1.x` and this has never been implemented in `2.x`, at least for now

Issue #391, opened 2016-12-19
Fatal errors are not recorded with the curl method async


e98cuenc:
Sentry flushes the unsent messages and calls curl join in the onShutdown method, which is correctly registered to run on shutdown.

However, fatal errors are logged on the ErrorHandler shutdown callback, which is registered after `onShutdown`. (`onShutdown` is registered on the Client constructor, and `error_handler->registerShutdownFunction` is call after you build the Client).

The end result is that a Fatal error gets recorded internally, but it's not send as `onShutdown` has already been called.

You need to make sure `onShutdown` runs after all other shutdown handlers.

mfb:
Although some work was done on this last year, I believe this is still an issue, see https://github.com/getsentry/sentry-php/issues/552#issuecomment-366473263

Jean85:
Isn't `onShutdown` called on the client destructor?

BTW yes, this is surely related to #552 

@nokitakaze IIRC you wrote this part, what do you think?

nokitakaze:
I'm afraid I don't know. This case must be tested in separate process to test the behaviour.

mfb:
Hmm not sure off the top of my head, if you had access to $client->_curl_handler->requests you could see if the requests were sent off?

mfb:
here's code to reproduce the problem, just need a way to detect it:
```php
include './vendor/autoload.php';
ini_set('memory_limit', '8M');
$client = new Raven_Client('...', ['curl_method' => 'async']);
$client->install();
while (TRUE) {
  $a[] = 'b';
}
```

Jean85:
I think that the only way to detect it would be adding first an other exception handler which could output the result... I'll try that now.

Jean85:
I cannot reproduce the issue, at least not under PHP 7.
This is my test script:

```php
<?php
include '../vendor/autoload.php';

$dsnPublicSite = 'https://*****:******@sentry.io/*****';
$client = new Raven_Client($dsnPublicSite, ['curl_method' => 'async']);
$client->install();

ini_set('memory_limit', '8M');
while (TRUE) {
    $a[] = 'b';
}
```
... and this is the correct result:

![schermata del 2018-03-20 23-21-39](https://user-images.githubusercontent.com/6729988/37685906-8cb7eb2e-2c95-11e8-878a-b39773c9793e.png)

I think that curl is called correctly because it's registered as a shutdown function here: 

[...]

PR #571, opened 2018-03-19
Revert ignoring fatal errors on PHP 7+

### LINKED ISSUE #552: Capturing E_ERROR fatal errors broken on PHP 7
As commented by @mfb on #514:

> What are you all doing to capture fatal PHP errors, such as allowed memory size exhausted?  To capture this fatal error successfully, I had to comment out the `return FALSE;` in shouldCaptureFatalError():
>
> ```php
> if (PHP_VERSION_ID >= 70000 && $type === E_ERROR) {
>   // return false;
> }
> ```
>
> sample code:
>
> ```php
> <?php
> include '../vendor/autoload.php';
> $client = new Raven_Client('...');
> $client->install();
> while (TRUE) {
>   $a[] = 11111111111;
> }
> ```

It seems like this does prevent those errors from being reporting, however removing that seems to bring us back to the original issue causing 2 reports for the same error, likely caused by https://github.com/getsentry/sentry-php/blob/25a2134ad9461e375baa22fe1e186796edab60a9/lib/Raven/ErrorHandler.php#L84 which causes the following error in my PHP 7.2 installation:

```
Parse error: syntax error, unexpected '}' in /private/var/www/sentry-php/error.php on line 5

Fatal error: Exception thrown without a stack frame in Unknown on line 0
```

Which are both reported, which is ofcourse incorrect.

Jean85:
This PR attempts to fix #552 allowing all fatals to be captured, with no exception anymore. This is because further investigations lead to discover that the issues in double or loop reporting of fatals where due to a bug on Symfony: see the issue https://github.com/symfony/symfony/issues/26438 and the fix https://github.com/symfony/symfony/pull/26568.

Eventual double reporting should be not fixed here in the client, but on the native integrations, like in the Symfony Bundle.

This PR also adds a regression test with a PHPT test, which could be run with PHPUnit with no issues.

### DIFF
--- a/lib/Raven/ErrorHandler.php
+++ b/lib/Raven/ErrorHandler.php
@@ -140,13 +140,6 @@ class Raven_ErrorHandler
 
     public function shouldCaptureFatalError($type)
     {
-        // Do not capture E_ERROR since those can be caught by userland since PHP 7.0
-        // E_ERROR should already be handled by the exception handler
-        // This prevents duplicated exceptions in PHP 7.0+
-        if (PHP_VERSION_ID >= 70000 && $type === E_ERROR) {
-            return false;
-        }
-
         return $type & $this->fatal_error_types;
     }
 
--- a/test/Raven/Tests/ErrorHandlerTest.php
+++ b/test/Raven/Tests/ErrorHandlerTest.php
@@ -207,8 +207,6 @@ class Raven_Tests_ErrorHandlerTest extends \PHPUnit\Framework\TestCase
                        ->getMock();
         $handler = new Raven_ErrorHandler($client);
 
-        $this->assertEquals($handler->shouldCaptureFatalError(E_ERROR), PHP_VERSION_ID < 70000);
-
         $this->assertEquals($handler->shouldCaptureFatalError(E_WARNING), false);
     }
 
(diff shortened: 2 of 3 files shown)


dcramer:
fwiw the core client should handle double reporting when it can (I think it does/did?). No reason to re-write the same thing.

Jean85:
@dcramer that kind of stuff is requested in #563. The kind of duplication that I'm talking about here is different, is due to third party integrations, and the scenario is like:

 * a fatal error is thrown
 * Sentry captures it correctly through the `ErrorHandler::handleError` function
 * Sentry the passes the error to the previous error handler
 * The previous error handler is Symfony's, which rethrows errors as `FatalThrowableError` exceptions
 * Sentry's exception handler kicks in...

In this way the error is reported twice, once as an error, once as a `FatalThrowableError`. Since those are different, the core cannot handle it in itself, but a simple `ignore_fatal_throwables` option on the Symfony bundle could fix this easily.

dcramer:
@Jean85 ah nevermind then - agreed.

Jean85:
With this I think it's time to release 1.8.4... @stayallive what do you think? We need to update the changelog too!

stayallive:
Release in https://github.com/getsentry/sentry-php/releases/tag/1.8.4.

mfb:
Another bug I'd suggest looking at re: fatal errors is #391 - I pushed some code that can fix the bug, or perhaps some other refactoring is desired

PR #575, opened 2018-03-21
Force event send after fatal error with curl_method async

Jean85:
This is to address #391: when using the `'curl_method' => 'async'` option, in the case of a fatal error the event is not sent to Sentry, at least in PHP < 7.

For starter, it adds a regression test that checks the order of method calls of `ErrorHandler::handleFatal()` and `Raven_CurlHandler::join()`.

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -221,6 +221,13 @@ class Raven_Client
         if (Raven_Util::get($options, 'install_shutdown_handler', true)) {
             $this->registerShutdownFunction();
         }
+
+        // manually trigger autoloading, as it cannot be done during error handling in some edge cases due to PHP (see #60149)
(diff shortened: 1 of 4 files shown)


Jean85:
Ok I think I've found the issue! Basically we were triggering the autoloader manually inside the `ErrorHandler::handleException()` here:
https://github.com/getsentry/sentry-php/pull/575/commits/1a2022ba226da1bf5315607118d4887ee396a4c0#diff-11a1d8314f22caf150ad02a00b294b58L636

The issue (present only in PHP > 7) is that that function is called also during fatal error handling, during which the autoloader doesn't work anymore; this halted the script completely, without invoking the CurlHandler's shutdown function.

Jean85:
Thanks! I forgot to remove a constant that I used in one of the previous attempts, I'll merge anyway!

PR #576, opened 2018-03-21
Apply other fix for #391

Jean85:
This implements the regression suggested in https://github.com/getsentry/sentry-php/issues/391#issuecomment-374925533 and should finally fix the issue with fatals + curl async.

Thanks again to @mfb for helping me debugging this issue.

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -131,7 +131,7 @@ class Raven_Client
     /**
      * @var bool
      */
-    private $useCompression;
+    public $useCompression;
 
     public function __construct($options_or_dsn = null, $options = array())
     {
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -128,6 +128,11 @@ class Raven_Client
      */
     protected $_shutdown_function_has_been_set;
 
+    /**
+     * @var bool
+     */
+    private $useCompression;
+
     public function __construct($options_or_dsn = null, $options = array())
     {
         if (is_array($options_or_dsn)) {
@@ -195,6 +200,7 @@ class Raven_Client
         $this->context = new Raven_Context();
         $this->breadcrumbs = new Raven_Breadcrumbs();
         $this->_shutdown_function_has_been_set = false;
+        $this->useCompression = function_exists('gzcompress') && extension_loaded('zip');
 
         $this->sdk = Raven_Util::get($options, 'sdk', array(
             'name' => 'sentry-php',
@@ -995,7 +1001,7 @@ class Raven_Client
             return false;
         }
 
-        if (function_exists("gzcompress")) {
+        if ($this->useCompression) {
             $message = gzcompress($message);
         }
 
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -228,12 +228,7 @@ class Raven_Client
             $this->registerShutdownFunction();
         }
 
-        // manually trigger autoloading, as it cannot be done during error handling in some edge cases due to PHP (see #60149)
-        if (!class_exists('Raven_Stacktrace')) {
-            // @codeCoverageIgnoreStart
-            spl_autoload_call('Raven_Stacktrace');
-            // @codeCoverageIgnoreEnd
-        }
+        $this->triggerAutoload();
     }
 
     public function __destruct()
@@ -1516,5 +1511,24 @@ class Raven_Client
     public function setReprSerializer(Raven_ReprSerializer $reprSerializer)
     {
         $this->reprSerializer = $reprSerializer;
+    }
+
+    private function triggerAutoload()
+    {
+        // manually trigger autoloading, as it cannot be done during error handling in some edge cases due to PHP (see #60149)
+       
+        if (! class_exists('Raven_Stacktrace')) {
+            spl_autoload_call('Raven_Stacktrace');
+        }
+
+        if (function_exists('mb_detect_encoding')) {
+            mb_detect_encoding('string');
+        }
+        if (function_exists('mb_convert_encoding')) {
+            mb_convert_encoding('string', 'UTF8');
+        }
+        if (function_exists('mb_convert_encoding')) {
+            mb_convert_encoding('string', 'UTF8');
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


test/Raven/phpt/fatal_reported_with_async.phpt
```diff
@@ -21,7 +22,7 @@ $client->setSendCallback(function () {
 
 $client->install();
 
-register_shutdown_function(function () use ($pendingEvents, $pendingRequests) {
+register_shutdown_function(function () use (&$pendingEvents, &$pendingRequests) {
```
mfb: You are using a reference for `&$pendingRequests` but I think this is still pointing at the list of requests when the script starts rather than actual client state?

Jean85: Nope, it's getting the actual client state at the end, since the test was failing until I applied 6eb54f2

Jean85:
I've decided to avoid fixing this under PHP 5.4 and 5.5. Those are long gone EOLed versions, and the issue is located in the JSON serialization of the exception during fatal error handling, so it's nearly impossible to fix it now.

I think this PR should be merged anyway, so we at least have some value in all the other use cases, and we can bring these regression tests along into 2.x. Can you re-review it?

test/Raven/phpt/fatal_reported_with_async.phpt
mfb: I just double-checked and no, the test will pass even if I break the code such that there are still pending requests.  $pendingRequests is a reference to the initial state of there being no pending requests, not the actual client's pending requests.

Jean85: Damn you're right once again :smile: 

Ok, so is 01a26d3 enough to fix it? It seems to fail now if I comment stuff inside the client...

lib/Raven/Client.php
```diff
@@ -195,6 +200,7 @@ public function __construct($options_or_dsn = null, $options = array())
         $this->context = new Raven_Context();
         $this->breadcrumbs = new Raven_Breadcrumbs();
         $this->_shutdown_function_has_been_set = false;
+        $this->useCompression = function_exists('gzcompress') && extension_loaded('zip');
```
mfb: Why does zip extension matter here? Zip extension doesn't provide gzcompress().

lib/Raven/ErrorHandler.php
```diff
@@ -138,6 +138,8 @@ public function handleFatalError()
                 @$error['message'], 0, @$error['type'],
                 @$error['file'], @$error['line']
             );
+            
+            $this->client->useCompression &= PHP_VERSION_ID > 70000;
```
mfb: Maybe it's just me, but using bitwise operator on boolean feels wrong to me :)

lib/Raven/Client.php
```diff
@@ -1506,4 +1512,23 @@ public function setReprSerializer(Raven_ReprSerializer $reprSerializer)
...
+            spl_autoload_call('Raven_Stacktrace');
+        }
+
+        if (function_exists('mb_detect_encoding')) {
+            mb_detect_encoding('string');
+        }
+        if (function_exists('mb_convert_encoding')) {
+            mb_convert_encoding('string', 'UTF8');
+        }
+        if (function_exists('mb_convert_encoding')) {
```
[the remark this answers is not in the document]
Jean85: The fact that they are there twice is a mistake, I'll delete them ASAP.

As for the rest, functions are autoloaded too, so I need to trigger that in advance, to avoid triggering the autoload during the fatal error handling phase, where it would crash hard.

test/Raven/Tests/ClientTest.php
```diff
@@ -2072,7 +2072,7 @@ public function testEncode()
         $this->assertRegExp('_^[a-zA-Z0-9/=]+$_', $value, 'Raven_Client::encode returned malformed data');
         $decoded = base64_decode($value);
         $this->assertInternalType('string', $decoded, 'Can not use base64 decode on the encoded blob');
-        if (function_exists("gzcompress")) {
+        if (function_exists('gzcompress') && extension_loaded('zip')) {
```
mfb: Another case where I think zip extension doesn't matter?

lib/Raven/Client.php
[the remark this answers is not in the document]
mfb: Oh interesting! Where can I read about autoload for functions?  I thought you couldn't autoload functions..

lib/Raven/Client.php
Jean85: In older PHP versions (5.4 to 5.6, see [this build](https://travis-ci.org/getsentry/sentry-php/jobs/356383426) of 68eb97c, which is the commit before ce2b5b5 which introduced this fix) the zip extension is not always present... I tried to not check the extension but it wasn't working.

lib/Raven/ErrorHandler.php
Jean85: Wops! That works by mere coincidence I think :sweat_smile: fixing now!

test/Raven/Tests/ClientTest.php
Jean85: It's the same stuff, but on the other side, the test... I'll try to change this and see if it still works

lib/Raven/ErrorHandler.php
```diff
@@ -138,6 +138,8 @@ public function handleFatalError()
                 @$error['message'], 0, @$error['type'],
                 @$error['file'], @$error['line']
             );
+            
+            $this->client->useCompression = $this->client->useCompression && PHP_VERSION_ID > 70000;
```
mfb: Would be good to comment why PHP version matters here.

Jean85: Well, now we have a regression test with this PR, so removing that check will make the test fail under PHP 5. 

Basically, the reason is that under PHP 7 handling errors is no more a "special place" (thanks to `\Throwabled` probably), so autoloading and other normal stuff work there too.

lib/Raven/Client.php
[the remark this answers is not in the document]
Jean85: Sorry, I used the wrong term. Functions are not "autoloaded" like classes, but still they are searched by namespace and loaded with a similar logic. So calling a function for the first time only during error handling makes everything crash under PHP 5, and that's why I'm forced to do this workaround and call them at least once before proceeding.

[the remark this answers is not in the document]
mfb: eesh! I had never seen a requirement to call functions first, but happily I don't use PHP 5 anymore so can't dig into this :)

Issue #175, opened 2014-10-02
Support modules key


mal:
`raven-ruby` logs gem versions to Sentry, it'd be great if `raven-php` could do the same with modules installed via composer.

If there's no nice way of getting the information from composer directly, a path setting for locating the `composer.lock` file (which stores the names and version of all installed modules) might be a good alternative.

It would also be useful to be able to include the versions of our own software.

ste93cry:
This feature has finally been implemented in #577 and will be available since version `2.0`. The `composer.json` file will be loaded from the path configured in the `project_root` option of the client

>It would also be useful to be able to include the versions of our own software.

This can be done too by using the `Event::withModules` method. Please note that you will have to manually merge any existing value with the new one as calling that method will replace the existing data

PR #577, opened 2018-03-22
[2.0] Support modules attribute of the event by logging Composer packages

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

A long standing issue (#175) is open to request the implementation of the support for the `modules` attribute of the event. In Sentry this means that the name and versions of the Composer packages used by the project where the SDK is used would be logged and displayed. This PR adds support for this feature

### DIFF
--- a/lib/Raven/Middleware/ModulesMiddleware.php
+++ b/lib/Raven/Middleware/ModulesMiddleware.php
@@ -2,6 +2,7 @@
 
 namespace Raven\Middleware;
 
+use Composer\Composer;
 use Composer\Factory;
 use Composer\IO\NullIO;
 use Psr\Http\Message\ServerRequestInterface;
@@ -28,6 +29,10 @@ final class ModulesMiddleware
      */
     public function __construct(Configuration $config)
     {
+        if (!class_exists(Composer::class)) {
+            throw new \LogicException('You need the "composer/composer" package in order to use this middleware.');
+        }
+
         $this->config = $config;
     }
 
--- a/lib/Raven/Middleware/ModulesMiddleware.php
+++ b/lib/Raven/Middleware/ModulesMiddleware.php
@@ -1,0 +1,66 @@
+<?php
+
+namespace Raven\Middleware;
+
+use Composer\Factory;
+use Composer\IO\NullIO;
+use Psr\Http\Message\ServerRequestInterface;
+use Raven\Configuration;
+use Raven\Event;
+
+/**
+ * This middleware logs with the event details all the versions of the packages
+ * installed with Composer, if any.
+ *
+ * @author Stefano Arlandini <sarlandini@alice.it>
+ */
+final class ModulesMiddleware
+{
+    /**
+     * @var Configuration The Raven client configuration
+     */
+    private $config;
+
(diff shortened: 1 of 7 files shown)


lib/Raven/Middleware/ModulesMiddleware.php
```diff
@@ -0,0 +1,66 @@
...
+
+        if (file_exists($composerFilePath)) {
+            $composer = Factory::create(new NullIO(), $composerFilePath, true);
+            $locker = $composer->getLocker();
+
+            if ($locker->isLocked()) {
+                $modules = [];
+
+                foreach ($locker->getLockedRepository()->getPackages() as $package) {
+                    $modules[$package->getName()] = $package->getVersion();
```
Jean85: It's better to use the `getPrettyVersion()` here, it has a more readable format

ste93cry:
>I'm not sure if we should enable this by default

I don't think we should enable it by default as it could slow down the whole sending of an event because it has to read the filesystem, parse the lock file (which can be huge) and probably Composer itself does a lot of other things

>we should add a config option to enable this

Actually the middleware does nothing if the `composer.json` file does not exists or cannot be found in the root of the project. If this is not enough, having the middleware disabled by default would save us from having to add a new option just for this feature as the config options are already too much

lib/Raven/Middleware/ModulesMiddleware.php
ste93cry: `getPrettyVersion()` formats the version in a different way, e.g. for the `clue/stream-filter` package the version `1.4.0` would appear as `v1.4.0` (this is what is tagged on GitHub) vs `1.4.0.0`. Basically `getVersion` normalizes the string and ensures that the format is consistent for all packages regardeless of how the version was really named. Wouldn't it be better to use this standard format?

ste93cry:
>we should provide and document this middleware

As no documentation has been written yet for any of the new features I would postpone this task to a later stage where everything is complete and we can focus on writing it

ste93cry:
Should I leave the `composer/composer` dependency in the `composer.json`? As the middleware will not be enabled by default we could avoid users users to download the dependency and maybe throw an error if the `Composer` class does not exists with an hint to install it to use the middleware

Jean85:
I think we shouldn't add anything to the `composer.json`, also because the right dependency would be `composer-plugin-api` [like here](https://github.com/Ocramius/PackageVersions/blob/4489d5002c49d55576fa0ba786f42dbb009be46f/composer.json#L14).

I simple `class_exists` with an exception throw should suffice.

ste93cry:
`composer-plugin-api` seems a virtual package for authors that want to provide a plugin for Composer, our middleware is not a plugin so I'm not sure we should use that dependency. Also, for tests we should require `composer/composer` at least in `require-dev`, right?

Jean85:
You're right, the example that I brought is a plugin. Yes, requiring it in dev should be enough.

PR #578, opened 2018-03-26
Fix x-forwarded-proto detection

AD7six:
Forwarded-proto detection is currently looking for headers that aren't set - because:

 * HTTP headers have the prefix HTTP_
 * php's env variables are underscored, not hyphenated

Related: https://github.com/getsentry/sentry-php/pull/251


Jean85:
Build failure is unrelated, it will be probably fixed in #576. Merging!

PR #581, opened 2018-04-18
[2.0] Exclude subclasses and implementations of exceptions

mcfedr:
Allows much better control over which exceptions are excluded
For example now you can exclude all `Http_Exception` subclasses

Previously if I wrote `exclude => ["Exception"]` I would still get all subclasses of exception

### DIFF
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -5,7 +5,8 @@
 - The `environment` option has been renamed to `current_environment`.
 - The `http_proxy` option has been renamed to `proxy`.
 - The `processorOptions` option has been renamed to `processors_options`.
-- The `exclude` option has been renamed to `excluded_exceptions`.
+- The `exclude` option has been renamed to `excluded_exceptions`. Unlike `exclude`,
+   `excluded_exceptions` will also exclude all subclasses. 
 - The `send_callback` option has been renamed to `should_capture`.
 - The `name` option has been renamed to `server_name`.
 - The `project` option has been removed.
--- a/tests/ConfigurationTest.php
+++ b/tests/ConfigurationTest.php
@@ -240,6 +240,11 @@ class ConfigurationTest extends TestCase
                 new \Exception(),
                 false,
             ],
+            [
+                [\Exception::class],
+                new \BadFunctionCallException(),
+                true,
+            ],
         ];
     }
 


lib/Raven/Client.php
```diff
@@ -610,8 +610,10 @@ public function captureException($exception, $data = null, $logger = null, $vars
     {
         $has_chained_exceptions = PHP_VERSION_ID >= 50300;
 
-        if (in_array(get_class($exception), $this->exclude)) {
-            return null;
+        foreach ($this->exclude as $exclude) {
+            if (is_a($exception, $exclude)) {
```
Jean85: Why not use `instanceof`?

test/Raven/Tests/ClientTest.php
```diff
@@ -694,6 +701,34 @@ public function testCaptureExceptionHandlesExcludeOption()
...
+    {
+        $client = new Dummy_Raven_Client(array(
+            'exclude' => array('Dummy_Exception_Interface'),
+        ));
+        $ex = new Dummy_Exception();
+        $client->captureException($ex, 'test');
+        $events = $client->getSentEvents();
+        $this->assertEquals(0, count($events));
+    }
+
```
Jean85: Great, you added tests upfront! :+1: 
You should add a test about "malformed" strings in the `excepts` option, though; I'm not sure how that would be handled.

lib/Raven/Client.php
mcfedr: Because `instanceof` works with the literal class name, not a string.

i.e. `$a instanceof Exception` works but `$a instance of 'Exception'` doesnt

test/Raven/Tests/ClientTest.php
mcfedr: Well `is_a` just returns `false` when you pass it stuff that isn't a class name,

mcfedr:
@stayallive I'll rebase it on the `2.0` branch then

lib/Raven/Client.php
mcfedr: ha, I'm wrong, 

`$a instanceof 'Exception'` doesnt work work,
but `$c = 'Exception'; $a instanceof $c` does... php...

mcfedr:
Lol, so this is already implemented in 2.0, as it uses `instanceof`

mcfedr:
i also added a change to the UPGRADE file that notes that `excluded_exceptions` is a bit different.

mcfedr:
@Jean85 can I use 2.0? is it stable? whats the status?

Jean85:
@mcfedr is currently under development, you can check the progress under the related milestone: https://github.com/getsentry/sentry-php/milestone/1

It should be near completion, but I'm not sure if it's in an usable state..

ste93cry:
>can I use 2.0? is it stable? whats the status?

Although `2.0` works it's still under heavy development and so using it in production is strongly discouraged. We don't have any ETA on when it will be ready for release as there are still a few key places we need to finish to refactor and/or improve

PR #583, opened 2018-04-19
Add excluded_exceptions that is compatible with 2.0

mcfedr:
So, given that I'm sticking with 1.x for the moment, I have made this solution, and I thought I'd share it as it might be interesting to you. I'm probably going to be deploying this on my server now.

Basically, instead of changing how `exclude` works, I've added a new option, `excluded_exceptions` that works just like it does in 2.0, and how I need it to work, excluding all subclasses.

As a bonus might make upgrade path simpler for people as they can start using `excluded_exceptions` in 1.X

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -89,6 +89,7 @@ class Raven_Client
     public $timeout;
     public $message_limit;
     public $exclude;
+    public $excluded_exceptions;
     public $http_proxy;
     protected $send_callback;
     public $curl_method;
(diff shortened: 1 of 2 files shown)


ste93cry:
As 2.0 is still in development and there is time to make the things well, if anyone has a better name for the option I'm open to suggestions. Please note that the name (`excluded_exceptions`) is the [same](https://github.com/getsentry/raven-ruby/blob/a365e13b5f1e948dd3702b1caa651d41e7d4aabd/lib/raven/configuration.rb#L32) for the Ruby SDK

stayallive:
Ah did not notice that, I think it's fine then 👍

Issue #579, opened 2018-04-04
Setting event_id on Exceptions can cause exceptions


leongersen:
I know this is very much an edge case, but I ran into the following problem:

A third party library I'm using throws a custom exception that has a `__set` method. This set method isn't intended to be used with unknown properties, and throws when Sentry/Raven tries to set `$e->event_id` on [this line](https://github.com/getsentry/sentry-php/blob/4b4131a65976cd3f8e320bed2b224474743b6e91/lib/Raven/ErrorHandler.php#L81).

For reproduction:

```PHP
Class CarelessSetException extends \Exception {

    public function __set ( $var, $value ) {
        if ( $var === 'event_id' ) throw new \Exception('Loop!');
    }
}
```

Jean85:
Thanks for reporting this issue! What's the behavior of the client in this circumstance? It does report just one event that traces back to this? Or it creates a proper loop sending multiple events?

leongersen:
The new exception seems to interupt the logging process, so neither: I just end up with no logging.

stayallive:
@leongersen on what PHP version are you? I tried to reproduce using the following but got the `Loop!` exception reported just fine, the `CarelessSetException` is discarded, but there is definitly something sent!

```php
<?php

require __DIR__ . '/vendor/autoload.php';

ini_set('memory_limit', '8M');

$sentry = new Raven_Client('valid_dsn_trust_me');
$sentry->install();

Class CarelessSetException extends \Exception
{
    public function __set($var, $value)
    {
        if ($var === 'event_id') {
            throw new \Exception('Loop!');
        }
    }
}


function throw_me($name)
{
    throw new CarelessSetException("I am throwed: {$name}!");
}

throw_me('throwable');
```

stayallive:
Although reporting the inner exception (which is not really important here since caused by the error tracking) is also not a good idea I created #587 but I am not sure if we should, although I don't see any reason against it 👍

Issue #343, opened 2016-09-02
Raven always reports errors as being in "handleError" or "handleFatalError"


mnpenner:
See screenshots below. The error clearly happened in `main()` but the description always says `handleError` because that's the name of the function in `Raven_ErrorHandler`.

![image](https://cloud.githubusercontent.com/assets/1112998/18218021/e7f52486-7114-11e6-8c4d-9e494f70a34f.png)

![image](https://cloud.githubusercontent.com/assets/1112998/18218072/3d995ed4-7115-11e6-9841-d17c31699711.png)

Would be nice if this reported correctly.

dcramer:
We could def improve this but you want to set your app root (see config docs) as that will improve stacks.

mnpenner:
You mean `app_path`? I did already.

dcramer:
@mnpenner I would have expected ServerConfig.php (assuming thats your code) to be annotated in this case. Is this stacktrace entirely third party code?

I do think that outside of this there might still be a concern with how we're generating the culprit in PHP.

mnpenner:
What do you mean by "annotated"? 

`ServerConfig` is my code. I set `app_path` to `/usr/local/webenginex`.

dcramer:
We split frames between app and "system" (non-app) when the data is available:

![screenshot 2016-09-06 11 48 28](https://cloud.githubusercontent.com/assets/23610/18286475/da98f504-7427-11e6-9a03-38aa2a63f824.png)

mnpenner:
That's a nice feature. I wonder why it's not working?

Is the splitting done in the PHP lib or on the server? Maybe I can track it down if it's in the PHP lib. I'm using `captureException` instead of the registering the error handler.

mnpenner:
I traced it down to `sentry/sentry/lib/Raven/Stacktrace.php` line 97:

```php
            // detect in_app based on app path
            if ($app_path) {
                $data['in_app'] = (bool)(substr($abs_path, 0, strlen($app_path)) === $app_path);
            }
```

That's coming back as `true` so it *is* flagging my internal stuff as "in_app", but it's not showing that way in the web app. Is that a new feature? Is it available in the self-hosted version?

dcramer:
@mnpenner good catch on trailing slash -- we def need to fix that. 

If all frames are in_app then we actually disable the split frames feature. If that sounds like the case here, then we'll chock this one up to needing to improve that detection.

torotil:
[...]

PR #584, opened 2018-04-24
[2.0] Refactor the error and exception handlers

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

This PR is a huge refactoring of the error and exception handlers. I tried to make them as transparent as possible in regards to other handlers, so the error handler converts the errors to exceptions to be able to log them but never throws them unless necessary. The aim of my work is to fix the many bugs that we're experiencing in `1.x`, in particular #579, #552, #391, #343. I'm currently tracking an issue with Zend Diactoros (zendframework/zend-diactoros#298) that prevents any error past the first one from being logged through custom error handlers after usage of the `Stream` class.


lib/Raven/ErrorHandler.php
```diff
@@ -11,203 +9,295 @@
...
-    }
 
-    public function shouldCaptureFatalError($type)
-    {
-        // Do not capture E_ERROR since those can be caught by userland since PHP 7.0
-        // E_ERROR should already be handled by the exception handler
-        // This prevents duplicated exceptions in PHP 7.0+
-        if (PHP_VERSION_ID >= 70000 && E_ERROR === $type) {
-            return false;
+        if (!empty($error) && $error['type'] & (E_ERROR | E_PARSE | E_CORE_ERROR | E_CORE_WARNING | E_COMPILE_ERROR | E_COMPILE_WARNING)) {
```
Jean85: Maybe you should wrap those two lines with a private method to give it a name?

tests/ErrorHandlerTest.php
```diff
@@ -12,239 +12,311 @@
...
     }
 
-    public function tearDown()
+    /**
+     * @dataProvider constructorThrowsWhenReservedMemorySizeIsWrongDataProvider
+     *
+     * @expectedException \UnexpectedValueException
+     * @expectedExceptionMessage The value of the $reservedMemorySize argument must be an integer greater than 0.
+     */
+    public function testConstructorThrowsWhenReservedMemorySizeIsWrong()
```
Jean85: You're not using the data from the provider

lib/Raven/ErrorHandler.php
ste93cry: Extracting the code that just checks whether the error is a fatal error does not makes much sense imho, and there would also be the cost of calling one more function

lib/Raven/ErrorHandler.php
```diff
@@ -11,203 +9,331 @@
...
+    public function handleException($exception)
     {
-        $this->old_exception_handler = set_exception_handler([$this, 'handleException']);
-        $this->call_existing_exception_handler = $call_existing;
+        $this->client->captureException($exception);
+
+        // If there is no exception handler that can run after the current one
+        // give back the exception to the native PHP handler
+        if (null === $this->previousExceptionHandler) {
+            throw $exception;
```
Jean85: Doesn't re-throwing change the stack trace of the exception?

lib/Raven/ErrorHandler.php
```diff
@@ -11,203 +9,331 @@
...
+
+        $previousExceptionHandlerException = $exception;
+
+        // Unset the previous exception handler to prevent infinite loop in case
+        // we need to handle an exception thrown from it
+        $previousExceptionHandler = $this->previousExceptionHandler;
+        $this->previousExceptionHandler = null;
+
+        try {
+            call_user_func($previousExceptionHandler, $exception);
```
Jean85: `$previousExceptionHandler($exception)` is maybe slightly faster?

lib/Raven/ErrorHandler.php
```diff
@@ -11,203 +9,331 @@
...
+
+        // If the exception object instance is the same as the one catched from
+        // the previous exception handler, if any, give it back to the native
+        // PHP handler to prevent infinite circular loop
+        if ($exception === $previousExceptionHandlerException) {
+            throw $exception;
+        }
 
-        return $this;
+        $this->handleException($previousExceptionHandlerException);
```
Jean85: Is this loop-prone?

tests/phpt/fatal_error.phpt
```diff
@@ -0,0 +1,40 @@
...
+
+ErrorHandler::register($client);
+
+register_shutdown_function('register_shutdown_function', function () use ($client) {
+    /** @var \Raven\Transport\HttpTransport $transport */
+    $transport = Assert::getObjectAttribute($client, 'transport');
+    $pendingRequests = Assert::getObjectAttribute($transport, 'pendingRequests');
+
+    Assert::assertEmpty($pendingRequests);
+});
```
Jean85: Those asserts can trigger the autoloading. If this is done under PHP 5 inside an error handling context, it fails silently. You should add an `echo` at the end and expect it to be sure that all those assertions are executed correctly.

tests/phpt/out_of_memory.phpt
```diff
@@ -0,0 +1,28 @@
...
+require $vendor . '/vendor/autoload.php';
+
+$client = ClientBuilder::create()->getClient();
+
+ErrorHandler::register($client);
+
+$foo = str_repeat('x', 1024 * 1024 * 30);
+?>
+--EXPECTF--
+Fatal error: Allowed memory size of %d bytes exhausted (tried to allocate %d bytes) in %s on line %d
```
Jean85: Shouldn't you add something to be sure that you really handled the error?

lib/Raven/ErrorHandler.php
ste93cry: Yes it should be as the next time the function will be called `$this->previousExceptionHandler` will be `null` and after sending the exception to Sentry the exception itself will be rethrown

lib/Raven/ErrorHandler.php
ste93cry: After a few tests it seems it doesn't

tests/phpt/out_of_memory.phpt
ste93cry: Done, I've added an `echo` statement after all the assertions whose existence is checked in the `--EXPECTF--` section of the PHPT test

ste93cry:
I still have to complete the refactor of the ErrorHandler for the breadcrumbs. Please don't merge this PR yet until I remove the WIP prefix

Jean85:
Sorry, I thought it was complete!

ste93cry:
>so an end user would still be able to customize the breadcrumbs.

Would it makes sense for an user to use only the error handler that registers the breadcrumbs and not the handler that captures the error itself?

Jean85:
No, obviously not, why?

ste93cry:
Because then it would not make sense to have two error handlers, one for capturing the errors and one for registering the breadcrumb. Or at least, I don't see the sense behind this. Having two handlers means  reserving memory for both of them (to handle fatal errors) and duplicating the logic for each of them. Also, I don't see why an user would want to not register the breadcrumbs but instead capture the errors

Jean85:
The scenario that I want to preserve is the possibility of replacing (or enhancing) the breadcrumb handler. If you collapse both handler this would not be possible anymore.

ste93cry:
I updated the PR by refactoring the breadcrumb handler and it's ready for another round of code reviews. The [zendframework/zend-diactoros#298](https://github.com/zendframework/zend-diactoros/issues/298) issue persists and if nothing changes before releasing `2.0` we will have to switch PSR-7 library or no errors after the first one will be logged and sent to the Sentry server

ste93cry:
I fixed the fact that fatal errors were reported twice in PHP 7+ (yes, again...) and this should be the good time. Before merging, should I switch PSR-7 library in this PR or do you prefer to postpone it later?

PR #586, opened 2018-04-29
Fix monolog handler not accepting Throwable

### LINKED ISSUE #561: Monolog Breadcrumbs handler does not support Throwable
It seems to me the Monolog breadcrumbs handler does not support `Throwable`, but only `Exception`:

https://github.com/getsentry/sentry-php/blob/512661046e403dc5eb5ae192d5a6cda10e068a95/lib/Raven/Breadcrumbs/MonologHandler.php#L63-L68

In contrast the `RavenHandler` in Monolog itself supports both in `context[exception]`:

https://github.com/seldaek/monolog/blob/fd8c787753b3a2ad11bc60c063cff1358a32a3b4/src/Monolog/Handler/RavenHandler.php#L182-L187

stayallive:
The monolog handler should accept a Throwable as an exception.

Fixes #561.

### DIFF
--- a/lib/Raven/Breadcrumbs/MonologHandler.php
+++ b/lib/Raven/Breadcrumbs/MonologHandler.php
@@ -61,9 +61,9 @@ class Raven_Breadcrumbs_MonologHandler extends AbstractProcessingHandler
             return;
         }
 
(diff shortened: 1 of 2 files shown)


stayallive:
Added test for both `Exception` and `Throwable` (since there was no test for `Exception`).

PR #588, opened 2018-05-02
Fixed undefined variable

Toflar:
Fixed PHP notice `Notice: Undefined variable: active in ... lib/Raven/CurlHandler.php on line 96`.

Depending on your setup, notices are converted to exceptions so you don't forget about them ;-) This notice currently causes my test environment to fail.

### DIFF
--- a/lib/Raven/CurlHandler.php
+++ b/lib/Raven/CurlHandler.php
@@ -94,6 +94,8 @@ class Raven_CurlHandler
      */
     protected function select()
     {
+        $active = false;
+        
         do {
             $mrc = curl_multi_exec($this->multi_handle, $active);
         } while ($mrc == CURLM_CALL_MULTI_PERFORM);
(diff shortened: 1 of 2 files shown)


Toflar:
Failing tests unrelated.

Jean85:
Unfortunately the failure is somewhat related... You're consuming more memory, hence changing the result of the (flaky) PHPT regression test. Can you try to change the memory limit to `16M` here?https://github.com/getsentry/sentry-php/blob/61c862beca7ff8f44ab1636a15e54df5c44de51d/test/Raven/phpt/fatal_reported_with_async.phpt#L44

Toflar:
No idea how that can use more memory but well, done :)

stayallive:
It does seem wierd to consume more memory by just adding a variable (that will be defined by php anyway)... still failing with the memory change though...

Jean85:
Damn, now an other regression test is failing, this time under 5.3... I have to investigate this further.

Toflar:
Yeah, I actually just wanted to get rid of the notice and I think what I did is pretty straightforward, isn't it? 😄

stayallive:
It should, but those tests are really flaky and testing some arcane edge cases so who knows... maybe this is even a cause somehow... 😜

Jean85:
I'm currently unable to reproduce the failure locally :cry:

PR #590, opened 2018-05-03
Releasenotes 1.9.0

stayallive:
Release notes and doc updates for 1.9.0.

> https://github.com/getsentry/sentry-php/compare/1.8.4...master

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,17 @@
 # CHANGELOG
 
-## Unreleased
+## 1.9.0 (2018-05-03)
 
-- ...
+- Fixed undefined variable (#588)
+- Fix for exceptions throwing exceptions when setting event id (#587)
+- Fix monolog handler not accepting Throwable (#586)
+- Add `excluded_exceptions` option to exclude exceptions and their extending exceptions (#583)
+- Fix `HTTP_X_FORWARDED_PROTO` header detection (#578)
+- Fix sending events async in PHP 5 (#576)
+- Avoid double reporting due to `ErrorException`s (#574)
+- Make it possible to overwrite serializer message limit of 1024 (#559)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,8 +1,17 @@
 # CHANGELOG
 
-## Unreleased
```
Jean85: Please leave the `unreleased` header at the top.

CHANGELOG.md
```diff
@@ -1,8 +1,17 @@
...
-
-- ...
+## 1.9.0 (2018-05-03)
+
+- Fixed undefined variable (#588)
+- Fix for exceptions throwing exceptions when setting event id (#587)
+- Fix monolog handler not accepting Throwable (#586)
+- Add `excluded_exceptions` option to exclude exceptions and their extending exceptions (#583)
```
Jean85: I would add here that is a 2.x forward compat functionality.

docs/config.rst
```diff
@@ -241,6 +241,26 @@ The following settings are available for the client:
...
+
+    The timeout for sending requests to the Sentry server in seconds, default is 2 seconds.
+
+    .. code-block:: php
+
+        'timeout' => 2,
+
+.. describe:: excluded_exceptions
+
+    Exception that should not be reported, exceptions extending exceptions in this list will also
```
Jean85: Here too, so it will be clear.

CHANGELOG.md
stayallive: I don't think we should do that yet. We don't know how and if when we release (bit of expectation management), when we do we can create the upgrade guide with all options that are forward compatible and which need removing / updating.

Notice that `excluded` is currently not documented for `1.x` and I intend to leave it that way (but ofcourse do mention it in the upgrade guide later).

PR #553, opened 2018-02-21
Update Serializer.php

truechudo:
The long text was clipped incorrectly. An error "Malformed UTF-8 characters, possibly incorrectly encoded" occurs in the method  encode in class Raven_Client,  when trying to decode the cropped text.
Sample text:
```
<p><strong class="h2">Борис Васильев. «Не стреляйте в белых лебедей»</strong><br></p>
<p><br></p>
<p> Человек и природа – вот основная тема произведения Бориса Васильева «Не стреляйте в белых лебедей».<br></p>
<p><br></p>
<p> Кто-то, как Егор Полушкин и его сынишка, стараются видеть вокруг красоту и помочь братьям нашим меньшим. А кто-то, как Федор Бурьянов и его сын Вова, смотрят на природу через прицел выгоды.<br></p>
<p><br></p>
<p> Вот из-за таких людей осталось от местного озера только название – «Лебяжье», а когда-то здесь действительно обитали лебеди. Егор Полушкин, поехав на рынок, покупает прекрасных белых лебедей для озера, чтобы возродить его красоту.<br></p>
<p><br></p>
<p> Но если для Егора лебеди - это красота, то для жителей деревни – только гастрономический изыск. <br></p>
<p><br></p>
<p> Беззаветно служа природе, Егор Полушкин отдает ей все, даже свою жизнь. Он не выдал своих обидчиков, наверное, в тайне надеясь на дальнейшее их исправление.<br></p>
<p><br></p>
<attach-media class="center" data-type="images" data-position="1"></attach-media>
<p><br></p>
```

PR #592, opened 2018-05-03
[2.0] Merge up to 1.9

Jean85:
This is a continuation of the work of #526, continuosly porting all the new stuff from the master branch into the 2.0. This is the list of merged PRs on master that I've ported here (by merge order on master):

- [x] #543: docs fix
- [x] #546: adding `application/json` input handling
- [x] #547: docs fix bis
- [x] #548: fix Client PHPDoc
- [x] #551: fix syntax highlighting in the README.md
- [x] #556: docs fix ter
- [x] #557: use badge/poser in the README.md
- [x] ~~#555: Cleanup the 'site' value if it's empty~~ `site` is a deprecated field, no longer present
- [x] ~~#564: add runtime info~~ already covered by #562
- [x] #570: fix official docs definition
- [x] #572: handle non-default host port 
- [x] #573: prepare 1.8.4 release
- [x] ~~#578: Fix x-forwarded-proto detection~~ will be handled with #594
- [x] ~~#583: `exclude_exceptions` with child types~~ already handled in #487
- [x] #554: allow serialization 5 levels deep
- [x] #559: make string limit customizable
- [x] #587: avoid looping when exception throws exception
- [x] #586: Fix monolog handler not accepting Throwable
- [x] #553: fix serializer to account for non-UTF-8 chars
- [x] ~~#588: fix for undefined variable~~ not applicable, CurlHandler no longer present
- [x] #590: update for 1.9.0 release
- [x] ~~#591: prepare 1.10.x release~~ not applicable

I will handle those PHPT regression tests separately:
- [ ] #571: add regression tests for fatal handling
- [ ] #574: avoid double reporting
- [ ] #575: force send with curl_method async
- [ ] #576: more regression PHPT tests and fixes
- [ ] #589: fix regression test

### DIFF
--- a/tests/SerializerAbstractTest.php
+++ b/tests/SerializerAbstractTest.php
@@ -366,16 +366,27 @@ abstract class SerializerAbstractTest extends TestCase
         if ($serialize_all_objects) {
             $serializer->setAllObjectSerialize(true);
         }
-        for ($i = 0; $i < 100; ++$i) {
-            foreach ([100, 1000, 1010, 1024, 1050, 1100, 10000] as $length) {
-                $input = '';
-                for ($i = 0; $i < $length; ++$i) {
-                    $input .= chr(mt_rand(0, 255));
-                }
-                $result = $serializer->serialize($input);
-                $this->assertInternalType('string', $result);
-                $this->assertLessThanOrEqual(1024, strlen($result));
-            }
+
+        foreach ([100, 1000, 1010, 1024, 1050, 1100, 10000] as $length) {
+            $input = str_repeat('x', $length);
+            $result = $serializer->serialize($input);
+            $this->assertInternalType('string', $result);
+            $this->assertLessThanOrEqual(1024, strlen($result));
+        }
+    }
+
+    public function testLongStringWithOverwrittenMessageLength()
+    {
+        $class_name = static::get_test_class();
+        /** @var \Raven\Serializer $serializer */
+        $serializer = new $class_name();
+        $serializer->setMessageLimit(500);
+
+        foreach (array(100, 490, 499, 500, 501, 1000, 10000) as $length) {
+            $input = str_repeat('x', $length);
+            $result = $serializer->serialize($input);
+            $this->assertInternalType('string', $result);
+            $this->assertLessThanOrEqual(500, strlen($result));
         }
     }
 
--- a/tests/SerializerAbstractTest.php
+++ b/tests/SerializerAbstractTest.php
@@ -420,6 +420,24 @@ abstract class SerializerAbstractTest extends TestCase
         $serializer->setAllObjectSerialize(false);
         $this->assertFalse($serializer->getAllObjectSerialize());
     }
+
+    public function testClippingUTF8Characters()
+    {
+        if (!extension_loaded('mbstring')) {
+            $this->markTestSkipped('mbstring extension is not enabled.');
+        }
+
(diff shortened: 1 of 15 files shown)


Jean85:
@ste93cry while trying to port #555 I discovered that the `site` field got lost during the development of the 2.x. Is it a custom field that we are ditching since we already have the standard `server_name`?

ste93cry:
It was removed because it's a deprecated field

Jean85:
@ste93cry looking at #578 I discovered that the `trust_x_forwarded_proto` option is handled but never used... what should we do?

ste93cry:
According to the security issues that could arise by using the `X_FORWARDED_*` headers I would remove the configuration option entirely and not take any action regarding support of this particular feature in the PHP SDK

tests/SerializerAbstractTest.php
```diff
@@ -409,6 +420,24 @@ public function testSetAllObjectSerialize()
...
         $this->assertFalse($serializer->getAllObjectSerialize());
     }
+
+    public function testClippingUTF8Characters()
+    {
+        if (!extension_loaded('mbstring')) {
+            $this->markTestSkipped('mbstring extension is not enabled.');
+        }
+
+        $testString = 'Прекратите надеяться, что ваши пользователи будут сообщать об ошибках';
```
Jean85: This is so funny! It's the Google-translated version of `Stop hoping that your users will report bugs` in russian :trollface:

Jean85:
This is no longer WIP, I will handle the PHPT tests separately, please review!

ste93cry:
I know that this argument has already been discussed many times and widely, but I strongly disagree with gathering the POST data and deserialize it if it's JSON. Now that we have a well organized way to collect all information we and users need through middlewares everyone can easily add whatever they want to the event data, and the SDK specs also says:

>This data should not be provided by default as it can get quite large

This is the main reason I did not collect POST data by default in the `RequestInterfaceMiddleware` middleware. If we really want to keep going this way, we should at least use the PSR-7 methods to get the data instead of using `file_get_contents` directly.

tests/SerializerAbstractTest.php
stayallive: I couldn't just add some random text, needs to make a little bit sense at least 😉

Jean85:
> This data should not be provided by default as it can get quite large

Oh well, then I think we should create a separated middleware to do that!

> This is the main reason I did not collect POST data by default in the RequestInterfaceMiddleware middleware. If we really want to keep going this way, we should at least use the PSR-7 methods to get the data instead of using file_get_contents directly.

What's the correct method to do it?

ste93cry:
I think you should use the `getBody()` method of the `ServerRequestInterface` interface

Jean85:
Ok I've reverted the commit about #546, I will handle it in a separate PR where I will create a separated middleware to handle POST body, with JSON handling too.

ste93cry:
#587 should not apply to `2.x` as we don't set anymore the `event_id` property

Jean85:
The `event_id` was set in the code of 1.x. The core of the issue is an exception that throws an other exception on `__set`, which would be triggered in a number of ways; in 1.x was the `event_id`.

ste93cry:
What I meant is that I don't see anything in the test that tries to set the `event_id` property on the exception object (which in turn should throw another exception), so how can the test work? By the way, are you sure that the same case is not covered by my error handler PR?

Jean85:
It's possible that is already covered by your PR, yes.

PR #597, opened 2018-05-04
[WIP] [2.0] Post body middleware

Jean85:
This stems from #592, to handle separately POST bodies.

This middleware should not be added by default, but the user could add it to get the POST body when needed. I'm thinking about adding also JSON handling, and at least some doc reference.


Jean85:
Closing this as no longer usable after #682.

ste93cry:
For anyone landing here: `2.x` is more flexible than `1.x` version so you are still able to write your own integration that fetches the `POST` data and attach it to the event. We will just not provide this feature out of the box

PR #603, opened 2018-05-20
[2.0] Make the Client class implement an interface

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

I extracted an interface from the `Client` class so that third-parties classes can avoid to typehint the real implementation of the Raven client


lib/Raven/Client.php
```diff
@@ -155,7 +162,7 @@ public function __construct(Configuration $config, TransportInterface $transport
         $this->breadcrumbRecorder = new Recorder();
         $this->transactionStack = new TransactionStack();
         $this->serializer = new Serializer($this->config->getMbDetectOrder());
-        $this->reprSerializer = new ReprSerializer($this->config->getMbDetectOrder());
+        $this->representationSerializer = new ReprSerializer($this->config->getMbDetectOrder());
```
Jean85: :+1:

lib/Raven/ClientInterface.php
```diff
@@ -0,0 +1,234 @@
...
+
+    /**
+     * Adds a new middleware with the given priority to the stack.
+     *
+     * @param callable $middleware The middleware instance
+     * @param int      $priority   The priority. The higher this value, the
+     *                             earlier a processor will be executed in
+     *                             the chain (defaults to 0)
+     */
+    public function addMiddleware(callable $middleware, $priority = 0);
```
Jean85: I don't think we should force other implementation to use middlewares...

lib/Raven/Middleware/ExceptionInterfaceMiddleware.php
```diff
@@ -51,8 +51,9 @@ public function __construct(Client $client)
      */
     public function __invoke(Event $event, callable $next, ServerRequestInterface $request = null, $exception = null, array $payload = [])
     {
-        // Do not override the level if it was set explicitly by the user
-        if (!isset($payload['level']) && $exception instanceof \ErrorException) {
+        if (isset($payload['level'])) {
```
Jean85: Was this a bug? Can we put a test here?

ste93cry: It wasn't a bug, I moved some additional code from the `Client::capture` method to the `ExceptionMiddleware` because I think it belongs there and the existing tests still pass

lib/Raven/ClientInterface.php
ste93cry: Middleware are a central part of the new client implementation, why shouldn't these methods be in the interface?

PR #604, opened 2018-05-24
Cleanup the PHP version

stayallive:
As discussed with the SDK team (Slack) we might want to not include the extra part of the PHP version to cleanup the Sentry interface.

image

(working on the unknown browser)

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -900,7 +900,7 @@ class Raven_Client
         }
 
         $existing_runtime_context = isset($data['contexts']['runtime']) ? $data['contexts']['runtime'] : array();
-        $runtime_context = array('version' => PHP_VERSION, 'name' => 'php');
+        $runtime_context = array('version' => PHP_MAJOR_VERSION . '.' . PHP_MINOR_VERSION . '.' . PHP_RELEASE_VERSION, 'name' => 'php');
         $data['contexts']['runtime'] =  array_merge($runtime_context, $existing_runtime_context);
 
         if (!$this->breadcrumbs->is_empty()) {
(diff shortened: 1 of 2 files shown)


ste93cry:
The same thing should be done for the `2.0` branch

stayallive:
Yes, however this does remove the `alpha-1` if that's being used... 

If someone knows an easy way around it I am all ears but otherwise for the sake of simplicity I think we should go for this (you can always override this in your own rumtime context if you really wanted). Because those Ubuntu extra's are terrible.

stayallive:
Yes. But. As you can see in the screenshot that also includes other possible jibberish, I would like to keep that as clean as possible.

ste93cry:
Maybe I misunderstood, but I thought the string `alpha-1` was an example of text present after the PHP version (e.g. RC version which would get reported as `7.2.6RC1`) and before the gibberish of the OS build. In that case it's correct to report `PHP 7.2.6RC1` as runtime version, but I agree that the rest should be removed (`-1+ubuntu14.04.1+deb.sury.org+1` in the case of the screenshot you uploaded). Unfortunately using the constants `PHP_*_VERSION` is not enough as we lose the stability part of the release (alpha, beta, etc), so I think that the only solution to the problem would be to parse out from the `PHP_VERSION` constant the portion we need

stayallive:
Yeah I also, I thought the "extra" version was always behind a "-" however for RC's that's not the case (`7.2.0RC6`) but dev (`7.2.6-dev`) is...

https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2942#issuecomment-318706969

This is a bit of a deep hole by the looks of it. We could take that snippet ^ and modify it for the nice version.

nokitakaze:
Are you sure that `7.0.0` and `7.0.0-RC2` are the same versions?

stayallive:
@Jean85 I added some processing from the snippet in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2942#issuecomment-318706969 so we have the beta/rc version appended to the version. Still okay with it now?

PR #606, opened 2018-05-27
Add `exclude some exceptions` feature

wi1dcard:
In some of frameworks, we usually need to ignore some exceptions like `HTTP Exception` or `Not Found Exception` etc. So I add this feature to php-sdk.

And sometimes, I extend this error handler and check if needed before the exception being captured. I found if the ex is not to be captured, but should be shown to users, I have to copy the `call existing exception handler` codes to my class, I think that's not pretty. So I make this to an single function.

My english is not professional, but I'm trying my best.

### DIFF
--- a/lib/Raven/ErrorHandler.php
+++ b/lib/Raven/ErrorHandler.php
@@ -43,6 +43,7 @@ class Raven_ErrorHandler
         E_COMPILE_WARNING,
         E_STRICT,
     );
+    protected $exclude_exceptions = array();
 
     /**
      * @var array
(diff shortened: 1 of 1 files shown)


wi1dcard:
Hope this will be merged. It's really useful for my team.

stayallive:
Hey @wi1dcard, thanks for making this PR.

In 1.9 we merged #583 which should do exactly the same as your PR if I am understanding it correctly.

Can you confirm if this is any different?

PR #609, opened 2018-06-05
[2.0] Transports documentation for v2

ste93cry:
This PR adds the documentation for the transports of the v2.

### DIFF
--- a/docs/transport.rst
+++ b/docs/transport.rst
@@ -1,0 +1,86 @@
+Sending events
+##############
+
+Sending an event is very straightforward: you create an instance of the client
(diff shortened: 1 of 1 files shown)


ste93cry:
Creating a transport is as simple as implementing the `TransportInterface` interface, however should I add anyway a section that shows how to do it?

Jean85:
> Creating a transport is as simple as implementing the TransportInterface interface, however should I add anyway a section that shows how to do it?

I don't think that would be needed. The interface is clear, and we provide multiple implementations ourselves.

Jean85:
Close/reopen to force new build, it should be fixed now thanks to #614.

PR #611, opened 2018-06-08
[2.0] Add the documentation about the processors

ste93cry:
This PR adds the documentation about the processors. The documentation is similar to the one of the middlewares and we could even think about dropping the concept of processor and replace them with middlewares in the future.


Jean85:
Close/reopen to force new build, it should be fixed now thanks to #614.

PR #610, opened 2018-06-06
[2.0] Add the documentation about the middlewares

ste93cry:
This PR adds the documentation about the middlewares, explaining what they do and how to add or remove them. It also explains briefly how to write your own.

PR #608, opened 2018-06-04
[2.0] Error handling documentation for v2

ste93cry:
This pull request adds some documentation for the error handling in the `2.0` version. I'm not sure if I should open a PR that tracks all documentation at once or if it's better to open a new one for each feature we want to document

PR #614, opened 2018-06-15
[2.0] Fix build cs namespaced constants

Jean85:
Due to a new rule in PHPCSFixer (see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3127), the 2.0 build was failing in recent PRs like #608 #609 #610 #611. This should fix them all.


Jean85:
Well, to be fair, I would prefer to tackle this differently. Maybe we could check the CS in a separate stage in the build, [like I do in the sentry-symfony build](https://travis-ci.org/getsentry/sentry-symfony/builds/382585349?utm_source=github_status&utm_medium=notification). In this way, we would be checking it only once, and just later, with a specific failure, not making it fail alltogether.

I'll try to write a PR in the next days.

PR #615, opened 2018-06-18
Allow public DSN

stayallive:
The secret part of the DSN is deprecated since Sentry 9 which is why this PR is allowing a secretless DSN to be used.

This should probably be released as a "bugfix" sooner than later (I don't see any BC problems).


ste93cry:
I will open the relevant PR for `2.0` as soon as possible

PR #618, opened 2018-06-19
[2.0] Use jobs in Travis build

Jean85:
This PR stems from https://github.com/getsentry/sentry-php/pull/614#issuecomment-397888515.

 - use Travis new Job feature to split tests from other checks like CS
 - add `--prefer-lowest` job
 - report coverage only once, in the last job, using PHPDBG for faster execution

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -5,7 +5,7 @@ tools:
     php_code_coverage: true
     external_code_coverage:
         timeout: 2400 # There can be another pull request in progress
-        runs: 4 # PHP 5.6 + PHP 7.0 + PHP 7.1 + PHP 7.2
+        runs: 1
 
 build:
     environment:
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,17 +22,29 @@ cache:
 
 before_install:
   - if [ "$REMOVE_XDEBUG" = "1" ]; then phpenv config-rm xdebug.ini; fi
-  - composer self-update
 
 install: travis_retry composer install --no-interaction
 
 script:
-  - composer phpcs
-  - composer tests-travis
+  - composer tests
 
-after_script:
-  - wget https://scrutinizer-ci.com/ocular.phar
-  - if [ $(phpenv version-name) = "5.6" ] && [ "$REMOVE_XDEBUG" = "0" ]; then php ocular.phar code-coverage:upload --format=php-clover tests/clover.xml --revision=$TRAVIS_COMMIT; fi
-  - if [ $(phpenv version-name) = "7.0" ] && [ "$REMOVE_XDEBUG" = "0" ]; then php ocular.phar code-coverage:upload --format=php-clover tests/clover.xml --revision=$TRAVIS_COMMIT; fi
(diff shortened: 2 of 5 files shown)


Jean85:
@ste93cry we have a problem with your tests:

https://travis-ci.org/getsentry/sentry-php/jobs/394188014

You used multiple times `error_clear_last()` but it's available only from PHP 7+!

https://github.com/getsentry/sentry-php/blob/4d6ff9297f94f89d4dd01024c4eb65f5fab7607d/tests/ClientTest.php#L201

Why did you need it?

ste93cry:
>Why did you need it?

I needed it because there are some tests that test the catching of errors by throwing them and you have use the `error_clear_last()` function to clear the environment or the next unit test not expecting an error or checking for another error will fail

>You used multiple times error_clear_last() but it's available only from PHP 7+!

There should be the Symfony polyfill that loads on PHP 5.6 and fill the gap. How it's possible that the unit tests passed until now?

ste93cry:
>Why did you need it?

I needed it because there are some tests that test the catching of errors by throwing them and you have use the `error_clear_last()` function to clear the environment or the next unit test not expecting an error or checking for another error will fail

>You used multiple times `error_clear_last()` but it's available only from PHP 7+!

There should be the Symfony polyfill that loads on PHP 5.6 and fill the gap. How it's possible that the unit tests passed until now?

Jean85:
> There should be the Symfony polyfill that loads on PHP 5.6 and fill the gap. How it's possible that the unit tests passed until now?

I honestly don't know! But they are not failing without that call too!
BTW I don't think we should put the polyfill in there, it could change the test behavior.

ste93cry:
>BTW I don't think we should put the polyfill in there, it could change the test behavior.

I don't know why it doesn't fail anymore, I'm sure that it did at the time I put those calls. By the way it doesn't alter the test behavior, it ensures that it starts from a clear state. If anything else throws an error before any test not expecting an error from the `error_get_last()` function it will break the build

Jean85:
> By the way it doesn't alter the test behavior, it ensures that it starts from a clear state. 

I was not referring to the calls but to the polyfill. Making stuff available could create false green tests.

I removed all the calls, the build is green.

ste93cry:
According to the build the `phpdbg` command exists only from PHP 7. Maybe we can continue using xdebug, even though it's slower?

Jean85:
This is one of the reason why I don't like PHPUnit mocks and I prefer Prophecy... This stuff can't happen with it!

ste93cry:
>Ok, I've copied the polyfill code to resolve the issue in the tests

Any reason to not use directly the polyfill (in `require-dev`, of course)? Also, the `error_clear_last()` is called twice and as I said I'm sure that it's here for a reason, so if removing it makes the tests still pass there is something wrong with the test maybe and we should double check (I'm referring to the `ContextText` class where you didn't revert the change)

Jean85:
> Any reason to not use directly the polyfill (in require-dev, of course)?
As I stated before, I do not want to introduce functions that should not be available under PHP 5.6, invalidating the test environment.

> Also, the error_clear_last() is called twice and as I said I'm sure that it's here for a reason, so if removing it makes the tests still pass there is something wrong with the test maybe and we should double check (I'm referring to the ContextText class where you didn't revert the change)

I did a double check, in that case it was superfluous, the class under test doesn't use the error_get_last, only that single test, so no interference.

PR #619, opened 2018-06-20
[2.0] Add PHPStan

Jean85:
This is now easily doable in CI thanks to #618.

This adds an additional job in CI, alongside the CS check, to do static analysis of code. It requires the package using a shim with the PHAR inside, since it requires PHP 7 otherwise.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,13 +36,19 @@ jobs:
         COMPOSER_OPTIONS: "--prefer-lowest"
         REMOVE_XDEBUG: "1"
       install: travis_retry composer update --no-interaction --prefer-lowest
-    - stage: Code style
-      php: 7.1
+    - stage: Code style & static analysis
+    - php: 7.1
       env: 
         CS-FIXER: true
         REMOVE_XDEBUG: "1"
       script: 
         - composer phpcs
+    - php: 7.1
+      env: 
+        PHPSTAN: true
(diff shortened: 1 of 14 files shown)


ste93cry:
A big 👍  from me. Although PHPStan found some things to fix in the tests code (which is great, anyway), should we exclude that directory from the analysis?

Jean85:
I fixed all the stuff up to level 3. I normally prefer to include test code in the analysis, it's not a bad practice. I just excluded the `tests/resources` dir because it has just some small stub function that makes the analysis fail.

ste93cry:
I'm really happy to see that almost all errors are in the docblock and not in the code itself 😄 can we make PHPStan work under PHP 5.6 or there is no way to do it?

stayallive:
"PHPStan requires PHP >= 7.1. You have to run it in environment with PHP 7.x but the actual code does not have to use PHP 7.x features. (Code written for PHP 5.6 and earlier can run on 7.x mostly unmodified.)" (from https://github.com/phpstan/phpstan)

Sounds like we only need to run it on one version since the output (should) be the same always!?

Jean85:
In noticed just now that it's requiring PHPStan during tests too... I'll try to fix that ASAP.

> Sounds like we only need to run it on one version since the output (should) be the same always!?

Yes, exactly!

PR #630, opened 2018-07-13
Do not truncate strings if message limit is set to zero

ezintz:
This pull request fixes a simple issue where we actually would like to track larger strings, if needed. Right now it does not seem to be possible.


stayallive:
Thanks @ezintz!

ezintz:
I did not check the 2.0 branch yet.

PR #634, opened 2018-07-27
Use configured message limit when creating serializers

Wirone:
When `message_limit` is set in Raven client options, serializers should use this value instead of default message limit from `RAVEN_Client::MESSAGE_LIMIT`.


Jean85:
Do we need to fix the same thing on 2.0? @ste93cry ?

PR #637, opened 2018-08-03
[2.0] Realign docs and readme to master

Jean85:
This PR ports to 2.0 #591 #600 and #621.

I also changed the readme to avoid having to port each release edit on that too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ Tagging a Release
 2. Create a new branch for the minor version (if not present):
 
 ```
-$ git checkout -b releases/1.10.x
+$ git checkout -b releases/2.1.x
 ```
 
 3. Update the hardcoded version tag in ``Client.php``:
@@ -134,20 +134,20 @@ $ git checkout -b releases/1.10.x
 ```php
 class Raven_Client
 {
-    const VERSION = '1.10.0';
+    const VERSION = '2.1.0';
 }
 ```
 
 4. Commit the change:
 
 ```
-$ git commit -a -m "1.10.0"
+$ git commit -a -m "2.1.0"
 ```
 
 5. Tag the branch:
 
 ```
-git tag 1.10.0
+git tag 2.1.0
 ```
 
 6. Push the tag:
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ use Raven\ClientBuilder;
 require 'vendor/autoload.php';
 
 // Instantiate the SDK with your DSN
-$client = ClientBuilder::create(['server' => 'http://public:secret@example.com/1'])->getClient();
+$client = ClientBuilder::create(['server' => 'http://public@example.com/1'])->getClient();
 
 // Capture an exception
 $eventId = $client->captureException(new \RuntimeException('Hello World!'));
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ Tagging a Release
 2. Create a new branch for the minor version (if not present):
 
 ```
-$ git checkout -b releases/1.9.x
+$ git checkout -b releases/1.10.x
 ```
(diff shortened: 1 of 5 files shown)


ste93cry:
As we have to align the `docs/index.rst` file with the changes of the `2.0` branch, could we do it in this PR? I think the documentation of the new version is complete, the only thing left is documenting the contexts (if needed)

Jean85:
What work is needed there? I'm not sure if I will have the time...

ste93cry:
The only thing left do to would be to update the content of the `docs/index.rst` file and link all the other files I've added so that the documentation on http://docs.sentry.io/ can be navigate

Jean85:
I would like to avoid doing that here, since this is just a port to realign with master. That stuff can be done in a separate PR

README.md
```diff
@@ -165,15 +165,15 @@ git checkout master
 8. Add the next minor release to the ``CHANGES`` file:
 
 ```
-## 1.10.0 (unreleased)
+## 1.11.0 (unreleased)
```
ste93cry: On the section above this one the referenced version has been updated to `2.x.x` but here it's still `1.x.x`

README.md
```diff
@@ -165,15 +165,15 @@ git checkout master
...
 ```
 
 9. Update the version in ``Client.php``:
 
 ```php
 class Raven_Client
 {
-    const VERSION = '1.10.x-dev';
+    const VERSION = '1.11.x-dev';
```
ste93cry: On the section above this one the referenced version has been updated to `2.x.x` but here it's still `1.x.x`

README.md
```diff
@@ -182,7 +182,7 @@ class Raven_Client
 ```json
     "extra": {
         "branch-alias": {
-            "dev-master": "1.10.x-dev"
+            "dev-master": "1.11.x-dev"
```
ste93cry: On the section above this one the referenced version has been updated to `2.x.x` but here it's still `1.x.x`

README.md
```diff
@@ -126,28 +126,28 @@ Tagging a Release
...
 
 ```
-$ git checkout -b releases/1.9.x
+$ git checkout -b releases/2.1.x
 ```
 
 3. Update the hardcoded version tag in ``Client.php``:
 
 ```php
 class Raven_Client
```
ste93cry: This class does not exists anymore in `2.x` and should be updated to reflect PSR-4 changes

README.md
```diff
@@ -165,15 +165,15 @@ git checkout master
...
 
 ```
-## 1.10.0 (unreleased)
+## 1.11.0 (unreleased)
 ```
 
 9. Update the version in ``Client.php``:
 
 ```php
 class Raven_Client
```
ste93cry: This class does not exists anymore in `2.x` and should be updated to reflect PSR-4 changes

docs/index.rst
```diff
@@ -29,7 +29,7 @@ once and reference it from anywhere you want to interface with Sentry:
 
 .. code-block:: php
 
-    $client = new \Raven\Client('___DSN___');
+    $client = new \Raven\Client('___PUBLIC_DSN___');
```
ste93cry: This code is not valid anymore since `2.x` as the constructor of the class accepts a different number of parameters. You should use the client builder or inject directly the configuration object

Jean85: I will not address this here, it needs a full PR, opened #650 to track this

ste93cry: I'm not really convinced that another PR is the way to go as you already edited the sections above to refer to the new classes introduced in `2.0` and so I don't see the value in not addressing this here, but if you prefer to seperate the things I won't stop you

PR #635, opened 2018-07-31
Support longer creditcard numbers

153957:
See https://github.com/getsentry/raven-python/pull/1280

PR #640, opened 2018-08-03
[2.0] Support longer creditcard numbers

Jean85:
This ports #635 manually to 2.0

### DIFF
--- a/tests/Processor/SanitizeDataProcessorTest.php
+++ b/tests/Processor/SanitizeDataProcessorTest.php
@@ -125,7 +125,19 @@ class SanitizeDataProcessorTest extends TestCase
             [
                 [
                     'extra_context' => [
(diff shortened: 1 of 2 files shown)


tests/Processor/SanitizeDataProcessorTest.php
```diff
@@ -125,7 +125,19 @@ public function processDataProvider()
             [
                 [
                     'extra_context' => [
-                        'ccnumba' => '4242424242424242',
+                        'ccnumba' => str_repeat('9', 13),
```
ste93cry: Just a minor comment, feel free to ignore it: does it really makes sense to not write the string and instead use `str_repeat`?

Jean85: I did it just for readability, so the length of the test string is immediately clear.

PR #641, opened 2018-08-03
[2.0] Cleanup the PHP version

Jean85:
This is a (manual) porting of #604 to 2.0

### DIFF
--- a/lib/Raven/Util/PHPVersion.php
+++ b/lib/Raven/Util/PHPVersion.php
@@ -8,11 +8,12 @@ class PHPVersion
 
     /**
      * @param string $version
+     *
      * @return string
      */
     public static function parseVersion($version = PHP_VERSION)
     {
-        if (! preg_match(self::VERSION_PARSING_REGEX, $version, $matches)) {
+        if (!preg_match(self::VERSION_PARSING_REGEX, $version, $matches)) {
             return $version;
         }
 
--- a/lib/Raven/Util/PHPVersion.php
+++ b/lib/Raven/Util/PHPVersion.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace Raven\Util;
+
+class PHPVersion
+{
+    const VERSION_PARSING_REGEX = '/^(?<base>\d\.\d\.\d{1,2})(?<extra>-(beta|rc)-?(\d+)?(-dev)?)?/i';
+
+    /**
+     * @param string $version
+     * @return string
+     */
+    public static function getParsed($version = PHP_VERSION)
+    {
+        if (! preg_match(self::VERSION_PARSING_REGEX, $version, $matches)) {
+            return $version;
+        }
+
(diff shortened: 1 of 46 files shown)


lib/Raven/Util/PHPVersion.php
```diff
@@ -0,0 +1,26 @@
...
+     */
+    public static function getParsed($version = PHP_VERSION)
+    {
+        if (! preg_match(self::VERSION_PARSING_REGEX, $version, $matches)) {
+            return $version;
+        }
+
+        $version = $matches['base'];
+        if (isset($matches['extra'])) {
+            $version .=  $matches['extra'];
```
stayallive: Extra space here.

tests/Util/PHPVersionTest.php
```diff
@@ -0,0 +1,43 @@
...
+namespace Raven\Tests\Util;
+
+use PHPUnit\Framework\TestCase;
+use Raven\Util\PHPVersion;
+
+class PHPVersionTest extends TestCase
+{
+    /**
+     * @dataProvider versionProvider
+     * @param $expected
```
ste93cry: I would remove these lines as they are not useful at all (they do not specify a typehint nor a description of the parameter) and we don't usually add docblocks to the methods of the unit test classes

tests/Util/PHPVersionTest.php
```diff
@@ -0,0 +1,43 @@
...
+            '-rc10' => "{$baseVersion}-rc10",
+            '-RC10' => "{$baseVersion}-RC10",
+            '-rc2-dev' => "{$baseVersion}-rc2-dev",
+            '-beta-2-dev' => "{$baseVersion}-beta-2-dev",
+            '-beta2' => "{$baseVersion}-beta2",
+            '-beta-9' => "{$baseVersion}-beta-9",
+        ];
+
+        foreach ($phpExtraVersions as $fullVersion => $parsedVersion) {
+            yield [$parsedVersion, $baseVersion . $fullVersion];
```
ste93cry: Can we simplify this by returning the array directly as we did everywhere else?

lib/Raven/Util/PHPVersion.php
```diff
@@ -0,0 +1,26 @@
...
+
+class PHPVersion
+{
+    const VERSION_PARSING_REGEX = '/^(?<base>\d\.\d\.\d{1,2})(?<extra>-(beta|rc)-?(\d+)?(-dev)?)?/i';
+
+    /**
+     * @param string $version
+     * @return string
+     */
+    public static function getParsed($version = PHP_VERSION)
```
ste93cry: I would rename this method to `parseVersion` so that it's more explicit. Also, we should add documentation about what the method and the parameter do (just for consistency with all the other code)

Jean85:
I've solved the conflicts after #648, @ste93cry can I merge this?

PR #642, opened 2018-08-14
Proper case in Raven_Util class name usage

Wirone:
This was causing Psalm (static analysis tool) to complain that `Raven_Util` class does not exist.


mleczakm:
@Jean85 Is it possible to make new release with this fix ASAP? ;) It breaks our pipelines.

stayallive:
A new release will be tagged before the weekend 👍

PR #647, opened 2018-08-26
[2.0] Remove the processors and transform them to middlewares

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

When `2.0` began the middlewares concept was introduced after refactoring the processors. As both middlewares and processors are similar in how they are managed and executed, I've decided to remove the concept of processor. All the functionalities provided by the various processors are still available and added to the client as before. I've also fixed a bug: in the client builder it was possible to add the middlewares to the client, but in reality they were never added to the instance after it was created.


Jean85:
Conflicts here, needs to be resolved.

ste93cry:
Everything is now green 👍

docs/middleware.rst
```diff
@@ -79,11 +94,19 @@ have the following priorities:
 - ``ExceptionInterfaceMiddleware``: 0
 - ``MessageInterfaceMiddleware``: 0
 - ``ModulesMiddleware``: 0
-- ``ProcessorMiddleware``: -250 (this middleware should always be at the end of
-  the chain)
 - ``RequestInterfaceMiddleware``: 0
 - ``SanitizerMiddleware``: -255 (this middleware should always be the last one)
 - ``UserInterfaceMiddleware``: 0
+- ``SanitizeHttpBodyMiddleware``: -200 (this middleware should always be after
```
Jean85: Should we sort them by priority?

docs/middleware.rst
```diff
@@ -27,6 +27,21 @@ several built-in middlewares whose list is:
 - ``UserInterfaceMiddleware``: adds some user-related information like the client
   IP address to the event.
 
+There are also some "special" middlewares that should be executed after all the
+other middlewares so that they can sanitize and remove sensitive information before
+they reach the Sentry server (not all of them are enabled by default):
```
Jean85: Should we mark which one are enabled by default?

ste93cry: Yes it's a good idea, done!

docs/middleware.rst
ste93cry: I've sorted the list by middleware priorities (should I sort from the first one to the last one or viceversa?) and I've added the `ContextMiddleware` as many times as it's really used taking care of describing which context is filled

ste93cry:
I've rebased and addressed all the code review issues, feel free to review again

ste93cry:
Please remember to squash commits before merge or the Git history will have a lot of commits that are not useful

PR #648, opened 2018-08-27
[2.0] Use the Context classes inside the event object

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

After the `*Context` classes have been introduced I started thinking we could use them to enforce data validation inside the events objects too instead of relying on plain arrays. However this led to problems because the context classes were mutable while the event was not. Plus, from a technical point of view we should ensure that if an object is immutable than everything inside itself is immutable too, which is too complicated, so I took the occasion to drop the concept of immutability of the event as I see no benefit anyway in having it for this particular case to simplify the code

### DIFF
--- a/lib/Raven/Middleware/UserInterfaceMiddleware.php
+++ b/lib/Raven/Middleware/UserInterfaceMiddleware.php
@@ -40,7 +40,7 @@ final class UserInterfaceMiddleware
             $userContext['ip_address'] = $request->getHeaderLine('REMOTE_ADDR');
         }
 
-        $event = $event->withUserContext($userContext);
+        $event->setUserContext($userContext);
 
         return $next($event, $request, $exception, $payload);
     }
--- a/lib/Raven/Middleware/UserInterfaceMiddleware.php
+++ b/lib/Raven/Middleware/UserInterfaceMiddleware.php
@@ -40,8 +40,6 @@ final class UserInterfaceMiddleware
             $userContext['ip_address'] = $request->getHeaderLine('REMOTE_ADDR');
         }
 
-        $event->setUserContext($userContext);
-
         return $next($event, $request, $exception, $payload);
     }
 }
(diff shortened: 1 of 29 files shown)


lib/Raven/Middleware/UserInterfaceMiddleware.php
```diff
@@ -40,8 +40,6 @@ public function __invoke(Event $event, callable $next, ServerRequestInterface $r
             $userContext['ip_address'] = $request->getHeaderLine('REMOTE_ADDR');
         }
 
-        $event = $event->withUserContext($userContext);
```
Jean85: Is this lost?

ste93cry:
>You also included CS changes everywhere, but well, they're good.

I think that unfortunately the PHPCS Fixer guys changed again the defaults, so our build was broken. If we can prepare a separate commit (no need for a PR but I don't have direct write access to the branch) which fixes the build so that I don't have to do it here it would be awesome

lib/Raven/Middleware/UserInterfaceMiddleware.php
ste93cry: No it's not, I'm using the `offsetSet` magic method of the `Context` class a few lines above

Jean85:
Done in #649

ste93cry:
I've rebased the branch, PR is ready for review once again

Jean85:
This will conflict with #647 and any other PR that touches the event... should we wait?

ste93cry:
I think that merging one or the other will result in pretty much the same amount of work needed to fix the conflicts, so let's go with this one if it's ready and I will then work on #647 to rebase it on top of these changes

PR #651, opened 2018-08-29
[2.0] Update the documentation index

ste93cry:
This PR updates the documentation index and align it with the README. I've also decided to remove the `examples` folder as I think the docs are enough and I've also removed the `bin` directory because I think that its usefulness is modest


docs/index.rst
```diff
@@ -10,53 +10,49 @@
...
-    $ composer require "sentry/sentry"
+If you just want to get started quickly you should run the following command:
 
-Configuration
--------------
+.. code-block:: bash
 
-The most important part is the creation of the raven client.  Create it
-once and reference it from anywhere you want to interface with Sentry:
+    php composer.phar require sentry/sentry php-http/curl-client guzzlehttp/psr7
```
Jean85: I would add an example on how to use Guzzle as a transport too, since it's so popular that a lot of people already use it.

docs/index.rst
```diff
@@ -10,53 +10,49 @@
...
-Much of the usefulness of Sentry comes from additional context data with
-the events.  The PHP client makes this very convenient by providing
-methods to set thread local context data that is then submitted
-automatically with all events.  For instance you can use the
-``setUserContext`` method to add information about the current user:
+    require 'vendor/autoload.php';
 
-.. sourcecode:: php
+    // Instantiate the SDK with your DSN
+    $client = ClientBuilder::create(['server' => 'http://___PUBLIC_DSN___@example.com/1'])->getClient();
```
Jean85: This makes me think: maybe we should add a separate fast-track method to create a client with just the DSN set? In an other PR obviously... @stayallive @dcramer what do you think?

docs/index.rst
```diff
@@ -10,53 +10,49 @@
...
 
-    $client->setUserContext(array(
-        'email' => $USER->getEmail()
-    ));
+    // Capture an exception
+    $eventId = $client->captureException(new \RuntimeException('Hello World!'));
```
Jean85: What about automatic capturing? How is it done?

ste93cry: Actually everything about the error handling is explained in its own dedicated page (which corresponds to the file `error_handling.rst`). Putting it here too means content duplication, and the `quickstart.rst` already explains everything in more details. As this is the "homepage" of the docs I don't know what we should really put into it, but for sure not everything

PR #652, opened 2018-08-30
Change specification for autoloading classes

samnela:
Use PSR-4 for autoloading, PSR-0 is deprecated

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,8 +33,8 @@
         "bin/sentry"
     ],
     "autoload": {
-        "psr-0" : {
-            "Raven_" : "lib/"
+        "psr-4" : {
+            "Raven\\" : "lib/"
         }
     },
     "scripts": {


stayallive:
Thanks for your contribution, however if you make this change you would also have to namespace all classes and rename them correctly. Currently `Raven_Client` cannot be autoloaded with PSR-4 (because it's not namespaced and following the correct naming convention) as far as I know.

Yes it's currently deprecated but it still works fine everywhere (not seen any composer warnings about it yet). Also there is a lot of work going on in the 2.0 branch that will address this 👍

Jean85:
Yes, 2.0 is already using PSR-4 correctly, see #414 & #440:

https://github.com/getsentry/sentry-php/blob/2.0/composer.json#L46-L53

samnela:
Oh Yes , I hadn't seen  branch 2.0 , really sorry for the noise  .

stayallive:
@samnela don't worry 👍 happy you took the time to make the PR in the first place.

PR #654, opened 2018-09-10
Do not force IPv4 protocol by default

skrivy:
Todays world is moving to IPv6 fast. Let's not be tied to IPv4 and let the system to choose, which protocol to use.

For networks with broken IPv6 implementation - the ability to force IPv4 only connection still remains.


stayallive:
Although I agree forcing IPv4 by default is not a good thing (we have DNS and system configs for that) what prompted you to make this change?

stayallive:
I see you updated the description, in that case 💯

skrivy:
Ok. Would that be a problem to merge this change first and then create Issue for the second part?

Jean85:
No problem!

PR #658, opened 2018-09-26
[2.0] Rename middlewares

Jean85:
This follows up a small discussion with @ste93cry, where we noted that a middleware had a name that didn't reflet its true purpose. I've found also an other one, and I've realigned a doc file about that.


Jean85:
I had to add a conflict with `php-http/client-common` 1.8 due to https://github.com/php-http/client-common/issues/109. I've left the conflict version pretty wide so we can restrict it as soon as it's fixed.

PR #659, opened 2018-09-27
[2.0] Change vendor from Raven to Sentry

Jean85:
This is an other PR that I discussed with @ste93cry: changing the namespace from `Raven` to `Sentry`.

This is a big change code-wise, since it impacts every single file, and it's doable only as a BC while releasing a new major, so it should be done now or never. But it's pretty straightforward, since it's just a mass `s/Raven/Sentry/` replace; it also gives us the freedom of removing one level of folder, and moving everything into a more standard `src` folder.


nokitakaze:
I think it is okay, until @dcramer blocks this PR

ste93cry:
Why would he bother blocking this PR when it's just part of the huge refactoring done for the 2.0 version? It also aligns the folder structure to many other opensource libraries out of here and I don't really see any problem in doing it

dcramer:
I'm fine with this, though in the future you can tag @mitsuhiko since I've been pretty removed from sdks lately

Jean85:
Thanks, will do in the future.

PR #663, opened 2018-10-03
[2.0] Fix examples using `server` instead of `dsn`

mindplay-dk:
it's `dsn`, not `server` (unless maybe you mean to change it and didn't do it yet - but the current version expected `dsn`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ use Sentry\ClientBuilder;
 require 'vendor/autoload.php';
 
 // Instantiate the SDK with your DSN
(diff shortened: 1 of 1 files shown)


ste93cry:
Thank you for noticing it, would you mind making the same change in [docs/index.rst](https://github.com/getsentry/sentry-php/blob/2.0/docs/index.rst)?

ste93cry:
I'm going to close this because the option is now called again `dsn`. Even though the docs may not be updated AFAIK they will be deleted from this repository and transferred to [docs.sentry.io](https://docs.sentry.io)

PR #667, opened 2018-10-06
Do not return error id if we know we did not send the error

stayallive:
If we know we did not sent the error to a Sentry server we should not return the error id which could like like it would have been sent.

This makes it easier to debug and easier to show an error id (or not in this case) to users or error pages. If no error was sent we should not have an ID set.

This ofcourse cannot be done when errors are stored for bulk sending.

For example: In Laravel we check if the event id is not null before showing it to the user on the 500 screen, better not show it if we are not sending it to not get error id's from users that are never sent.


stayallive:
@Jean85 [yes](https://github.com/getsentry/sentry-php/blob/2.0/src/Client.php#L339-L350), I will make a PR for that now :).

stayallive:
Doesn't matter, it's there to keep track of it. Merged or not we are going to need to take a look at it again (or not). Better to have it than not.

Jean85:
It would make sense only if a test is added, that wouldn't need to be changed much.
Otherwise, IMHO an issue is better.

Jean85:
BTW @stayallive already did it in #671

ste93cry:
As far as I saw in the unified API a `bool` is not enough. Instead in the Javascript SDK they return the status which is an enum of values

PR #668, opened 2018-10-06
Fix PHP version not matching when running on php master

stayallive:
This should fix the builds on php nightly


stayallive:
Yeah I agree, however since that function itself is already tested I thought it would be fine.

PR #671, opened 2018-10-12
[2.0] Do not return error id if we know we did not send the error

stayallive:
If we know we did not sent the error to a Sentry server we should not return the error id or keep track of that event as the last event.


HazAT:
Closing this, this is already in `2.0`, we always return `?string` now.

PR #674, opened 2018-10-15
[2.0] feat: Add Scope & Hub

HazAT:
This PR introduces the `Hub` and `Scope` to the SDK.
Additionally, it adds `User` & `Severity`.

While there a still some TODOs in the code and if used together with the current Client it will not work yet.
Next step would be to refactor the `Client` to work together with the `Hub` and `Scope`.


src/Context/DefaultContextTrait.php
```diff
@@ -0,0 +1,92 @@
...
+     * @throws InvalidOptionsException   If any of the options don't fulfill the
+     *                                   specified validation rules
+     */
+    public function __construct(array $data = [])
+    {
+        $this->resolver = new OptionsResolver();
+
+        $this->configureOptions($this->resolver);
+
+        parent::__construct($this->resolver->resolve($data));
```
Jean85: I don't like this trait a lot, because it calls `parent` many times, so it's making assumptions.

Maybe here we're missing an `abstract` intermediate class?

Jean85: Or even we can move this to the base `Context` class?

src/Interfaces/Severity.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Sentry\Interfaces;
+
+final class Severity
```
Jean85: This will replace `Client::translateSeverity` and some more in the future, right?

src/State/Hub.php
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Class Hub.
+ *
+ * Responsible for maintaining a stack of Client <-> Scope. Basically it's the main entry point to talk to the Client.
+ */
+final class Hub
+{
+    /**
+     * @var array Containing {@link Layer}
```
Jean85: This will become `Layer[]`

src/State/Hub.php
```diff
@@ -0,0 +1,147 @@
...
+            return $client->captureMessage($message);
+        }
+    }
+
+    /**
+     * @param mixed $exception
+     *
+     * @return null|string
+     */
+    public function captureException($exception): ?string
```
Jean85: We can typehint against `\Throwable`

Jean85: What is the returned string here?

src/State/Layer.php
```diff
@@ -0,0 +1,57 @@
...
+namespace Sentry\State;
+
+use Sentry\Client;
+
+final class Layer
+{
+    private $client;
+    private $scope;
+
+    public function __construct(?Client $client, ?Scope $scope)
```
Jean85: Does it make sense to have a Layer without a client? And one without a scope?

src/State/Scope.php
```diff
@@ -0,0 +1,202 @@
...
+use Sentry\Context\UserContext;
+use Sentry\Event;
+use Sentry\Interfaces\Severity;
+
+final class Scope
+{
+    /**
+     * Array holding all breadcrumbs.
+     *
+     * @var array
```
Jean85: `Breadcrumb[]`?

src/State/Scope.php
```diff
@@ -0,0 +1,202 @@
...
+     * @var null|Severity
+     */
+    private $level = null;
+
+    /**
+     * Array of eventProcessors. Closure receiving the event, they should return an event or null.
+     *
+     * @var array
+     */
+    private $eventProcessors;
```
Jean85: We can probably recycle the middleware here, in the future.

src/State/Scope.php
```diff
@@ -0,0 +1,202 @@
...
+        return $this;
+    }
+
+    /**
+     * @param Breadcrumb $crumb
+     * @param int        $maxBreadcrumbs
+     *
+     * @return Scope
+     */
+    public function addBreadcrumb(Breadcrumb $crumb, int $maxBreadcrumbs = 100): self
```
Jean85: Does it make sense to pass `$maxBreadcrumbs` each time? Maybe we can truncate it during serialization?

src/State/Scope.php
```diff
@@ -0,0 +1,202 @@
...
+    /**
+     * @param string $key
+     * @param string $value
+     *
+     * @return Scope
+     */
+    public function setTag(string $key, string $value): self
+    {
+        if (null === $this->tags) {
+            $this->tags = new TagsContext();
```
Jean85: Why are you not instantiating those at construction?

tests/State/HubTest.php
```diff
@@ -0,0 +1,124 @@
...
+use Sentry\Context\TagsContext;
+use Sentry\Event;
+use Sentry\State\Hub;
+use Sentry\State\Scope;
+
+class HubTest extends TestCase
+{
+    public function testWithScope()
+    {
+        $reflectionProperty = new \ReflectionProperty(Scope::class, 'tags');
```
Jean85: Why do you need reflection for this test?

tests/State/HubTest.php
```diff
@@ -0,0 +1,124 @@
...
+        $reflectionProperty->setAccessible(false);
+    }
+
+    public function testPop()
+    {
+        $reflectionProperty = new \ReflectionProperty(Hub::class, 'stack');
+        $reflectionProperty->setAccessible(true);
+        $hub = new Hub();
+        $hub->pushScope();
+        $this->assertCount(2, $reflectionProperty->getValue($hub));
```
Jean85: You can use `assertPropertyCount` to avoid using reflection directly.

I will not comment on other tests, there are similar fixes doable.

src/Interfaces/Severity.php
HazAT: Yes, that's the idea.

src/State/Hub.php
HazAT: It should always return the `eventId`

src/State/Scope.php
HazAT: I am not sure if it will stay like this, I basically "copied" the functions and logic flow from JavaScript. 
When we change the client later this will be used.

src/State/Scope.php
HazAT: Yes, kind of.

src/Context/DefaultContextTrait.php
ste93cry: Just a note: I didn't put this in the base `Context` class because not all contexts need validation and it would be useless to use the `OptionsResolver` component in those cases

src/State/Hub.php
ste93cry: If we really want to return the ID maybe we should return an instance of the `Ramsey\UuidInterface` interface

HazAT: We decided to just return `id` in case for unified API. 
Sometimes the event will be generated async in the background, enriched with some data. We want that this function synchronously always returns the `id` for the event.
While returning the event may work in PHP it does not for all other languages.

HazAT: https://github.com/getsentry/sentry-php/pull/674#discussion_r226196114

HazAT: @ste93cry It should be really easy to get the `id` without calling any getter or type casting. `string` is just fine.

PR #679, opened 2018-10-22
[2.0] Implement the Scope and Hub of the unified API

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

This PR aims at providing the Scope and Hub features of the unified SDK API. It's based on the work of @HazAT and improves it by adding all the needed docblocks and refactoring the code while improving at the same time the unit tests. There are a few inconsistencies of the API which I would like to discuss before merging:

- the `capture*` methods of the `Scope` do not follow the same pattern for the function definition: some let you set the severity level of the event, others don't. Also there is no way to pass a customized payload to the client (see the `Client::capture*` methods to understand what I'm referring to)
- the `Scope` class does not allows to set data on all contexts: runtime and server OS contexts are missing. Also the API of the scope doesn't expose the underlying context object which means that the methods developed to easy the management of the context data are not available to the end user.
- the `Hub::applyToEvent` method behaves differently from data to data being merged into the event. As an example, the breadcrumbs are not merged with the existing ones but the severity level always override the one set in the event. Why?


HazAT:
When talking about `capture*` on the `Scope` you actually meant `Hub`, right?
So I know this is one of the bigger changes of the API. Before our `capture*` methods took a `payload` which can be anything the event takes? 
While this maybe worked in PHP in other SDKs you were not able to pass anything here the event takes, only for example `tags` `extra` and so on, but not an `exception` with for example.

That's why we went with the approach to remove this "magic" payload parameter and introduced `Hub::withScope` and `EventProcessors` as a workaround for this.

---

If someone wants to change `server` or `os` context of an event, they should write their own integration or add an `EventProcessor`. We do not want to expose all possible getter/setter that is on the event. Instead, if someone wants to fundamentally modify the event they should, as said before an `EventProcessor`, an `Integration` or `beforeSend` callback.

---

`Scope::applyToEvent` wasn't just fully implemented by me, ultimately the `Scope` should be the single source of truth when it comes to the "State". Which means the current `Scope` holds all breadcrumbs, context data (most used) and `EventProcessors`. Before an event will be sent off to Sentry if should go through `Scope::applyToEvent` where we merge the "State" of the `Scope` onto the event.

src/State/Scope.php
```diff
@@ -0,0 +1,310 @@
...
+
+        return $this;
+    }
+
+    /**
+     * Gets the tags contained in the tags context.
+     *
+     * @return array<string, string>
+     */
+    public function getTags(): array
```
HazAT: This applies to all `getters` in the `Scope`. I on purpose did not expose the `getters` since we do not want the user to hold on to references of the `context` and modifies them outside of `configureScope` or `withScope`.
Also there should actually never be a reason for the user the call a `getter` of the `Scope`. Again for this reason we have `beforeSend` and `EventProcessors` / `Integrations`.

src/BreadcrumbErrorHandler.php
```diff
@@ -44,7 +44,7 @@ public function doHandleException($exception)
         }
 
         $this->client->leaveBreadcrumb(new Breadcrumb(
-            $this->client->translateSeverity($exception->getSeverity()),
+            Severity::fromString($this->client->translateSeverity($exception->getSeverity())),
```
Jean85: This seems a not useful double-run... Shouldn't we strip the severity translation from the client and put it in completely inside `Severity`?

src/Breadcrumbs/Breadcrumb.php
```diff
@@ -97,15 +92,15 @@ public function __construct($level, $type, $category, $message = null, array $me
...
+     * @param Severity    $level    The error level of the breadcrumb
      * @param string      $type     The type of the breadcrumb
      * @param string      $category The category of the breadcrumb
      * @param string|null $message  Optional text message
      * @param array       $metadata Additional information about the breadcrumb
      *
      * @return static
      */
-    public static function create($level, $type, $category, $message = null, array $metadata = [])
+    public static function create(Severity $level, $type, $category, $message = null, array $metadata = [])
```
Jean85: Is this static method still useful, since it's the same as the constructor?

src/Breadcrumbs/Breadcrumb.php
```diff
@@ -142,27 +137,23 @@ public function withType($type)
     /**
      * Gets the breadcrumb level.
      *
-     * @return string
+     * @return Severity
      */
-    public function getLevel()
+    public function getLevel(): Severity
```
Jean85: Maybe we can rename it to `getSeverity()` and rename the property too, for consistency?

tests/State/LayerTest.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+/*
+ * This file is part of Raven.
```
HazAT: Not really specific to this file, but we should, in the end, decide what to do with this `License` headers. I don't really find the necessary and helpful.
Also we do not want to use `Raven` as a term anywhere ^^

src/Severity.php
```diff
@@ -0,0 +1,157 @@
...
+ * to.
+ *
+ * @author Stefano Arlandini <sarlandini@alice.it>
+ */
+final class Severity
+{
+    /**
+     * This constant represents the "debug" severity level.
+     */
+    public const SEVERITY_DEBUG = 'debug';
```
Jean85: Isn't the `SEVERITY_` prefix a little redundant? It's `Severity:SEVERITY_DEBUG` like this...

src/State/Hub.php
```diff
@@ -0,0 +1,228 @@
...
+     * actions prior to an error or crash.
+     *
+     * @param Breadcrumb $breadcrumb The breadcrumb to record
+     */
+    public function addBreadcrumb(Breadcrumb $breadcrumb): void
+    {
+        $client = $this->getClient();
+
+        if (null !== $client) {
+            $client->leaveBreadcrumb($breadcrumb);
```
Jean85: This should be done through the scope directly, right?

src/State/Layer.php
```diff
@@ -0,0 +1,95 @@
...
+     */
+    private $scope;
+
+    /**
+     * Constructor.
+     *
+     * @param ClientInterface|null $client The client held by this layer
+     * @param Scope                $scope  The scope held by this layer
+     */
+    public function __construct(?ClientInterface $client, Scope $scope)
```
Jean85: What's the reason behind having a layer without a client? To disable sending events?

src/State/Scope.php
```diff
@@ -0,0 +1,310 @@
...
+    {
+        return $this->tags->toArray();
```
Jean85: Why all contexts are returned as array? To avoid editing from a distance?

src/State/Scope.php
```diff
@@ -0,0 +1,310 @@
...
+     *
+     * @param Event $event          The event object
+     * @param int   $maxBreadcrumbs The maximum number of breadcrumbs to add to
+     *                              the event
+     *
+     * @return Event|null
+     */
+    public function applyToEvent(Event $event, int $maxBreadcrumbs = 100): ?Event
+    {
+        if (!empty($this->fingerprint) && empty($event->getFingerprint())) {
```
Jean85: To make those conditions more readable, maybe we can skip checking if the `$event` is empty? At that point it doesn't matter any more...

src/State/Scope.php
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        $event->getTagsContext()->merge($this->tags->toArray());
+        $event->getExtraContext()->merge($this->extra->toArray());
+        $event->getUserContext()->merge($this->user->toArray());
+
+        foreach ($this->eventProcessors as $processor) {
+            $event = $processor($event);
+
+            if (null !== $event && !($event instanceof Event)) {
```
Jean85: If you put this check below the other, you can skip the first check.

tests/Context/AbstractContextTest.php
```diff
@@ -0,0 +1,109 @@
...
+
+use PHPUnit\Framework\TestCase;
+use Sentry\Context\Context;
+
+abstract class AbstractContextTest extends TestCase
+{
+    /**
+     * @dataProvider valuesDataProvider
+     */
+    public function testConstructor($initialData, $expectedData, $expectedExceptionClass, $expectedExceptionMessage)
```
Jean85: Can we put some types to those arguments?

ste93cry:
> When talking about `capture*` on the `Scope` you actually meant `Hub`, right?
> So I know this is one of the bigger changes of the API. Before our `capture*` methods took a `payload` which can be anything the event takes?
> While this maybe worked in PHP in other SDKs you were not able to pass anything here the event takes, only for example `tags` `extra` and so on, but not an `exception` with for example.
> 
> That's why we went with the approach to remove this "magic" payload parameter and introduced `Hub::withScope` and `EventProcessors` as a workaround for this.

Yes, I indeed meant `Hub`. It's fine if you can't use anymore the payload, but the methods are inconsistent anyway because for example the `captureMessage` accepts a `level` argument but the `captureException` not. I think we should normalize them (and in case normalize the API across all SDKs). For the same reason you should use `Hub::withScope` or event processors to customize the data of the event you should in case use them to customize the level. Does it makes sense?

> If someone wants to change `server` or `os` context of an event, they should write their own integration or add an `EventProcessor`. We do not want to expose all possible getter/setter that is on the event. Instead, if someone wants to fundamentally modify the event they should, as said before an `EventProcessor`, an `Integration` or `beforeSend` callback.

Even though I understand that exposing all the getters/setter of the event on the `Scope` object is out of discussion, on which base we decide that `setTag` and `getTags` are fine to be on the scope but not the getters/setters for the other contexts (this is just an example, I could have written the same thing for any property of the event)? Probably the unified API should also unify the way you interact with the event and should be consistent without putting half of the setters/getters on the scope and half on the event

tests/State/ScopeTest.php
```diff
@@ -0,0 +1,200 @@
...
+    public function testAddEventProcessor(): void
+    {
+        $callback1Called = false;
+        $callback2Called = false;
+        $callback3Called = false;
+
+        $event = new Event(new Configuration());
+        $scope = new Scope();
+
+        $scope->addEventProcessor(function (Event $eventArg) use (&$callback1Called): ?Event {
```
Jean85: Should we test in some way the order of invocation of the callbacks?

src/Severity.php
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * Returns whether two object instances of this class are equal.
+     *
+     * @param self $other The object to compare
+     *
+     * @return bool
+     */
+    public function isEqualTo(self $other): bool
```
Jean85: This shouldn't be needed, isn't `==` enough?

tests/State/LayerTest.php
Jean85: Agree on this, but since that's enforced by the code style fixer, it should be done in a separate PR, since it will generate a huge diff and build failures on any new file from other PRs...

ste93cry: At the moment I would like to add it to all files because this is what we have done until now. Before finalizing `2.0` we could simply change all the references of `Raven` to `Sentry` or simply drop the license headers all together as part of a big CS cleanup. WDYT?

Said that, I think that the license header is fine as is not too big and helps understanding that the source code is provided under a license. However I'm not a layer, so I don't know if it's realyl necessary to have it or the LICENSE file in the root of the project is enough. Looking at the other projects there doesn't seems to be a standard, everyone does what he prefers

src/State/Scope.php
ste93cry: > This applies to all `getters` in the `Scope`. I on purpose did not expose the `getters` since we do not want the user to hold on to references of the `context` and modifies them outside of `configureScope` or `withScope`.

It seems legit. The only downside is that the context objects provide a nice API to manipulate the data while here you can't for example remove a value or merge existing data recursively

> Also there should actually never be a reason for the user the call a `getter` of the `Scope`. Again for this reason we have `beforeSend` and `EventProcessors` / `Integrations`.

I exposed them because I saw the unified API on the Javascript package does the same thing. If the purpose of this new API is to be the same across all SDKs we should have the same methods

src/BreadcrumbErrorHandler.php
ste93cry: I completely agree but the client at the moment has a map (a part that we didn't refactor yet) that can be configured to customize which event severity gets assigned to an event according to the error level. Making this change here means refactoring that whole part and maybe a separate PR would better fit this purpose. It's something I want to do anyway soon or later

HazAT:
@ste93cry The basic idea is that the use case for `captureMessage` is for "logging".
For logging, it makes sense to define the `level` easily since it would be really annoying to do 

```php
Sentry.withScope(function($scope) {
  $scope->setLevel(Severity::debug());
  Sentry.captureMessage("test");
});
```
vs.
```php
  Sentry.captureMessage("test", Severity::debug());
```

In general, we want that the simplest use case of the SDK doesn't need any deep knowledge about Sentry at all. That's why we removed all the `payload` and second "magic" args from all other `capture*` methods.

There is `setLevel` on the Scope if you ever want to change the Level when capturing an Exception with `captureException`.

---

The idea of the unified API is to abstract away the complexity of the Client and the Event behind the Hub / Scope. 
We still want users to be able to just create a new Client, instead of calling `init`, but then they are responsible for managing the State by themselves. If someone wants to fundamentally change the payload they have to go through an `EventProcessor`.

tests/State/LayerTest.php
HazAT: I am +1 one for just removing them, doesn't really add any value in my eyes. We have a License and Contributors file in the Repo. I will add it to the todo list before we release 2.0.

src/State/Scope.php
HazAT: If someone wants to work with the context they can still modify it in `beforeSend` of an `EventProcessor` directly on the event.

That they are public in JS is also just because of the tests, we will remove the `getters` (that was a mistake on our side).

src/Breadcrumbs/Breadcrumb.php
ste93cry: When I created it its purpose was to let users have a way for example to use it as callback of functions that accept a `callable` (e.g. `array_map`). Maybe this use case is not really useful, so I will remove it

src/State/Layer.php
ste93cry: I'm not sure, the unified API of the [Javascript API ](https://github.com/getsentry/sentry-javascript/blob/d67ad66183649c8c09182ebaae773ad40aed532f/packages/hub/src/interfaces.ts#L7)is done in this way (the scope is `nullable` too) even though the [constructor](https://github.com/getsentry/sentry-javascript/blob/d67ad66183649c8c09182ebaae773ad40aed532f/packages/hub/src/hub.ts#L53) of the `Hub` does not allow passing `null` as scope, so it seems to be an inconsistency of the implementation. ping @HazAT as this code was originally written by himself

src/State/Scope.php
ste93cry: We're discussing about this here: https://github.com/getsentry/sentry-php/pull/679#discussion_r227275312

src/State/Scope.php
HazAT: 😬 https://github.com/getsentry/sentry-javascript/pull/1671/files#diff-9149adc4b7fd7b844f7874022e9d197d

ste93cry: >That they are public in JS is also just because of the tests, we will remove the getters (that was a mistake on our side).

They are useful for testing too in this package, otherwise we have to rely on reflection (the `assertAttribute* methods of PHPUnit are going [to be removed](https://github.com/sebastianbergmann/phpunit/issues/3339) so I would not rely on them anymore)

src/State/Scope.php
ste93cry: The code resembles what is done in the Javascript SDK. I think that the check is there because that way the scope won't override any data already set in the event (this is valid for fingerprint and breadcrumbs as written in the docblock)

src/Severity.php
ste93cry: Yes it is, but I think that having an `isEqualTo` method provides a nicer API than comparing two objects to check their internal value is equal

```php
$enum1 = Severity::error();
$enum2 = Severity::fatal();

var_dump($enum1 == $enum2);
var_dump($enum1->isEqualTo($enum2));
```

If someone for some reason uses `===` will always get false from the assertion while using `isEqualTo` is more expressive

src/State/Scope.php
Jean85: Yes, I got that; I was suggesting that `empty($event->getFingerprint())` is enough of a check to avoid overriding existing data. The other check is to avoid setting null, but I think is superfluous and hinders readability.

src/State/Scope.php
HazAT: Then we at least should document them with `@internal`.

src/Breadcrumbs/Breadcrumb.php
ste93cry: It was named this way because the SDK docs refer to the `level` attribute. Another inconsistency that should be fixed @HazAT?

src/State/Scope.php
ste93cry: Got it now, I will change it as it is superfluous to not call the `Event::setFingerprint` method even though `$this->fingerprint` is empty. I will change it as you recommended!

src/State/Hub.php
ste93cry: Theoretically yes, but looking at the source code of the Javascript SDK they [call the client passing to it the scope](https://github.com/getsentry/sentry-javascript/blob/65965ff30b4c978c4f200ef8bb5c88d220059324/packages/hub/src/hub.ts#L65), do some operations I don't really understand and then add the breadcrumb to the given scope. I've changed this to directly push the breadcrumb to the scope but I would like an opinion on the matter from @HazAT

tests/State/ScopeTest.php
ste93cry: Seems fair, any idea how to do it?

ste93cry:
Imho soon or later someone will start asking why on the scope there are some methods and not others... Also the API doesn't provide any way except having to use an event processor to remove a tag or any data from the contexts exposed by the scope. It seems to me that the API aims to be simple and this is good, but half of the things can be set in some way and the other half can't. As a user my first question would be on what basis those methods were chosen and not others.

tests/State/ScopeTest.php
Jean85: You probably have to pass all the booleans to all the callbacks and check true/false accordingly.

HazAT:
@ste93cry So we decided on the public API what we think would be the most common use case for >90% of people using Sentry.
Most people install it, call `init` and that's it.
Maybe they set some Tags to filter or add extra data to have more context.

Just to elaborate on the advantages of small public API again:
We can easily change the internals of events and how we send them.

While we still wanted to maintain great flexibility if your usage of our SDK isn't the norm, it's fine, we want users to have ways around our limited public API but we will not deliver guarantees that the internal will work like this in a future version of our SDK.

The Public API we deliver is a contract we tend to keep, but if we someday decide for example to completely drop the OS Context (not that we actually want to do this) from the event and people where using `setOSContext` across all SDKs, we have to update all SDKs to be on par with our decision.

Ultimately it's about the smaller the public documented API the less maintenance we have to do.

src/Breadcrumbs/Breadcrumb.php
HazAT: Actually, the levels of a breadcrumb are not 1:1 to event Severity.
https://docs.sentry.io/clientdev/interfaces/breadcrumbs/
Maybe we shouldn't use Severity for breadcrumbs.

ste93cry: Oh, didn't notice that. I will then revert to the change to the constants

src/State/Hub.php
HazAT: So we pass it through the Client because the options do live on the Client.
One of the options we support is `beforeBreadcrumb`. `beforeBreadcrumb` receives the breadcrumb before it will be added, user can decide if they want to "add" it or "drop" it.
The Client also receives the current scope on `leaveBreadcrumb` (which should also be renamed to `addBreadcrumb`) and adds it to the `Scope` if `beforeBreadcrumb` says so.

Jean85:
@HazAT we should probably state this (maybe concisely) on the README to clarify what's covered and considered as public API, in relation to BC and semantic versioning then.

src/State/Layer.php
HazAT: Main Reason for allowing `null` Client is for to make testing easier. Not really necessary in PHP if we can mock it easily.

`Scope` is just optional which means if you do not pass an explicit `Scope` you created then the ctor will create a `new Scope` for you.

ste93cry: As we can easily mock the client thanks for its interface and a layer does not makes sense to exist without it, I will change the constructor definition to `__construct(ClientInterface $client, ?Scope $scope = null)`

src/Client.php
```diff
@@ -166,25 +161,20 @@ public function __construct(Configuration $config, TransportInterface $transport
     /**
      * {@inheritdoc}
      */
-    public function getBreadcrumbsRecorder()
+    public function addBreadcrumb(Breadcrumb $breadcrumb, ?Scope $scope = null)
```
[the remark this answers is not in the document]
ste93cry: Actually it's needed because of the need to check the configuration options. The problem is that I forgot to call it from the `Hub` instead of directly calling the `Scope`. Said that, given that the `Hub` knows about the client and the scope and that that method cannot work without the scope instead I think we can safely remove it. @HazAT WDYT?

src/State/Hub.php
```diff
@@ -0,0 +1,225 @@
...
+     * Calls the given callback passing to it the current scope so that any
+     * operation can be run within its context.
+     *
+     * @param callable $callback The callback to be executed
+     */
+    public function configureScope(callable $callback): void
+    {
+        $layer = $this->getStackTop();
+
+        if (null !== $layer->getClient()) {
```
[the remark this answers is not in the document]
HazAT: Because if no client is set on the hub, it should be a noop.

src/Client.php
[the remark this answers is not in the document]
Jean85: The way I see it, the client should become just a sender of events through the Transport... Maybe we should move the config elsewhere, or use it in the hub too?

Or maybe we can split it in two, with client/transport related stuff in one, and data/event related stuff in the other?

src/Severity.php
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * Creates a new instance of this class with the given severity value.
+     *
+     * @param string $severity The severity
+     *
+     * @return self
+     */
+    public static function fromString(string $severity): self
```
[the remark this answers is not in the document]
Jean85: Well but it's the same of the constructor, it doesn't give you any addition nor translation to it..

src/State/Hub.php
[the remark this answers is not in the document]
ste93cry: I'm not sure, I just ported the code from the Javascript SDK which does the following check:

```js
if (top.scope && top.client) {
   ...
}
```

Actually the first check is not needed because there will always be a scope (or you pass it explicitly when creating a nwe `Hub` instance or it will be created for you internally in the constructor so the other check left was the one on the client. Maybe @HazAT can explain us why it's needed, I see no reason regarding the code as we never use the client so I agree that we shouldn't care about it

src/Severity.php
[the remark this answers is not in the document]
ste93cry: I think that `Severity::fromString(...)` is more readable than having to do in-place (because most of the cases you will need to create this object imho is inline code) `new Severity(...)`. I will remove this method anyway

src/State/Hub.php
```diff
@@ -179,7 +175,7 @@ public function bindClient(ClientInterface $client): void
      * @param string   $message The message
      * @param Severity $level   The severity level of the message
      *
-     * @return null|string
+     * @return string
```
Jean85: This diff seems wrong to me, it contradicts the signature.

src/Client.php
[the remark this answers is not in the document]
HazAT: So we leave it as it is right now, you are right, the client is basically nothing more than an API over a transport. The Client should be responsible for building together an event and delegate the sending to a transport.

PR #681, opened 2018-10-29
[2.0] Restrict conflict with php-http/client-common

Jean85:
In #658 I did found an issue with php-http/client-common 1.8.0, precisely https://github.com/php-http/client-common/issues/109. The problematic code is now reverted, so we can reduce the conflict constraint to the only bugged version.


Jean85:
Damn, build is now failing due to https://github.com/php-http/message/pull/100

PR #682, opened 2018-10-30
[2.0] Public API changes

HazAT:
This is only a part of changes if I do everything in one go it's probably too much to review.


src/Client.php
```diff
@@ -75,7 +77,7 @@ class Client implements ClientInterface
     private $representationSerializer;
 
     /**
-     * @var Configuration The client configuration
+     * @var Options The client configuration
      */
     private $config;
```
Jean85: Should we rename the variable too?

src/ClientBuilder.php
```diff
@@ -84,7 +84,7 @@
 final class ClientBuilder implements ClientBuilderInterface
 {
     /**
-     * @var Configuration The client configuration
+     * @var Options The client configuration
      */
     private $configuration;
```
Jean85: Here too... should we change the var name?

src/Middleware/ModulesMiddleware.php
```diff
@@ -18,16 +18,16 @@
 final class ModulesMiddleware
 {
     /**
-     * @var Configuration The Raven client configuration
+     * @var Options The Raven client configuration
      */
     private $config;
```
Jean85: Again...

tests/ClientTest.php
```diff
@@ -554,36 +419,23 @@ public function testSetReprSerializer()
...
-    {
-        $handler = function () {
-            return false;
-        };
-
-        set_error_handler($handler);
-        @trigger_error('');
-        restore_error_handler();
-    }
+//    TODO
```
Jean85: WIP

src/ClientInterface.php
```diff
@@ -26,192 +23,59 @@
...
-    /**
-     * Translate a PHP Error constant into a Sentry log level group.
-     *
-     * @param int $severity PHP E_$x error constant
      *
-     * @return string Sentry log level group
+     * @return null|string
      */
-    public function translateSeverity(int $severity);
+    public function send(Event $event): ?string;
```
Jean85: Does it make sense to have this public?

src/Event.php
```diff
@@ -132,33 +133,33 @@
...
+     * Event constructor.
      */
-    public function __construct(Configuration $config)
+    public function __construct()
     {
-        $this->id = Uuid::uuid4();
+        try {
+            $this->id = Uuid::uuid4();
+        } catch (\Exception $e) {
+            // TODO: This happens when Uuid::uuid4 has issues, fallback to something else?
```
Jean85: That is a very reliable library, I think we can trust it.

src/Integration/ExceptionIntegration.php
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Applies exception to the passed event
+     *
+     * @param ExceptionIntegration $self
+     * @param Event $event
+     * @param null|$exception
+     * @return null|Event
+     */
+    public static function applyToEvent(ExceptionIntegration $self, Event $event, $exception = null): ?Event
```
Jean85: exception should be a `\Throwable`?

src/Integration/ExceptionIntegration.php
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    /**
+     * Translate a PHP Error constant into a Sentry log level group.
+     *
+     * @param int $severity PHP E_$x error constant
+     *
+     * @return Severity
+     */
+    private function translateSeverity(int $severity): Severity
```
Jean85: I would move this inside a `Severity` named constructor.

src/Integration/Handler.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    private static $integrations = [];
+
+    public static function setupIntegrations(array $integrations): array
+    {
+        $integrationIndex = [];
+        foreach ($integrations as $integration) {
+            $class = \get_class($integration);
+            if (!\array_key_exists($class, self::$integrations)) {
+                /* @var Integration $integration */
```
Jean85: We should check if it's an instance of that interface explicitly

tests/AbstractErrorHandlerTest.php
```diff
@@ -90,7 +86,7 @@ public function handleErrorShouldNotCaptureDataProvider()
     {
         return [
             [false, E_ERROR, E_ERROR],
-            [false, E_ALL, E_ERROR],
+//            [false, E_ALL, E_ERROR], This tests breaks
```
HazAT: This test fails now, apparently, it didn't before but I debugged it and capture was also called before. So I am not sure what's going on here.

Jean85: If I'm not mistaken, the error was about the call not matching... Maybe the argument is changed someway?

src/Integration/ExceptionIntegration.php
```diff
@@ -0,0 +1,104 @@
...
+            $currentException = $exception;
+
+            do {
+                if ($self->options->isExcludedException($currentException)) {
+                    continue;
+                }
+
+                $data = [
+                    'type' => \get_class($currentException),
+                    'value' => (new Serializer($self->options->getMbDetectOrder()))->serialize($currentException->getMessage()),
```
Jean85: We should avoid instantiating the serializer at every loop

src/Integration/ExceptionIntegration.php
```diff
@@ -0,0 +1,104 @@
...
+                ];
+
+                if ($self->options->getAutoLogStacks()) {
+                    $data['stacktrace'] = Stacktrace::createFromBacktrace($self->options, $currentException->getTrace(), $currentException->getFile(), $currentException->getLine());
+                }
+
+                $exceptions[] = $data;
+            } while ($currentException = $currentException->getPrevious());
+
+            $exceptions = [
```
Jean85: Can we avoid this reassignment of value to the same var? Maybe we can inline this with the setter below?

src/Integration/Handler.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    private static $integrations = [];
+
+    public static function setupIntegrations(array $integrations): array
+    {
+        $integrationIndex = [];
+        foreach ($integrations as $integration) {
+            $class = \get_class($integration);
+            if (!$integration instanceof IntegrationInterface) {
+                throw new \InvalidArgumentException(sprintf('IntegrationInterface must implement %s interface', IntegrationInterface::class));
```
Jean85:
```suggestion
                throw new \InvalidArgumentException(sprintf('Expecting integration implementing %s interface, got %s', IntegrationInterface::class, $class));
```

src/Integration/Handler.php
```diff
@@ -0,0 +1,27 @@
...
+            }
+            if (!\array_key_exists($class, self::$integrations)) {
```
Jean85: Maybe this is more flexible?
```suggestion
            if (self::$integrations[$class] ?? false) {
```

src/Integration/ModulesIntegration.php
```diff
@@ -0,0 +1,85 @@
...
+        });
+    }
+
+    /**
+     * @param ModulesIntegration $self
+     * @param Event              $event
+     *
+     * @return null|Event
+     */
+    public static function applyToEvent(self $self, Event $event): ?Event
```
Jean85: Do we really need this (and similar) methods to be static?

src/Integration/RequestIntegration.php
```diff
@@ -0,0 +1,87 @@
...
+     * @param Event                       $event
+     * @param null|ServerRequestInterface $request
+     *
+     * @return Event
+     */
+    public static function applyToEvent(Event $event, ?ServerRequestInterface $request = null): Event
+    {
+        if (null == $request) {
+            /** @var ServerRequestInterface $request */
+            $request = isset($_SERVER['REQUEST_METHOD']) && \PHP_SAPI !== 'cli' ? ServerRequestFactory::fromGlobals() : null;
```
Jean85: Suggestion for the future: can we use a generic PSR-7 implementation with HTTPlug here, and ditch Diactoros as direct dependency? @ste93cry ?

src/Integration/RequestIntegration.php
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        $requestData = [
+            'url' => (string) $request->getUri(),
+            'method' => $request->getMethod(),
+            'headers' => $request->getHeaders(),
+            'cookies' => $request->getCookieParams(),
+        ];
+
+        if ('' !== $request->getUri()->getQuery()) {
```
Jean85: What if it's `null`? We should use loose comparison here
```suggestion
        if ($request->getUri()->getQuery()) {
```

src/Options.php
```diff
@@ -642,6 +632,42 @@ public function setBeforeBreadcrumbCallback(callable $callback): void
...
+    {
+        $options = array_merge($this->options, ['transport' => $transport]);
+
+        $this->options = $this->resolver->resolve($options);
+    }
+
+    /**
+     * @return null|TransportInterface
+     */
+    public function getTransport(): ?TransportInterface
```
Jean85: As said by @ste93cry, this is a dependency, not an option... It doesn't belong here

src/Options.php
```diff
@@ -653,6 +679,8 @@ public function setBeforeBreadcrumbCallback(callable $callback): void
     private function configureOptions(OptionsResolver $resolver)
     {
         $resolver->setDefaults([
+            'transport' => null,
+            'integrations' => null,
```
Jean85: Maybe empy array?

Jean85: And does this add or substitute all the integrations?

src/Sdk.php
```diff
@@ -52,6 +52,24 @@ function captureEvent(array $payload): ?string
...
+ */
+function captureLastError(): ?string
+{
+    $error = error_get_last();
+
+    if (null === $error || !isset($error['message'][0])) {
+        return null;
+    }
+
+    $exception = new \ErrorException(@$error['message'], 0, @$error['type'], @$error['file'], @$error['line']);
```
Jean85: Maybe we should pass default values instead of silencing errors?
```suggestion
    $exception = new \ErrorException(
        $error['message'] ?? null,
        0, 
        $error['type'] ?? null, // severity -- maybe E_ERROR?
        $error['file'] ?? null, // this by default risks being the file of this class, which would be wrong!
        $error['line'] ?? null // this too
    );
```

src/State/Scope.php
```diff
@@ -298,8 +309,8 @@ public function applyToEvent(Event $event, int $maxBreadcrumbs = 100): ?Event
         $event->getExtraContext()->merge($this->extra->toArray());
         $event->getUserContext()->merge($this->user->toArray());
 
-        foreach ($this->eventProcessors as $processor) {
-            $event = $processor($event);
+        foreach (array_merge(self::$globalEventProcessors, $this->eventProcessors) as $processor) {
```
Jean85: Why do we have two of them?

tests/AbstractErrorHandlerTest.php
```diff
@@ -12,27 +12,23 @@
...
-    protected $client;
+    protected $callbackMock;
 
     protected function setUp()
     {
-        $this->client = $this->getMockBuilder(Client::class)
-            ->disableOriginalConstructor()
-            ->setMethodsExcept(['translateSeverity'])
-            ->getMock();
+        $this->callbackMock = $this->createPartialMock(\stdClass::class, ['__invoke']);
```
Jean85: As said elsewhere, you should use a data provider here to test every possible kind of callback

tests/AbstractSerializerTest.php
```diff
@@ -447,9 +447,6 @@ public function serializableCallableProvider()
                 ], [
                     'callable' => [$this, 'serializableCallableProvider'],
                     'expected' => 'Callable Sentry\Tests\AbstractSerializerTest::serializableCallableProvider []',
-                ], [
-                    'callable' => [Client::class, 'getConfig'],
```
Jean85: You should replace this kind of callback with an other (non deprecated, sure) one

tests/ClientTest.php
```diff
@@ -38,65 +31,53 @@ public function testConstructorInitializesTransactionStack()
...
 
-        /** @var MiddlewareStack|\PHPUnit_Framework_MockObject_MockObject $middlewareStack */
-        $middlewareStack = $this->createMock(MiddlewareStack::class);
-        $middlewareStack->expects($this->once())
-            ->method('removeMiddleware')
-            ->with($middleware);
+        /** @var Client|\PHPUnit_Framework_MockObject_MockObject $client */
+        $client = $this->getMockBuilder(Client::class)
+            ->setConstructorArgs([new Options(), $transport])
+            ->setMethodsExcept(['captureMessage', 'prepareEvent', 'captureEvent', 'getOptions'])
```
Jean85: A partial mock is a bad way to proceed. You should set expectation on the transport dependency instead.

tests/ClientTest.php
```diff
@@ -333,72 +185,22 @@ private function assertMixedValueAndArray($expected_value, $actual_value)
...
 
         $client = ClientBuilder::create([
             'dsn' => 'http://public:secret@example.com/1',
-            'should_capture' => function () use (&$shouldCaptureCalled) {
-                $shouldCaptureCalled = true;
+            'before_send' => function () use (&$beforeSendCalled) {
+                $beforeSendCalled = true;
 
-                return false;
+                return null;
```
Jean85: Since you're not returning the event, maybe you should assert that nothing is being sent too?

tests/ClientTest.php
```diff
@@ -534,34 +321,31 @@ public function testAddBreadcrumbStoresBreadcrumbReturnedByBeforeBreadcrumbCallb
...
-        $client->setSerializer($serializer);
+        /** @var TransportInterface|\PHPUnit_Framework_MockObject_MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
 
-        $this->assertSame($serializer, $client->getSerializer());
-    }
+        /** @var Client|\PHPUnit_Framework_MockObject_MockObject $client */
+        $client = $this->getMockBuilder(Client::class)
+            ->setConstructorArgs([new Options(), $transport, [new ExceptionIntegration(new Options())]])
+            ->setMethodsExcept(['captureException', 'prepareEvent', 'captureEvent', 'getIntegration', 'getOptions'])
```
Jean85: Here too, we should avoid partial mocks.

tests/phpt/fatal_error.phpt
```diff
@@ -17,18 +17,16 @@ while (!file_exists($vendor . '/vendor')) {
...
 
-ErrorHandler::register($client);
-
-register_shutdown_function('register_shutdown_function', function () use ($client) {
+register_shutdown_function('register_shutdown_function', function () {
+    $client = Hub::getCurrent()->getClient();
     /** @var \Sentry\Transport\HttpTransport $transport */
     $transport = Assert::getObjectAttribute($client, 'transport');
 
-    Assert::assertNotNull($client->getLastEvent());
```
Jean85: Shouldn't we replace this with a different assertion?

composer.json
```diff
@@ -40,7 +40,7 @@
         "monolog/monolog": "Automatically capture Monolog events as breadcrumbs"
     },
     "conflict": {
-        "php-http/client-common": "^1.8.0",
+        "php-http/client-common": "1.8.0",
```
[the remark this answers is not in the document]
HazAT: Forgot to rebase the one commit 🙈

src/Context/UserContext.php
```diff
@@ -78,4 +78,24 @@ public function setEmail(?string $email): void
...
     }
+
+    /**
+     * Gets the ip address of the user.
+     *
+     * @return null|string
+     */
+    public function getIpAddress(): ?string
+    {
+        return $this->data['email'] ?? null;
```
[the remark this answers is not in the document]
HazAT: 🙈 Also added test.

src/Integration/ExceptionIntegration.php
HazAT: I refactored the code, now if we use `setExceptions` we do not by default add `values`. Instead we just add it in `toArray` because that's where we need it.

src/Integration/ModulesIntegration.php
HazAT: The reason for this to be static is if there are multiple clients, with multiple `ModulesIntegration` we have `self::$loadedModules` as a cache.

src/State/Scope.php
HazAT: Because there are "global" and scope related EventProcessors.
Integrations use the "global" event processors but as an example if an user wants to change an event in the context of a scope, the can 
```
withScope(function($scope) {
  $scope->addEventProcessors(function ($event) {
      // Changing $event
     return null; // or dropping it
  });
});

File: tests/Integration/ExceptionIntegrationTest.php [hunk 30]
```diff
@@ -124,85 +128,76 @@ public function invokeDataProvider()
...
+        $options = new Options();
+        $event = new Event();
 
-        $middleware = new ExceptionInterfaceMiddleware($client);
+        $integration = new ExceptionIntegration($options);
 
         $malformedString = "\xC2\xA2\xC2"; // ill-formed 2-byte character U+00A2 (CENT SIGN)
-        $returnedEvent = $this->assertMiddlewareInvokesNext($middleware, $event, null, new \Exception($malformedString));
+        $returnedEvent = ExceptionIntegration::applyToEvent($integration, $event, new \Exception($malformedString));
+        $this->assertNotNull($returnedEvent);
```
ste93cry: Please put a blank line above this one to improve readability

tests/Integration/ModulesIntegrationTest.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    public function testInvoke()
+    {
+        $options = new Options(['project_root' => __DIR__ . '/../Fixtures']);
+        $event = new Event();
+
+        $integration = new ModulesIntegration($options);
+
+        $returnedEvent = ModulesIntegration::applyToEvent($integration, $event);
+        $this->assertNotNull($returnedEvent);
```
ste93cry: Please put a blank line above this one to improve readability

tests/Integration/ModulesIntegrationTest.php
```diff
@@ -0,0 +1,33 @@
...
+
```
ste93cry: Please remove this blank line

tests/Integration/ModulesIntegrationTest.php
```diff
@@ -0,0 +1,33 @@
...
+class ModulesIntegrationTest extends TestCase
+{
+    public function testInvoke()
+    {
+        $options = new Options(['project_root' => __DIR__ . '/../Fixtures']);
+        $event = new Event();
+
+        $integration = new ModulesIntegration($options);
+
+        $returnedEvent = ModulesIntegration::applyToEvent($integration, $event);
```
ste93cry: You can rename this variable `$event` and instantiate the event object in-place

tests/Transport/NullTransportTest.php
```diff
@@ -21,8 +20,8 @@ class NullTransportTest extends TestCase
     public function testSend()
     {
         $transport = new NullTransport();
-        $event = new Event(new Configuration());
+        $event = new Event();
 
-        $this->assertTrue($transport->send($event));
+        $this->assertNotNull($transport->send($event));
```
ste93cry: You should test that the returned event ID is the same as the one of the passed event

src/Event.php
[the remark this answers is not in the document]
HazAT: The problem is that if we do that, we add `@throws` everywhere imho it's really annoying to have a `@throws` in the ctor of `Event`.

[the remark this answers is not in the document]
ste93cry: I understand that it may be annoying to have IDEs warn that an exception could be thrown by the constructor but it's foundamentally broken to catch a failed attempt to create an UUID as an event without it is not valid. Imho btw the ID should be generated server-side (what happens if someone generates an already existing UUID?) and then be returned by the request so that it can be set back to the event object in the client. Also if in the future you want to change the format of the field every SDK will have to handle this. I understand that anyway this is out of scope for this PR

tests/ClientTest.php
```diff
@@ -534,34 +319,27 @@ public function testAddBreadcrumbStoresBreadcrumbReturnedByBeforeBreadcrumbCallb
         $this->assertSame([$breadcrumb2], $scope->getBreadcrumbs());
     }
 
-    public function testSetSerializer()
+    public function testHandlingExceptionThrowingAnException()
```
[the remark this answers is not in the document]
HazAT: I think it's a good test since it tests that just using a client with this integration also works.

src/State/Scope.php
```diff
@@ -270,24 +286,19 @@ public function clear(): self
...
      */
-    public function applyToEvent(Event $event, int $maxBreadcrumbs = 100): ?Event
+    public function applyToEvent(Event $event, array $payload): ?Event
     {
         if (empty($event->getFingerprint())) {
             $event->setFingerprint($this->fingerprint);
         }
 
         if (empty($event->getBreadcrumbs())) {
-            $breadcrumbs = \array_slice($this->breadcrumbs, -$maxBreadcrumbs);
```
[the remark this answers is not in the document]
HazAT: Yes, it also doesn't really make sense there.

src/AbstractErrorHandler.php
```diff
@@ -80,16 +80,18 @@
     /**
      * Constructor.
      *
-     * @param ClientInterface $client             The Raven client
-     * @param int             $reservedMemorySize The amount of memory to reserve for the fatal error handler
+     * @param callable $callback           The callback that will be invoked in case of an exception
+     * @param int      $reservedMemorySize The amount of memory to reserve for the fatal error handler
+     *
+     * @throws \Exception|\Throwable|\ErrorException
```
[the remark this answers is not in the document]
HazAT: screenshot 2018-11-13 11 19 42
It can

src/Client.php
```diff
@@ -51,391 +52,164 @@ class Client implements ClientInterface
...
+     *
+     * @return null|Event
      */
-    public function getLastEventId()
+    protected function prepareEvent(array $payload, ?Scope $scope = null): ?Event
     {
-        @trigger_error(sprintf('The %s() method is deprecated since version 2.0. Use getLastEvent() instead.', __METHOD__), E_USER_DEPRECATED);
-
-        if (null === $this->lastEvent) {
+        if (mt_rand(1, 100) / 100.0 > $this->getOptions()->getSampleRate()) {
```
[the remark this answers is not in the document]
ste93cry: >even though just discarding random events seems really wrong experience showed that it still works surprisingly well

I can't see any way this can work surprisingly well :open_mouth: I mean, you're discarding events in a totally random order and you could miss some important events. Imho you should not provide this feature because it encourages a bad habit

src/Client.php
```diff
@@ -178,183 +165,157 @@ public function addBreadcrumb(Breadcrumb $breadcrumb, ?Scope $scope = null)
...
-        $payload['exception'] = $exception;
-
-        return $this->capture($payload);
+        $this->middlewareStack->removeMiddleware($middleware);
     }
 
     /**
      * {@inheritdoc}
      */
-    public function captureLastError(array $payload = [])
```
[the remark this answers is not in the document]
HazAT: I removed the `ExceptionIntegration` converted it to a core part of the SDK and moved this function back into the client.

src/Event.php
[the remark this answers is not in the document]
HazAT: I removed the `try/catch`

src/Stacktrace.php
```diff
@@ -56,42 +56,47 @@ class Stacktrace implements \JsonSerializable
...
+     * @param Options $options The client options
      */
-    public function __construct(ClientInterface $client)
+    public function __construct(Options $options)
     {
-        $this->client = $client;
-        $this->serializer = $client->getSerializer();
-        $this->reprSerializer = $client->getRepresentationSerializer();
+        $this->options = $options;
+        $this->serializer = new Serializer($this->options->getMbDetectOrder());
```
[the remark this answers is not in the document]
HazAT: I started to refactor this now, `Serializer` now only lives on the client.

src/Options.php
```diff
@@ -698,11 +728,13 @@ private function configureOptions(OptionsResolver $resolver)
...
         $resolver->setAllowedTypes('dsn', ['null', 'boolean', 'string']);
         $resolver->setAllowedTypes('server_name', 'string');
-        $resolver->setAllowedTypes('should_capture', ['null', 'callable']);
+        $resolver->setAllowedTypes('before_send', ['callable']);
         $resolver->setAllowedTypes('tags', 'array');
         $resolver->setAllowedTypes('error_types', ['null', 'int']);
         $resolver->setAllowedTypes('max_breadcrumbs', 'int');
         $resolver->setAllowedTypes('before_breadcrumb', ['callable']);
+        $resolver->setAllowedTypes('transport', ['null', 'Sentry\Transport\TransportInterface']);
+        $resolver->setAllowedTypes('integrations', ['null', 'Sentry\Integration\IntegrationInterface']);
```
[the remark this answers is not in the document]
HazAT: Fixed!

src/AbstractErrorHandler.php
```diff
@@ -80,16 +80,18 @@
...
      *
-     * @param ClientInterface $client             The Raven client
-     * @param int             $reservedMemorySize The amount of memory to reserve for the fatal error handler
+     * @param callable $callback           The callback that will be invoked in case an error is caught
+     * @param int      $reservedMemorySize The amount of memory to reserve for the fatal error handler
+     *
+     * @throws \ReflectionException
      */
-    protected function __construct(ClientInterface $client, $reservedMemorySize = 10240)
+    protected function __construct(callable $callback, $reservedMemorySize = 10240)
```
Jean85:
```suggestion
    protected function __construct(callable $callback, int $reservedMemorySize = 10240)
```

src/AbstractErrorHandler.php
```diff
@@ -186,6 +190,8 @@ public function handleError($level, $message, $file, $line)
      *
      * @param array|null $error The error details as returned by error_get_last()
      *
+     * @throws \Throwable
```
Jean85: Since you're catching it, this is wrong, should be removed.

src/Client.php
```diff
@@ -32,139 +28,89 @@ class Client implements ClientInterface
...
 
     /**
      * {@inheritdoc}
      */
-    public function addBreadcrumb(Breadcrumb $breadcrumb, ?Scope $scope = null)
+    public function addBreadcrumb(Breadcrumb $breadcrumb, ?Scope $scope = null): void
     {
-        $beforeBreadcrumbCallback = $this->config->getBeforeBreadcrumbCallback();
-        $maxBreadcrumbs = $this->config->getMaxBreadcrumbs();
+        $beforeBreadcrumbCallback = $this->options->getBeforeBreadcrumbCallback();
```
Jean85: We're still not handling all the possible types of callback. I've opened #692 to track it after this PR.

src/Client.php
```diff
@@ -284,158 +169,128 @@ public function captureLastError(array $payload = [])
...
+     * @param null|Scope $scope   optional scope which enriches the Event
+     *
+     * @return null|Event returns ready to send Event, however depending on options it can be discarded
      */
-    public function capture(array $payload)
+    private function prepareEvent(array $payload, ?Scope $scope = null): ?Event
     {
-        $event = new Event($this->config);
+        $sampleRate = $this->getOptions()->getSampleRate();
+        if ($sampleRate < 1 && mt_rand(1, 100) / 100.0 > $sampleRate) {
```
Jean85: Isn't this wrong? With a sample rate of 0.3 I want a third of the events, I'm getting two thirds like this. Or am I getting this backwards?

src/Client.php
```diff
@@ -284,158 +169,128 @@ public function captureLastError(array $payload = [])
...
-        if (!$this->config->shouldCapture($event)) {
-            return;
+        if (null !== $message) {
+            $event->setMessage(substr($message, 0, self::MESSAGE_MAX_LENGTH_LIMIT), $messageParams);
         }
 
-        // should this event be sampled?
-        if (mt_rand(1, 100) / 100.0 > $this->config->getSampleRate()) {
-            return;
+        if (isset($payload['exception'])) {
```
Jean85: You're assuming this contains a `\Throwable`. Can we assume that?

src/Client.php
```diff
@@ -284,158 +169,128 @@ public function captureLastError(array $payload = [])
...
+                    $currentException->getTrace(),
+                    $currentException->getFile(),
+                    $currentException->getLine()
+                );
+            }
+
+            $exceptions[] = $data;
+        } while ($currentException = $currentException->getPrevious());
+
+        $event->setException($exceptions);
```
Jean85: Shouldn't it be `setExceptions`?

src/Event.php
```diff
@@ -448,13 +449,13 @@ public function getBreadcrumbs()
...
 
     /**
-     * Adds a new breadcrumb to the event.
+     * Set new breadcrumbs to the event.
      *
-     * @param Breadcrumb $breadcrumb The breadcrumb
+     * @param Breadcrumb[] $breadcrumbs The breadcrumb array
      */
-    public function setBreadcrumb(Breadcrumb $breadcrumb)
+    public function setBreadcrumb(array $breadcrumbs)
```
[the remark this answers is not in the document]
Jean85: It's done at the hub level, the event is prepared with all the stuff at once when sending it. See https://github.com/getsentry/sentry-php/pull/682/files#diff-9a950f4db299479d9933d02a78100059R45

src/Integration/RequestIntegration.php
Jean85: To be tracked in #693

tests/AbstractErrorHandlerTest.php
Jean85: This still needs to be resolved.

tests/AbstractSerializerTest.php
```diff
@@ -545,4 +545,9 @@ class SerializerTestObject
     private $foo = 'bar';
 
     public $key = 'value';
+
+    public static function testy(): array
+    {
+        return [];
```
Jean85: Throw an exception here: we should be sure of not invoking it while serializing it.

src/Client.php
HazAT: It's always weird to wrap my head around this, even though it seems to be so simple.
You are right, if it's 0.3, in two-thirds of the cases this comparison is `true`, which means we return `null` which means we discard it ^^
So only in one-third this evaluation is `false` and we continue to create/send the event.

src/Client.php
HazAT: I've added a check in `addThrowableToEvent`

`if (null === $exception || !$exception instanceof \Throwable) {`

tests/ClientTest.php
```diff
@@ -12,91 +12,60 @@
...
 
-        /** @var MiddlewareStack|\PHPUnit_Framework_MockObject_MockObject $middlewareStack */
-        $middlewareStack = $this->createMock(MiddlewareStack::class);
-        $middlewareStack->expects($this->once())
-            ->method('addMiddleware')
-            ->with($middleware, -10);
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(function (Event $event) {
+                $this->assertInstanceOf(Event::class, $event);
```
ste93cry: This assert is useless since the argument is typehinted 😄

tests/ClientTest.php
```diff
@@ -12,91 +12,60 @@
...
-            ->method('removeMiddleware')
-            ->with($middleware);
+        /** @var TransportInterface|\PHPUnit_Framework_MockObject_MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
 
-        $client = ClientBuilder::create()->getClient();
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(function (Event $event) {
+                $this->assertInstanceOf(Event::class, $event);
```
ste93cry: This assert is useless since the argument is typehinted 😄

tests/ClientTest.php
```diff
@@ -534,43 +357,268 @@ public function testAddBreadcrumbStoresBreadcrumbReturnedByBeforeBreadcrumbCallb
...
+        $assertHasStacktrace = $options->getAutoLogStacks();
 
-        $this->assertSame($serializer, $client->getRepresentationSerializer());
+        /** @var TransportInterface|\PHPUnit_Framework_MockObject_MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
+
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(function (Event $event) use ($expectedResult, $assertHasStacktrace) {
+                $this->assertNotNull($event);
```
ste93cry: This assert is unnecessary because the argument is typehinted to not accept `null` values

tests/ClientTest.php
```diff
@@ -534,43 +357,268 @@ public function testAddBreadcrumbStoresBreadcrumbReturnedByBeforeBreadcrumbCallb
...
+        $utf8String = 'äöü';
+        $latin1String = utf8_decode($utf8String);
+
+        /** @var TransportInterface|\PHPUnit_Framework_MockObject_MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
+
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(function (Event $event) use ($utf8String) {
+                $this->assertNotNull($event);
```
ste93cry: This assert is unnecessary because the argument is typehinted to not accept `null` values

tests/ClientTest.php
```diff
@@ -534,43 +357,268 @@ public function testAddBreadcrumbStoresBreadcrumbReturnedByBeforeBreadcrumbCallb
...
+    {
+        $malformedString = "\xC2\xA2\xC2"; // ill-formed 2-byte character U+00A2 (CENT SIGN)
+
+        /** @var TransportInterface|\PHPUnit_Framework_MockObject_MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
+
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(function (Event $event) {
+                $this->assertNotNull($event);
```
ste93cry: This assert is unnecessary because the argument is typehinted to not accept `null` values

tests/ClientTest.php
```diff
@@ -534,43 +357,268 @@ public function testAddBreadcrumbStoresBreadcrumbReturnedByBeforeBreadcrumbCallb
...
+            'mb_detect_order' => ['ISO-8859-1', 'ASCII', 'UTF-8'],
+        ]);
+
+        /** @var TransportInterface|\PHPUnit_Framework_MockObject_MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
+
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(function (Event $event) {
+                $this->assertNotNull($event);
```
ste93cry: This assert is unnecessary because the argument is typehinted to not accept `null` values

tests/ClientTest.php
```diff
@@ -534,43 +357,268 @@ public function testAddBreadcrumbStoresBreadcrumbReturnedByBeforeBreadcrumbCallb
...
+    {
+        $options = new Options(['auto_log_stacks' => false]);
+
+        /** @var TransportInterface|\PHPUnit_Framework_MockObject_MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
+
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(function (Event $event) {
+                $this->assertNotNull($event);
```
ste93cry: This assert is unnecessary because the argument is typehinted to not accept `null` values

tests/ClientTest.php
```diff
@@ -534,43 +357,268 @@ public function testAddBreadcrumbStoresBreadcrumbReturnedByBeforeBreadcrumbCallb
...
+    public function testConvertExceptionWithAutoLogStacksDisabled()
+    {
+        $options = new Options(['auto_log_stacks' => false]);
+
+        /** @var TransportInterface|\PHPUnit_Framework_MockObject_MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
+
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(function (Event $event) {
```
ste93cry: Please add the typehint for the return value (`bool`)

tests/ClientTest.php
```diff
@@ -534,43 +357,268 @@ public function testAddBreadcrumbStoresBreadcrumbReturnedByBeforeBreadcrumbCallb
...
+            'auto_log_stacks' => true,
+            'mb_detect_order' => ['ISO-8859-1', 'ASCII', 'UTF-8'],
+        ]);
+
+        /** @var TransportInterface|\PHPUnit_Framework_MockObject_MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
+
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(function (Event $event) {
```
ste93cry: Please add the typehint for the return value (`bool`)

tests/ClientTest.php
```diff
@@ -534,43 +357,268 @@ public function testAddBreadcrumbStoresBreadcrumbReturnedByBeforeBreadcrumbCallb
...
+    public function testConvertExceptionContainingInvalidUtf8Characters()
+    {
+        $malformedString = "\xC2\xA2\xC2"; // ill-formed 2-byte character U+00A2 (CENT SIGN)
+
+        /** @var TransportInterface|\PHPUnit_Framework_MockObject_MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
+
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(function (Event $event) {
```
ste93cry: Please add the typehint for the return value (`bool`)

tests/ClientTest.php
```diff
@@ -534,43 +357,268 @@ public function testAddBreadcrumbStoresBreadcrumbReturnedByBeforeBreadcrumbCallb
...
+
+        $utf8String = 'äöü';
+        $latin1String = utf8_decode($utf8String);
+
+        /** @var TransportInterface|\PHPUnit_Framework_MockObject_MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
+
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(function (Event $event) use ($utf8String) {
```
ste93cry: Please add the typehint for the return value (`bool`)

tests/ClientTest.php
```diff
@@ -534,43 +357,268 @@ public function testAddBreadcrumbStoresBreadcrumbReturnedByBeforeBreadcrumbCallb
...
-        $client->setRepresentationSerializer($serializer);
+        $assertHasStacktrace = $options->getAutoLogStacks();
 
-        $this->assertSame($serializer, $client->getRepresentationSerializer());
+        /** @var TransportInterface|\PHPUnit_Framework_MockObject_MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
+
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(function (Event $event) use ($expectedResult, $assertHasStacktrace) {
```
ste93cry: Please add the typehint for the return value (`bool`)

tests/ClientTest.php
```diff
@@ -534,43 +357,268 @@ public function testAddBreadcrumbStoresBreadcrumbReturnedByBeforeBreadcrumbCallb
...
+        /** @var TransportInterface|\PHPUnit_Framework_MockObject_MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
+
+        $client = new Client(new Options(), $transport, []);
+
+        Hub::getCurrent()->bindClient($client);
+
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(function ($event) {
```
ste93cry: Please add the typehints for both the argument and the return value (`bool`)

src/Event.php
```diff
@@ -134,15 +134,14 @@
 
     /**
      * Event constructor.
+     *
+     * @throws \Ramsey\Uuid\Exception\UnsatisfiedDependencyException if `Moontoast\Math\BigNumber` is not present
```
ste93cry: You should import the class as everything else and not use the FQCN here

src/ClientInterface.php
```diff
@@ -26,192 +23,66 @@
...
-
-    /**
-     * Sets the representation serializer.
+     * @param Scope|null $scope An optional scope keeping the state
      *
-     * @param ReprSerializer $representationSerializer The serializer instance
+     * @return null|string
      */
-    public function setRepresentationSerializer(ReprSerializer $representationSerializer);
+    public function captureLastError(?Scope $scope = null): ?string;
```
ste93cry: I would move back the method to its original position for two reasons: it simpliefies the diff of this file (and so its review) and it stays with all the other `capture*` events before the generic one

src/ErrorHandler.php
```diff
@@ -25,21 +25,24 @@ class ErrorHandler extends AbstractErrorHandler
...
     }
 
     /**
      * {@inheritdoc}
      */
     protected function doHandleException($exception)
     {
-        $this->client->captureException($exception);
+        $callback = $this->callback;
+        $callback($exception);
```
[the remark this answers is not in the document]
ste93cry: This has not been resolved yet but the fix is really easy so we should do it

```diff
-$callback = $this->callback;
-$callback($exception)
+call_user_func($this->callback, $exception);
```

src/Integration/Handler.php
```diff
@@ -0,0 +1,47 @@
...
+ * Class Handler
+ */
+final class Handler
+{
+    private static $integrations = [];
+
+    /**
+     * Calls {@link IntegrationInterface::setupOnce} for all passed integrations if it hasn't been called yet.
+     *
+     * @param array $integrations
```
ste93cry: Missing description of this parameter

src/Integration/Handler.php
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * @return array
```
ste93cry: We can describe the return type in a better way than a generic array: `array<string, IntegrationInterface>`

src/Integration/IntegrationInterface.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\Integration;
+
+/**
+ * This interface describes an Integration.
```
ste93cry: I propose this description: `This interface defines a contract that must be implemented by middlewares, bindings or hooks that integrate certain frameworks or environments with the SDK.`

src/Sdk.php
```diff
@@ -8,11 +8,13 @@
...
- * @param array $options
+ * @param array              $options       the Client options
+ * @param null|ClientBuilder $clientBuilder Use this instance of ClientBuilder instead of creating a new one
  */
-function init(array $options = []): void
+function init(array $options = [], ?ClientBuilder $clientBuilder = null): void
 {
-    Hub::setCurrent(new Hub(ClientBuilder::create($options)->getClient()));
+    $builder = $clientBuilder ?? ClientBuilder::create($options);
+    Hub::setCurrent(new Hub($builder->getClient()));
```
ste93cry: Please add a blank line above this one to improve readability. One thing that puzzles me is that accepting an instance of the client builder means that it has to be created using certain options but the `init` method accept the options too. This in turn means that the first argument is useless if you're using the second argument and this may not be a clear enough API for the user... Just wondering though, we may leave it there and just postpone the problem to a later PR

src/Stacktrace.php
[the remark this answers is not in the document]
ste93cry: Well done, we should add an interface for the serializer (in another PR, obviously) so that we can work on a new implementation of the serializer (the current one is buggy and too limited in features) that can be switched later at any time if needed

tests/Breadcrumbs/BreadcrumbTest.php
```diff
@@ -47,7 +47,7 @@ public function testConstructor()
         $this->assertEquals('foo bar', $breadcrumb->getMessage());
         $this->assertEquals(Breadcrumb::TYPE_USER, $breadcrumb->getType());
         $this->assertEquals(['baz'], $breadcrumb->getMetadata());
-        $this->assertEquals(microtime(true), $breadcrumb->getTimestamp());
+        $this->assertNotNull($breadcrumb->getTimestamp());
```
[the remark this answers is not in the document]
HazAT: Reverted, it's just that when I ran the tests from PHPStrom, the time mismatched.

src/Sdk.php
HazAT: Created a new issue for this: https://github.com/getsentry/sentry-php/issues/695

src/Client.php
```diff
@@ -51,391 +52,164 @@ class Client implements ClientInterface
...
-            return null;
+        if ($maxBreadcrumbs <= 0) {
+            return;
         }
 
-        $exception = new \ErrorException(@$error['message'], 0, @$error['type'], @$error['file'], @$error['line']);
-
-        return $this->captureException($exception, $payload);
-    }
+        $breadcrumb = $beforeBreadcrumbCallback($breadcrumb);
```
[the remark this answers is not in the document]
ste93cry: @HazAT this has not been resolved yet although the fix is really simple

```diff
-$beforeBreadcrumbCallback = $this->options->getBeforeBreadcrumbCallback();
+$breadcrumb = call_user_func($beforeBreadcrumbCallback, $breadcrumb);
```

tests/AbstractSerializerTest.php
ste93cry: What's the purpose of this function having a return type that is not `void`?

tests/AbstractErrorHandlerTest.php
```diff
@@ -121,5 +118,10 @@ public function captureAtDataProvider()
         ];
     }
 
+    /**
```
ste93cry: Useless doc in PHPUnit tests, we never use docblocks there

tests/AbstractErrorHandlerTest.php
```diff
@@ -52,47 +49,47 @@ public function testConstructor()
...
-        $prevErrorReporting = error_reporting($errorReporting);
+        $prevErrorReporting = error_reporting(E_ERROR); // to avoid making the test error bubble up and make the test fail
 
         try {
             $this->assertFalse($errorHandler->handleError(E_WARNING, 'Test', __FILE__, __LINE__));
         } finally {
             error_reporting($prevErrorReporting);
+            restore_error_handler();
+            restore_exception_handler();
+            $this->addToAssertionCount(1);
```
ste93cry: What was the intention of this line of code? Is it a workaround for something?

tests/AbstractErrorHandlerTest.php
```diff
@@ -52,47 +49,47 @@ public function testConstructor()
...
+            $this->callbackMock->expects($this->never())
+                ->method('__invoke');
         }
 
-        $errorHandler = $this->createErrorHandler($this->client);
+        $errorHandler = $this->createErrorHandler($this->callbackMock);
         $errorHandler->captureAt($captureAt, true);
 
-        $prevErrorReporting = error_reporting($errorReporting);
+        $prevErrorReporting = error_reporting(E_ERROR); // to avoid making the test error bubble up and make the test fail
```
ste93cry: Move the comment the line above as it's done in other places

src/State/Scope.php
```diff
@@ -298,8 +309,8 @@ public function applyToEvent(Event $event, int $maxBreadcrumbs = 100): ?Event
         $event->getExtraContext()->merge($this->extra->toArray());
         $event->getUserContext()->merge($this->user->toArray());
 
-        foreach ($this->eventProcessors as $processor) {
-            $event = $processor($event);
+        foreach (array_merge(self::$globalEventProcessors, $this->eventProcessors) as $processor) {
+            $event = $processor($event, $payload);
```
ste93cry: Please use `call_user_func` or not all types of `callable` will work

src/State/Hub.php
```diff
@@ -274,4 +291,20 @@ public static function setCurrent(self $hub): self
...
+    /**
+     * Gets the integration whose FQCN matches the given one if it's available on the current client.
+     *
+     * @param string $className The FQCN of the integration
+     *
+     * @return null|IntegrationInterface
+     */
+    public function getIntegration(string $className): ?IntegrationInterface
+    {
+        if ($client = $this->getClient()) {
```
ste93cry: Please split this condition so that it explicitly check for `null` like it's done in the other methods

src/Integration/ErrorHandlerIntegration.php
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * This integration hooks into the global error handlers and emits events to Sentry.
+ */
+final class ErrorHandlerIntegration implements IntegrationInterface
+{
+    public function setupOnce(): void
+    {
+        ErrorHandler::register(function ($exception) {
+            $self = Hub::getCurrent()->getIntegration(\get_class($this));
```
ste93cry: We can replace the `get_class` call with `self::class`

src/Integration/ModulesIntegration.php
```diff
@@ -0,0 +1,84 @@
...
+        $this->options = $options;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setupOnce(): void
+    {
+        Scope::addGlobalEventProcessor(function (Event $event) {
+            $self = Hub::getCurrent()->getIntegration(\get_class($this));
```
ste93cry: We can replace the `get_class` call with `self::class`

src/Integration/RequestIntegration.php
```diff
@@ -0,0 +1,77 @@
...
+ */
+final class RequestIntegration implements IntegrationInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function setupOnce(): void
+    {
+        Scope::addGlobalEventProcessor(function (Event $event) {
+            $self = Hub::getCurrent()->getIntegration(\get_class($this));
```
ste93cry: We can replace the `get_class` call with `self::class`

src/ClientBuilder.php
```diff
@@ -125,7 +117,16 @@
      */
     public function __construct(array $options = [])
     {
-        $this->configuration = new Configuration($options);
+        $this->options = new Options($options);
+
+        if (null === $this->options->getIntegrations()) {
+            $this->integrations = [];
```
ste93cry: What about initializing the variable where it's declared with an empty `array` and invert this `if` statement so that it gets overwritten if needed?

HazAT: We do not gain anything from this, the if does not really change.

PR #683, opened 2018-10-31
Added passing data from context in monolog breadcrumb handler

dmyers:
This simple change is very useful for breadcrumbs when using monolog. My app uses Monolog via Laravel and I think others will want this behavior too.

Before:
screen shot 2018-10-31 at 2 31 58 pm

After:
screen shot 2018-10-31 at 2 32 21 pm

### DIFF
--- a/lib/Raven/Breadcrumbs/MonologHandler.php
+++ b/lib/Raven/Breadcrumbs/MonologHandler.php
@@ -92,6 +92,7 @@ class Raven_Breadcrumbs_MonologHandler extends AbstractProcessingHandler
                     'level' => $this->logLevels[$record['level']],
                     'category' => $record['channel'],
                     'message' => $record['message'],
+                    'data' => $record['context'],
                 );
             }
         }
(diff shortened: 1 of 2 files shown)


Jean85:
This is a very good catch! Can you add tests for this change?

dmyers:
Thanks, glad you like it. I've been working with the breadcrumbs and it's a very powerful feature.

Does this test seem good? I tested on my computer and made it similar to the other one with the added context assertion.

Jean85:
Nope, it's good, let's merge!

ruudk:
@Jean85 For my info, what is the release process for new tagged releases? Then I know if I can wait or need to use my own fork on production.

Jean85:
I suggest you to use this fork, since it's the only difference from the last tagged version and I can't guarantee when the next tag will be done.

ruudk:
I see 18 changes? https://github.com/getsentry/sentry-php/compare/1.9.2...master

Jean85:
Oh sorry, I was mixing up with the Symfony integration that I just released. In this case I think that we can ask @stayallive that did the last 2 or 3 tags.

ruudk:
Thanks guys !! 💪

PR #684, opened 2018-11-07
[2.0] Do not break in sanitize

tacovandenbroek:
The breaks in the sanitize method will stop sanitizing at the first array and/or sensitive data match failing to replace any sensitive data further down the data being sanitized.


HazAT:
Thanks for the PR but I am closing this.
We will not support any sanitization by default in `2.x` Sentry will introduce a new system for this in general that work across all SDKs.
If you need to sanitize anything we guide people to use `before_send` option.

PR #685, opened 2018-11-07
Monolog > Pass `context` with breadcrumb

ruudk:
Because this is not very useful:
screenshot 2018-11-07 at 12 56 16


Jean85:
Thansk for the contribution! This was recently changed in monolog too, see https://github.com/Seldaek/monolog/pull/1214#issuecomment-435683324. How does this interact with that?

ruudk:
It's unrelated. Because Seldaek/monolog#1214 is about the RavenHandler inside Monolog and this PR changes Sentry's Breadcrumb Handler.

Jean85:
Mh yeah right, they're both handler but they do different stuff.

I'm :+1: on this, but could you add tests?

ste93cry:
This is a duplicate of #683

ruudk:
Lets get #683 merged then? :)

PR #687, opened 2018-11-09
Prepare for 1.11.x

stayallive:
You can merge when reviewed 👍

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -16,7 +16,7 @@
 
 class Raven_Client
 {
-    const VERSION = '1.10.x-dev';
+    const VERSION = '1.11.x-dev';
 
     const PROTOCOL = '6';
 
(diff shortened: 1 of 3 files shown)


lib/Raven/Client.php
```diff
@@ -16,7 +16,7 @@
 
 class Raven_Client
 {
-    const VERSION = '1.10.x-dev';
+    const VERSION = '1.11.x-dev';
```
Jean85: We forgot to fix this on the tag, right?

stayallive: No? https://github.com/getsentry/sentry-php/blob/1.10.0/lib/Raven/Client.php#L19

Jean85: Oh you did it without a PR! :+1: Missed it

stayallive: Yeah it can be done on the release branch itself so don’t need to PR to do things over there :eyes:

PR #690, opened 2018-11-12
Merge extra contexts to contexts

dercodercom:
This Fix solved the Problem with extra data to contexts.
See also: https://github.com/Seldaek/monolog/pull/1214/files/4a86052a6e621284ca19eec4e9c94f647f0dd16d

Example: 
$sentry->extra_context([
    'contexts' => ['browser' => [
        'version' => '32.0',
        'name' => 'Firefox'
    ]]

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -975,6 +975,9 @@ class Raven_Client
                 $data['contexts'] = $data['extra']['contexts'];
             }
             unset($data['extra']['contexts']);
+            if (empty($data['extra'])) {
+                unset($data['extra']);
+            }
         }
         if (!empty($data['breadcrumbs'])) {
             $data['breadcrumbs'] = $this->serializer->serialize($data['breadcrumbs'], 5);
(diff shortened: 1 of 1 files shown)


stayallive:
Hey @dercodercom, could you also add some tests to verify this is working as intended?

dercodercom:
Here the Tests => https://github.com/getsentry/sentry-php/pull/691

Jean85:
Test are green in #691 even without this patch. Are we sure that we need this?

dercodercom:
> Test are green in #691 even without this patch. Are we sure that we need this?

Yes, because at the moment no browser or OS data are taken over in Sentry.
Enclosed two screenshots without fix and with fix.

bildschirmfoto 2018-11-14 um 11 10 32
bildschirmfoto 2018-11-14 um 11 10 11

Jean85:
That's not correct. If the test is not red without the patch, the test is not useful. That's why you have to submit both.

HazAT:
I will close this issue since we actually do not want to merge extra into "global" contexts.
The Browser context has been fixed on sentry.io itself.
Thanks for your contirubution but the root cause of this bug should be fixed now.

PR #691, opened 2018-11-13
Add tests for OS and browser in extra contexts

dercodercom:


Jean85:
Please do not submit this as a separate PR from #690.

Also, since the tests are green even without #690, they are either wrong or the other PR is not needed at all.

stayallive:
Please add this commit to #690 and check to make sure you are actually testing your code (the merging of an context in the extra context) :)

I think you are missing that the php runtime should still be in the context after the browser is added, you are just testing here that the full `contexts` is overridden not that it is merged (what your PR is doing).

PR #632, opened 2018-07-18
Make it possible to changeable default_max_depth in Raven_Serializer

howyi:
I'd like to refer to the complicated nested JSON request in Sentry.

PR #697, opened 2018-11-16
Serialize arrays in a readable format instead of just counting the items.

ArSn:
We do currently have the problem that we are getting an exception where we can not really reproduce the issue but some of our users are running into the issue. Even though we are able to fail gracefully here, we'd like to resolve the issue.

However, the raven client for PHP serializes arrays in a not very helpful way by just counting the array items, which leads to event data like this:

With the proposed changes this changes to (amongst other points where arrays show up):

Which is evidently much more helpful as we can see the user input and reproduce the problem now.

Would be awesome if we could put that into one of the next releases for the official client so we can switch back to that from our forked one. 

I adjusted the tests accordingly, let me know if I should change anything else and keep up the good work!

### DIFF
--- a/lib/Raven/ReprSerializer.php
+++ b/lib/Raven/ReprSerializer.php
@@ -29,11 +29,11 @@ class Raven_ReprSerializer extends Raven_Serializer
         } elseif (is_integer($value) || is_float($value)) {
             return (string) $value;
         } elseif (is_object($value) || gettype($value) == 'object') {
-            return 'Object '.$this->serializeObject($value);
+	        return 'Object '.get_class($value);
         } elseif (is_resource($value)) {
             return 'Resource '.get_resource_type($value);
         } elseif (is_array($value)) {
-            return 'Array '.$this->serializeArray($value);
+            return $this->serializeArray($value);
         } else {
             return $this->serializeString($value);
         }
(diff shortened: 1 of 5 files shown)


Jean85:
There are already numerous attempts at refactoring this issue. We currently are working on the 2.0 branch for a new release, and we would like to address this issue thoroughly.

This approach seems dangerous and not practical to me, since it doesn't take into account the fact that an array can contain any number of things, especially objects, so this PR doesn't seem feasible to me.

ArSn:
@Jean85 You have a valid point, this is however the reason that I am using print_r and not var_export so one does not run into problems with circular references. I appreciate that object content may be serialized here as well, but how would this be a problem? Is your point that the request might be too big?

Regardless, do you have any advice on how we could analyze our issue differently in the meantime before 2.0 is finished?

Jean85:
The first thing that you can track down is which version of the client are you using, since #554 should have reduced this issue a bit. For anything else, there's #632 which would make the limit customizable.

ArSn:
We were in fact using an old version and the level was just cut off too early. Bummer I did not see this before trying to cook my own soup. Well, you never step learning I guess.

Thanks for the hint @Jean85 - I'm closing this PR as I agree it is nonsense then.

Issue #693, opened 2018-11-14
Leverage PSR-17 to use generic PSR-7 factories


Jean85:
As reported in https://github.com/getsentry/sentry-php/pull/682/files#r231547783, we can leverage PSR-17 to not rely on Diactoros directly to obtain the PSR-7 request object during event generation.

Jean85:
I've tried multimple times, but I can't require the virtual package...
```
$ composer require psr/http-factory-implementation:^1.0                            
                                                                                                                                                                                                       
  [InvalidArgumentException]                                                                                                                                                                               
  Could not find a matching version of package psr/http-factory-implementation. Check the package spelling, your version constraint and that the package is available in a stability which matches your minimum-stability (stable).```

@ste93cry you where able to do it with PSR-7, how did you do it?

Jean85:
I talked to @ste93cry and checked on HTTPlug progress. It seems that PSR-17 is too young for now, and they don't support it for now: https://github.com/php-http/message-factory/issues/37

We should leave this issue open, but remove the milestone since it's not feasible now, in the short term.

SlvrEagle23:
Hello all! The PSR-17 and PSR-18 specifications for factories and HTTP clients respectively have seen improved adoption since this issue was created, with Guzzle now directly implementing PSR-18 and the `http-interop/http-factory-x` libraries implementing PSR-17.

It would be really nice to see a version of this library that's decoupled from httplug and supports using any implementation of the PSR spec instead, so we can bring our own versions of each if our application already uses them.

ste93cry:
>It would be really nice to see a version of this library that's decoupled from httplug and supports using any implementation of the PSR spec instead, so we can bring our own versions of each if our application already uses them.

Even though we didn't talk about this yet, I strongly think that we are gonna do this in `3.0` which should be released by the end of the year. You can track the progress of that version on the [`3.x`](https://github.com/getsentry/sentry-php/tree/3.x) branch

PR #698, opened 2018-11-16
[2.0] Allow Symfony 4 & Diactoros 2

Jean85:
This PR allows Symfony 4 by allowing 4.x versions of the only component required by us, which is `symfony/options-resolver`. I also took the liberty of switching from `~` to `^` in all the constraints, because it's [the right one to use when following SemVer](https://getcomposer.org/doc/articles/versions.md#caret-version-range-):

> This is the recommended operator for maximum interoperability when writing library code.


Jean85:
I've also added support for Diactoros 2, since #693 is not feasible for now.

PR #699, opened 2018-11-17
[2.0] Add PHP 7 typehints

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.0
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | no
| License       | MIT

Big PR, big changes: I've added the typehints in the code where it was possible to do it (almost everywhere). Unfortunately I spotted a problem with the [`symfony/phpunit-bridge`](https://github.com/symfony/phpunit-bridge) package that breaks our build: they have an error handler that catches and reports deprecation errors when the test suite ends and in some cases it doesn't return a `boolean` (expected to be returned from any error handler). This breaks our code now that it's typehinted. This is clearly a problem in their implementation and I'm going to open an issue soon to try getting this fixed. In the meanwhile I will leave this PR in WIP but I will still ask for your review as it's ready anyway. If things goes too much long in time we may have to choose to remove the return type typehint from our error handler and introduce it later (not a big problem after all, it's just makes me sad 😢)


Jean85:
Maybe we could cast the return value to bool? In that case, a `null`/`void` return would turn into `false`, which would be a correct assumption I think, since the documented signature reports `bool` as return value: http://php.net/manual/en/function.set-error-handler.php

But my assumption should be tested, because returning `false` means calling the standard PHP error handler, so we need to check PHP behavior on `null`

Jean85:
Nope. As I feared, native PHP doesn't cast, but checks if `false ===` against the error handler return, see the notice from the nativa handler appearing here: https://3v4l.org/rSesI

 So we can circumvent the bug by `return false === $this->previousHandler...`.

src/ErrorHandler.php
```diff
@@ -19,29 +12,322 @@
...
+        $this->exceptionReflection->setValue($errorAsException, $backtrace);
+
+        try {
+            $this->handleException($errorAsException);
+        } catch (\Throwable $exception) {
+            // Do nothing as this error handler should be as transparent as possible
+        }
+
+        if (null !== $this->previousErrorHandler) {
+            return \call_user_func($this->previousErrorHandler, $level, $message, $file, $line);
```
Jean85: As suggested in https://github.com/getsentry/sentry-php/pull/699#issuecomment-439650920
```suggestion
            return false !== \call_user_func($this->previousErrorHandler, $level, $message, $file, $line);
```

Jean85: We should also add a unit test for this.

ste93cry:
> So we can circumvent the bug by `return false === $this->previousHandler...`.

I don't think we should. It's a bug in the implementation of a third party dependency and we should report and imho fix that at the origin of the problem

ste93cry:
I've opened an issue in the Symfony repository to address the problem: symfony/symfony#29253. After looking at the PHP documentation there is nothing written regarding what an error handler should return to stop the normal error handling process so I'm going to apply your suggestion to workaround any other possible problem with third party handlers

tests/State/ScopeTest.php
```diff
@@ -122,7 +113,7 @@ public function testAddEventProcessor(): void
         $this->assertSame($event, $scope->applyToEvent($event, []));
         $this->assertTrue($callback1Called);
 
-        $scope->addEventProcessor(function () use ($callback1Called, &$callback2Called, $callback3Called): ?Event {
+        $scope->addEventProcessor(function () use ($callback1Called, &$callback2Called, $callback3Called) {
```
HazAT: why did you remove `?Event` return value?

tests/State/ScopeTest.php
```diff
@@ -131,7 +122,7 @@ public function testAddEventProcessor(): void
             return null;
         });
 
-        $scope->addEventProcessor(function () use (&$callback3Called): ?Event {
+        $scope->addEventProcessor(function () use (&$callback3Called) {
```
HazAT: why did you remove ?Event return value?

src/Integration/RequestIntegration.php
```diff
@@ -23,7 +23,7 @@
      */
     public function setupOnce(): void
     {
-        Scope::addGlobalEventProcessor(function (Event $event) {
+        Scope::addGlobalEventProcessor(function (Event $event): Event {
```
HazAT: nit: Not really relevant but an `EventProcessor` could also return `null` so while in this case this will not happen it is possible for an `EventProcessor` to discard an `Event` by returning `null`.

src/Integration/ModulesIntegration.php
```diff
@@ -46,6 +45,8 @@ public function setupOnce(): void
     }
 
     /**
+     * Applies the information gathered by the this integration to the event.
```
HazAT: `- the`
`Applies the information gathered by this integration to the event.`

src/Integration/RequestIntegration.php
ste93cry: Indeed, but the current function can't return `null` so it makes no sense to typehint the return type against a `nullable` value

tests/State/ScopeTest.php
ste93cry: I said to you in another PR to add them but rethinking about it it makes no sense if the only return type of the lambda is an hardcoded `null` value

tests/State/ScopeTest.php
ste93cry: I said to you in another PR to add them but rethinking about it it makes no sense if the only return type of the lambda is an hardcoded `null` value

src/ErrorHandler.php
ste93cry: Done. For the moment I've added a unit test in PHPUnit but I was thinking about transforming all these tests into PHPT tests so that we can make the `handle*` methods of the error handler `private` and stop having to rely to be `public` just to test them.

.php_cs
```diff
@@ -8,6 +8,7 @@ return PhpCsFixer\Config::create()
         'array_syntax' => ['syntax' => 'short'],
         'concat_space' => ['spacing' => 'one'],
         'ordered_imports' => true,
+        'declare_strict_types' => true,
```
Jean85: Totally :+1: on this!

PR #702, opened 2018-11-21
[2.0] fix: gzip compressed requests

HazAT:
I think this broke when I removed the base64 encoding.
Before we were compressing the JSON manually and then encoding it with base64, basically sending an uncompressed base64 request.

This now properly sets all requests header and compresses the whole request.

Headers before:
```
POST /api/15/store/ HTTP/1.1
Host: x.eu.ngrok.io
Accept: */*
Transfer-Encoding: chunked
Content-Type: application/octet-stream
User-Agent: sentry.php/2.0.x-dev
X-Sentry-Auth: Sentry sentry_version=7, sentry_client=sentry.php/2.0.x-dev, sentry_timestamp=1542796018.247622, sentry_key=45cca635c50a4e4695681086588e6454
```

Headers with this PR:
```
POST /api/15/store/ HTTP/1.1
Host: x.eu.ngrok.io
Accept: */*
Content-Type: application/json
User-Agent: sentry.php/2.0.x-dev
X-Sentry-Auth: Sentry sentry_version=7, sentry_client=sentry.php/2.0.x-dev, sentry_timestamp=1542797229.347195, sentry_key=45cca635c50a4e4695681086588e6454
Accept-Encoding: gzip
Accept-Encoding: deflate
TE: gzip
TE: deflate
TE: chunked
Content-Length: 14752
```

### DIFF
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -14,6 +14,7 @@ namespace Sentry;
 use Http\Client\Common\Plugin;
 use Http\Client\Common\Plugin\AuthenticationPlugin;
 use Http\Client\Common\Plugin\BaseUriPlugin;
+use Http\Client\Common\Plugin\DecoderPlugin;
 use Http\Client\Common\Plugin\ErrorPlugin;
 use Http\Client\Common\Plugin\HeaderSetPlugin;
 use Http\Client\Common\Plugin\RetryPlugin;
@@ -262,6 +263,10 @@ final class ClientBuilder implements ClientBuilderInterface
         $this->addHttpClientPlugin(new AuthenticationPlugin(new SentryAuth($this->options)));
         $this->addHttpClientPlugin(new RetryPlugin(['retries' => $this->options->getSendAttempts()]));
         $this->addHttpClientPlugin(new ErrorPlugin());
+
+        if ('gzip' === $this->options->getEncoding()) {
+            $this->addHttpClientPlugin(new DecoderPlugin());
+        }
 
         return new PluginClient($this->httpClient, $this->httpClientPlugins);
     }
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -264,7 +264,7 @@ final class ClientBuilder implements ClientBuilderInterface
         $this->addHttpClientPlugin(new RetryPlugin(['retries' => $this->options->getSendAttempts()]));
         $this->addHttpClientPlugin(new ErrorPlugin());
 
-        if ('gzip' === $this->options->getEncoding()) {
+        if ($this->options->isCompressionEnabled()) {
             $this->addHttpClientPlugin(new DecoderPlugin());
         }
 
--- a/tests/ClientBuilderTest.php
+++ b/tests/ClientBuilderTest.php
@@ -214,7 +214,7 @@ class ClientBuilderTest extends TestCase
     /**
      * @dataProvider getClientTogglesCompressionPluginInHttpClientDataProvider
      */
-    public function testGetClientTogglesCompressionPluginInHttpClient($enabled): void
+    public function testGetClientTogglesCompressionPluginInHttpClient(bool $enabled): void
     {
         $builder = ClientBuilder::create(['enable_compression' => $enabled, 'dsn' => 'http://public:secret@example.com/sentry/1']);
         $builder->getClient();
--- a/tests/ClientBuilderTest.php
+++ b/tests/ClientBuilderTest.php
@@ -196,7 +196,7 @@ class ClientBuilderTest extends TestCase
             ['setMbDetectOrder', ['foo', 'bar']],
             ['setAutoLogStacks', false],
             ['setContextLines', 0],
-            ['setEncoding', 'gzip'],
+            ['setEnableCompression', false],
             ['setCurrentEnvironment', 'test'],
             ['setEnvironments', ['default']],
             ['setExcludedLoggers', ['foo', 'bar']],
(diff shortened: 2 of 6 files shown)


tests/Transport/HttpTransportTest.php
```diff
@@ -114,40 +114,6 @@ public function testSendWithoutCompressedEncoding()
         $reflectionMethod->setAccessible(false);
     }
 
-    public function testSendWithCompressedEncoding()
```
Jean85: Shouldn't we replace with test with new ones?

HazAT:
Added a test + fixed a bug with the url:

screenshot 2018-11-21 15 37 51

tests/ClientBuilderTest.php
```diff
@@ -210,6 +210,38 @@ public function optionsDataProvider()
             ['setErrorTypes', 0],
         ];
     }
+
+    public function testCompressionPluginAdded()
+    {
+        $builder = ClientBuilder::create(['encoding' => 'gzip', 'dsn' => 'https://completelyrandom@dsn.asdf/42']);
+        $builder->getClient();
+        $reflectionProperty = new \ReflectionProperty(ClientBuilder::class, 'httpClientPlugins');
```
Jean85: I don't like testing this with reflection, but I see that there isn't another pretty way to do it.
Let's hope to be able to use a stub server ASAP for this kind of stuff.

src/ClientBuilder.php
```diff
@@ -263,6 +264,10 @@ private function createHttpClientInstance()
         $this->addHttpClientPlugin(new RetryPlugin(['retries' => $this->options->getSendAttempts()]));
         $this->addHttpClientPlugin(new ErrorPlugin());
 
+        if ('gzip' === $this->options->getEncoding()) {
```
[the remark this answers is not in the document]
HazAT: Agreed!

tests/ClientBuilderTest.php
ste93cry: Instead of manually using the `ReflectionProperty` class we could simplify a bit the test by using the helper method of PHPUnit `getObjectAttribute`

tests/ClientBuilderTest.php
```diff
@@ -210,6 +210,38 @@ public function optionsDataProvider()
             ['setErrorTypes', 0],
         ];
     }
+
+    public function testCompressionPluginAdded()
+    {
+        $builder = ClientBuilder::create(['enableCompression' => true, 'dsn' => 'https://completelyrandom@dsn.asdf/42']);
```
ste93cry: Not that it changes something or is important at all, but we used through the codebase this DSN for the tests: `http://public:secret@example.com/sentry/1`

tests/ClientBuilderTest.php
```diff
@@ -210,6 +210,38 @@ public function optionsDataProvider()
...
+        $builder->getClient();
+        $reflectionProperty = new \ReflectionProperty(ClientBuilder::class, 'httpClientPlugins');
+        $reflectionProperty->setAccessible(true);
+        $foundDecoderPlugin = false;
```
ste93cry: Add a blank line above and below this line to improve readability. Also I would rename the variable to `$decoderPluginFound` because I beleive it's grammatically more correct (or not?)

tests/ClientBuilderTest.php
```diff
@@ -210,6 +210,38 @@ public function optionsDataProvider()
...
+        foreach ($reflectionProperty->getValue($builder) as $plugin) {
+            if ($plugin instanceof Plugin\DecoderPlugin) {
```
ste93cry: We can break after the first time this condition is `true`

tests/ClientBuilderTest.php
```diff
@@ -210,6 +210,38 @@ public function optionsDataProvider()
...
+                $foundDecoderPlugin = true;
+            }
+        }
+        $this->assertTrue($foundDecoderPlugin);
```
ste93cry: Add a blank line above this one to improve readability

tests/ClientBuilderTest.php
```diff
@@ -210,6 +210,38 @@ public function optionsDataProvider()
...
+        $reflectionProperty->setAccessible(false);
+    }
+
+    public function testCompressionPluginNotAdded()
```
ste93cry: If we use a data provider for the test above we could just accept as parameter of the test if we expect or not the plugin instance to exists in the list of plugins and then assert against that

tests/ClientBuilderTest.php
```diff
@@ -210,6 +210,33 @@ public function optionsDataProvider()
...
+
+    /**
+     * @dataProvider compressionOptionDataProvider
+     */
+    public function testCompressionPlugin($enabled)
+    {
+        $builder = ClientBuilder::create(['enable_compression' => $enabled, 'dsn' => 'http://public:secret@example.com/sentry/1']);
+        $builder->getClient();
+
+        $decoderPluginFound = false;
```
ste93cry: Please add a blank line below this one to improve readability

tests/ClientBuilderTest.php
```diff
@@ -210,6 +210,33 @@ public function optionsDataProvider()
...
+        foreach ($this->getObjectAttribute($builder, 'httpClientPlugins') as $plugin) {
+            if ($plugin instanceof Plugin\DecoderPlugin) {
+                $decoderPluginFound = true;
+                break;
```
ste93cry: Please add a blank line above this one to improve readability

tests/ClientBuilderTest.php
```diff
@@ -210,6 +210,33 @@ public function optionsDataProvider()
...
+            }
+        }
+
+        $this->assertEquals($enabled, $decoderPluginFound);
+    }
+
+    public function compressionOptionDataProvider()
```
ste93cry: Typehint the return type (`array`) and change the method name to `getClientTogglesCompressionPluginInHttpClientDataProvider`

tests/ClientBuilderTest.php
```diff
@@ -210,6 +210,33 @@ public function optionsDataProvider()
             ['setErrorTypes', 0],
         ];
     }
+
+    /**
+     * @dataProvider compressionOptionDataProvider
+     */
+    public function testCompressionPlugin($enabled)
```
ste93cry: Please typehint the return type (`void`) and the parameter. Also I would change the method name to `testGetClientTogglesCompressionPluginInHttpClient`

Issue #695, opened 2018-11-15
[2.0] Revisit `init` parameters


HazAT:
Right now init looks like this

```php
function init(array $options = [], ?ClientBuilder $clientBuilder = null): void
```

as mentioned https://github.com/getsentry/sentry-php/pull/682#discussion_r233652663
this is a bit confusing.

ste93cry:
>One possible good solution that I see here would be reducing the scope of the ClientBuilder to a simpler TransportFactory, because that is the only proper dependency that needs to be handled outside of simple options.

As you said the serializer is another dependency that should not be handled as an option, so we still have the need of having a builder if we want to provide users with something they can quickly use to build up a client instance

Jean85:
I've put the serializer as an other dep of the client and as a property of the builder in #701

PR #703, opened 2018-11-22
[2.0] `init` remove ClientBuilder arg

HazAT:
This PR removes the ambiguous `ClientBuilder` arg.
We do not really need it since we can bind a new Client to the Hub anytime -> see phpt test

Fixes https://github.com/getsentry/sentry-php/issues/695

### DIFF
--- a/tests/phpt/fatal_error_not_captured_twice.phpt
+++ b/tests/phpt/fatal_error_not_captured_twice.phpt
@@ -10,7 +10,6 @@ use Sentry\Spool\MemorySpool;
 use Sentry\Transport\SpoolTransport;
 use Sentry\ClientBuilder;
 use Sentry\State\Hub;
-use function Sentry\init;
 
 $vendor = __DIR__;
 
@@ -23,9 +22,8 @@ require $vendor . '/vendor/autoload.php';
 $spool = new MemorySpool();
 $transport = new SpoolTransport($spool);
 
-init([]);
-
 $builder = ClientBuilder::create()->setTransport($transport);
+
 Hub::getCurrent()->bindClient($builder->getClient());
 
 register_shutdown_function('register_shutdown_function', function () use ($spool) {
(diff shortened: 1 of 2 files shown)


Jean85:
It seems fine to me, but I would like to see and evaluate the "alternative path" first, since it will have to be documented too.

What should the user do if he want to go deeper and use the `ClientBuilder` to customize the client further? Should he skip the `init()` altogether? Like:

```php
$client = ClientBuilder::create($options)
    // ...
    ->getClient();

Hub::setCurrent(new Hub($client));
```

So I see some suggestions:
 * we can call `Hub::setCurrent()` in the `Hub` constructor if there's no current hub
 * we can expose a `ClientBuilder::createFromOptions(Options $options)` so users can use a typed class instead of a free array, to be more strict

Jean85:
The first suggestion is to skip a call when working with a single Hub, which in PHP should be 99% of the cases.

The second one is to encourage the usage of typed code, instead of a free array that, with a typo in key values, can lead to undesired errors. If it wasn't for the unified API, I would ditch the array completely...

ste93cry:
>If it wasn't for the unified API, I would ditch the array completely...

The array is validated anyway inside the `Options` class (both the type of each value assigned to each option and the value itself) so it's fine to use a generic array, you will get an error the first time you instantiate the client (not a compile time, but as soon as your code runs so better than nothing). The only thing that leaves me to think is that there is no way to configure the transport or the serializer, but I also believe that a user that wants to do it is now a newbie and wants to specifically be in control of what he does/runs so it's fine for him to construct the client instance itself and then bind it to the hub. Probably not a single framework integration will use the `init` method (I strongly believe it's more useful in other SDKs like the Javascript one) but we need it no way around so...

tests/phpt/fatal_error_not_captured_twice.phpt
```diff
@@ -22,7 +23,10 @@ require $vendor . '/vendor/autoload.php';
 $spool = new MemorySpool();
 $transport = new SpoolTransport($spool);
 
-init([], ClientBuilder::create()->setTransport($transport));
+init([]);
```
ste93cry: I don't find very useful to use the `init` method that internally creates an instance of the client to just override it one line below. Please just use the `bindClient` method

PR #705, opened 2018-11-22
[2.0] Rework Options

HazAT:
This brings all the options to the unified API.
The only option missing is `debug` but we do not have an internal debug logger yet, not really crucial for the first 2.0 release.


src/Client.php
```diff
@@ -124,6 +124,10 @@ public function captureMessage(string $message, ?Severity $level = null, ?Scope
         $payload['message'] = $message;
         $payload['level'] = $level;
 
+        if ($this->getOptions()->getAttachStacktrace()) {
+            $payload['exception'] = new \Exception('Sentry Synthetic Exception');
```
Jean85: What's this for?

src/Client.php
```diff
@@ -264,16 +271,14 @@ private function addThrowableToEvent(Event $event, \Throwable $exception): void
...
-                $data['stacktrace'] = Stacktrace::createFromBacktrace(
-                    $this->getOptions(),
-                    $this->serializer,
-                    $this->representationSerializer,
-                    $currentException->getTrace(),
-                    $currentException->getFile(),
-                    $currentException->getLine()
-                );
-            }
+            $data['stacktrace'] = Stacktrace::createFromBacktrace(
```
Jean85: Why did we loose the `if` check here?

tests/ClientBuilderTest.php
```diff
@@ -188,13 +188,10 @@ public function optionsDataProvider(): array
...
             ['setSampleRate', 0.5],
             ['setMbDetectOrder', ['foo', 'bar']],
-            ['setAutoLogStacks', false],
+            ['setAttachStacktrace', true],
             ['setContextLines', 0],
             ['setEnableCompression', false],
-            ['setCurrentEnvironment', 'test'],
-            ['setEnvironments', ['default']],
-            ['setExcludedLoggers', ['foo', 'bar']],
-            ['setExcludedExceptions', ['foo', 'bar']],
```
Jean85: You didn't remove `setExcludedExceptions`, that's an error?

src/ClientBuilder.php
```diff
@@ -41,18 +41,12 @@
...
+ * @method setAttachStacktrace(bool $enable)
  * @method int getContextLines()
  * @method setContextLines(int $contextLines)
- * @method string getCurrentEnvironment()
- * @method setCurrentEnvironment(string $environment)
- * @method string[] getEnvironments()
- * @method setEnvironments(string[] $environments)
- * @method string[] getExcludedLoggers()
- * @method setExcludedLoggers(string[] $loggers)
- * @method string[] getExcludedExceptions()
```
Jean85: You didn't remove `get`/`setExcludedExceptions`

tests/ClientTest.php
```diff
@@ -395,23 +389,6 @@ public function convertExceptionDataProvider()
                     ],
                 ],
             ],
-            [
-                new \RuntimeException('foo'),
-                [
-                    'auto_log_stacks' => false,
```
Jean85: You're no longer testing for this case, why?

src/Options.php
```diff
@@ -179,23 +179,23 @@ public function setMbDetectOrder($detectOrder): void
     }
 
     /**
-     * Gets whether the stacktrace must be auto-filled.
+     * Gets whether the stacktrace will be attached on captureMessage.
      *
      * @return bool
      */
-    public function getAutoLogStacks(): bool
+    public function getAttachStacktrace(): bool
```
ste93cry: I would name this method `shouldAttachStacktrace` because it's not really a getter

src/Options.php
```diff
@@ -179,23 +179,23 @@ public function setMbDetectOrder($detectOrder): void
     }
 
     /**
-     * Gets whether the stacktrace must be auto-filled.
+     * Gets whether the stacktrace will be attached on captureMessage.
```
ste93cry: Is this option specifically designed to work only in the `captureMessage` method? It seems a bit strange...

src/ClientBuilder.php
```diff
@@ -41,18 +41,12 @@
...
- * @method setExcludedExceptions(string[] $exceptions)
+ * @method null|string getEnvironment()
+ * @method setEnvironment(?string $environment)
```
ste93cry: The PHPDoc standard doesn't allow `nullable` values to be defined with the syntax PHP uses (the exclamation mark before the type). You always have to use the union of the type and the `null` value

src/Options.php
```diff
@@ -179,23 +179,23 @@ public function setMbDetectOrder($detectOrder): void
...
-        return $this->options['auto_log_stacks'];
+        return $this->options['attach_stacktrace'];
     }
 
     /**
-     * Sets whether the stacktrace must be auto-filled.
+     * Sets whether the stacktrace will be attached on captureMessage.
      *
-     * @param bool $enable Flag indicating if the stacktrace must be auto-filled
+     * @param bool $enable flag indicating if the stacktrace will be attached to captureMessage calls
```
ste93cry: Not that important after all, but the description of the parameter should always start with a capitalized letter

src/Options.php
```diff
@@ -660,14 +614,10 @@ private function configureOptions(OptionsResolver $resolver)
...
             'sample_rate' => 1,
             'mb_detect_order' => null,
-            'auto_log_stacks' => true,
+            'attach_stacktrace' => false,
             'context_lines' => 3,
             'enable_compression' => true,
-            'current_environment' => 'default',
-            'environments' => [],
-            'excluded_loggers' => [],
-            'excluded_exceptions' => [],
```
ste93cry: Not that important after all but I'm going to ask anyway: is there any reason regarding these options being moved to the end of the list?

src/Client.php
HazAT: We don't need this check, if there is an exception we always want the stacktrace.

HazAT: `attachStacktrace` only determines if a stacktrace is attached to `captureMessage`

src/ClientBuilder.php
HazAT: fixed

tests/ClientTest.php
HazAT: Explained above, for exceptions we always want a stacktrace.

tests/ClientBuilderTest.php
HazAT: Fixed

PR #707, opened 2018-11-26
[2.0] send_default_pii option

HazAT:
This PR adds `send_default_pii` as an option to the SDK.
Purpose of this is that it must be opt-in to send any user-related information such as IP address or cookies.
Frames or anything more sophisticated than this will not be done for now since in future "Relay" will handle PII stripping.

This though is the bare minimum we have to support in each SDK.

### DIFF
--- a/src/Integration/RequestIntegration.php
+++ b/src/Integration/RequestIntegration.php
@@ -100,14 +100,9 @@ final class RequestIntegration implements IntegrationInterface
     private function removePiiFromHeaders(array $headers): array
     {
         $keysToRemove = ['authorization', 'cookie', 'set-cookie', 'remote_addr'];
-        $sanitizedHeaders = [];
 
-        foreach ($headers as $key => $value) {
-            if (!\in_array(\strtolower($key), $keysToRemove)) {
-                $sanitizedHeaders[$key] = $value;
-            }
-        }
-
-        return $sanitizedHeaders;
+        return array_filter($headers, function ($key) use ($keysToRemove) {
+            return !\in_array(\strtolower($key), $keysToRemove);
+        }, ARRAY_FILTER_USE_KEY);
     }
 }
--- a/src/Integration/RequestIntegration.php
+++ b/src/Integration/RequestIntegration.php
@@ -6,6 +6,7 @@ namespace Sentry\Integration;
 
 use Psr\Http\Message\ServerRequestInterface;
 use Sentry\Event;
+use Sentry\Options;
 use Sentry\State\Hub;
 use Sentry\State\Scope;
 use Zend\Diactoros\ServerRequestFactory;
@@ -19,6 +20,21 @@ use Zend\Diactoros\ServerRequestFactory;
 final class RequestIntegration implements IntegrationInterface
 {
     /**
+     * @var Options The client options
+     */
+    private $options;
+
+    /**
+     * RequestIntegration constructor.
+     *
+     * @param Options $options The Client Options
+     */
+    public function __construct(Options $options)
+    {
+        $this->options = $options;
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function setupOnce(): void
@@ -30,7 +46,7 @@ final class RequestIntegration implements IntegrationInterface
                 return $event;
             }
 
-            self::applyToEvent($event);
+            self::applyToEvent($self, $event);
 
             return $event;
         });
@@ -39,10 +55,11 @@ final class RequestIntegration implements IntegrationInterface
     /**
      * Applies the information gathered by the this integration to the event.
      *
+     * @param self                        $self    The current instance of RequestIntegration
      * @param Event                       $event   The event that will be enriched with a request
      * @param null|ServerRequestInterface $request The Request that will be processed and added to the event
      */
-    public static function applyToEvent(Event $event, ?ServerRequestInterface $request = null): void
+    public static function applyToEvent(self $self, Event $event, ?ServerRequestInterface $request = null): void
     {
         if (null === $request) {
             $request = isset($_SERVER['REQUEST_METHOD']) && \PHP_SAPI !== 'cli' ? ServerRequestFactory::fromGlobals() : null;
@@ -55,24 +72,42 @@ final class RequestIntegration implements IntegrationInterface
         $requestData = [
             'url' => (string) $request->getUri(),
             'method' => $request->getMethod(),
-            'headers' => $request->getHeaders(),
-            'cookies' => $request->getCookieParams(),
         ];
 
         if ($request->getUri()->getQuery()) {
             $requestData['query_string'] = $request->getUri()->getQuery();
         }
 
-        if ($request->hasHeader('REMOTE_ADDR')) {
-            $requestData['env']['REMOTE_ADDR'] = $request->getHeaderLine('REMOTE_ADDR');
+        if ($self->options->shouldSendDefaultPii()) {
+            if ($request->hasHeader('REMOTE_ADDR')) {
+                $requestData['env']['REMOTE_ADDR'] = $request->getHeaderLine('REMOTE_ADDR');
+            }
+
+            $requestData['cookies'] = $request->getCookieParams();
+            $requestData['headers'] = $request->getHeaders();
+
+            $userContext = $event->getUserContext();
+            if (null === $userContext->getIpAddress() && $request->hasHeader('REMOTE_ADDR')) {
+                $userContext->setIpAddress($request->getHeaderLine('REMOTE_ADDR'));
+            }
+        } else {
+            $requestData['headers'] = $self->removePiiFromHeaders($request->getHeaders());
         }
 
         $event->setRequest($requestData);
+    }
 
-        $userContext = $event->getUserContext();
+    private function removePiiFromHeaders(array $headers): array
+    {
+        $keysToRemove = ['authorization', 'cookie', 'set-cookie', 'remote_addr'];
+        $sanitizedHeaders = [];
 
-        if (null === $userContext->getIpAddress() && $request->hasHeader('REMOTE_ADDR')) {
-            $userContext->setIpAddress($request->getHeaderLine('REMOTE_ADDR'));
+        foreach ($headers as $key => $value) {
+            if (!\in_array(\strtolower($key), $keysToRemove)) {
+                $sanitizedHeaders[$key] = $value;
+            }
         }
+
+        return $sanitizedHeaders;
     }
 }
--- a/src/Integration/RequestIntegration.php
+++ b/src/Integration/RequestIntegration.php
@@ -87,6 +87,7 @@ final class RequestIntegration implements IntegrationInterface
             $requestData['headers'] = $request->getHeaders();
 
             $userContext = $event->getUserContext();
+
             if (null === $userContext->getIpAddress() && $request->hasHeader('REMOTE_ADDR')) {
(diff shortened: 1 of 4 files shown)


src/Integration/RequestIntegration.php
```diff
@@ -55,24 +72,42 @@ public static function applyToEvent(Event $event, ?ServerRequestInterface $reque
...
 
-        $userContext = $event->getUserContext();
+    private function removePiiFromHeaders(array $headers): array
+    {
+        $keysToRemove = ['authorization', 'cookie', 'set-cookie', 'remote_addr'];
+        $sanitizedHeaders = [];
 
-        if (null === $userContext->getIpAddress() && $request->hasHeader('REMOTE_ADDR')) {
-            $userContext->setIpAddress($request->getHeaderLine('REMOTE_ADDR'));
+        foreach ($headers as $key => $value) {
```
Jean85: Maybe we can simplify and optimize this with `array_filter`?

ste93cry:
I'm totally against this implementation as is because I see several problems with it:

- the PII can be several data, ranging from IP addresses to user data like emails, etc while the option only takes the simplest cases into account
- we are overcomplicating the `RequestIntegration` integration with code that doesn't belong there and makes the testing much more difficult. Instead there should be an integration that does the job of saniting the data and only that
- when middlewares were still a thing, we had several of them to sanitize different things like cookies with a really clean code. I strongly believe we should reuse that code because it's clearer, more readable, more reliable and can be configured. For example the `Authorization` header can have different names depending on the platform and hardcoding all of them is impossible

In the end, I perfectly understand the need for this PR and the fact that we don't want to include anymore much sanitization because doing that job is the responsability of [Relay](https://docs.sentry.io/data-management/relay/) but I believe this is not the way to do basic sanitization at all because it's so simplistic that it's almost useless

tests/Integration/RequestIntegrationTest.php
```diff
@@ -8,46 +8,68 @@
...
+            foreach ($requestData['headers'] as $name => $value) {
+                $request = $request->withHeader($name, $value);
+            }
 
-        RequestIntegration::applyToEvent($event, $request);
+            $this->assertInstanceOf(ServerRequestInterface::class, $request);
 
-        $this->assertEquals(['ip_address' => '127.0.0.1', 'foo' => 'bar'], $event->getUserContext()->toArray());
+            $integration = new RequestIntegration(new Options(['send_default_pii' => $pii]));
+            RequestIntegration::applyToEvent($integration, $event, $request);
```
ste93cry: Add a blank line above this one to improve readability

tests/Integration/RequestIntegrationTest.php
```diff
@@ -8,46 +8,68 @@
...
 
-        $request = new ServerRequest();
-        $request = $request->withHeader('REMOTE_ADDR', '127.0.0.1');
-        $this->assertInstanceOf(ServerRequestInterface::class, $request);
+    /**
+     * @dataProvider invokeDataProvider
+     */
+    public function testInvoke(array $requestData, array $expectedValueWithPii, array $expectedValueWithoutPii): void
+    {
+        foreach ([true, false] as $pii) {
```
ste93cry: Instead of using a `foreach` you should use the data provider as you do with the rest of the data

tests/Integration/RequestIntegrationTest.php
```diff
@@ -8,46 +8,68 @@
...
-        foreach ($requestData['headers'] as $name => $value) {
-            $request = $request->withHeader($name, $value);
-        }
-
+        $request = $request->withHeader('REMOTE_ADDR', '127.0.0.1');
         $this->assertInstanceOf(ServerRequestInterface::class, $request);
 
-        RequestIntegration::applyToEvent($event, $request);
+        $integration = new RequestIntegration(new Options(['send_default_pii' => $pii]));
+        RequestIntegration::applyToEvent($integration, $event, $request);
```
ste93cry: Add a blank line above this one to improve readability

tests/Integration/RequestIntegrationTest.php
```diff
@@ -8,46 +8,68 @@
...
         $request = new ServerRequest();
-        $request = $request->withCookieParams($requestData['cookies']);
-        $request = $request->withUri(new Uri($requestData['uri']));
-        $request = $request->withMethod($requestData['method']);
-
-        foreach ($requestData['headers'] as $name => $value) {
-            $request = $request->withHeader($name, $value);
-        }
-
+        $request = $request->withHeader('REMOTE_ADDR', '127.0.0.1');
```
ste93cry: Add a blank line below this one to improve readability

src/Integration/RequestIntegration.php
```diff
@@ -55,24 +72,37 @@ public static function applyToEvent(Event $event, ?ServerRequestInterface $reque
...
-        $userContext = $event->getUserContext();
+    private function removePiiFromHeaders(array $headers): array
+    {
+        $keysToRemove = ['authorization', 'cookie', 'set-cookie', 'remote_addr'];
 
-        if (null === $userContext->getIpAddress() && $request->hasHeader('REMOTE_ADDR')) {
-            $userContext->setIpAddress($request->getHeaderLine('REMOTE_ADDR'));
-        }
+        return array_filter($headers, function ($key) use ($keysToRemove) {
+            return !\in_array(\strtolower($key), $keysToRemove);
```
ste93cry: You should use strict comparison here by setting the third parameter of the function to `true`

src/Integration/RequestIntegration.php
```diff
@@ -55,24 +72,37 @@ public static function applyToEvent(Event $event, ?ServerRequestInterface $reque
...
+        if ($self->options->shouldSendDefaultPii()) {
+            if ($request->hasHeader('REMOTE_ADDR')) {
+                $requestData['env']['REMOTE_ADDR'] = $request->getHeaderLine('REMOTE_ADDR');
+            }
+
+            $requestData['cookies'] = $request->getCookieParams();
+            $requestData['headers'] = $request->getHeaders();
+
+            $userContext = $event->getUserContext();
+            if (null === $userContext->getIpAddress() && $request->hasHeader('REMOTE_ADDR')) {
```
ste93cry: Add a blank line above this one to improve readability

tests/Integration/RequestIntegrationTest.php
HazAT: I didn't want to duplicate the data in the `dataProvider` and just toggle the flag true / false.
Maybe if I box the data differently this will work.

ste93cry: The data provider should provide the data for a complete test case, if half of the data is coming from the provider and half is inside the test it's confusing and it could means that either the method is testing too much things together or the provided data is not enough for the test-case

tests/Integration/RequestIntegrationTest.php
```diff
@@ -8,52 +8,77 @@
...
+                false,
+                ['foo' => 'bar'],
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider invokeDataProvider
+     */
+    public function testInvoke(bool $pii, array $requestData, array $expectedResult): void
```
ste93cry: Can we rename the variable `$pii` to something that better describes its purpose (e.g. `$shouldSendPii`)?

tests/Integration/RequestIntegrationTest.php
```diff
@@ -8,52 +8,77 @@
...
-        $request = $request->withHeader('REMOTE_ADDR', '127.0.0.1');
+        $request = $request->withCookieParams($requestData['cookies']);
+        $request = $request->withUri(new Uri($requestData['uri']));
+        $request = $request->withMethod($requestData['method']);
+
+        foreach ($requestData['headers'] as $name => $value) {
+            $request = $request->withHeader($name, $value);
+        }
+
         $this->assertInstanceOf(ServerRequestInterface::class, $request);
```
ste93cry: This assert is not needed since the `$request` object will always be an instance of `ServerRequestInterface`

tests/Integration/RequestIntegrationTest.php
```diff
@@ -8,52 +8,77 @@
...
 
-        $this->assertEquals(['ip_address' => '127.0.0.1', 'foo' => 'bar'], $event->getUserContext()->toArray());
+        $this->assertEquals($expectedResult, $event->getRequest());
     }
 
     public function invokeDataProvider(): array
     {
         return [
             [
+                true, // send_default_pii
```
ste93cry: In all data providers we never commented anything. If you think that the test case is not enough clear, a better way to describe it is to use as key a little phrase describing what it does. See the example below:

```php
return [
    'description' => [
        ...
    ]
];
```

You can also mix test cases that have a name with cases that don't have one, even though I don't think is nice to see 😄

tests/Integration/RequestIntegrationTest.php
```diff
@@ -8,52 +8,77 @@
...
 use Zend\Diactoros\Uri;
 
 final class RequestIntegrationTest extends TestCase
 {
     /**
-     * @dataProvider invokeDataProvider
+     * @dataProvider invokeUserContextPiiDataProvider
      */
-    public function testInvoke(array $requestData, array $expectedValue): void
+    public function testInvokeWithRequestHavingIpAddress(bool $pii, array $expectedValue): void
```
ste93cry: Can we rename the variable `$pii` to something that better describes its purpose (e.g. `$shouldSendPii`)?

src/Options.php
```diff
@@ -634,6 +656,7 @@ private function configureOptions(OptionsResolver $resolver)
             },
             'excluded_exceptions' => [],
             'excluded_app_paths' => [],
+            'send_default_pii' => false,
```
ste93cry: Wouldn't be `send_pii` or `send_pii_by_default` better names? Maybe you can find a better one, what makes me wondering is that the current one seems to tell users that the SDK will send some default PII which to me means nothing, or at least I don't understand what it means

src/Options.php
```diff
@@ -597,6 +597,28 @@ public function getIntegrations(): ?array
...
+     */
+    public function shouldSendDefaultPii(): bool
+    {
+        return $this->options['send_default_pii'];
+    }
+
+    /**
+     * Sets if default PII should be sent with every event (if possible).
+     *
+     * @param bool $enable flag indicating if default PII will be sent
```
ste93cry: Capitalize the first word of the description pelase

src/Integration/RequestIntegration.php
```diff
@@ -55,24 +72,38 @@ public static function applyToEvent(Event $event, ?ServerRequestInterface $reque
...
+            }
+        } else {
+            $requestData['headers'] = $self->removePiiFromHeaders($request->getHeaders());
         }
 
         $event->setRequest($requestData);
+    }
 
-        $userContext = $event->getUserContext();
+    private function removePiiFromHeaders(array $headers): array
```
ste93cry: This function is missing the docblock (description of both the method and the parameter)

tests/Integration/RequestIntegrationTest.php
HazAT: We need this because of https://travis-ci.org/getsentry/sentry-php/jobs/460676193

src/Options.php
ste93cry: May I suggest to change this in the unified API then? I won't block this PR, just suggesting a change :wink:

PR #708, opened 2018-11-26
Minimal code refactors

Jean85:


src/Stacktrace.php
```diff
@@ -305,8 +305,8 @@ protected static function getFrameArgumentsValues(array $frame, int $maxValueLen
 
         $result = [];
 
-        for ($i = 0; $i < \count($frame['args']); ++$i) {
-            $result['param' . ($i + 1)] = self::serializeArgument($frame['args'][$i], $maxValueLength);
+        foreach ($frame['args'] as $i => $argument) {
```
ste93cry: Rename this variable `$index`

src/Context/ServerOsContext.php
```diff
@@ -186,7 +186,7 @@ public function setKernelVersion(string $kernelVersion): void
     private function configureOptions(OptionsResolver $resolver): void
     {
         $resolver->setDefaults([
-            'name' => php_uname('s'),
+            'name' => PHP_OS,
```
ste93cry: Does it makes sense to change this instead of being coherent with the rest of the values using the `php_uname` function?

ste93cry: Yea indeed using a constant is better than calling a runtime function, I'm not sure it was worth the change since we have anyway other calls `php_uname` and so removing one imho won't affect the performance. Either ways are fine by the way

PR #709, opened 2018-11-27
Remove addBreadcrumb from client

Jean85:
This spurred while analyzing the migration for the Symfony integration to the 2.0 client. It should simplify usage, avoiding the mirroring of the `addBreadcrumb' method between the client and the hub.

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -662,7 +662,7 @@ class Options
         $resolver->setAllowedValues('dsn', \Closure::fromCallable([$this, 'validateDsnOption']));
         $resolver->setAllowedValues('integrations', \Closure::fromCallable([$this, 'validateIntegrationsOption']));
         $resolver->setAllowedValues('max_breadcrumbs', function ($value) {
-            return $value <= self::DEFAULT_MAX_BREADCRUMBS;
+            return $value >= 0 && $value <= self::DEFAULT_MAX_BREADCRUMBS;
         });
 
         $resolver->setNormalizer('dsn', \Closure::fromCallable([$this, 'normalizeDsnOption']));
--- a/src/Options.php
+++ b/src/Options.php
@@ -661,9 +661,7 @@ class Options
 
         $resolver->setAllowedValues('dsn', \Closure::fromCallable([$this, 'validateDsnOption']));
         $resolver->setAllowedValues('integrations', \Closure::fromCallable([$this, 'validateIntegrationsOption']));
-        $resolver->setAllowedValues('max_breadcrumbs', function ($value) {
-            return $value >= 0 && $value <= self::DEFAULT_MAX_BREADCRUMBS;
-        });
+        $resolver->setAllowedValues('max_breadcrumbs', \Closure::fromCallable([$this, 'validateMaxBreadcrumbsOptions']));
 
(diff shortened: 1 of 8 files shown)


ste93cry:
I think we can just merge this and then rebase the other one? This way the other PR is just about adding the interface to the `Hub` class and fixing its tests so that we separate the two things

tests/State/HubTest.php
```diff
@@ -240,17 +241,83 @@ public function testAddBreadcrumbDoesNothingIfClientIsNotBinded(): void
...
+        $hub->addBreadcrumb(new Breadcrumb(Breadcrumb::LEVEL_ERROR, Breadcrumb::TYPE_ERROR, 'error_reporting'));
+
+        $this->assertEmpty($hub->getScope()->getBreadcrumbs());
+    }
+
+    public function addBreadcrumbDoesNothingIfMaxBreadcrumbsLimitIsTooLowDataProvider(): array
+    {
+        return [
+            [0],
+            [-1],
```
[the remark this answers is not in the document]
Jean85: That would be a good idea. I'll try to do that.

src/Options.php
```diff
@@ -662,7 +662,7 @@ private function configureOptions(OptionsResolver $resolver): void
         $resolver->setAllowedValues('dsn', \Closure::fromCallable([$this, 'validateDsnOption']));
         $resolver->setAllowedValues('integrations', \Closure::fromCallable([$this, 'validateIntegrationsOption']));
         $resolver->setAllowedValues('max_breadcrumbs', function ($value) {
```
ste93cry: Please add a typehint to this function for the return type and also extract its code to a `private function` like I started doing for the other validation callables

tests/OptionsTest.php
```diff
@@ -243,4 +244,24 @@ public function excludedPathProviders()
...
+     * @dataProvider maxBreadCrumbsInvalidValuesDataProvider
+     */
+    public function testMaxBreadCrumbsInvalidValues($invalidValue): void
+    {
+        $this->expectException(InvalidOptionsException::class);
+
+        new Options(['max_breadcrumbs' => [$invalidValue]]);
+    }
+
+    public function maxBreadCrumbsInvalidValuesDataProvider()
```
ste93cry: Missing return type typehint

tests/OptionsTest.php
```diff
@@ -243,4 +244,24 @@ public function excludedPathProviders()
             [__FILE__, __FILE__],
         ];
     }
+
+    /**
+     * @dataProvider maxBreadCrumbsInvalidValuesDataProvider
+     */
+    public function testMaxBreadCrumbsInvalidValues($invalidValue): void
```
ste93cry: This test is only testing the invalid values but we should at least test the min/max allowed values which are `0` and `Options::DEFAULT_MAX_BREADCRUMBS`. I would then rename this method something like `testMaxBreadcrumbsOptionIsValidatedCorrectly` and its data provider `maxBreadcrumbsOptionIsValidatedCorrectlyDataProvider`. The `expectException` should be then wrapped in an `if` statement according to the validity of the input value

ste93cry: > it helped me spotting that the test was also invalid because it was wrapping the value inside an array

Nice I didn't even notice that 😄

PR #711, opened 2018-11-27
Cleanup SentryAuth class

Jean85:
Just a small fix


ste93cry:
Nope, this function is not useless. When the second argument is not passed to `array_filter` it removes all values that pass the `empty(...)` check. In this specific case it's useful because the `secret` could be null. I think when I coded it I wanted to avoid an `if` statement, but I'm not sure if there is some other reason

PR #713, opened 2018-11-27
[WIP] Add option to client for SDK specific identifier

Jean85:
This is a WIP to propose an alternative approach to #710. I have to add a lot of stuff (tests, same stuff in the client builder) before proceeding, but I would like to ask if this approach seems good.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -69,19 +69,26 @@ class Client implements ClientInterface
     private $representationSerializer;
 
     /**
+     * @var string|null The suffix to be appended to the SDK_IDENTIFIER in events for official integrations
+     */
+    private $sdkIntegrationIdentifier;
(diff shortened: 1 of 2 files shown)


HazAT:
Hmm, I am not sure about this.
I don't really want that the user is able to overwrite the sdk identifer when initalizing a Client, it seems odd. Also we need to be able to change the version.

What about if we do something similiar like in #710 but instead of changing ClientBuilder so deeply
we are able to pass a `ClientClass` into `ClientBuilder::create()->getClient($class)`?

I really don't want that the sdk identifer / version is a option the user can set.
I am not seeing any other simple way than inheritance of Clients.

HazAT:
I think we can close this right?

PR #714, opened 2018-11-28
[WIP] feat: Extensible Client + Options

HazAT:
This PR enables https://github.com/getsentry/sentry-laravel/pull/166 the Laravel SDK.

~~This PR also fixes `in_app` flag in frames.~~

PR #716, opened 2018-11-28
Event factory

Jean85:
~WIP, this is only a PoC for now. Please review to tell me if I should proceed further with this.~

This supersedes #714 & #713.

In a single swoop, I've refactored the client extracting the serializers and the creation of events, and made possible to inject the SDK identifier in the new `EventFactory`.

Next steps:
- [x] Instantiate the `EventFactory` inside the `ClientBuilder`
- [x] Inject the SDK identifier (and the other dependencies) into the `EventFactory`
- [x] Make the SDK identifier a constructor argument of the `Event`, into an immutable property, and use it in `Event::toArray()`
- [x] Pass the same SDK identifier from the `ClientBuilder` into `SentryAuth` for consistency
- [x] Add `ClientBuilder::setSdkIdentifierSuffix(string)` marked as internal, to be used by official integrations
- [x] Fix all the tests and add new ones where necessary

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -93,7 +93,7 @@ class Client implements ClientInterface
         if ($event = $this->prepareEvent($payload, $scope, $this->options->shouldAttachStacktrace())) {
             return $this->send($event);
         }
-    
+
         return null;
     }
 
--- a/src/Client.php
+++ b/src/Client.php
@@ -6,13 +6,8 @@ namespace Sentry;
 
 use Sentry\Integration\Handler;
 use Sentry\Integration\IntegrationInterface;
-use Sentry\Serializer\RepresentationSerializer;
-use Sentry\Serializer\RepresentationSerializerInterface;
-use Sentry\Serializer\Serializer;
-use Sentry\Serializer\SerializerInterface;
 use Sentry\State\Scope;
 use Sentry\Transport\TransportInterface;
-use Zend\Diactoros\ServerRequestFactory;
 
 /**
  * Default implementation of the {@see ClientInterface} interface.
@@ -68,33 +63,19 @@ class Client implements ClientInterface
     private $integrations;
 
     /**
-     * @var SerializerInterface The serializer
-     */
-    private $serializer;
-
-    /**
-     * @var RepresentationSerializerInterface The representation serializer
-     */
-    private $representationSerializer;
-
-    /**
      * Constructor.
      *
-     * @param Options $options The client configuration
-     * @param TransportInterface $transport The transport
-     * @param EventFactory $eventFactory
-     * @param SerializerInterface $serializer The serializer used for events
-     * @param RepresentationSerializerInterface $representationSerializer The representation serializer to be used with stacktrace frames
+     * @param Options                $options      The client configuration
+     * @param TransportInterface     $transport    The transport
+     * @param EventFactory           $eventFactory The factory for events
      * @param IntegrationInterface[] $integrations The integrations used by the client
      */
-    public function __construct(Options $options, TransportInterface $transport, EventFactory $eventFactory, SerializerInterface $serializer, RepresentationSerializerInterface $representationSerializer, array $integrations = [])
+    public function __construct(Options $options, TransportInterface $transport, EventFactory $eventFactory, array $integrations = [])
     {
         $this->options = $options;
         $this->transport = $transport;
         $this->eventFactory = $eventFactory;
         $this->integrations = Handler::setupIntegrations($integrations);
-        $this->serializer = $serializer;
-        $this->representationSerializer = $representationSerializer;
     }
 
     /**
@@ -110,14 +91,10 @@ class Client implements ClientInterface
      */
     public function captureMessage(string $message, ?Severity $level = null, ?Scope $scope = null): ?string
     {
-        $payload['message'] = $message;
-        $payload['level'] = $level;
-
-        if ($this->getOptions()->shouldAttachStacktrace()) {
-            $payload['stacktrace'] = Stacktrace::createFromBacktrace($this->getOptions(), $this->serializer, $this->representationSerializer, \debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS), __FILE__, __LINE__);
-        }
-
-        return $this->captureEvent($payload, $scope);
+        return $this->captureEvent([
+            'message' => $message,
+            'level' => $level,
+        ], $scope);
     }
 
     /**
@@ -185,6 +162,8 @@ class Client implements ClientInterface
      * @param null|Scope $scope   optional scope which enriches the Event
      *
      * @return null|Event returns ready to send Event, however depending on options it can be discarded
+     *
+     * @throws \Exception
      */
     private function prepareEvent(array $payload, ?Scope $scope = null): ?Event
     {
--- a/src/Client.php
+++ b/src/Client.php
@@ -58,6 +58,11 @@ class Client implements ClientInterface
     private $transport;
 
     /**
+     * @var EventFactory The factory to create {@see Event} from raw data
+     */
+    private $eventFactory;
+
+    /**
      * @var IntegrationInterface[] The stack of integrations
      */
     private $integrations;
@@ -75,16 +80,18 @@ class Client implements ClientInterface
     /**
      * Constructor.
      *
-     * @param Options                           $options                  The client configuration
-     * @param TransportInterface                $transport                The transport
-     * @param SerializerInterface               $serializer               The serializer used for events
+     * @param Options $options The client configuration
+     * @param TransportInterface $transport The transport
+     * @param EventFactory $eventFactory
+     * @param SerializerInterface $serializer The serializer used for events
      * @param RepresentationSerializerInterface $representationSerializer The representation serializer to be used with stacktrace frames
-     * @param IntegrationInterface[]            $integrations             The integrations used by the client
+     * @param IntegrationInterface[] $integrations The integrations used by the client
      */
-    public function __construct(Options $options, TransportInterface $transport, SerializerInterface $serializer, RepresentationSerializerInterface $representationSerializer, array $integrations = [])
+    public function __construct(Options $options, TransportInterface $transport, EventFactory $eventFactory, SerializerInterface $serializer, RepresentationSerializerInterface $representationSerializer, array $integrations = [])
     {
         $this->options = $options;
         $this->transport = $transport;
+        $this->eventFactory = $eventFactory;
         $this->integrations = Handler::setupIntegrations($integrations);
         $this->serializer = $serializer;
         $this->representationSerializer = $representationSerializer;
@@ -186,86 +193,12 @@ class Client implements ClientInterface
             return null;
         }
 
-        $event = new Event();
-        $event->setServerName($this->getOptions()->getServerName());
-        $event->setRelease($this->getOptions()->getRelease());
-        $event->getTagsContext()->merge($this->getOptions()->getTags());
-        $event->setEnvironment($this->getOptions()->getEnvironment());
-
-        if (isset($payload['transaction'])) {
-            $event->setTransaction($payload['transaction']);
-        } else {
-            $request = ServerRequestFactory::fromGlobals();
-            $serverParams = $request->getServerParams();
-
-            if (isset($serverParams['PATH_INFO'])) {
-                $event->setTransaction($serverParams['PATH_INFO']);
-            }
-        }
-
-        if (isset($payload['logger'])) {
-            $event->setLogger($payload['logger']);
-        }
-
-        $message = $payload['message'] ?? null;
-        $messageParams = $payload['message_params'] ?? [];
-
-        if (null !== $message) {
-            $event->setMessage(substr($message, 0, self::MESSAGE_MAX_LENGTH_LIMIT), $messageParams);
-        }
-
-        if (isset($payload['exception']) && $payload['exception'] instanceof \Throwable) {
-            $this->addThrowableToEvent($event, $payload['exception']);
-        }
-
-        if (isset($payload['stacktrace']) && $payload['stacktrace'] instanceof Stacktrace) {
-            $event->setStacktrace($payload['stacktrace']);
-        }
+        $event = $this->eventFactory->create($payload);
 
         if (null !== $scope) {
             $event = $scope->applyToEvent($event, $payload);
         }
 
         return \call_user_func($this->options->getBeforeSendCallback(), $event);
-    }
-
-    /**
-     * Stores the given exception in the passed event.
-     *
-     * @param Event      $event     The event that will be enriched with the exception
-     * @param \Throwable $exception The exception that will be processed and added to the event
-     */
-    private function addThrowableToEvent(Event $event, \Throwable $exception): void
-    {
-        if ($exception instanceof \ErrorException) {
-            $event->setLevel(Severity::fromError($exception->getSeverity()));
-        }
-
-        $exceptions = [];
-        $currentException = $exception;
-
-        do {
-            if ($this->getOptions()->isExcludedException($currentException)) {
-                continue;
-            }
-
-            $data = [
-                'type' => \get_class($currentException),
-                'value' => $this->serializer->serialize($currentException->getMessage()),
-            ];
-
-            $data['stacktrace'] = Stacktrace::createFromBacktrace(
-                $this->getOptions(),
-                $this->serializer,
-                $this->representationSerializer,
-                $currentException->getTrace(),
-                $currentException->getFile(),
-                $currentException->getLine()
-            );
-
-            $exceptions[] = $data;
-        } while ($currentException = $currentException->getPrevious());
-
-        $event->setExceptions($exceptions);
     }
 }
--- a/src/Client.php
+++ b/src/Client.php
@@ -86,7 +86,7 @@ class Client implements ClientInterface
         ];
 
         if ($event = $this->prepareEvent($payload, $scope, $this->options->shouldAttachStacktrace())) {
-            return $this->send($event);
+            return $this->transport->send($event);
         }
 
         return null;
@@ -108,7 +108,7 @@ class Client implements ClientInterface
     public function captureEvent(array $payload, ?Scope $scope = null): ?string
     {
         if ($event = $this->prepareEvent($payload, $scope)) {
-            return $this->send($event);
+            return $this->transport->send($event);
         }
 
         return null;
@@ -136,18 +136,6 @@ class Client implements ClientInterface
     public function getIntegration(string $className): ?IntegrationInterface
     {
         return $this->integrations[$className] ?? null;
-    }
-
-    /**
-     * Sends the given event to the Sentry server.
-     *
-     * @param Event $event The event to send
-     *
-     * @return null|string
-     */
-    private function send(Event $event): ?string
-    {
-        return $this->transport->send($event);
     }
 
     /**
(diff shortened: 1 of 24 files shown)


Jean85:
I can try to solve this ASAP, so we don't delay this further.

For the naming part, it's using two different names because the `ClientBuilder` is [a different design pattern](https://en.wikipedia.org/wiki/Builder_pattern) from the factory, since it has multiple methods to configure how the client will be composed. The `EventFactory` instead has just a straight method that encapsulate a semi-rigid logic, without needing further instructions.

Jean85:
No longer WIP, this should be complete now! Please review.

src/EventFactory.php
```diff
@@ -0,0 +1,141 @@
...
+        $message = $payload['message'] ?? null;
+        $messageParams = $payload['message_params'] ?? [];
+
+        if (null !== $message) {
+            $event->setMessage(substr($message, 0, Client::MESSAGE_MAX_LENGTH_LIMIT), $messageParams);
+        }
+
+        if (isset($payload['exception']) && $payload['exception'] instanceof \Throwable) {
+            $this->addThrowableToEvent($event, $payload['exception']);
+        } elseif ($this->options->shouldAttachStacktrace()) {
```
HazAT: We have to be careful here, the docs say that this really only works for `captureMessage`. In theory if someone calls `captureEvent` this will also trigger.
So maybe this has to go back into `captureMessage`.

Jean85: That's not an issue here. There is only `captureMessage` and `captureException`, and the latter sets the `exception` key, so it will trigger the first `if` branch.

`captureLastError` goes through `captureException` too, and `captureEvent` would behave like `captureMessage`.

Jean85:
Answering point by point:

> I don't really like that $sdkIdentifer is a constructor parameter for an Event. Imagine a User creating an event and calling captureEvent which should be perfectly fine. How should the User know what to put there? I would not make it a ctor parameter instead, use a setter and set it in the event factory. It's just annoying that you have to set it when creating an event.

The `Event` is in fact an internal object, it's not even used in `captureEvent`, which accepts an array, not `Event`. Maybe we should mark `Event` as `@internal` then?

> We can't set the version yet, this is basically similar to the identifier. It's perfectly fine for an "sub" SDK to have it's own version. So maybe we also need to set this.

Umh, that seems trick to me. We would miss the base SDK version then!

> We have to be able to extend the options, we can solve this in a later PR though but I think we need it. E.g.: Laravel has sql_breadcrumbs as an option, I can't tell the ClientBuilder to use the laravel options instead of the default Client Options.

Umh, an other tricky part, but probably out of scope for this PR. We should probably move the instantiation of `Options` outside the `ClientBuilder` then, maybe move it to the `init`, so the integrations would avoid using it and ho straight to the `ClientBuilder`.

src/EventFactory.php
ste93cry: >`captureEvent` would behave like `captureMessage`

If I don't recall wrong it means that `captureMessage` will use the `shouldAttachStacktrace` option which is wrong

Jean85: Moving that piece of code back would mean loosing the refactoring in term of deps splitting. I'll think about it a bit more...

HazAT:
@Jean85 

### Event

I still think having `sdkIdentifier` as ctor argument is no ideal, why should the event / the one who creates the event be responsible for knowing / setting the sdkIdentifier?
I really don't see the benefit of adding it there, I would really go back to setter.

---

### Version

>Umh, that seems trick to me. We would miss the base SDK version then!

We added this additional information into SDK which is not used yet but some SDKs already send it (like JS). See example below:

```

{
  name: "sentry.php.laravel",
  version: "x.x.x",
  packages: [
    {
      name: "composer:sentry-php",
      version: "x.x.x"
    },
    {
      name: "composer:sentry-larvel",
      version: "x.x.x"
    }
  ]
},
```

Version of sentry-laravel is ambiguous but we can't change this for now, since `package` is completly optional.

---

### Options

It would help to inject an `Options` object into ClientBuilder instead of passing an array and creating options in there. I agree lets not do this in this PR but we still need to make some stuff in `Options` protected instead of private, then this should just work fine.

src/Client.php
```diff
@@ -103,14 +86,10 @@ public function getOptions(): Options
...
-
-        if ($this->getOptions()->shouldAttachStacktrace()) {
-            $payload['stacktrace'] = Stacktrace::createFromBacktrace($this->getOptions(), $this->serializer, $this->representationSerializer, \debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS), __FILE__, __LINE__);
-        }
-
-        return $this->captureEvent($payload, $scope);
+        return $this->captureEvent([
+            'message' => $message,
+            'level' => $level,
+        ], $scope);
```
HazAT: What if you do something like:

```
$payload = ['message' => $message, 'level' => $level];
if ($this->getOptions()->shouldAttachStacktrace()) {
  $payload['stacktrace'] = Stacktrace::createFromBacktrace($this->getOptions(), $this->serializer, $this->representationSerializer, \debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS), __FILE__, __LINE__);
}
return $this->captureEvent($payload, $scope);
```

Jean85:
* `Event` and identifier in setter, not ctor: no problem
 * Version part: the `packages` part seems interesting, but at that point it would be better to use info from composer instead of leaving that to the user/lib; IMHO we should use the same stuff used in `ModulesIntegration`:

https://github.com/getsentry/sentry-php/blob/f01a5fe22034cf75d1ba7d7dcd23f79856e104ef/src/Integration/ModulesIntegration.php#L56-L58

We would have just to filter for `sentry/*` packages and list them all! It would be easy,straightforward and not open to modifications by the end user. WDYT?

ste93cry:
I've looked quickly at the code and I don't see a real usefulness in this change. I like the fact that we are trying to separate the code that instantiates and populates the data of an event from the client but as is this code is pretty limited because the factory is `final` and doesn't implement an interface. This means that a user cannot extend/decorate/change the implementation of the factory we provide and there is anyway no way to set it from the outside of the client builder: basically it seems to me we just moved around some code with no real benefit apart from hiding a few dependencies into another class instead of the `Client` class. Also I can't see how this is going to help us with setting the SDK identifier attribute of the event. I would rather suggest to have a default value set directly in the constructor of the event and then provide a `setter` that allows to override the attribute like any other one and use it from each specific SDK (laravel, symfony, etc) through an integration. The only thing that would be left is to think how to change the `User-Agent` in the authentication header, even though I strongly believe it shouldn't be there

Jean85:
As for making the factory `final`, it is intended: if the end user wants to alter the `Event`, there's the `before_send` callback already.

As for the setter, it's the same suggestions as Daniel, so I agree. Will do ASAP.

src/Client.php
```diff
@@ -102,14 +85,16 @@ public function getOptions(): Options
...
-        $payload['level'] = $level;
+        $payload = [
+            'message' => $message,
+            'level' => $level,
+        ];
 
-        if ($this->getOptions()->shouldAttachStacktrace()) {
-            $payload['stacktrace'] = Stacktrace::createFromBacktrace($this->getOptions(), $this->serializer, $this->representationSerializer, \debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS), __FILE__, __LINE__);
+        if ($event = $this->prepareEvent($payload, $scope, $this->options->shouldAttachStacktrace())) {
+            return $this->send($event);
```
ste93cry: Since the `send` function is just a small method that just calls in turn the transport we can inline its code here

src/ClientBuilder.php
```diff
@@ -219,6 +224,18 @@ public function setRepresentationSerializer(RepresentationSerializerInterface $r
         return $this;
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function setSdkIdentifierSuffix(?string $sdkIdentifierSuffix): void
```
ste93cry: I don't think we should expose a method to set just the suffix on the SDK identifier. I would prefer to be able to set the whole name (there is no server-side validation regarding this value).

src/ClientBuilder.php
```diff
@@ -305,4 +320,12 @@ private function createTransportInstance(): TransportInterface
 
         return new HttpTransport($this->options, $this->createHttpClientInstance(), $this->messageFactory);
     }
+
+    private function createEventFactory(): EventFactory
```
ste93cry: Missing docblock with the description of what this function does

src/ClientBuilderInterface.php
```diff
@@ -108,4 +108,9 @@ public function setSerializer(SerializerInterface $serializer);
      * @return $this
      */
     public function setRepresentationSerializer(RepresentationSerializerInterface $representationSerializer);
+
+    /**
+     * @param string|null $sdkIdentifierSuffix The suffix for the SDK identifier, to be used by official SDK integrations
```
ste93cry: Missing the description of what this method does (something can be taken from the description of the parameter that can then be reduced)

src/EventFactory.php
```diff
@@ -0,0 +1,161 @@
...
+
+declare(strict_types=1);
+
+namespace Sentry;
+
+use Sentry\Serializer\RepresentationSerializerInterface;
+use Sentry\Serializer\SerializerInterface;
+use Zend\Diactoros\ServerRequestFactory;
+
+final class EventFactory
```
ste93cry: Missing the description of what this class does

src/EventFactory.php
```diff
@@ -0,0 +1,161 @@
...
+    public function __construct(SerializerInterface $serializer, RepresentationSerializerInterface $representationSerializer, Options $options, string $sdkIdentifier)
+    {
+        $this->serializer = $serializer;
+        $this->representationSerializer = $representationSerializer;
+        $this->options = $options;
+        $this->sdkIdentifier = $sdkIdentifier;
+    }
+
+    /**
+     * @param array $payload The data to be attached to the Event
```
ste93cry: Missing the description of what this method does

src/EventFactory.php
```diff
@@ -0,0 +1,161 @@
...
+        if (!$event->getStacktrace()) {
+            $stacktrace = Stacktrace::createFromBacktrace($this->options, $this->serializer, $this->representationSerializer, \debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS), __FILE__, __LINE__);
+            $event->setStacktrace($stacktrace);
+        }
+
+        return $event;
+    }
+
+    /**
+     * @param array $payload The data to be attached to the Event
```
ste93cry: Missing the description of what this method does

src/EventFactory.php
```diff
@@ -0,0 +1,161 @@
...
+     *
+     * @return Event
+     */
+    public function createWithStacktrace(array $payload): Event
+    {
+        $event = $this->create($payload);
+
+        if (!$event->getStacktrace()) {
+            $stacktrace = Stacktrace::createFromBacktrace($this->options, $this->serializer, $this->representationSerializer, \debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS), __FILE__, __LINE__);
+            $event->setStacktrace($stacktrace);
```
ste93cry: Add a blank line above this one to improve readability

tests/EventFactoryTest.php
```diff
@@ -0,0 +1,175 @@
...
+            [
+                'type' => \Exception::class,
+                'value' => 'testMessage',
+            ],
+            [
+                'type' => \RuntimeException::class,
+                'value' => 'testMessage2',
+            ],
+        ];
+        $this->assertArraySubset($expectedData, $event->getExceptions());
```
ste93cry: Add a blank line above this one to improve readability

tests/EventFactoryTest.php
```diff
@@ -0,0 +1,175 @@
...
+            $this->createMock(SerializerInterface::class),
+            $this->createMock(RepresentationSerializerInterface::class),
+            $options,
+            'sentry.sdk.identifier'
+        );
+
+        $event = $eventFactory->createWithStacktrace([]);
+
+        $stacktrace = $event->getStacktrace();
+        $this->assertInstanceOf(Stacktrace::class, $stacktrace);
```
ste93cry: Add blank line above this line to improve readability

tests/EventFactoryTest.php
```diff
@@ -0,0 +1,175 @@
...
+        /** @var Frame $lastFrame */
+        $lastFrame = array_reverse($stacktrace->getFrames())[0];
```
ste93cry: Add blank lines around this line to improve readability

src/Client.php
```diff
@@ -58,35 +48,28 @@ class Client implements ClientInterface
...
-     * @param Options                           $options                  The client configuration
-     * @param TransportInterface                $transport                The transport
-     * @param SerializerInterface               $serializer               The serializer used for events
-     * @param RepresentationSerializerInterface $representationSerializer The representation serializer to be used with stacktrace frames
+     * @param Options            $options      The client configuration
+     * @param TransportInterface $transport    The transport
+     * @param EventFactory       $eventFactory The factory for events
      */
-    public function __construct(Options $options, TransportInterface $transport, SerializerInterface $serializer, RepresentationSerializerInterface $representationSerializer)
+    public function __construct(Options $options, TransportInterface $transport, EventFactory $eventFactory)
```
ste93cry: I don't like at all that we're passing to the client a dependency that can't be mocked because it's `final` and not has an interface

src/EventFactory.php
```diff
@@ -54,11 +54,12 @@ public function __construct(SerializerInterface $serializer, RepresentationSeria
     public function create(array $payload): Event
     {
         try {
-            $event = new Event($this->sdkIdentifier);
+            $event = new Event();
         } catch (\Throwable $error) {
             throw new \RuntimeException('Unable to instantiate an event', 0, $error);
```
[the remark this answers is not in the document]
Jean85: Changing this to a more specific exception to make catching easier if needed.

src/Client.php
Jean85: I've avoided making an interface for that because I wouldn't like the end user extending that. The factory in itself does just instantiation and population of the event, it shouldn't be an issue not mocking it.

ste93cry: My thinking is that this is a dependency, so it should be mockable (you have otherwise to mock all its dependencies). Also a class should be `final` when it implements an interface or is a VO

src/Event.php
```diff
@@ -133,15 +133,13 @@
...
-     *
      * @throws UnsatisfiedDependencyException if `Moontoast\Math\BigNumber` is not present
      * @throws \InvalidArgumentException
      * @throws \Exception
      */
-    public function __construct(string $sdkIdentifier)
+    public function __construct()
     {
-        $this->sdkIdentifier = $sdkIdentifier;
+        $this->sdkIdentifier = Client::SDK_IDENTIFIER;
```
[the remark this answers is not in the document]
ste93cry: This conversation has been marked as resolved without being really resolved

src/Event.php
```diff
@@ -154,6 +166,58 @@ public function getId(): string
...
+     * Gets the version of the SDK package that generated this Event.
+     *
+     * @return string
+     *
+     * @internal
+     */
+    public function getSdkVersion(): string
+    {
+        if (null === $this->sdkVersion) {
+            $this->sdkVersion = PrettyVersions::getVersion('sentry/sentry')->getPrettyVersion();
```
ste93cry: I don't understand why this exact code is duplicated in the `ClientBuilder` class. I thought that the purpose of the event factory was also to set this attribute, we should not have this logic inside the `Event` class (the attribute is not even required so it can be left `nullable` without any sort of problem)

src/Exception/EventCreationException.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Sentry\Exception;
+
+use Throwable;
+
+class EventCreationException extends \RuntimeException implements ExceptionInterface
```
ste93cry: Missing the description of this class

src/Exception/EventCreationException.php
```diff
@@ -0,0 +1,13 @@
...
+{
+    public function __construct(Throwable $previous)
```
ste93cry: Missing docblock of this method

tests/ClientBuilderTest.php
```diff
@@ -246,6 +248,26 @@ public function optionsDataProvider(): array
         ];
     }
 
+    public function testSdkIdentifierAndVersionDefaultValues(): void
```
ste93cry: Can we rename this test to `testClientBuilderFallbacksToDefaultSdkIdentifierAndVersion` please? Also it seems to me that there is a missing test that assert that those values can be overwritten using the setters

tests/ClientTest.php
```diff
@@ -58,48 +59,55 @@ public function testTransactionEventAttributeIsNotPopulatedInCli()
...
-            ->disableOriginalConstructor()
-            ->setMethodsExcept(['captureException'])
-            ->getMock();
+        /** @var TransportInterface|MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(function (Event $event) use ($exception): bool {
+                $this->assertCount(1, $event->getExceptions());
+                $exceptionData = $event->getExceptions()[0];
```
ste93cry: Add some blank lines around this line to improve readability of the code

tests/EventFactoryTest.php
```diff
@@ -0,0 +1,187 @@
...
+        $eventFactory = new EventFactory(
+            $this->createMock(SerializerInterface::class),
+            $this->createMock(RepresentationSerializerInterface::class),
+            $options,
+            'sentry.sdk.identifier',
+            '1.2.3'
+        );
+
+        $event = $eventFactory->createWithStacktrace([]);
+
```
ste93cry: Remove this blank line

tests/EventFactoryTest.php
```diff
@@ -0,0 +1,187 @@
...
+        foreach ($event->getExceptions() as $exceptionData) {
+            $this->assertInstanceOf(Stacktrace::class, $exceptionData['stacktrace']);
+        }
+    }
+
+    public function testCreateWithErrorException(): void
+    {
+        $exception = new \ErrorException('testMessage', 0, E_USER_ERROR);
+
```
ste93cry: Remove this blank line

tests/EventFactoryTest.php
```diff
@@ -0,0 +1,187 @@
...
+        $eventFactory = new EventFactory(
+            new Serializer(),
+            $this->createMock(RepresentationSerializerInterface::class),
+            new Options(),
+            'sentry.sdk.identifier',
+            '1.2.3'
+        );
+
+        $event = $eventFactory->create(['exception' => $exception]);
+
```
ste93cry: Remove this blank line

tests/EventFactoryTest.php
```diff
@@ -0,0 +1,187 @@
...
+        $event = $eventFactory->create([]);
+
+        $this->assertNull($event->getTransaction());
+    }
+
+    public function testCreateWithException(): void
+    {
+        $previousException = new \RuntimeException('testMessage2');
+        $exception = new \Exception('testMessage', 0, $previousException);
+
```
ste93cry: Remove this blank line

src/Event.php
Jean85: I left it here to provide a sensible default. If this is not required, I can leave it as nullable and remove this piece. @HazAT WDYT?

PR #717, opened 2018-12-03
feat: Make options extensible

HazAT:
This makes it possible for "sub" SDKs to have their own options while still using the `ClientBuilder`.

### DIFF
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -115,11 +115,11 @@ final class ClientBuilder implements ClientBuilderInterface
     /**
      * Class constructor.
      *
-     * @param array $options The client options
+     * @param Options $options The client options
      */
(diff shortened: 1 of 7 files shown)


ste93cry:
As a general comment I think we are complicating a bit too much the instantiation of the client builder that now requires an `Options` object instead of a plain array. For example what's the point now to proxy the getters/setters of the config object through the client builder if the user has to pass the instance from the outside?. Probably if this is the way we want to take then adding at least a `createFromArray` method would ease the instantiation of the builder. Said this, I'm against coupling the configuration/code of the base SDK with the `sentry-laravel` or whatever other integration we have/will develop in the future as it's not something that can scale well. I'm not sure about what options we need for each integration but probably it's something that doesn't belong to the base SDK anyway

PR #718, opened 2018-12-04
feat: Add default_integrations option

HazAT:
I realized while writing the docs that just `integrations` allowing `null` isn't enough.
Added `default_integrations` : `bool` and restricted `integrations` option to `IntegrationInterface[]`.

### DIFF
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -64,7 +64,7 @@ use Sentry\Transport\TransportInterface;
  * @method setTags(string[] $tags)
  * @method bool shouldSendDefaultPii()
  * @method setSendDefaultPii(bool $enable)
- * @method bool getDefaultIntegrations()
+ * @method bool hasDefaultIntegrations()
  * @method setDefaultIntegrations(bool $enable)
  */
 final class ClientBuilder implements ClientBuilderInterface
@@ -123,13 +123,13 @@ final class ClientBuilder implements ClientBuilderInterface
     {
         $this->options = new Options($options);
 
-        if ($this->options->getDefaultIntegrations()) {
+        $this->integrations = $this->options->getIntegrations();
+
+        if ($this->options->hasDefaultIntegrations()) {
             $this->integrations = \array_merge([
                 new ErrorHandlerIntegration(),
                 new RequestIntegration($this->options),
             ], $this->options->getIntegrations());
-        } else {
-            $this->integrations = $this->options->getIntegrations();
         }
     }
 
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -20,7 +20,6 @@ use Http\Message\MessageFactory;
 use Http\Message\UriFactory;
 use Sentry\HttpClient\Authentication\SentryAuth;
 use Sentry\Integration\ErrorHandlerIntegration;
-use Sentry\Integration\IntegrationInterface;
 use Sentry\Integration\RequestIntegration;
 use Sentry\Serializer\RepresentationSerializer;
 use Sentry\Serializer\RepresentationSerializerInterface;
@@ -110,11 +109,6 @@ final class ClientBuilder implements ClientBuilderInterface
     private $representationSerializer;
 
     /**
-     * @var IntegrationInterface[] List of default integrations
-     */
-    private $integrations = [];
-
-    /**
      * Class constructor.
      *
      * @param array $options The client options
@@ -123,13 +117,11 @@ final class ClientBuilder implements ClientBuilderInterface
     {
         $this->options = new Options($options);
 
-        $this->integrations = $this->options->getIntegrations();
-
         if ($this->options->hasDefaultIntegrations()) {
-            $this->integrations = \array_merge([
+            $this->options->setIntegrations(\array_merge([
                 new ErrorHandlerIntegration(),
                 new RequestIntegration($this->options),
-            ], $this->options->getIntegrations());
+            ], $this->options->getIntegrations()));
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/ClientBuilder.php
```diff
@@ -121,11 +123,13 @@ public function __construct(array $options = [])
     {
         $this->options = new Options($options);
 
-        if (null !== $this->options->getIntegrations()) {
+        if ($this->options->getDefaultIntegrations()) {
```
ste93cry: Instead of having a pair of `if`/`else` maybe we can improve the readability of the code a bit by editing it like the following one:

```php
$this->integrations = $this->options->getIntegrations();

if ($this->options->getDefaultIntegrations()) {
    // Note: does it matter if default integrations are registered before the custom ones?
    $this->integrations = array_merge($this->integrations, [...]);
}
```

Note that you should check it still covers the cases you're trying to fix

src/Options.php
```diff
@@ -636,8 +659,9 @@ private function configureOptions(OptionsResolver $resolver): void
         $resolver->setAllowedTypes('error_types', ['int']);
         $resolver->setAllowedTypes('max_breadcrumbs', 'int');
         $resolver->setAllowedTypes('before_breadcrumb', ['callable']);
-        $resolver->setAllowedTypes('integrations', ['null', 'array']);
+        $resolver->setAllowedTypes('integrations', 'array');
         $resolver->setAllowedTypes('send_default_pii', 'bool');
+        $resolver->setAllowedTypes('default_integrations', 'bool');
```
ste93cry: I think that the name of this option cames from the unified API so this suggestion is not related to this particular change/package: wouldn't be better and more clear to call it `use_default_integrations`?

ste93cry:
As a side note: I think that the name of this option comes from the unified API so this suggestion is not related to this particular PR/package: wouldn't be better and more clear to call it `use_default_integrations`? The current name seems to indicate that it's a list of integrations instead of a flag toggling whether the default integrations are automatically registered or not

src/ClientBuilder.php
HazAT: Changed, it's better to be initialized after the default ones in case your custom integration wants to change something.

HazAT:
Good point, `use_default_integrations` seems to be more fitting to what it does. It is in the unified API though, we will put this on our things to discuss.

tests/ClientBuilderTest.php
```diff
@@ -151,6 +151,26 @@ public function testGetClient(): void
         $this->assertAttributeSame($this->getObjectAttribute($clientBuilder, 'httpClient'), 'client', $httpClientPlugin);
     }
 
+    public function testClientHasDefaultIntegration(): void
```
ste93cry: Rethinking about this test it seems to me that it's not as useful as it should since it's only checking that the list of integrations is empty/not empty but not whether it contains the right list of integrations. We should change the code to something like:

`$this->assertEquals($expectedIntegrations, $client->getOptions()->getIntegrations(), '', 0.0, 10, e)`

HazAT: What about `$this->assertEquals(\array_values($client->getOptions()->getIntegrations()), \array_values($integrations));` to not overcomplicate things?

ste93cry: I agree that what I wrote is not user-friendly but we don't have named parameters in PHP unfortunately, so to use the last argument of the function `assertEquals` we must specify all the arguments that comes before too. By the way I wrote the code that way because that `boolean` parameter normalizes both the input and expected arrays by sorting them so that the test doesn't fail if their values are not in the same exact order. `array_values` instead returns all the values of an array reindexing but not sorting them. We should use `sort` on both arrays but at this point I would simply use the convenience of the `assertEquals` parameters to leave to PHPUnit the burden of normalization

Jean85: Maybe doing the asserts in a foreach is easier?

ste93cry: We would still keep track of the fact that some keys may exists in an array but not in the other (so the test must fail) and viceversa. I don't think it's worth to create a custom code to do the assertion when PHPUnit already provides the way to do it even though it's not user friendly, we just have to write some readable code that uses its methods, e.g.

```php
$integrationsClassNames = array_map('get_class', $client->getOptions()->getIntegrations());

$this->assertEquals($expectedIntegrationsClassNames, $integrationsClassNames, '', 0, 10, true)
```

tests/ClientBuilderTest.php
```diff
@@ -151,6 +151,26 @@ public function testGetClient(): void
...
+    {
+        $clientBuilder = new ClientBuilder(['dsn' => 'http://public:secret@example.com/sentry/1']);
+        $client = $clientBuilder->getClient();
+
+        $integrations = $this->getObjectAttribute($client, 'integrations');
+
+        $this->assertNotEmpty($integrations);
+    }
+
+    public function testClientHasNoDefaultIntegration(): void
```
ste93cry: We can remove this test and just leave the one above this that should use a data provider that defines the list of expected integrations, the list of custom integrations and the value of the `default_integrations` option. This way we can build a matrix of test cases that should be similar to this one:

| `default_integrations` option | `integrations` option | expected integrations                     |
|-------------------------------|-----------------------|-------------------------------------------|
| `false`                       | `[]`                  | `[]`                                      |
| `false`                       | `[Foo::class]`        | `[Foo::class]`                            |
| `true`                        | `[]`                  | `[DefaultIntegration::class]`             |
| `true`                        | `[Foo::class]`        | `[DefaultIntegration::class, Foo::class]` |

To get the list of integrations from the client we should use `$client->getOptions()->getIntegrations()` instead of fetching the list from the `private` variable using the reflection since we are now sure that after the recent changes what is in the options is the only source of truth for the list of integrations

PR #719, opened 2018-12-04
Event factory with packages

Jean85:
THIS IS A WIP/PoC, it starts from #716 (and targets that PR, not the base branch).

This is to show how I would like to address https://github.com/getsentry/sentry-php/pull/716#issuecomment-443620896, so we can send more reliable info about SDK and integrations package versions. This would work out of the box in official integrations too.

I'm thinking about adding an official maps of integrations, so we could have something like:
```php
switch($packageName) {
    case 'sentry/sentry':
        return 'sentry.php';
    case 'sentry/symfony':
        return 'sentry.php.symfony';
```

etc. WDYT?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,7 @@
         "php": "^7.1",
         "ext-json": "*",
         "ext-mbstring": "*",
+        "jean85/pretty-package-versions": "^1.2",
         "php-http/async-client-implementation": "^1.0",
         "php-http/client-common": "^1.5",
         "php-http/discovery": "^1.2",
@@ -26,7 +27,6 @@
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.13",
-        "jean85/pretty-package-versions": "^1.2",
         "monolog/monolog": "^1.3",
         "php-http/curl-client": "^1.7.1",
         "php-http/message": "^1.5",
@@ -35,9 +35,6 @@
         "phpstan/phpstan": "^0.10.3",
         "phpunit/phpunit": "^7.0",
         "symfony/phpunit-bridge": "^4.1.6"
-    },
-    "suggest": {
-        "jean85/pretty-package-versions": "To use the ModulesIntegration, that lists all installed dependencies in every sent event"
     },
     "conflict": {
         "php-http/client-common": "1.8.0",
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -18,6 +18,7 @@ use Http\Discovery\MessageFactoryDiscovery;
 use Http\Discovery\UriFactoryDiscovery;
 use Http\Message\MessageFactory;
 use Http\Message\UriFactory;
+use Jean85\PrettyVersions;
 use Sentry\HttpClient\Authentication\SentryAuth;
 use Sentry\Integration\ErrorHandlerIntegration;
 use Sentry\Integration\RequestIntegration;
@@ -111,7 +112,12 @@ final class ClientBuilder implements ClientBuilderInterface
     /**
      * @var string The SDK identifier, to be used in {@see Event} and {@see SentryAuth}
      */
-    private $sdkIdentifier = Client::SDK_IDENTIFIER;
+    private $sdkIdentifier;
+
+    /**
+     * @var string the SDK version of the Client
+     */
+    private $sdkVersion;
 
     /**
      * Class constructor.
@@ -128,6 +134,8 @@ final class ClientBuilder implements ClientBuilderInterface
                 new RequestIntegration($this->options),
             ], $this->options->getIntegrations()));
         }
+
+        $this->sdkIdentifier = Client::SDK_IDENTIFIER;
     }
 
     /**
@@ -227,13 +235,29 @@ final class ClientBuilder implements ClientBuilderInterface
     /**
      * {@inheritdoc}
      */
-    public function setSdkIdentifierSuffix(?string $sdkIdentifierSuffix): void
+    public function setSdkIdentifier(string $sdkIdentifier): void
     {
-        $this->sdkIdentifier = Client::SDK_IDENTIFIER;
+        $this->sdkIdentifier = $sdkIdentifier;
+    }
 
-        if ($sdkIdentifierSuffix) {
-            $this->sdkIdentifier .= '.' . $sdkIdentifierSuffix;
+    /**
+     * @return string
+     */
+    private function getSdkVersion(): string
+    {
+        if (null === $this->sdkVersion) {
+            $this->sdkVersion = PrettyVersions::getVersion('sentry/sentry')->getPrettyVersion();
         }
+
+        return $this->sdkVersion;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setSdkVersion(string $sdkVersion): void
+    {
+        $this->sdkVersion = $sdkVersion;
     }
 
     /**
@@ -287,8 +311,8 @@ final class ClientBuilder implements ClientBuilderInterface
             $this->addHttpClientPlugin(new BaseUriPlugin($this->uriFactory->createUri($this->options->getDsn())));
         }
 
-        $this->addHttpClientPlugin(new HeaderSetPlugin(['User-Agent' => $this->sdkIdentifier . '/' . Client::VERSION]));
-        $this->addHttpClientPlugin(new AuthenticationPlugin(new SentryAuth($this->options, $this->sdkIdentifier)));
+        $this->addHttpClientPlugin(new HeaderSetPlugin(['User-Agent' => $this->sdkIdentifier . '/' . $this->getSdkVersion()]));
+        $this->addHttpClientPlugin(new AuthenticationPlugin(new SentryAuth($this->options, $this->sdkIdentifier, $this->getSdkVersion())));
         $this->addHttpClientPlugin(new RetryPlugin(['retries' => $this->options->getSendAttempts()]));
         $this->addHttpClientPlugin(new ErrorPlugin());
 
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -243,7 +243,7 @@ final class ClientBuilder implements ClientBuilderInterface
     }
 
     /**
-     * @return string
+     * @return string Sets the SDK version to be passed onto {@see Event} and HTTP User-Agent header
      */
     private function getSdkVersion(): string
     {
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -66,6 +66,8 @@ use Sentry\Transport\TransportInterface;
  * @method setSendDefaultPii(bool $enable)
  * @method bool hasDefaultIntegrations()
  * @method setDefaultIntegrations(bool $enable)
+ * @method callable getBeforeSendCallback()
+ * @method setBeforeSendCallback(callable $beforeSend)
  */
 final class ClientBuilder implements ClientBuilderInterface
 {
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -246,7 +246,7 @@ final class ClientBuilder implements ClientBuilderInterface
     private function getSdkVersion(): string
     {
         if (null === $this->sdkVersion) {
-            $this->sdkVersion = PrettyVersions::getVersion('sentry/sentry')->getPrettyVersion();
+            $this->setSdkVersionByPackageName('sentry/sentry');
         }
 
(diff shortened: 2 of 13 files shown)


ste93cry:
Not that I don't like the purpose of this PR, but making the main SDK library aware of other packages that exists in the world (e.g. the `sentry-laravel` package) doesn't seems to me a great idea. I would rather expose as usual some getters and setters and then enrich the event using integrations specific for each framework. The main reason is that the other SDKs are aware of this package because they depend from it to work but the opposite is not true

Jean85:
The point of making it aware is to avoid modifications from end users to use non-official SDK identifiers; IIRC, @HazAT told us that it would be better to lock down that; if that isn't true, we can avoid that stuff.

Also, it would be helpful to "elect" integrations as official in the base SDK.

ste93cry:
But as a user I could have extended the SDK to make a customized version and I may want to report this to Sentry (afaik the details of the SDK are shown in the event details). I don't see any reason why I should need a custom SDK, but it may happen so the attribute should not be locked down. Plus there is no real way to prevent the setter from being called, putting an annotation won't stop users from using it

Jean85:
We should ask this to @HazAT, because I don't know if that value is parsed and/or validated in any way on the server side.

HazAT:
We do not parse or validate this value there, so there is no need of locking it down. Also if someone wants to send something else here we can't prevent them from doing so. I am more on @ste93cry that it's probably not the best idea to have all "sub" sdks referenced in the main sdk.

composer.json
```diff
@@ -15,6 +15,7 @@
         "php": "^7.1",
         "ext-json": "*",
         "ext-mbstring": "*",
+        "jean85/pretty-package-versions": "^1.2",
```
Jean85: This is now a root requirement. It has the side-effect of making the `ModulesIntegration` work out-of-the-box, and it could be also leveraged by other integrations to set the version in a faster and more reliable way too.

src/ClientBuilder.php
```diff
@@ -115,7 +116,12 @@
     /**
      * @var string The SDK identifier, to be used in {@see Event} and {@see SentryAuth}
      */
-    private $sdkIdentifier = Client::SDK_IDENTIFIER;
+    private $sdkIdentifier;
+
+    /**
+     * @var string the SDK version of the Client
+     */
+    private $sdkVersion;
```
Jean85: All handling of identifier and version is now moved here, into the `ClientBuilder`

src/ClientBuilder.php
```diff
@@ -231,13 +239,29 @@ public function setRepresentationSerializer(RepresentationSerializerInterface $r
...
 
-        if ($sdkIdentifierSuffix) {
-            $this->sdkIdentifier .= '.' . $sdkIdentifierSuffix;
+    /**
+     * @return string
+     */
+    private function getSdkVersion(): string
+    {
+        if (null === $this->sdkVersion) {
+            $this->sdkVersion = PrettyVersions::getVersion('sentry/sentry')->getPrettyVersion();
```
Jean85: This is just a lazy way of fetching the default value. Notice that his getter is private, it's just an internal implementation detail.

src/ClientBuilder.php
```diff
@@ -64,6 +65,8 @@
  * @method setTags(string[] $tags)
  * @method bool shouldSendDefaultPii()
  * @method setSendDefaultPii(bool $enable)
+ * @method callable getBeforeSendCallback()
```
Jean85: This got out-of-sync to the `Options` definition, so I got unjustified PHPStan errors here.

IMHO we should ditch this `__call` magic, and use the `Options` object directly. As discussed elsewhere, I would like to pass the `Options` in the ctor, so the naked array requìred only with the `init()`. @ste93cry @HazAT WDYT?

src/Event.php
```diff
@@ -514,6 +552,24 @@ public function setStacktrace(Stacktrace $stacktrace): void
         $this->stacktrace = $stacktrace;
     }
 
+    private static function getSdkPackages(): array
+    {
+        if (null === self::$sdkPackages) {
+            self::$sdkPackages = [];
+
+            foreach (Versions::VERSIONS as $package => $version) {
+                if (0 === strpos($package, 'sentry/')) {
```
Jean85: This is another nice side-effect: all Sentry official packages will be listed in the `sdk.packages` field of the event, with precise versions (up to tag or commit) and very little overhead.

src/ClientBuilder.php
ste93cry: I'm +1 on this change, my intentions of making the builder a proxy for the options did not bring the desired results so let's remove everything and simplify things. I would just suggest to do this kind of change in a separate PR since it's not the main purpose of this one

src/Event.php
HazAT: tbh I am not happy that this logic lives inside of the `Event`, the `Event` shouldn't know anything about packages or how/where to fetch them.
It should rather be a simple object holding data.

Also a side note the `sdk.packages` has no prio, we don't do anything with it yet. 
We discard it on the server right now.
While we probably want to have this at some point we don't need it now.

composer.json
HazAT: I don't have anything against your package but I really don't want to add another required dep for the SDK to work.
We already have this `php composer.phar require sentry/sentry php-http/curl-client guzzlehttp/psr7`
Ideally we would only have `sentry/sentry`.

src/Event.php
```diff
@@ -528,7 +584,8 @@ public function toArray(): array
             'platform' => 'php',
             'sdk' => [
                 'name' => $this->sdkIdentifier,
-                'version' => Client::VERSION,
+                'version' => $this->getSdkVersion(),
+                'packages' => self::getSdkPackages(),
```
HazAT: See comment above.

composer.json
Jean85: This isn't to be added anywhere while installing, is required so it would be implied in the `composer require sentry/sentry`. The other package is needed to specify which HTTP client to use, since it's up to the user to chose that.

src/Event.php
Jean85: Ok I'll discard it for now, we can still add it later. BTW it could be even moved inside the `EventFactory` to avoid logic inside the `Event`.

src/Event.php
```diff
@@ -4,6 +4,8 @@
 
 namespace Sentry;
 
+use Jean85\PrettyVersions;
```
HazAT:
```suggestion
```

src/Event.php
```diff
@@ -4,6 +4,8 @@
...
+use PackageVersions\Versions;
```
HazAT:
```suggestion
```

src/Event.php
```diff
@@ -178,6 +187,30 @@ public function setSdkIdentifier(string $sdkIdentifier): void
...
+            $this->sdkVersion = PrettyVersions::getVersion('sentry/sentry')->getPrettyVersion();
+        }
+
+        return $this->sdkVersion;
+    }
+
+    /**
+     * Sets the version of the SDK package that generated this Event.
+     *
+     * @param string $sdkVersion
```
HazAT:
```suggestion
     * @param string $sdkVersion
     *
     * @internal
```

Let's make these functions internal.

src/Event.php
```diff
@@ -178,6 +187,30 @@ public function setSdkIdentifier(string $sdkIdentifier): void
         $this->sdkIdentifier = $sdkIdentifier;
     }
 
+    /**
+     * Gets the version of the SDK package that generated this Event.
+     *
+     * @return string
```
HazAT:
```suggestion
     * @return string
     *
     * @internal
```
Let's make these functions internal.

src/Event.php
```diff
@@ -161,6 +168,8 @@ public function getId(): string
     }
 
     /**
+     * Gets the SDK identifier for the SDK package that generated this Event.
+     *
      * @return string
```
HazAT:
```suggestion
     * @return string
     *
     * @internal
```
     Let's make these functions internal.

src/ClientBuilder.php
```diff
@@ -231,13 +241,39 @@ public function setRepresentationSerializer(RepresentationSerializerInterface $r
...
+     */
+    public function setSdkVersion(string $sdkVersion): void
+    {
+        $this->sdkVersion = $sdkVersion;
+    }
+
+    /**
+     * Sets the version of the SDK package that generated this Event using the Packagist name.
+     *
+     * @param string $packageName
```
HazAT:
```suggestion
     * @param string $packageName
     *
     * @internal
```
     Let's make these functions internal.

src/ClientBuilder.php
```diff
@@ -231,13 +241,39 @@ public function setRepresentationSerializer(RepresentationSerializerInterface $r
...
+    public function setSdkIdentifier(string $sdkIdentifier): void
     {
-        $this->sdkIdentifier = Client::SDK_IDENTIFIER;
+        $this->sdkIdentifier = $sdkIdentifier;
+    }
 
-        if ($sdkIdentifierSuffix) {
-            $this->sdkIdentifier .= '.' . $sdkIdentifierSuffix;
+    /**
+     * @return string
```
HazAT: Comment + internal plz.

src/ClientBuilder.php
```diff
@@ -231,13 +241,39 @@ public function setRepresentationSerializer(RepresentationSerializerInterface $r
     /**
      * {@inheritdoc}
      */
-    public function setSdkIdentifierSuffix(?string $sdkIdentifierSuffix): void
+    public function setSdkIdentifier(string $sdkIdentifier): void
```
HazAT: `@internal`

src/ClientBuilder.php
```diff
@@ -231,13 +241,39 @@ public function setRepresentationSerializer(RepresentationSerializerInterface $r
...
+            $this->setSdkVersionByPackageName('sentry/sentry');
         }
+
+        return $this->sdkVersion;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setSdkVersion(string $sdkVersion): void
```
HazAT: `@internal`

src/Event.php
Jean85: This is used, cannot be removed. I removed the line after this one though.

Jean85:
Rebased on top of new commits from #716 and resolved all comments.

tests/EventTest.php
```diff
@@ -91,6 +92,7 @@ public function testEventIsGeneratedWithUniqueIdentifier(): void
     public function testToArray(): void
     {
         $this->options->setRelease('1.2.3-dev');
```
ste93cry: Is this line really needed? I can't see how this affect the the last assertion since we're never checking the `release` attribute of the returned payload

tests/EventTest.php
```diff
@@ -91,6 +92,7 @@ public function testEventIsGeneratedWithUniqueIdentifier(): void
...
+        $sentryPrettyVersion = PrettyVersions::getVersion('sentry/sentry')->getPrettyVersion();
```
ste93cry: This can be inlined directly in the array, there is no reason to declare a temporary variable to hold a value used only in one place

composer.json
ste93cry: I agree with @HazAT that we should avoid introducing other dependencies not strictly needed. In this case I see the utility of the package but I believe that it's not essential and that remembering to bump a constant before release would not be such a drama

src/Event.php
```diff
@@ -161,23 +167,57 @@ public function getId(): string
...
         return $this->sdkIdentifier;
     }
 
     /**
-     * @internal
+     * Sets the SDK identifier for the SDK package that generated this Event.
      *
      * @param string $sdkIdentifier
+     *
+     * @internal
```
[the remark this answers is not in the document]
HazAT: We can argue about this, we can still make it non `@internal` later, it's not blocking imho
This goes for all `@internal`

composer.json
HazAT: So tbh if we don't need to tell people that they have to add the package to their `composer` call like: 
`php composer.phar require sentry/sentry php-http/curl-client guzzlehttp/psr7 jean85/pretty-package-versions` then this can stay. If they have to add it I vote for removing it.

Jean85: No, it's not needed to be added manually, it's installed automatically as as sub-dependency of `sentry/sentry`. On the contrary, installation would have been needed manually if you wanted to use the `ModulesIntegration`; not anymore with this change.

Jean85:
I've addressed all comments, apart from the `@internal` stuff, where I agree with @HazAT

PR #723, opened 2018-12-08
Allow configure breadcrumb amount

christophlehmann:


christophlehmann:
The current limit of 100 seems a little bit high to me and i bet no one checks the last 100 error possibilities so lets make it configurable

Jean85:
If I'm not mistaken, that limit is related to the fact that it's a server-side limitation.

BTW, I wouldn't add this option due to the fact that it shouldn't be an issue, since breacrumbs are collapsed in Sentry's web interface. Also, we are currently nearing an alpha release of the 2.0 SDK, we should avoid adding new (undocumented) features to the old version.

HazAT:
Heads up, I change the base to `1.x` since this is the old `master`.
Version `1.x` will receive minor and patch releases from this branch from now on.

ste93cry:
Since `2.0` has the `max_breadcrumbs` option that does exactly this and given that it's still in development and this change is really simple, I would simply adjust this PR backporting what we do in the newer version of the SDK (validating the option value and calling it the same between the two versions) and merging it

ste93cry:
@christophlehmann  are you still interested in finishing this PR or can we close it?

christophlehmann:
Lets close it.

PR #728, opened 2018-12-13
Improve client builder and options

Jean85:
This PR addresses a few last things:

- [X] `ClientBuilder` ctor now accepts `Options`, not a naked array
- [X] Adds `Options:setDsn`
- [X] Removes the magic method `ClientBuilder::__call` that forwarded calls to the options
- [X] Adds return types to `ClientBuilderInterface`

Last thing: should we make `Options` `final?

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ use Sentry\ClientBuilder;
 require 'vendor/autoload.php';
 
 // Instantiate the SDK with your DSN
-$client = ClientBuilder::create(['server' => 'http://public@example.com/1'])->getClient();
+$client = ClientBuilder::create(new Options(['server' => 'http://public@example.com/1']))->getClient();
 
 // Capture an exception
 $eventId = $client->captureException(new \RuntimeException('Hello World!'));
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ use Sentry\ClientBuilder;
 require 'vendor/autoload.php';
 
 // Instantiate the SDK with your DSN
-$client = ClientBuilder::create(new Options(['server' => 'http://public@example.com/1']))->getClient();
+$client = ClientBuilder::create(['server' => 'http://public@example.com/1'])->getClient();
 
 // Capture an exception
 $eventId = $client->captureException(new \RuntimeException('Hello World!'));
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -456,7 +456,7 @@
 
   // or
 
-  $client = ClientBuilder::create([...])->getClient();
+  $client = ClientBuilder::create(new Options([...]))->getClient();
   ```
 
 ### Processors
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -456,7 +456,7 @@
 
   // or
 
-  $client = ClientBuilder::create(new Options([...]))->getClient();
+  $client = ClientBuilder::create([...])->getClient();
   ```
 
 ### Processors
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -34,40 +34,6 @@ use Sentry\Transport\TransportInterface;
  * The default implementation of {@link ClientBuilderInterface}.
  *
  * @author Stefano Arlandini <sarlandini@alice.it>
- *
- * @method int getSendAttempts()
- * @method setSendAttempts(int $attemptsCount)
- * @method string[] getPrefixes()
- * @method setPrefixes(array $prefixes)
- * @method float getSampleRate()
- * @method setSampleRate(float $sampleRate)
- * @method bool shouldAttachStacktrace()
- * @method setAttachStacktrace(bool $enable)
- * @method int getContextLines()
- * @method setContextLines(int $contextLines)
- * @method null|string getEnvironment()
- * @method setEnvironment(null|string $environment)
- * @method string[] getExcludedProjectPaths()
- * @method setExcludedProjectPaths(string[] $paths)
- * @method setExcludedLoggers(string[] $loggers)
- * @method string[] getExcludedExceptions()
- * @method string getProjectRoot()
- * @method setProjectRoot(string $path)
- * @method string getLogger()
- * @method setLogger(string $logger)
- * @method string getRelease()
- * @method setRelease(string $release)
- * @method string getDsn()
- * @method string getServerName()
- * @method setServerName(string $serverName)
- * @method string[] getTags()
- * @method setTags(string[] $tags)
- * @method bool shouldSendDefaultPii()
- * @method setSendDefaultPii(bool $enable)
- * @method bool hasDefaultIntegrations()
- * @method setDefaultIntegrations(bool $enable)
- * @method callable getBeforeSendCallback()
- * @method setBeforeSendCallback(callable $beforeSend)
  */
 final class ClientBuilder implements ClientBuilderInterface
 {
@@ -299,25 +265,6 @@ final class ClientBuilder implements ClientBuilderInterface
         $this->transport = $this->transport ?? $this->createTransportInstance();
 
         return new Client($this->options, $this->transport, $this->createEventFactory());
-    }
-
-    /**
-     * This method forwards all methods calls to the options object.
-     *
-     * @param string $name      The name of the method being called
-     * @param array  $arguments Parameters passed to the $name'ed method
-     *
-     * @return $this
-     *
-     * @throws \BadMethodCallException If the called method does not exists
-     */
-    public function __call($name, $arguments)
-    {
-        if (!method_exists($this->options, $name)) {
-            throw new \BadMethodCallException(sprintf('The method named "%s" does not exists.', $name));
-        }
-
-        return $this->options->$name(...$arguments);
     }
 
     /**
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -107,7 +107,7 @@ final class ClientBuilder implements ClientBuilderInterface
     /**
      * {@inheritdoc}
      */
-    public static function create(Options $options = null): self
+    public static function create(Options $options = null): ClientBuilderInterface
     {
         return new static($options);
     }
@@ -123,7 +123,7 @@ final class ClientBuilder implements ClientBuilderInterface
     /**
      * {@inheritdoc}
      */
-    public function setUriFactory(UriFactory $uriFactory): self
+    public function setUriFactory(UriFactory $uriFactory): ClientBuilderInterface
     {
         $this->uriFactory = $uriFactory;
 
@@ -133,7 +133,7 @@ final class ClientBuilder implements ClientBuilderInterface
     /**
      * {@inheritdoc}
      */
-    public function setMessageFactory(MessageFactory $messageFactory): self
+    public function setMessageFactory(MessageFactory $messageFactory): ClientBuilderInterface
     {
         $this->messageFactory = $messageFactory;
 
@@ -143,7 +143,7 @@ final class ClientBuilder implements ClientBuilderInterface
     /**
      * {@inheritdoc}
      */
-    public function setTransport(TransportInterface $transport): self
+    public function setTransport(TransportInterface $transport): ClientBuilderInterface
     {
         $this->transport = $transport;
 
@@ -153,7 +153,7 @@ final class ClientBuilder implements ClientBuilderInterface
     /**
      * {@inheritdoc}
      */
-    public function setHttpClient(HttpAsyncClient $httpClient): self
+    public function setHttpClient(HttpAsyncClient $httpClient): ClientBuilderInterface
     {
         $this->httpClient = $httpClient;
 
@@ -163,7 +163,7 @@ final class ClientBuilder implements ClientBuilderInterface
     /**
      * {@inheritdoc}
      */
-    public function addHttpClientPlugin(Plugin $plugin): self
+    public function addHttpClientPlugin(Plugin $plugin): ClientBuilderInterface
     {
         $this->httpClientPlugins[] = $plugin;
 
@@ -173,7 +173,7 @@ final class ClientBuilder implements ClientBuilderInterface
     /**
      * {@inheritdoc}
      */
-    public function removeHttpClientPlugin(string $className): self
+    public function removeHttpClientPlugin(string $className): ClientBuilderInterface
     {
         foreach ($this->httpClientPlugins as $index => $httpClientPlugin) {
             if (!$httpClientPlugin instanceof $className) {
@@ -189,7 +189,7 @@ final class ClientBuilder implements ClientBuilderInterface
     /**
      * {@inheritdoc}
      */
-    public function setSerializer(SerializerInterface $serializer): self
+    public function setSerializer(SerializerInterface $serializer): ClientBuilderInterface
     {
         $this->serializer = $serializer;
 
@@ -199,7 +199,7 @@ final class ClientBuilder implements ClientBuilderInterface
     /**
      * {@inheritdoc}
      */
-    public function setRepresentationSerializer(RepresentationSerializerInterface $representationSerializer): self
+    public function setRepresentationSerializer(RepresentationSerializerInterface $representationSerializer): ClientBuilderInterface
     {
         $this->representationSerializer = $representationSerializer;
 
@@ -209,7 +209,7 @@ final class ClientBuilder implements ClientBuilderInterface
     /**
      * {@inheritdoc}
      */
-    public function setSdkIdentifier(string $sdkIdentifier): self
+    public function setSdkIdentifier(string $sdkIdentifier): ClientBuilderInterface
     {
         $this->sdkIdentifier = $sdkIdentifier;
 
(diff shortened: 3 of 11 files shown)


tests/ClientBuilderTest.php
```diff
@@ -161,12 +161,13 @@ public function testGetClient(): void
...
     {
-        $clientBuilder = new ClientBuilder([
-            'default_integrations' => $defaultIntegrations,
-            'integrations' => $integrations,
-        ]);
+        $options = new Options();
+        $options->setDefaultIntegrations($defaultIntegrations);
+        $options->setIntegrations($integrations);
 
+        $clientBuilder = new ClientBuilder($options);
```
ste93cry: We can just use `ClientBuilder::create` and get the client in one-line instead of saving the instance of the builder which we don't use anymore in a temporary variable

tests/ClientBuilderTest.php
```diff
@@ -295,7 +247,9 @@ public function testClientBuilderSetsSdkIdentifierAndVersion(): void
      */
     public function testGetClientTogglesCompressionPluginInHttpClient(bool $enabled): void
     {
-        $builder = ClientBuilder::create(['enable_compression' => $enabled, 'dsn' => 'http://public:secret@example.com/sentry/1']);
+        $options = new Options(['dsn' => 'http://public:secret@example.com/sentry/1']);
+        $options->setEnableCompression($enabled);
+        $builder = ClientBuilder::create($options);
```
ste93cry: Add a new line above this one to improve readability

src/ClientBuilderInterface.php
```diff
@@ -35,7 +42,7 @@ public static function create(array $options = []);
      *
      * @return $this
      */
-    public function setUriFactory(UriFactory $uriFactory);
+    public function setUriFactory(UriFactory $uriFactory): self;
```
ste93cry: I explicitly avoided to use `self` in the interfaces because PHP doesn't have covariant returns and this forces anyone implementing this interface to define that the methods returns the interface and not the type of the class itself which can cause issues with fluent interfaces like this one. Please revert everything back

ste93cry:
So, for the question related to making the `Options` class final :+1: from me. For the other points I see no real benefit in most of the changes. The one that concerns me most is that to instantiate the builder you now have to first instantiate the `Options` class and you lose a lot of simplicity for no real benefit since that class has to be created from an array of params anyway. We can provide a `createFromOptions` method if you really think that configuring the client options outside the builder is something really needed but I want to absolutely avoid over-complicating the builder instantiation

src/Options.php
```diff
@@ -393,6 +393,18 @@ public function getDsn(): ?string
         return $this->dsn;
     }
 
+    /**
+     * Sets the DSN of the Sentry server the authenticated user is bound to.
+     *
+     * @param string $dsn
+     */
+    public function setDsn(string $dsn): void
```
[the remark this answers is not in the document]
Jean85: This was needed while passing directly through the `Option` object. But still, we need to decide if we want to go further.

Jean85:
I can see your points. If we want to reduce the impact, we may reduce this PR to:
 * making `Options` `final`
 * add a `ClientBuilder::createFromOptions()`
 * Add a way to pass the `dsn` to Options at construction apart from the naked array, something like `Options::createWithDsn()`.

@HazAT @ste93cry WDYT?

ste93cry:
Both the first and second point of the list are fine. The last point makes me wonder if we really need it. What's the improvement of having a parameter dedicated to just the DSN instead of passing it directly together with all the other options?

Jean85:
The only bonus for the third point is that I'm forced to pass it with an array, always, contrary to all other options with have a setter instead.

src/ClientBuilder.php
```diff
@@ -141,15 +107,23 @@ public function __construct(array $options = [])
     /**
      * {@inheritdoc}
      */
-    public static function create(array $options = []): self
+    public static function create(Options $options = null): ClientBuilderInterface
```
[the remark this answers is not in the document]
HazAT: My opinion on this is twofold.
I would leave `create(array $options = [])` but in the ctor we should use `__construct(Options $options = null)` and create the options object in `create.

ste93cry:
>contrary to all other options with have a setter instead.

This is because you can't change the DSN after creating a client since the transport won't be updated to reflect the new value. The only downside I see in making this change is that you're treating the DSN option differently from the others

src/ClientBuilder.php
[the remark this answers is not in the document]
ste93cry: We can also do that, I'm not against this as long as there is a way to create the builder from a simple array of options so that the user won't need to instantiate himself another object

Jean85:
I've reverted the `setDns` commit, and applied @HazAT suggestion from https://github.com/getsentry/sentry-php/pull/728#discussion_r242126295. Apart from that, IMHO we should leave the return types in the `ClientBuilderInterface`, since it's better that nothing, and `@return $this` is documented on each method to underline that is a fluent interface.

src/ClientBuilder.php
```diff
@@ -124,11 +90,11 @@
     /**
      * Class constructor.
      *
-     * @param array $options The client options
+     * @param Options $options The client options
      */
-    public function __construct(array $options = [])
+    public function __construct(Options $options = null)
```
ste93cry: Since this argument is `nullable` you should use the new PHP notation with the question mark before the typehint

src/ClientBuilder.php
```diff
@@ -124,11 +90,11 @@
     /**
      * Class constructor.
      *
-     * @param array $options The client options
+     * @param Options $options The client options
```
ste93cry: This docblock doesn't reflect the nullability of the parameter

src/ClientBuilderInterface.php
```diff
@@ -22,11 +22,18 @@
     /**
      * Creates a new instance of this builder.
      *
-     * @param array $options The client options
+     * @param array $options The client options, in naked array form; it will be used to construct {@see Options}
```
ste93cry: I don't think that it's useful to know that the array will be used to construct an object instance, I would remove this part of the description

tests/ClientBuilderTest.php
```diff
@@ -35,7 +35,7 @@ public function testCreate(): void
 
     public function testHttpTransportIsUsedWhenServeIsConfigured(): void
     {
-        $clientBuilder = new ClientBuilder(['dsn' => 'http://public:secret@example.com/sentry/1']);
+        $clientBuilder = new ClientBuilder(new Options(['dsn' => 'http://public:secret@example.com/sentry/1']));
```
ste93cry: To simplify the tests I would use the `create` method so that we avoid creating the `Options` object everywhere

tests/ClientBuilderTest.php
```diff
@@ -56,7 +56,7 @@ public function testSetUriFactory(): void
         /** @var UriFactory|MockObject $uriFactory */
         $uriFactory = $this->createMock(UriFactory::class);
 
-        $clientBuilder = new ClientBuilder(['dsn' => 'http://public:secret@example.com/sentry/1']);
+        $clientBuilder = new ClientBuilder(new Options(['dsn' => 'http://public:secret@example.com/sentry/1']));
```
ste93cry: To simplify the tests I would use the `create` method so that we avoid creating the `Options` object everywhere

tests/ClientBuilderTest.php
```diff
@@ -67,7 +67,7 @@ public function testSetMessageFactory(): void
         /** @var MessageFactory|MockObject $messageFactory */
         $messageFactory = $this->createMock(MessageFactory::class);
 
-        $clientBuilder = new ClientBuilder(['dsn' => 'http://public:secret@example.com/sentry/1']);
+        $clientBuilder = new ClientBuilder(new Options(['dsn' => 'http://public:secret@example.com/sentry/1']));
```
ste93cry: To simplify the tests I would use the `create` method so that we avoid creating the `Options` object everywhere

tests/ClientBuilderTest.php
```diff
@@ -82,7 +82,7 @@ public function testSetTransport(): void
         /** @var TransportInterface|MockObject $transport */
         $transport = $this->createMock(TransportInterface::class);
 
-        $clientBuilder = new ClientBuilder(['dsn' => 'http://public:secret@example.com/sentry/1']);
+        $clientBuilder = new ClientBuilder(new Options(['dsn' => 'http://public:secret@example.com/sentry/1']));
```
ste93cry: To simplify the tests I would use the `create` method so that we avoid creating the `Options` object everywhere

tests/ClientBuilderTest.php
```diff
@@ -94,7 +94,7 @@ public function testSetHttpClient(): void
         /** @var HttpAsyncClient|MockObject $httpClient */
         $httpClient = $this->createMock(HttpAsyncClient::class);
 
-        $clientBuilder = new ClientBuilder(['dsn' => 'http://public:secret@example.com/sentry/1']);
+        $clientBuilder = new ClientBuilder(new Options(['dsn' => 'http://public:secret@example.com/sentry/1']));
```
ste93cry: To simplify the tests I would use the `create` method so that we avoid creating the `Options` object everywhere

tests/ClientBuilderTest.php
```diff
@@ -140,7 +140,7 @@ public function testRemoveHttpClientPlugin(): void
 
     public function testGetClient(): void
     {
-        $clientBuilder = new ClientBuilder(['dsn' => 'http://public:secret@example.com/sentry/1']);
+        $clientBuilder = new ClientBuilder(new Options(['dsn' => 'http://public:secret@example.com/sentry/1']));
```
ste93cry: To simplify the tests I would use the `create` method so that we avoid creating the `Options` object everywhere

tests/ClientBuilderTest.php
```diff
@@ -161,12 +161,13 @@ public function testGetClient(): void
      */
     public function testIntegrationsAreAddedToClientCorrectly(bool $defaultIntegrations, array $integrations, array $expectedIntegrations): void
     {
-        $clientBuilder = new ClientBuilder([
```
ste93cry: I would keep using the `array` of options so that we don't have to instantiate the `Options` object ourselves

tests/ClientBuilderTest.php
```diff
@@ -198,63 +199,13 @@ public function integrationsAreAddedToClientCorrectlyDataProvider(): array
...
-    }
-
     public function testClientBuilderFallbacksToDefaultSdkIdentifierAndVersion(): void
     {
         $callbackCalled = false;
         $expectedVersion = PrettyVersions::getVersion('sentry/sentry')->getPrettyVersion();
 
-        $clientBuilder = new ClientBuilder();
-        $clientBuilder->setBeforeSendCallback(function (Event $event) use ($expectedVersion, &$callbackCalled) {
+        $options = new Options();
```
ste93cry: Instead of instantiating the `Options` object I would prefer to stick with the plain `array` of options

tests/ClientBuilderTest.php
```diff
@@ -272,8 +223,8 @@ public function testClientBuilderSetsSdkIdentifierAndVersion(): void
     {
         $callbackCalled = false;
 
-        $clientBuilder = new ClientBuilder();
-        $clientBuilder->setBeforeSendCallback(function (Event $event) use (&$callbackCalled) {
+        $options = new Options();
```
ste93cry: Instead of instantiating the `Options` object I would prefer to stick with the plain `array` of options

tests/ClientBuilderTest.php
```diff
@@ -295,7 +247,9 @@ public function testClientBuilderSetsSdkIdentifierAndVersion(): void
      */
     public function testGetClientTogglesCompressionPluginInHttpClient(bool $enabled): void
     {
-        $builder = ClientBuilder::create(['enable_compression' => $enabled, 'dsn' => 'http://public:secret@example.com/sentry/1']);
+        $options = new Options(['dsn' => 'http://public:secret@example.com/sentry/1']);
```
ste93cry: Instead of instantiating the `Options` object I would prefer to stick with the plain `array` of options

tests/ClientTest.php
```diff
@@ -202,14 +202,16 @@ public function testSendChecksBeforeSendOption(): void
         $transport->expects($this->never())
             ->method('send');
 
-        $client = ClientBuilder::create([
-            'dsn' => 'http://public:secret@example.com/1',
-            'before_send' => function () use (&$beforeSendCalled) {
-                $beforeSendCalled = true;
+        $options = new Options(['dsn' => 'http://public:secret@example.com/1']);
```
ste93cry: Instead of instantiating the `Options` object I would prefer to stick with the plain `array` of options

tests/ClientTest.php
```diff
@@ -219,19 +221,22 @@ public function testSendChecksBeforeSendOption(): void
     /**
      * @dataProvider sampleRateAbsoluteDataProvider
      */
-    public function testSampleRateAbsolute($options): void
+    public function testSampleRateAbsolute(float $sampleRate): void
     {
         $httpClient = new MockClient();
 
-        $client = ClientBuilder::create($options)
+        $options = new Options(['dsn' => 'http://public:secret@example.com/1']);
```
ste93cry: Instead of instantiating the `Options` object I would prefer to stick with the plain `array` of options. Also I would check if we really need a valid HTTP client or we can just set a mocked transport to assert that the `sample_rate` option works

Jean85:
In that regard, The PHPDoc says `@return $this`, so the IDE should pick that up to infer the actual subtype of the builder in use.

Yes, we don't have contravariance in PHP so it's not solvable at the language level, but at the IDE level it seems fine to me.

UPGRADE-2.0.md
```diff
@@ -456,7 +456,7 @@
 
   // or
 
-  $client = ClientBuilder::create([...])->getClient();
+  $client = ClientBuilder::create(new Options([...]))->getClient();
```
ste93cry: Please revert this change since the `create` method still accepts an `array`

README.md
```diff
@@ -58,7 +58,7 @@ use Sentry\ClientBuilder;
 require 'vendor/autoload.php';
 
 // Instantiate the SDK with your DSN
-$client = ClientBuilder::create(['server' => 'http://public@example.com/1'])->getClient();
+$client = ClientBuilder::create(new Options(['server' => 'http://public@example.com/1']))->getClient();
```
ste93cry: Please revert this change since the `create` method still accepts an `array`

ste93cry:
>Yes, we don't have contravariance in PHP so it's not solvable at the language level, but at the IDE level it seems fine to me.

Yea probably IDEs will be able to work if they give priority to the annotation instead of the typehint, but as you said yourself this problem is in the PHP language itself and we should not try to workaround it using a docblock since it's misleading and doesn't respect the real code

src/ClientBuilder.php
Jean85: That's not the same.
 * `Options $options = null` accepts `Options` or nothing, and defaults to `null`
 * `?Options $options` accepts `Options` or `null`, doesn't accept nothing
 * `?Options $options = null` accepts `Options`, `null` or nothing, and defaults to `null`

ste93cry: I was thinking more about the tests where we may want to test all the cases (passing options or using the default ones) and using a dataprovider this would not be possible if the argument is not `nullable` :thinking:

tests/ClientBuilderTest.php
Jean85: I replaced it whenever possible due to the presence of the DSN only. Everywhere else, it may be better to have setters, which are easier to find if needed.

Jean85:
I've addressed or answered every comment from the review.

> we should not try to workaround it using a docblock since it's misleading and doesn't respect the real code

I disagree on this. PHPDoc here helps to narrow the type that  can be inferred from the code alone, in the same way we document arrays as `@var string[]`. PHPStan and other static analysis tools like IDEs do give priority to PHPDoc over code for this exact reason; so in this way we have the stricter language barrier, and the right documentation with the help of PHPDoc.

src/ClientBuilder.php
Jean85: That's not an issue. I've added such test in 1e2c94d

ste93cry:
I hope that with this example the reasons why I am against this change are clearer. Let's take as an example the following code:

```php
interface HelloWorldInterface
{
    /**
     * @return $this
     */
    public function foo(): self;
}

class HelloWorld implements HelloWorldInterface
{
    /**
     * {@inheritdoc}
     */
    public function foo(): HelloWorldInterface
    {
        echo 'foo method called' . PHP_EOL;

        return $this;
    }
    
    public function bar(): self
    {
        echo 'bar method called' . PHP_EOL;

        return $this;
    }
}
    
(new HelloWorld())->foo()->bar()->foo();
```

This is perfectly fine and runs on PHP, however PHPStan reports an error. Infact the only reason the build is green is because you're ignoring the error in the `phpstan.neon` config file, so you are actually bypassing a language limitation using a docblock that doesn't match the real return typehint and its meaning by tricking the configuration of the static checker.

Jean85:
This seems a bug to me: I've moved `@return $this` directly on the implementation and PHPStorm stopped complaining. As per PHPStan, is a tracked bug: https://github.com/phpstan/phpstan/issues/1648

So I would like to avoid changing the correctness of the code basing our choiches on tools that will be fixed in the future. Unfortunately PSR-5 is still under discussion, so it's not reliable for now.

PR #732, opened 2019-01-02
feat: Add spool drain test

HazAT:


HazAT:
Not sure why the tests are failing, it works locally. I guess it's because it's using a newer deps.

PR #733, opened 2019-01-02
feat: Add .craft.yml

HazAT:
This adds the config file https://github.com/getsentry/craft needs to make releases.

We have the composer integration in this repo is setup so as soon as a tag is pushed to this repo it will sync with https://packagist.org/packages/sentry/sentry so for now we don't need composer support in craft.

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -1,0 +1,11 @@
+minVersion: '0.7.0'
+github:
+  owner: getsentry
+  repo: sentry-php
+changelogPolicy: simple
+targets:
+  - name: github
+  - name: registry
+    type: sdk
+    config:
+      canonical: 'composer:sentry/sentry'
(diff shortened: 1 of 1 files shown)


HazAT:
The CI fail is because of this https://github.com/getsentry/sentry-php/pull/732#issuecomment-450821991

.craft.yml
```diff
@@ -0,0 +1,11 @@
...
+github:
+  owner: getsentry
+  repo: sentry-php
+changelogPolicy: simple
+targets:
+  - name: github
+  - name: registry
+    type: sdk
+    config:
+      canonical: 'composer:sentry/sentry'
```
ste93cry: Missing EOL here, the rest looks good to me apart from the indentation that I would change to 4 spaces to be consistent with everything else in this repository

Jean85:
#732 merged, restarting CI...

PR #734, opened 2019-01-05
Raven_Compat::substr() method for calling mb_substr() if available

### LINKED ISSUE #727: mb_substr() should be used on strings, if available (2.0 and master branches)
There are places in the 2.0 and master branches where strings are truncated using substr(), which results in invalid UTF-8 if a multibyte character is truncated.  Strings should be truncated by mb_substr() if the function is available.

See e.g.

```php
if (isset($data['message'])) {
  $data['message'] = substr($data['message'], 0, $this->message_limit);
}
```

```php
$event->setMessage(substr($message, 0, Client::MESSAGE_MAX_LENGTH_LIMIT), $messageParams);
```

mfb:
Fixes #727

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -862,7 +862,7 @@ class Raven_Client
         }
 
         if (isset($data['message'])) {
-            $data['message'] = substr($data['message'], 0, $this->message_limit);
+            $data['message'] = Raven_Compat::substr($data['message'], 0, $this->message_limit);
         }
 
         $data = array_merge($this->get_default_data(), $data);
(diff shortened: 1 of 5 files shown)


mfb:
Well that's interesting, the test loads a bunch of binary gibberish into the `$event['message']` :p

stayallive:
Would be nice to update `Serializer.php` too with this helper and cleanup some code there.

https://github.com/getsentry/sentry-php/blob/82c6d9ad6c59a8648a9fe1730a08e6e22e9a476a/lib/Raven/Serializer.php#L94-L117

To not have any `mb_substr` calls in the code left.

(looks like the `Serializer.php` has the only `mb_substr` call: https://github.com/getsentry/sentry-php/search?classic=1&q=mb_substr%28)

HazAT:
Heads up, I change the base to `1.x` since this is the old `master`.

ste93cry:
@HazAT can we get this merged?

PR #735, opened 2019-01-10
meta: Update readme, remove upgrade guide for now

HazAT:


README.md
```diff
@@ -51,30 +44,18 @@ and [`http-message-implementation`](https://packagist.org/providers/psr/http-mes
...
-
-use Sentry\ClientBuilder;
-
-require 'vendor/autoload.php';
+use function Sentry\init;
+use function Sentry\captureException;
 
-// Instantiate the SDK with your DSN
-$client = ClientBuilder::create(['server' => 'http://public@example.com/1'])->getClient();
+init(['dsn' => '___PUBLIC_DSN___' ]);
```
Jean85: Should we also add an example on how to register the error handler for automatic error reporting?

ste93cry:
I see you ran the PHP-CS-Fixer and fixed the build failing due to CS, it's fine for me but the only thing I would check is why the `\` has been removed from certain core functions (AFAIK they were adding it to improve performances and I don't get why only certain functions keep it now)

Jean85:
>  (AFAIK they were adding it to improve performances and I don't get why only certain functions keep it now)

I think they changed the default to which functions are included in that regard or not. Probably those other functions do not benefit from the perf improvement?

We can choose to prefix all functions and be done with it...

ste93cry:
>We can choose to prefix all functions and be done with it...

I'm fine with how it is, I believe that the new default is anyway good or it would not have been changed, I was just curious about why such change has been made

tests/ErrorHandlerTest.php
```diff
@@ -316,8 +316,8 @@ public function testHandleExceptionWithPreviousExceptionHandler(): void
                 $errorHandler->handleException($exception);
 
                 $this->fail('Exception expected');
-            } catch (\Exception $catchedException) {
```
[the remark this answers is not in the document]
HazAT: Fixed it and corrected wording.

README.md
ste93cry: Isn't it registered automatically when using the default integrations?

HazAT: Right

PR #736, opened 2019-01-13
Make Raven_Breadcrumbs_MonologHandler::write() compatible with AbstractProcessingHandler

davidsneighbour:
Annotation of return type missing in function write().

Declaration of Raven_Breadcrumbs_MonologHandler::write(array $record) must be compatible with Monolog\Handler\AbstractProcessingHandler::write(array $record): void in /home/patrick/Projects/Booka/vendor/sentry/sentry/lib/Raven/Breadcrumbs/MonologHandler.php


davidsneighbour:
Looks like a deeper issue with a more current monolog version. Let's close it for now.

Jean85:
This probably means that the handler is incompatible with Monolog 2.0. It's not a problem with PHP 7.2+ BTW.

PR #742, opened 2019-01-22
Add PHP 7.3 to Travis CI build and support EditorConfig

ste93cry:
Since PHP 7.3 is out we can now add it to the TravisCI pipeline. I've also added support for EditorConfig and changed the steps order to ensure that PHPStan runs first. The reason why I decided to do it is that it makes sense for a static analysis tool to runs first since things that it reports are likely to be bugs that should be fixed regardless of the CI build result, so knowing that there is something wrong as soon as possible is better

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+; top-most EditorConfig file
+root = true
+
+; Unix-style newlines
+[*]
+charset = utf-8
+end_of_line = LF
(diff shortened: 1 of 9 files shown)


ste93cry:
I believe it is because otherwise you could not be sure that the lowest dependencies actually runs on a specific version of PHP

PR #743, opened 2019-01-22
Ensure context lines option is used (fixes #740)

### LINKED ISSUE #740: context_lines setting is never used
I was attempting to limit the lines of context by setting `context_lines => 1`, but it looks like it's hard coded for 5 lines of pre and post context: https://github.com/getsentry/sentry-php/blob/master/src/Stacktrace.php#L134

mattzuba:
Tests included, ensures that the context options that are set are used, rather than being hard-coded to a constant.

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -159,9 +159,9 @@ final class Options
     /**
      * Gets the number of lines of code context to capture, or null if none.
      *
-     * @return int|null
+     * @return int
      */
-    public function getContextLines(): ?int
+    public function getContextLines(): int
     {
         return $this->options['context_lines'];
     }
@@ -169,9 +169,9 @@ final class Options
     /**
      * Sets the number of lines of code context to capture, or null if none.
      *
-     * @param int|null $contextLines The number of lines of code
+     * @param int $contextLines The number of lines of code
      */
-    public function setContextLines(?int $contextLines): void
+    public function setContextLines(int $contextLines): void
     {
         $options = array_merge($this->options, ['context_lines' => $contextLines]);
 
@@ -636,7 +636,7 @@ final class Options
             'prefixes' => explode(PATH_SEPARATOR, get_include_path()),
             'sample_rate' => 1,
             'attach_stacktrace' => false,
-            'context_lines' => 3,
+            'context_lines' => 5,
             'enable_compression' => true,
             'environment' => null,
             'project_root' => null,
--- a/tests/StacktraceTest.php
+++ b/tests/StacktraceTest.php
@@ -143,7 +143,7 @@ final class StacktraceTest extends TestCase
      */
     public function testAddFrameRespectsContextLinesOption(string $fixture, int $lineNumber, ?int $contextLines, int $preContextCount, int $postContextCount): void
     {
-        if ($contextLines !== null) {
+        if (null !== $contextLines) {
             $this->options->setContextLines($contextLines);
         }
 
--- a/tests/StacktraceTest.php
+++ b/tests/StacktraceTest.php
@@ -188,6 +188,34 @@ final class StacktraceTest extends TestCase
         }
     }
 
+    public function testAddFrameUsesOptionsContext(): void
+    {
+        $this->options->setContextLines(2);
+
+        $fileContent = explode("\n", $this->getFixture('code/LongFile.php'));
+        $stacktrace = new Stacktrace($this->options, $this->serializer, $this->representationSerializer);
+
+        $stacktrace->addFrame($this->getFixturePath('code/LongFile.php'), 8, [
+            'function' => '[unknown]',
+        ]);
+
+        $frames = $stacktrace->getFrames();
+
+        $this->assertCount(1, $frames);
+        $this->assertCount(2, $frames[0]->getPreContext());
+        $this->assertCount(2, $frames[0]->getPostContext());
+
+        for ($i = 0; $i < 2; ++$i) {
+            $this->assertEquals(rtrim($fileContent[$i + 5]), $frames[0]->getPreContext()[$i]);
+        }
+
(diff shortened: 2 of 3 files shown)


mattzuba:
Looks like the get/setContextLines methods allow for possible null, but the OptionsResolver only allows an int, do you mind if I change the method signatures of these to be int only?

ste93cry:
>Looks like the get/setContextLines methods allow for possible null, but the OptionsResolver only allows an int

That's indeed another bug and I agree that the parameter should not be `nullable`. Maybe we can also optimize a bit the code in `Stacktrace::getSourceCodeExcerpt()` by checking if `$maxLinesToFetch` is greater than 0 and if not return early instead of opening the file anyway to read nothing. There is also another bug I'm seeing right now that is being reported by Scrutinizer: we're calling that method as if it was a `static` one but it's not

https://github.com/getsentry/sentry-php/blob/5ec1fbec3ff7c32dda49f234a2d5db5759393676/src/Stacktrace.php#L134

mattzuba:
Even if `$maxLinesToFetch` was 0, you'd still want to open the file to read the line that was a part of the stack though, right?

src/Options.php
```diff
@@ -616,7 +616,7 @@ private function configureOptions(OptionsResolver $resolver): void
             'sample_rate' => 1,
             'mb_detect_order' => null,
             'attach_stacktrace' => false,
-            'context_lines' => 3,
+            'context_lines' => 5,
```
Jean85: @HazAT which should be the right default for this value? Which is the "normal" default on the server side?

tests/StacktraceTest.php
```diff
@@ -188,6 +188,34 @@ public function testAddFrameReadsCodeFromLongFile(): void
         }
     }
 
+    public function testAddFrameUsesOptionsContext(): void
+    {
+        $this->options->setContextLines(2);
```
Jean85: Can you add a data provider to this test so we can see it work with `0` and two other different values?

ste93cry:
>Even if $maxLinesToFetch was 0, you'd still want to open the file to read the line that was a part of the stack though, right?

As a user I would expect no lines to be sent to Sentry if the value of that option is set to 0. I may be wrong but I think that actually the number of lines also includes the line that originated the frame itself

HazAT:
Setting it to `0` should just remove the `pre` and `post` context lines, the line of the frame in the stacktrace should still be there.
btw thanks for the work 👍

mattzuba:
I didn't forget about this guys, just been busy at work.  I'll try to get a data provider setup on the test and merge the other test and update the PR today or tomorrow.

ste93cry:
@mattzuba are you willing to make the requested changes? We would like to release `beta 2` asap and it would be cool if this bugfix was part of that version

tests/StacktraceTest.php
mattzuba: Please review the latest changes - I moved the redunant tests that @ste93cry mentioned into the dataprovider and created a few additional ones (zero context, context near end and beginning of file).

tests/StacktraceTest.php
```diff
@@ -138,54 +138,47 @@ public function testAddFrameMarksAsInApp(): void
         $this->assertTrue($frames[1]->isInApp());
     }
 
-    public function testAddFrameReadsCodeFromShortFile(): void
+    /**
+     * @dataProvider contextDataProvider
+     */
+    public function testAddFrameWithDifferentContexts($fixture, $lineNumber, $contextLines, $preContextCount, $postContextCount)
```
ste93cry: Please rename this test method to `testAddFrameRespectsContextLinesOption`

tests/StacktraceTest.php
```diff
@@ -138,54 +138,47 @@ public function testAddFrameMarksAsInApp(): void
...
 
-        for ($i = 0; $i < 2; ++$i) {
-            $this->assertEquals(rtrim($fileContent[$i + 3]), $frames[0]->getPostContext()[$i]);
+        for ($i = 0; $i < $postContextCount; ++$i) {
+            $this->assertEquals(rtrim($fileContent[$i + $lineNumber]), $frames[0]->getPostContext()[$i]);
         }
     }
 
-    public function testAddFrameReadsCodeFromLongFile(): void
+    public function contextDataProvider()
```
ste93cry: As a convention used across the codebase we used for the name of the method of the data provider the name of the test method that uses it + the `DataProvider` suffix. The only exception is if the data provider is used from more than one test method

tests/StacktraceTest.php
ste93cry: Still some more minor fixes, the rest looks good!

tests/StacktraceTest.php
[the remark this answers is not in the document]
mattzuba: Added these to parameters and also added return types

tests/StacktraceTest.php
```diff
@@ -138,54 +138,47 @@ public function testAddFrameMarksAsInApp(): void
...
     }
 
-    public function testAddFrameReadsCodeFromShortFile(): void
+    /**
+     * @dataProvider contextDataProvider
+     */
+    public function testAddFrameWithDifferentContexts($fixture, $lineNumber, $contextLines, $preContextCount, $postContextCount)
     {
-        $fileContent = explode("\n", $this->getFixture('code/ShortFile.php'));
+        if (isset($contextLines)) {
```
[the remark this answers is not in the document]
mattzuba: Done

tests/StacktraceTest.php
mattzuba: Done

tests/StacktraceTest.php
mattzuba: Done

mattzuba:
Made the requested changes and rebased against master as well.

PR #744, opened 2019-01-22
Mention TYPO3 integration

christophlehmann:


Jean85:
Thanks for working on this, but right now the master branch contains the new 2.0 release, so your integration doesn't work with the current situation.

Maybe we should split/reedit that section to highlight which integrations are using an up to date SDK and which doesn't?

christophlehmann:
I would just add it. I will update our Integration as soon as a the new SDK has a non beta tag ;-)

ste93cry:
For me PR is ok, the only thing is that I agree with @Jean85 we should avoid publicizing packages that uses the old SDK version in the branch of the unified API

PR #745, opened 2019-01-23
feat: Add max_value_length option

HazAT:


tests/Serializer/SerializerTest.php
```diff
@@ -95,11 +96,45 @@ public function testNull(bool $serializeAllObjects): void
...
+            $result = $this->invokeSerialization($serializer, $input);
+
+            $this->assertInternalType('string', $result);
+            $this->assertLessThanOrEqual(500, \strlen($result));
+        }
+    }
+
+    public function testClippingUTF8Characters(): void
+    {
+        if (!\extension_loaded('mbstring')) {
```
ste93cry: Since we're requiring the `mbstring` extension in the `composer.json` we can safely remove this code

tests/Serializer/SerializerTest.php
```diff
@@ -95,11 +96,45 @@ public function testNull(bool $serializeAllObjects): void
...
+        $this->assertSame(JSON_ERROR_NONE, json_last_error());
+    }
+
     /**
+     * @param Options $options
+     *
      * @return Serializer
      */
-    protected function createSerializer(): AbstractSerializer
+    protected function createSerializer(Options $options = null): AbstractSerializer
```
ste93cry: Since this is a `nullable` argument with a default value it should be `?Options $options = null`

src/Options.php
```diff
@@ -597,6 +597,28 @@ public function setDefaultIntegrations(bool $enable): void
...
+     *
+     * @return int
+     */
+    public function getTruncationLength(): int
+    {
+        return $this->options['truncation_length'];
+    }
+
+    /**
+     * Sets the truncation length for specific values in the event payload.
```
ste93cry: Does this really apply to only some values? I thought it was applied to everything that is processed by the serializers

tests/Serializer/AbstractSerializerTest.php
```diff
@@ -317,20 +317,6 @@ public function testLongString(bool $serializeAllObjects): void
         }
     }
 
-    public function testLongStringWithOverwrittenMessageLength(): void
```
Jean85: Why did you move those?

src/Serializer/AbstractSerializer.php
```diff
@@ -76,15 +74,22 @@ abstract class AbstractSerializer
...
 
-    public function __construct(int $maxDepth = 3, ?string $mbDetectOrder = null, int $messageLimit = Client::MESSAGE_MAX_LENGTH_LIMIT)
+    /**
+     * AbstractSerializer constructor.
+     *
+     * @param Options     $options       The SDK configuration options
+     * @param int         $maxDepth
+     * @param string|null $mbDetectOrder
+     */
+    public function __construct(Options $options, int $maxDepth = 3, ?string $mbDetectOrder = null)
```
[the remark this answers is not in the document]
ste93cry: Does it really makes sense to have it as a client option? It seems to be something related to the serializer, imho we should drop it and just rely on the argument of this constructor instead

src/Options.php
HazAT: I quickly went over the code, we do not use this value for everything.
Maybe we should tackle this in a different PR.

tests/Serializer/AbstractSerializerTest.php
HazAT: Because the tests didn't run.

tests/EventFactoryTest.php
```diff
@@ -16,20 +16,29 @@
...
 class EventFactoryTest extends TestCase
 {
+    /**
+     * @var Options
+     */
+    private $options;
+
+    protected function setUp(): void
+    {
+        $this->options = new Options();
```
ste93cry: I don't see a real value in having this instance set as variable of the class, is there any reason you moved it here?

HazAT: Mostly for convenience, having `new Options()` in almost every test doesn't really make sense. Also, we've done this in other test classes.

tests/Serializer/AbstractSerializerTest.php
ste93cry: Wasn't it possible to just add the `$options` parameter to the function?

tests/EventFactoryTest.php
ste93cry: Not that it changes much, but only two tests are using the options so I don't think we should move the instantiation to the `setUp` method which creates an additional object each time a test case is executed even if there is no need for it

PR #747, opened 2019-01-24
Create necessary factories only when using HTTP transport

### LINKED ISSUE #741: Disable http plug discovery when using custom transport
I'm using this SDK (the 2.0 beta version) inside an asynchronous ReactPHP application, but due to the broken http-plug reactphp client I've made my own sentry transport which uses the already available HTTP client inside my application. But due to the way how the client is built inside the client builder, even when I set a custom transport, the client builder tries to find a http plug HTTP client using the http plug discovery mechanism.

As such I have to set a custom strategy to mitigate this.
```php
$class = \get_class((new class() {
    static function getCandidates() {
        return array(array('class' => function () {}));
    }
}));

\Http\Discovery\ClassDiscovery::setStrategies(array($class));
```

Would you be interested in a simple patch to not do the discovery if a transport is set? From what I've seen the HTTP client is only used when building a transport. So the patch would be to simply check whether a transport is set or not.

https://github.com/getsentry/sentry-php/blob/5ec1fbec3ff7c32dda49f234a2d5db5759393676/src/ClientBuilder.php#L260-L268

The simple patch would be to change
https://github.com/getsentry/sentry-php/blob/5ec1fbec3ff7c32dda49f234a2d5db5759393676/src/ClientBuilder.php#L264-L265
to
```php
if($this->transport === null) {
    $this->httpClient = $this->httpClient ?? HttpAsyncClientDiscovery::find();
    $this->transport = $this->createTransportInstance();
}
```

Of course you're welcome to do the simple change yourself. :)

CharlotteDunois:
This PR does the changes discussed in #741.

When setting a custom transport, the client builder will not create any factories used by the HTTP transport anymore. These prevented using a custom transport which did not rely on HTTPlug (because these relied on class discovery).

Closes #741.

### DIFF
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -259,9 +259,6 @@ final class ClientBuilder implements ClientBuilderInterface
      */
     public function getClient(): ClientInterface
     {
-        $this->messageFactory = $this->messageFactory ?? MessageFactoryDiscovery::find();
-        $this->uriFactory = $this->uriFactory ?? UriFactoryDiscovery::find();
-        $this->httpClient = $this->httpClient ?? HttpAsyncClientDiscovery::find();
         $this->transport = $this->transport ?? $this->createTransportInstance();
 
         return new Client($this->options, $this->transport, $this->createEventFactory());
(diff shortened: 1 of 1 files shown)


ste93cry:
>I would like to have a regression test for this

Some tests of the client builder class already checks whether the HTTP transport that is created uses the custom factories, so no need for a regression test. PR is fine for such change, the only thing I would like to explore (it's not blocking btw) is whether it makes sense to expose the `set*Factory` methods in the client builder. Since the client is agnostic of the transport, the builder should not know anything regarding a specific transport type which in this case is HTTP

Jean85:
I agree with the possibility of removing the factory setters. But we can address that in a separate PR.

Thanks @CharlotteDunois for this one, merging!

Issue #746, opened 2019-01-23
[2.0] error_types option doesn't work


zavoloklom:
When I set option error_types to (E_ALL & ~E_WARNING & ~E_NOTICE) I expect that types E_WARNING and E_NOTICE won't send information to server.

Code example:
```
<?php
require(__DIR__ . '/vendor/autoload.php');
use function Sentry\init;
init(['dsn' => __DSN___, 'error_types' => E_ALL & ~E_WARNING & ~E_NOTICE]));
$my_array = array();
$info_error= $my_array[2];
```

But when I execute it I receive:
`Notice: Undefined offset: 2`

It happened because of `capturedErrors` private property in ErrorHandler class.
https://github.com/getsentry/sentry-php/blob/master/src/ErrorHandler.php#L46

Because of that `shouldCaptureError` is always true.
https://github.com/getsentry/sentry-php/blob/master/src/ErrorHandler.php#L181

---
php version 7.1.24
sentry package 2.0.0-beta1

Jean85:
The base errors is in the fact that the error handler doesn't consider the options.

To try and fix that (and other lingering issues) I've started #748.

ste93cry:
Issue should have been fixed in the referenced PR. Feel free to reopen if you still have problems

PR #748, opened 2019-01-24
[WIP] Change the IntegrationInterface to pass around instances

Jean85:
This is a PoC to see if we can address #746. With this change every integration will be instantiated in each client, so we will be able to bind the client options to each of them.

With this in mind, we will be able to pass the options to the ErrorHandler, and have multiple clients with different error handling configured.


Jean85:
Closing in favor of #749

PR #749, opened 2019-01-25
[WIP] Integrations at the Hub level

Jean85:
This is a new PoC, more complete and probably more viable than #748, to solve #746.

In this way, integrations are bound to the Hub level, so the client is used indirectly, and the (default) integrations are generated inside the options, so they can carry those everywhere.

I hope this iteration is the correct solution. WDYT?


src/ErrorHandler.php
```diff
@@ -83,8 +83,6 @@ final class ErrorHandler
      *
      * @param callable $callback           The callback that will be invoked in case an error is caught
      * @param int      $reservedMemorySize The amount of memory to reserve for the fatal error handler
-     *
-     * @throws \ReflectionException
      */
     private function __construct(callable $callback, int $reservedMemorySize = self::DEFAULT_RESERVED_MEMORY_SIZE)
```
ste93cry: Since the constructor is `private` remove the default value for the `$reservedMemorySize` argument

src/Integration/ModulesIntegration.php
```diff
@@ -24,26 +24,19 @@ final class ModulesIntegration implements IntegrationInterface
...
     }
 
     /**
      * Applies the information gathered by this integration to the event.
      *
-     * @param self  $self  The instance of this integration
      * @param Event $event The event that will be enriched with the modules
      */
-    public static function applyToEvent(self $self, Event $event): void
+    public static function applyToEvent(Event $event): void
```
ste93cry: Since this method should not be called directly from the outside, make it `private`

src/Integration/RequestIntegration.php
```diff
@@ -37,29 +37,20 @@ public function __construct(Options $options)
...
 
     /**
      * Applies the information gathered by the this integration to the event.
      *
-     * @param self                        $self    The current instance of RequestIntegration
      * @param Event                       $event   The event that will be enriched with a request
      * @param ServerRequestInterface|null $request The Request that will be processed and added to the event
      */
-    public static function applyToEvent(self $self, Event $event, ?ServerRequestInterface $request = null): void
+    public function applyToEvent(Event $event, ?ServerRequestInterface $request = null): void
```
ste93cry: Since this method should not be called directly from the outside, make it `private`

src/Sdk.php
```diff
@@ -9,13 +9,23 @@
 /**
  * Creates a new Client and Hub which will be set as current.
  *
- * @param array $options The client options
+ * @param array $arrayOptions The client options
  */
-function init(array $options = []): void
+function init(array $arrayOptions = []): void
```
ste93cry: Please revert this change and just override the variable later when wrapping it with the `Options` class

src/Sdk.php
```diff
@@ -9,13 +9,23 @@
...
 {
-    $client = ClientBuilder::create($options)->getClient();
+    $options = new Options($arrayOptions);
+    $client = (new ClientBuilder($options))->getClient();
```
ste93cry: Why you changed this to `new` instead of using the static method `create` which makes the code nicer?

src/Sdk.php
```diff
@@ -9,13 +9,23 @@
...
 
-    Hub::setCurrent(new Hub($client));
+    $hub = new Hub($client);
+    Hub::setCurrent($hub);
```
ste93cry: I think you can just do this instead of initializing the hub yourself:

```php
$hub = Hub::getCurrent();
$hub->bindClient($client);
```

src/State/Hub.php
```diff
@@ -229,17 +233,9 @@ public static function setCurrent(HubInterface $hub): HubInterface
...
-    public function getIntegration(string $className): ?IntegrationInterface
+    public function addIntegration(IntegrationInterface $integration): void
     {
-        $client = $this->getClient();
-        if (null !== $client) {
-            return $client->getIntegration($className);
-        }
-
-        return null;
+        $this->integrations[] = $integration->bindToHub($this);
```
ste93cry: I don't like the fact that an integration could be bounded later to a different hub. Imho if this is the direction we are going to follow then an integration should require the hub to bind in the constructor

tests/Serializer/AbstractSerializerTest.php
```diff
@@ -346,7 +346,7 @@ public function testSerializeValueResource(bool $serializeAllObjects): void
 
         $this->assertNotFalse($filename, 'Temp file creation failed');
 
-        $resource = fopen($filename, 'wb');
```
ste93cry: Please check that your local dependencies are updated, with the latest version of PHPCS-Fixer this change should not happen (it was made by @HazAT when he committed the file, then with my latest commit it got reverted back and now you're again reverting it)

Jean85:
Thanks for the input @HazAT. 

I fear that currently the integrations are treated as singletons, due to the fact that the `Handler` is called statically in the client, and it has [a static property](https://github.com/getsentry/sentry-php/blob/679ffb26dfb461525bb4842a4dfd8aec9b7edc73/src/Integration/Handler.php#L13) to track which integrations are installed.

Also, I don't know if filtering by FQCN is a good way to ensure single usage of `setupOnce`, since I may want to use the same class with different options/dependencies to do different stuff. We should bind this things to the instance, not the class IMHO.

I agree that this approach is too radical, we can maybe solve it better just changing the `Handler` behavior.

HazAT:
@Jean85 This is right, but this static property is just used to track to `setupOnce` calls, this index is what the `Client` uses
https://github.com/getsentry/sentry-php/blob/679ffb26dfb461525bb4842a4dfd8aec9b7edc73/src/Integration/Handler.php#L24
and this is not a singleton, it returns the instance passed to it (or am I wrong)?

HazAT:
I think we can close this PR right?
Even though it might provide an example of how to move integrations to the hub level.
I think if we want to go down this route we should take a step back and talk about concepts first before implementing it.

Jean85:
Yes. Closing in favor of #762 which solves the same problem without too much changes (and with additional features too!)

PR #752, opened 2019-01-27
Allow formatted message to be passed in event payload

### LINKED ISSUE #750: Support named parameters for event messages
In version 1.x, we can send messages with named parameters to Sentry server, e.g.
```
$message = 'Hello %planet.';
$params = ['%planet' => 'world'];
$data['sentry.interfaces.Message'] = [
  'message' => $message,
  'params' => $params,
  'formatted' => strtr($message, $params),
];
```
I'm not sure how to do this for events in 2.x - it looks like it forcibly sets `formatted` to `vsprintf($this->message, $this->messageParams)`.

mfb:
Allow formatted message to be passed in event payload (fixes #750); use mb_substr (refs #727).

This PR will need to be updated after #745 is merged.

### DIFF
--- a/src/Event.php
+++ b/src/Event.php
@@ -62,6 +62,11 @@ final class Event implements \JsonSerializable
     private $message;
 
     /**
+     * @var string|null The formatted error message
+     */
+    private $messageFormatted;
+
+    /**
      * @var array The parameters to use to format the message
      */
     private $messageParams = [];
@@ -340,6 +345,16 @@ final class Event implements \JsonSerializable
     }
 
     /**
+     * Gets the formatted message.
+     *
+     * @return string|null
+     */
+    public function getMessageFormatted(): ?string
+    {
+        return $this->messageFormatted;
+    }
+
+    /**
      * Gets the parameters to use to format the message.
      *
      * @return string[]
@@ -352,13 +367,15 @@ final class Event implements \JsonSerializable
     /**
      * Sets the error message.
      *
-     * @param string $message The message
-     * @param array  $params  The parameters to use to format the message
+     * @param string      $message   The message
+     * @param array       $params    The parameters to use to format the message
+     * @param string|null $formatted The formatted message
      */
-    public function setMessage(string $message, array $params = []): void
+    public function setMessage(string $message, array $params = [], string $formatted = null): void
     {
         $this->message = $message;
         $this->messageParams = $params;
+        $this->messageFormatted = $formatted;
     }
 
     /**
@@ -659,7 +676,7 @@ final class Event implements \JsonSerializable
                 $data['message'] = [
                     'message' => $this->message,
                     'params' => $this->messageParams,
-                    'formatted' => vsprintf($this->message, $this->messageParams),
+                    'formatted' => $this->messageFormatted ?? vsprintf($this->message, $this->messageParams),
                 ];
             }
         }
--- a/tests/EventFactoryTest.php
+++ b/tests/EventFactoryTest.php
@@ -94,14 +94,14 @@ class EventFactoryTest extends TestCase
             ],
             [
                 [
-                    'message' => 'testMessage @param',
-                    'message_params' => ['@param' => 'param'],
+                    'message' => 'testMessage %foo',
+                    'message_params' => ['%foo' => 'param'],
                     'message_formatted' => 'testMessage param',
                 ],
                 [
                     'message' => [
-                        'message' => 'testMessage @param',
-                        'params' => ['@param' => 'param'],
+                        'message' => 'testMessage %foo',
+                        'params' => ['%foo' => 'param'],
                         'formatted' => 'testMessage param',
                     ],
                 ],
--- a/tests/EventFactoryTest.php
+++ b/tests/EventFactoryTest.php
@@ -92,6 +92,20 @@ class EventFactoryTest extends TestCase
                     ],
                 ],
             ],
+            [
+                [
+                    'message' => 'testMessage @param',
+                    'message_params' => ['@param' => 'param'],
+                    'message_formatted' => 'testMessage param',
+                ],
+                [
+                    'message' => [
+                        'message' => 'testMessage @param',
+                        'params' => ['@param' => 'param'],
+                        'formatted' => 'testMessage param',
+                    ],
+                ],
+            ],
         ];
     }
 
--- a/tests/EventTest.php
+++ b/tests/EventTest.php
@@ -122,10 +122,11 @@ final class EventTest extends TestCase
     /**
      * @dataProvider toArrayWithMessageDataProvider
      */
-    public function testToArrayWithMessage(array $message, $expectedValue): void
+    public function testToArrayWithMessage(array $setMessageArguments, $expectedValue): void
     {
         $event = new Event();
-        \call_user_func_array([$event, 'setMessage'], $message);
+
+        \call_user_func_array([$event, 'setMessage'], $setMessageArguments);
 
         $data = $event->toArray();
 
@@ -136,9 +137,18 @@ final class EventTest extends TestCase
     public function toArrayWithMessageDataProvider(): array
     {
         return [
-            [['foo bar'], 'foo bar'],
-            [['foo %s', ['bar']], ['message' => 'foo %s', 'params' => ['bar'], 'formatted' => 'foo bar']],
-            [['foo @bar', ['@bar' => 'bar'], 'foo bar'], ['message' => 'foo @bar', 'params' => ['@bar' => 'bar'], 'formatted' => 'foo bar']],
+            [
+                ['foo bar'],
+                'foo bar',
+            ],
+            [
+                ['foo %s', ['bar']],
+                ['message' => 'foo %s', 'params' => ['bar'], 'formatted' => 'foo bar'],
+            ],
+            [
+                ['foo %bar', ['%bar' => 'baz'], 'foo baz'],
+                ['message' => 'foo %bar', 'params' => ['%bar' => 'baz'], 'formatted' => 'foo baz'],
+            ],
         ];
     }
 
@@ -163,10 +173,11 @@ final class EventTest extends TestCase
     /**
      * @dataProvider getMessageDataProvider
      */
-    public function testGetMessage(array $message, array $expectedValue): void
+    public function testGetMessage(array $setMessageArguments, array $expectedValue): void
     {
         $event = new Event();
-        \call_user_func_array([$event, 'setMessage'], $message);
+
+        \call_user_func_array([$event, 'setMessage'], $setMessageArguments);
 
         $this->assertSame($expectedValue['message'], $event->getMessage());
         $this->assertSame($expectedValue['params'], $event->getMessageParams());
@@ -176,8 +187,14 @@ final class EventTest extends TestCase
     public function getMessageDataProvider(): array
     {
         return [
-            [['foo %s', ['bar']], ['message' => 'foo %s', 'params' => ['bar'], 'formatted' => null]],
-            [['foo @bar', ['@bar' => 'bar'], 'foo bar'], ['message' => 'foo @bar', 'params' => ['@bar' => 'bar'], 'formatted' => 'foo bar']],
+            [
+                ['foo %s', ['bar']],
+                ['message' => 'foo %s', 'params' => ['bar'], 'formatted' => null],
+            ],
+            [
+                ['foo %bar', ['%bar' => 'baz'], 'foo baz'],
+                ['message' => 'foo %bar', 'params' => ['%bar' => 'baz'], 'formatted' => 'foo baz'],
+            ],
         ];
     }
 
--- a/tests/EventTest.php
+++ b/tests/EventTest.php
@@ -138,16 +138,41 @@ final class EventTest extends TestCase
     {
         return [
             [
-                ['foo bar'],
+                [
+                    'foo bar',
+                ],
                 'foo bar',
             ],
             [
-                ['foo %s', ['bar']],
-                ['message' => 'foo %s', 'params' => ['bar'], 'formatted' => 'foo bar'],
+                [
+                    'foo %s',
+                    [
+                        'bar',
+                    ],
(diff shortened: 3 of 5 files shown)


mfb:
It would make sense to me if the payload 'message' element could be either a string or array; however we may need to continue to support 'message_params' element so I kept it as is.

ste93cry:
The `sentry.interfaces.Message.params` attribute AFAIK is not deprecated, so we have to keep supporting it, what doesn't convince me is to expose the `sentry.interfaces.Message.formatted` attribute since it seems to me that it's something "internal" that should be computed automatically from the pair of message text and message params. As I said in the issue I never seen in the documentation the use of named parameters, although it seems that they are supported, and I've never seen support for `sprintf`ing them in other SDKs

mfb:
Sorry if I was unclear, I was not suggesting that params is deprecated. I was just musing about how the event payload array could hypothetically be structured.

My two cents is that it's critical for apps to be able to send the message, formatted message and params to Sentry. The message is used to aggregate events together, and the formatted message is the message that actually appears in the Sentry UI (with the params below), so it's quite important for Sentry's core utility to get this right.

ste93cry:
What I don't like is that exposing the `formatted` attribute of the interface means that someone could pass a text, some params and a totally different formatted message and then things would be out of sync. Honestly I don't know if these fields should be coherent or if it makes sense to reason about this question

src/Event.php
```diff
@@ -352,13 +367,15 @@ public function getMessageParams(): array
...
      * Sets the error message.
      *
-     * @param string $message The message
-     * @param array  $params  The parameters to use to format the message
+     * @param string      $message   The message
+     * @param array       $params    The parameters to use to format the message
+     * @param string|null $formatted The formatted message
      */
-    public function setMessage(string $message, array $params = []): void
+    public function setMessage(string $message, array $params = [], string $formatted = null): void
```
ste93cry: Since the argument is `nullable` the correct typehint should be `?string $formatted = null`, otherwise it means only that its default value is `null`

ste93cry:
After talking with @untitaker we agreed that the `captureMessage` function won't change its signature to allow passing the `params` attribute of the `sentry.interfaces.Message` interface. To send the parameters you will have to use the `captureEvent` function and pass the raw payload data that will then be processed by the code of this PR. Can you please add some tests to verify that both the `getter`/`setter` method works and that the event data is computed correctly according to the new logic?

src/Event.php
Jean85: Nope, it's not the same. `string $formatted = null` means that IF you pass something, it MUST NOT be `null`. IMHO it's the right behavior, so if `$formatted === null`, we know the user didn't pass anything.

ste93cry: Yup you're right, in this case it makes no sense to accept `null` values beside the default value of the argument

mfb:
Resolved merge conflict and added test coverage.

tests/EventTest.php
```diff
@@ -165,6 +165,38 @@ public function testToArrayWithBreadcrumbs(): void
...
+        $event = new Event();
+        $event->setMessage('foo @bar', ['@bar' => 'bar'], 'foo bar');
+
+        $data = $event->toArray();
+
+        $this->assertArrayHasKey('message', $data);
+        $this->assertSame($expected, $data['message']);
+    }
+
+    public function testGetMessage(): void
```
ste93cry: Instead of having multiple test conditions into a single test case please use a data provider

tests/EventTest.php
```diff
@@ -165,6 +165,38 @@ public function testToArrayWithBreadcrumbs(): void
...
+    {
+        $event = new Event();
+        $event->setMessage('foo @bar', ['@bar' => 'bar'], 'foo bar');
+        $this->assertSame('foo @bar', $event->getMessage());
```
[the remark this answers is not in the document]
ste93cry: It doesn't change much in this case, but since we're comparing strings it makes no sense (in this case, in others it makes) to use strict comparison. It doesn't really change much, both are good but since in the codebase I believe that for the most times we did the same I've asked to change it to maintain consistency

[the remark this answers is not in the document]
mfb: Hmm I use strict comparison everywhere as a habit, seems much safer.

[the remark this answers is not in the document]
ste93cry: It is safer indeed, but in this case since we have typehinted the return type strictly comparing is not needed since we are sure that both arguments are strings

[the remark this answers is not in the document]
mfb: Of course.. but I guess my assumption re: testing is that if code is refactored the typing could be wrong, so tests can be overly defensive.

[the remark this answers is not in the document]
ste93cry: I could not think of any case where a wrong typehint while refactoring code even not using strict comparison could make the test not fail if both values are not equal (still, in this specific case). Anyway I get your point, if you prefer to leave it as is it's ok. The other requested changes are still needed of course

ste93cry:
@mfb are you willing to make the requested changes? We would like to release `beta 2` asap and we would like to include this feature in that version

tests/EventTest.php
```diff
@@ -119,32 +119,27 @@ public function testToArray(): void
...
 
-    public function testToArrayWithMessageWithParams(): void
+    public function toArrayWithMessageDataProvider(): array
     {
-        $expected = [
-            'message' => 'foo %s',
-            'params' => ['bar'],
-            'formatted' => 'foo bar',
+        return [
+            [['foo bar'], 'foo bar'],
```
ste93cry: Please split this array into multiple lines (have a look at how other data providers are formatted)

tests/EventTest.php
```diff
@@ -119,32 +119,27 @@ public function testToArray(): void
...
+            [['foo %s', ['bar']], ['message' => 'foo %s', 'params' => ['bar'], 'formatted' => 'foo bar']],
+            [['foo @bar', ['@bar' => 'bar'], 'foo bar'], ['message' => 'foo @bar', 'params' => ['@bar' => 'bar'], 'formatted' => 'foo bar']],
```
ste93cry: Please split this array into multiple lines (have a look at how other data providers are formatted). Also please always use the sprintf notation for the parameters, this way we can be sure that the function won't be called (if it does and the number of parameters is not equal or greater than the number of parameters used in the string it throws a warning)

tests/EventTest.php
```diff
@@ -119,32 +119,27 @@ public function testToArray(): void
...
-    public function testToArrayWithMessageWithParams(): void
+    public function toArrayWithMessageDataProvider(): array
     {
-        $expected = [
-            'message' => 'foo %s',
-            'params' => ['bar'],
-            'formatted' => 'foo bar',
+        return [
+            [['foo bar'], 'foo bar'],
+            [['foo %s', ['bar']], ['message' => 'foo %s', 'params' => ['bar'], 'formatted' => 'foo bar']],
```
ste93cry: Please split this array into multiple lines (have a look at how other data providers are formatted)

tests/EventTest.php
```diff
@@ -165,36 +160,25 @@ public function testToArrayWithBreadcrumbs(): void
...
-        $this->assertSame(['@bar' => 'bar'], $event->getMessageParams());
-        $this->assertSame('foo bar', $event->getMessageFormatted());
-
-        $event = new Event();
-        $event->setMessage('foo %s', ['bar']);
-        $this->assertSame('foo %s', $event->getMessage());
-        $this->assertSame(['bar'], $event->getMessageParams());
-        $this->assertNull($event->getMessageFormatted());
+        return [
+            [['foo %s', ['bar']], ['message' => 'foo %s', 'params' => ['bar'], 'formatted' => NULL]],
```
ste93cry: Please split this array into multiple lines (have a look at how other data providers are formatted)

tests/EventTest.php
```diff
@@ -165,36 +160,25 @@ public function testToArrayWithBreadcrumbs(): void
...
+            [['foo @bar', ['@bar' => 'bar'], 'foo bar'], ['message' => 'foo @bar', 'params' => ['@bar' => 'bar'], 'formatted' => 'foo bar']],
```
ste93cry: Please split this array into multiple lines (have a look at how other data providers are formatted). Also please always use the sprintf notation for the parameters, this way we can be sure that the function won't be called (if it does and the number of parameters is not equal or greater than the number of parameters used in the string it throws a warning)

tests/EventTest.php
```diff
@@ -119,32 +119,27 @@ public function testToArray(): void
         $this->assertEquals($expected, $event->toArray());
     }
 
-    public function testToArrayWithMessage(): void
+    /**
+     * @dataProvider toArrayWithMessageDataProvider
+     */
+    public function testToArrayWithMessage(array $message, $expectedValue): void
```
ste93cry: Please rename this variable to something that better reflect what it stores (e.g. `$setMessageMethodArguments`)

tests/EventTest.php
```diff
@@ -119,32 +119,27 @@ public function testToArray(): void
...
     {
         $event = new Event();
-        $event->setMessage('foo bar');
+        call_user_func_array([$event, 'setMessage'], $message);
```
ste93cry: Please add a blank line above this one to improve readability

tests/EventTest.php
```diff
@@ -165,36 +160,25 @@ public function testToArrayWithBreadcrumbs(): void
...
     {
-        $expected = [
-            'message' => 'foo @bar',
-            'params' => ['@bar' => 'bar'],
-            'formatted' => 'foo bar',
-        ];
-
         $event = new Event();
-        $event->setMessage('foo @bar', ['@bar' => 'bar'], 'foo bar');
+        call_user_func_array([$event, 'setMessage'], $message);
```
ste93cry: Please add a blank line above this one to improve readability

tests/EventFactoryTest.php
```diff
@@ -92,6 +92,20 @@ public function createWithPayloadDataProvider()
                     ],
                 ],
             ],
+            [
+                [
+                    'message' => 'testMessage @param',
```
ste93cry: Assuming that this test case covers the case in which a `message_formatted` param is specified in the payload, please use the `sprintf` notation for the value of the `message` key and a totally different message for the value of the `formatted` key so that we can catch if the override works

tests/EventTest.php
```diff
@@ -165,6 +160,27 @@ public function testToArrayWithBreadcrumbs(): void
         $this->assertSame($breadcrumbs, $data['breadcrumbs']['values']);
     }
 
+    /**
+     * @dataProvider getMessageDataProvider
+     */
+    public function testGetMessage(array $message, array $expectedValue): void
+    {
+        $event = new Event();
+        \call_user_func_array([$event, 'setMessage'], $message);
```
ste93cry: Please add a blank line above this one to improve readability

tests/EventTest.php
```diff
@@ -165,6 +160,27 @@ public function testToArrayWithBreadcrumbs(): void
         $this->assertSame($breadcrumbs, $data['breadcrumbs']['values']);
     }
 
+    /**
+     * @dataProvider getMessageDataProvider
+     */
+    public function testGetMessage(array $message, array $expectedValue): void
```
ste93cry: Please rename this variable to something that better reflect what it stores (e.g. `$setMessageMethodArguments`)

mfb:
Feel free to make changes to my branch if you need to, as I won't be able to work on this again until Sunday.

ste93cry:
friendly ping @mfb I would prefer the contributor to finish the PR if possible. Changes should be relatively quick and easy

tests/EventTest.php
mfb: It looks like some data provider arrays are split into multiple lines while some are not.  Cleaning all of those up seems out of scope for this PR though.

tests/EventTest.php
mfb: Not sure I'm following you, but by sprintf notation I'm guessing you mean e.g. `'foo %bar'` as the uninterpolated message (rather than `'foo @bar'`)?

ste93cry: Exactly, since we're using `vsprintf` to format the message, using the `sprintf` notation we can be sure of these things:

- if the test case is testing that the message should be autoformatted but there are no parameters then the `vsprintf` function is not called
- if the test case is testing that the message should be autoformatted the `sprintf` function is called with the correct number of arguments
- if the test case is testing that the message should be equal to the `formatted` message passed from the user and there are parameters the `vsprintf` function is not called anyway and the correct message is used

Does it makes sense?

tests/EventTest.php
ste93cry: I meant the data provider of this file. You should format the arrays one item per line if possible, e.g.

```php
[
    [
        'foo @bar',
        [
            '@bar' => 'bar',
        ],
        'foo bar',
    ],
    [
        'message' => 'foo @bar',
        'params' => [
            '@bar' => 'bar',
        ],
        'formatted' => 'foo bar',
    ],
]
```

ste93cry:
Build is failing due to CS issues :disappointed:

tests/EventTest.php
mfb: FYI there was a pre-existing data provider in this file that has "packed" array syntax:
```
            ['modules', ['foo' => '0.0.1', 'bar' => '0.0.2'], ['modules' => ['foo' => '0.0.1', 'bar' => '0.0.2']]],
```

mfb:
It looks like the code style issue is in a file I didn't modify..?

tests/EventTest.php
ste93cry: Thank you for reporting this, some inconsistencies are indeed possible since style is not forced by PHPCS-Fixer. However I would prefer usage of the "expanded" version

Jean85:
> It looks like the code style issue is in a file I didn't modify..?

It's possible that the defaults of CS Fixer changed somehow, you should be able to apply the same fix by doing `composer update` and re-running the fixer

ste93cry:
Thank you for your hard work and the patience, LGTM!

PR #753, opened 2019-01-28
Remove unused 'serialize_all_object' option

Jean85:
While working on the Symfony integration I've found this relic, left over from 1.x.


Jean85:
I think the upgrade guide is wrong then! At the very least, that's an option for the serializer, not a global one. I'll add an edit for that.

PR #755, opened 2019-01-28
Rename excluded_paths option to make it more uniform

Jean85:
This is to make the option name more uniform. The previous name was a remainder from 1.x.


ste93cry:
According to the [Unified API documentation](https://docs.sentry.io/error-reporting/configuration/?platform=php) the option should be called `in_app_exclude`. There is an open issue somewhere where it was suggested to have the ability to specify glob patterns for this option, however it's still pending implementation

Jean85:
Glob patterns would have to be implemented from scratch, it would be really hard.

BTW, since there is a common name specified from the unified API, I'll switch to that.

HazAT:
Thanks! Perfect!
We were also discussing implementing `in_app_include` to have a whitelist.
But for now this change is really good.

Issue #756, opened 2019-01-28
Options::normalizeAbsolutePath is wrong under Windows


Jean85:
I did not test this, but I believe we currently have this bug: `Options::normalizeAbsolutePath` is written with the (wrong) assumption that we are under a UNIX-style filesystem, which would brake under Windows, which do not use a starting dir separator for absolute paths.

Should we add a Windows CI job? IIRC Travis just added those...

CharlotteDunois:
The Windows filesystem normalizes `/` to `\`. But I would not rely on this and instead normalize the path correctly in `normalizeAbsolutePath`, since there *could*  be edge cases which don't get normalized by Windows. I'm not aware of any, but it's a good idea to be one step ahead nevertheless.

AndreKR:
What is the desired normalization result here? Why normalize at all?

My prefixes end with a slash/backslash, so that the reported paths start with `"src\..."`. In Raven client 1.10 this worked properly. Now the prefixes are normalized, cutting off the backslash, so the reported paths start with `"\src\..."`.

In general, I find it questionable to call `realpath()` on the provided paths. If I symlink my application, there's generally a reason for that and I want to see the symlinked path.

stayallive:
I believe we fixed this a long time ago already and we have Windows CI jobs running now.

So I think wel can close this as fixed in 2.0? @Jean85?

---

@AndreKR if the SDK is currently doing something weird with normalizing please create a new issue with some more info on what is happening and what you are seeing and on which SDK version you are so we can investigate?

Jean85:
@AndreKR matching relative paths is an issue and hard to do, and I'm not sure we ever did that. Your reporting was probably working fine due to other options, like the excludes... Paths are considered in as defaults.

I would also add that 1.x is no longer supported, and yes, we have a Windows CI job, so if there's an issue, you could simply submit a PR with a failing test, so we can take it from there. The test doesn't need to spin up the whole app, just using the paths as input of the `Options` class is enough, and reducible to a unit test case.

AndreKR:
@Jean85 Here it is: https://github.com/getsentry/sentry-php/pull/1163

PR #757, opened 2019-01-28
[WIP] Add Windows CI job

Jean85:
This WIP is to try and add a Windows CI job, and avoid issues like #756.

If I will not be able to have a green build, I'll add it to the allowed failures, and fix the SDK in a separate PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ php:
     - nightly
 
 env:
-    - dependencies=lowest
     - dependencies=highest
+    - dependencies=lowest
 
 matrix:
     fast_finish: true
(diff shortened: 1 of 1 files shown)


Jean85:
Damn, there's an open issue that make the install hang if there's a secret variable: https://travis-ci.community/t/windows-instances-hanging-before-install/250/4

And we have one of those...

Jean85:
Should we pursue (at least temporarely) a different platform for a Windows, like AppVeyor?

@HazAT @ste93cry @stayallive WDYT?

ste93cry:
Imho we should indeed have the Windows build. Travis CI had Windows support in beta since a few months, but I think we will need to wait way more to see stable support for it so probably going with AppVeyor is the best solution at the moment. I don't like to have builds splitted across different CI tools but I don't see any other way

Jean85:
Closing since we're now using AppVeyor from #758

PR #759, opened 2019-01-30
Pass options to error integration

### LINKED ISSUE #746: [2.0] error_types option doesn't work

HazAT:
Fixes #746

I hope I got the comparison right 🙈 
https://github.com/getsentry/sentry-php/compare/fix/error_types-option?expand=1#diff-9a950f4db299479d9933d02a78100059R40


ste93cry:
I totally disagree with the solution implemented in this PR, we are leaking part of the purpose of the error handler into an integration just because the current architecture has some limitations in this specific case. I would like to think a bit more about it since I'm working on a different solution in my free time

HazAT:
@ste93cry I do not fully understand your concerns.
We could probably remove the part from the `ErrorHandler` that lets you set which error it should handle.
I would make it that the `ErrorHandler` always receives all errors. 

Yes, the `ErrorHandler` should be a singleton, since dealing with multiple `set_error_handler` etc. is very complex so having one class responsible for it is enough.

Additionally what we could do is, instead of having `ErrorHandler::register` we could have something like `ErrorHandler::addCallback` so basically an array of "listeners". In case of an error the `ErrorHandler` calls all listeners with the error and level.

It's up to the listener (in our case the `ErrorHandlerIntegration`) to decide what to do with this error.

I think this approach is very straight forward and easy, we do similar stuff in other SDKs already.

tests/phpt/send_only_warning.phpt
```diff
@@ -0,0 +1,44 @@
...
+]);
+
+error_reporting(E_ALL);
+
+$client = Hub::getCurrent()->getClient();
+
+/** @var \Sentry\Transport\HttpTransport $transport */
+$transport = Assert::getObjectAttribute($client, 'transport');
+
+trigger_error("Cannot divide by zero", E_USER_ERROR);
```
Jean85: This test is wrong: no code below this point is ever executed, this error halts execution.

src/ErrorHandler.php
```diff
@@ -41,10 +41,15 @@ final class ErrorHandler
...
     /**
-     * @var int The errors that will be catched by the error handler
+     * @var int The errors that will be caught by the error handler
      */
     private $capturedErrors = E_ALL;
 
+    /**
+     * @var int The last caught error type
+     */
+    private $lastErrorType = 0;
```
Jean85: Why using a property and not passing it as every invocation of `handleException`?

PR #762, opened 2019-02-04
Error handler as singleton and with listeners

Jean85:
Sorry for opening a new PR, but this is a new approach to solve #746, different from #759.

In here, as discussed together, we use the integrations to attach two listeners to the error handler, moving away the issue of error level filtering from that class.

I like this approach very much:
 * it's cleaner
 * tests are better and cleaner
 * the end user can use the listeners to attach custom reporting
 * the end user can even use the default integrations to decide which auto-reporting to enable (errors or exceptions)


HazAT:
We should not forget to update docs: https://github.com/getsentry/sentry-docs/blob/master/src/collections/_documentation/platforms/php/default-integrations.md

src/ErrorHandler.php
```diff
@@ -337,4 +295,32 @@ private function cleanBacktraceFromErrorHandlerFrames(array $backtrace, string $
...
     }
+
+    /**
+     * @param \ErrorException $errorAsException The error to be passed onto listeners
+     */
+    private function invokeErrorListeners(\ErrorException $errorAsException): void
+    {
+        foreach ($this->errorListeners as $listener) {
+            try {
+                $listener($errorAsException);
```
HazAT: Shouldn't we use `call_user_func` here?

src/ErrorHandler.php
```diff
@@ -337,4 +295,32 @@ private function cleanBacktraceFromErrorHandlerFrames(array $backtrace, string $
...
+    }
+
+    /**
+     * @param \Throwable $throwable The exception to be passed onto listeners
+     */
+    private function invokeExceptionListeners(\Throwable $throwable): void
+    {
+        foreach ($this->exceptionListeners as $listener) {
+            try {
+                $listener($throwable);
```
HazAT: Shouldn't we use `call_user_func` here?

HazAT:
Otherwise, this PR really looks good, like the approach. I will close my PR :) 
Close https://github.com/getsentry/sentry-php/pull/759

src/ErrorHandler.php
Jean85: It's basically the same. We even skip a function invocation like this.

src/ErrorHandler.php
```diff
@@ -111,52 +100,72 @@ private function __construct(callable $callback, int $reservedMemorySize = self:
...
-     * @throws \ReflectionException
+     * @return ErrorHandler The ErrorHandler singleton
      */
-    public static function register(callable $callback, int $reservedMemorySize = self::DEFAULT_RESERVED_MEMORY_SIZE): self
+    public static function getRegisteredHandler(int $reservedMemorySize = null): self
     {
-        return new self($callback, $reservedMemorySize);
+        if (null === self::$registeredHandler) {
+            self::$registeredHandler = new self($reservedMemorySize ?? self::DEFAULT_RESERVED_MEMORY_SIZE);
+        } elseif (null !== $reservedMemorySize) {
```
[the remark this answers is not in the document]
Jean85: It's to avoid inconsistency. 

I need to expose the parameter at least for the first registration, so I cannot ignore it if the user calls it more than once. The user could not know that the error handler was registered before.

src/ErrorHandler.php
```diff
@@ -111,52 +100,72 @@ private function __construct(callable $callback, int $reservedMemorySize = self:
...
 
     /**
-     * Sets the PHP error levels that will be captured by the Raven client when
-     * a PHP error occurs.
+     * Adds a listener to the current error handler to be called upon each invoked captured error;
+     * if no handler is registered, this method will instantiate and register it.
      *
-     * @param int  $levels  A bit field of E_* constants for captured errors
-     * @param bool $replace Whether to replace or amend the previous value
+     * If you want to use a naked callable and not an object as a listener,
```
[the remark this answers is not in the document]
Jean85: I've removed that line because it was no true. I would prefer the interface to avoid issues that would arise only while handling errors, which would complicate the situation a lot.

I would move to `callable` only if we could validate the function signature, but that would be costly and error prone. @HazAT WDYT?

src/ErrorHandler.php
```diff
@@ -227,30 +226,23 @@ public function handleFatalError(array $error = null): void
 
         if (!empty($error) && $error['type'] & (E_ERROR | E_PARSE | E_CORE_ERROR | E_CORE_WARNING | E_COMPILE_ERROR | E_COMPILE_WARNING)) {
             $errorAsException = new \ErrorException(self::ERROR_LEVELS_DESCRIPTION[$error['type']] . ': ' . $error['message'], 0, $error['type'], $error['file'], $error['line']);
-        }
-
-        try {
```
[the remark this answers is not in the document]
Jean85: The if is not removed, I just simplified the flow. There's just one place where `$errorAsException` is instantiated, and it's at the line before the method call, already inside an `if`.

src/ErrorHandler.php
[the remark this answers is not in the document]
ste93cry: Renaming this method to something else like `register` or `registerOnce` along with a DocBlock description could help the user better understand what this method does and which rules it follows (e.g. it registers a singleton, so multiple calls have no effect). I fear that leaving open the possibility of changing after initialization the amount of reserved memory could have some implications that we don't know of and also I don't see any valid reason someone should do it

src/ErrorHandler.php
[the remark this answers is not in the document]
ste93cry: I don't think we should validate the signature since it would slow down the performance for nothing (it's up to the user to provide a valid callable). If in the future something like [HHVM fun](http://docs.hhvm.com/hack/reference/function/HH.fun/) will be available in PHP we will be able to check it. I like the interface concept but I fear that it could be limiting for simple use cases which doesn't require a whole class to be implemented. Also as said in another comment if we go that way then we cannot use `__invoke` or it will be impossible for a class to implement both the `ErrorHandlerInterface` and the `ExceptionHandlerInterface` or the methods will clash

src/ErrorHandler.php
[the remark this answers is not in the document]
Jean85: No issues whatsoever, the effect would be the same as registering once as with that memory. Only point where that changes is when setting to `0` while handling a fatal error, but that's a point of no return, PHP is crashing, so the user cannot make any mess anyway.

src/ErrorHandler.php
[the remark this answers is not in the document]
Jean85: I agree on changing `__invoke`; I will change it if we decide to maintain the interfaces, otherwise we're dropping them in favor of callables.

src/ErrorHandler.php
```diff
@@ -227,30 +226,23 @@ public function handleFatalError(array $error = null): void
...
-     * Handles the given exception by capturing it through the Raven client and
+     * Handles the given exception by passing it to all the listeners,
      * then forwarding it to another handler.
      *
      * @param \Throwable $exception The exception to handle
      *
      * @throws \Throwable
      *
-     * @internal
+     * @internal This method is public only because it's used with set_exception_handler
```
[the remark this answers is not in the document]
Jean85: I think it's out of scope here, refactoring this would require a big effort: the test is verifying that this method rethrows the exception, so doing that in a test without direct access to the method is complicated.

I would leave this out of this PR.

src/ErrorHandler.php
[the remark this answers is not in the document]
Jean85: @HazAT said that `callable` it's better. I've dropped the interfaces and switched to `callable`s.

src/ErrorHandler.php
```diff
@@ -283,31 +257,17 @@ public function handleException(\Throwable $exception): void
...
-     * @param int $previousThrownErrors The previous error mask
+     * @param int $reservedMemorySize The amount of memory to be reserved, is in char string length
      */
-    private function reRegister(int $previousThrownErrors): void
+    private static function setReservedMemory(int $reservedMemorySize): void
     {
-        if ($this->capturedErrors === $previousThrownErrors) {
-            return;
+        if ($reservedMemorySize <= 0) {
+            throw new \UnexpectedValueException('The $reservedMemorySize argument must be greater than 0.');
```
ste93cry: Isn't `InvalidArgumentException` a more appropriate exception in this case (since it talks about the argument and not a generic value)?

src/ErrorHandler.php
ste93cry: Since we switched to `callable` we now have to use `call_user_func`

src/ErrorHandler.php
ste93cry: I agree, since we switched to `callable` this is necessary

src/Integration/ErrorListenerIntegration.php
```diff
@@ -0,0 +1,49 @@
...
+{
+    /**
+     * @var Options The options, to know which error level to use
+     */
+    private $options;
+
+    /**
+     * ErrorListenerIntegration constructor.
+     *
+     * @param Options $options
```
ste93cry: Missing description of the parameter

src/ErrorHandler.php
```diff
@@ -337,4 +297,32 @@ private function cleanBacktraceFromErrorHandlerFrames(array $backtrace, string $
...
+            try {
+                $listener($errorAsException);
+            } catch (\Throwable $exception) {
+                // Do nothing as this should be as transparent as possible
+            }
+        }
+    }
+
+    /**
+     * @param \Throwable $throwable The exception to be passed onto listeners
```
ste93cry: Missing description of this method. Does it makes sense to have this code in a separate function (since it's used only one and the only thing it's doing is looping over the listeners)?

src/ErrorHandler.php
```diff
@@ -337,4 +297,32 @@ private function cleanBacktraceFromErrorHandlerFrames(array $backtrace, string $
 
         return $cleanedBacktrace;
     }
+
+    /**
+     * @param \ErrorException $errorAsException The error to be passed onto listeners
```
ste93cry: Missing description of the method

src/ErrorHandler.php
```diff
@@ -337,4 +297,32 @@ private function cleanBacktraceFromErrorHandlerFrames(array $backtrace, string $
...
+     */
+    private function invokeErrorListeners(\ErrorException $errorAsException): void
```
ste93cry: Since both the error and exception listeners have the same signature (they accept an exception instance) wouldn't it make sense to have just one method that iterates over a given parameter containing the list of callbacks? So instead of just the exception we also pass to this method the array of listeners to invoke

src/ErrorHandler.php
```diff
@@ -227,30 +208,23 @@ public function handleFatalError(array $error = null): void
...
             $errorAsException = new \ErrorException(self::ERROR_LEVELS_DESCRIPTION[$error['type']] . ': ' . $error['message'], 0, $error['type'], $error['file'], $error['line']);
-        }
-
-        try {
-            if (null !== $errorAsException) {
-                $this->handleException($errorAsException);
-            }
-        } catch (\Throwable $errorAsException) {
-            // Ignore this re-throw
+            $this->invokeErrorListeners($errorAsException);
```
ste93cry: Please add a blank line above this one to improve readability

src/ErrorHandler.php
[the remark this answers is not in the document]
ste93cry: One more thing that doesn't convince me at all is that since this is a getter I don't expect it to edit the object that it returns

src/ErrorHandler.php
Jean85: You're right, it's a reminder of the interfaces that I dropped before.

src/ErrorHandler.php
Jean85: Good suggestion. Will refactor.

src/ErrorHandler.php
Jean85: Nope it's not necessary; I've added a test to prove it: https://github.com/getsentry/sentry-php/pull/762/files#diff-c290a9b0e216f21a0de832495713aa14R313

src/ErrorHandler.php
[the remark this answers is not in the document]
Jean85: Well, it's already a magic getter since it triggers the singleton instantiation. Maybe we can rename it?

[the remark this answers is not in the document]
ste93cry: But triggering the singleton instantiation is different from being able to edit it on each subsequent call to the method, isn't it?

src/ErrorHandler.php
Jean85: I've added a test that proves that `call_user_func` is not needed.

src/ErrorHandler.php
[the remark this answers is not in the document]
HazAT: I agree with @ste93cry rename it to registerOnce and remove the setter. That way it make the most sense to me.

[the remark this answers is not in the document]
Jean85: Done. I had to move a test to a PHPT.

ste93cry:
Can you please also add an entry to the `CHANGELOG.md` file?

src/ErrorHandler.php
```diff
@@ -111,52 +100,53 @@ private function __construct(callable $callback, int $reservedMemorySize = self:
...
-        }
-
-        $this->reRegister($prev);
+        $handler = self::registerOnce();
+        $handler->errorListeners[] = $listener;
+    }
 
-        return $prev;
+    /**
+     * Adds a listener to the current error handler to be called upon each invoked captured exception;
```
ste93cry: As a "soft" limit usually the descriptions of the methods goes to a new line after ~ 80 chars

src/Integration/ExceptionListenerIntegration.php
```diff
@@ -0,0 +1,25 @@
...
+ * Sentry.
+ */
+final class ExceptionListenerIntegration implements IntegrationInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function setupOnce(): void
+    {
+        ErrorHandler::addExceptionListener(function (\Throwable $throwable): void {
```
ste93cry: Since this function doesn't need access to `$this` it can be `static`

tests/ErrorHandlerTest.php
```diff
@@ -6,153 +6,52 @@
...
-    }
+        $errorHandler = ErrorHandler::registerOnce();
 
-    public function captureAtDataProvider(): array
-    {
-        return [
-            [E_USER_NOTICE, false, E_ALL],
-            [E_USER_NOTICE, true, E_USER_NOTICE],
-        ];
+        $this->assertSame($errorHandler, ErrorHandler::registerOnce());
```
ste93cry: You can just inline the two calls to `ErrorHandler::registerOnce()`, we're not interested in reusing the instance so it makes no sense to store it into a local variable

tests/ErrorHandlerTest.php
```diff
@@ -6,153 +6,52 @@
...
-                return true;
-            }));
+        $listener = new StubErrorListener();
+        $errorLine = null;
 
         try {
-            $errorHandler = ErrorHandler::register($this->callbackMock);
-            $errorHandler->captureAt(0, true);
+            ErrorHandler::addErrorListener($listener);
+            $errorHandler = ErrorHandler::registerOnce();
```
ste93cry: Please add a blank line above this one

tests/ErrorHandlerTest.php
```diff
@@ -6,153 +6,52 @@
...
             $reflectionProperty->setAccessible(false);
 
-            $this->assertFalse($errorHandler->handleError(0, 'foo bar', __FILE__, __LINE__));
-
-            $errorHandler->captureAt(E_USER_NOTICE, true);
-
-            $this->assertFalse($errorHandler->handleError(E_USER_WARNING, 'foo bar', __FILE__, __LINE__));
-            $this->assertFalse($errorHandler->handleError(E_USER_NOTICE, 'foo bar', __FILE__, 123));
+            $errorLine = __LINE__ + 1;
+            $this->assertFalse($errorHandler->handleError(E_USER_NOTICE, 'foo bar', __FILE__, __LINE__));
```
ste93cry: Please add a blank line above this one

tests/ErrorHandlerTest.php
```diff
@@ -6,153 +6,52 @@
...
+            $this->assertInstanceOf(\ErrorException::class, $exception);
+            $this->assertEquals(__FILE__, $exception->getFile());
+            $this->assertEquals($errorLine, $exception->getLine());
+            $this->assertEquals('User Notice: foo bar', $exception->getMessage());
+            $this->assertEquals(E_USER_NOTICE, $exception->getSeverity());
+
+            $backtrace = $exception->getTrace();
+
+            $this->assertGreaterThanOrEqual(2, $backtrace);
+
```
ste93cry: Please remove this blank line

tests/ErrorHandlerTest.php
```diff
@@ -161,39 +60,17 @@ public function testHandleError(): void
...
             ->method('__invoke')
             ->with(E_USER_NOTICE, 'foo bar', __FILE__, 123)
             ->willReturn($previousErrorHandlerErrorReturnValue);
 
+        $listener = new StubErrorListener();
+
         try {
-            $errorHandler = ErrorHandler::register($this->callbackMock);
+            ErrorHandler::addErrorListener($listener);
+            $errorHandler = ErrorHandler::registerOnce();
```
ste93cry: Please add a blank line above this one

tests/ErrorHandlerTest.php
```diff
@@ -204,6 +81,26 @@ public function testHandleErrorWithPreviousErrorHandler($previousErrorHandlerErr
...
+            $this->assertEquals('User Notice: foo bar', $exception->getMessage());
+
+            $backtrace = $exception->getTrace();
+
+            $this->assertGreaterThanOrEqual(2, $backtrace);
+
+            $this->assertEquals('handleError', $backtrace[0]['function']);
+            $this->assertEquals(ErrorHandler::class, $backtrace[0]['class']);
+            $this->assertEquals('->', $backtrace[0]['type']);
+
```
ste93cry: Please remove this blank line

tests/ErrorHandlerTest.php
```diff
@@ -204,6 +81,26 @@ public function testHandleErrorWithPreviousErrorHandler($previousErrorHandlerErr
...
+            $this->assertInstanceOf(\ErrorException::class, $exception);
+            $this->assertEquals(__FILE__, $exception->getFile());
+            $this->assertEquals(123, $exception->getLine());
+            $this->assertEquals(E_USER_NOTICE, $exception->getSeverity());
+            $this->assertEquals('User Notice: foo bar', $exception->getMessage());
+
+            $backtrace = $exception->getTrace();
+
+            $this->assertGreaterThanOrEqual(2, $backtrace);
+
```
ste93cry: Please remove this blank line

tests/ErrorHandlerTest.php
```diff
@@ -221,21 +118,12 @@ public function handleErrorWithPreviousErrorHandlerDataProvider(): array
...
-                $this->assertEquals('Parse Error: foo bar', $exception->getMessage());
-
-                return true;
-            }));
+        $listener = new StubErrorListener();
 
         try {
-            $errorHandler = ErrorHandler::register($this->callbackMock);
+            ErrorHandler::addErrorListener($listener);
+            $errorHandler = ErrorHandler::registerOnce();
```
ste93cry: Please add a blank line above this one

tests/ErrorHandlerTest.php
```diff
@@ -245,16 +133,25 @@ public function testHandleFatalError(): void
...
     public function testHandleFatalErrorWithNonFatalErrorDoesNothing(): void
     {
-        $this->callbackMock->expects($this->never())
-            ->method('__invoke');
+        $listener = new StubErrorListener();
 
         try {
-            $errorHandler = ErrorHandler::register($this->callbackMock);
+            ErrorHandler::addErrorListener($listener);
+            $errorHandler = ErrorHandler::registerOnce();
```
ste93cry: Please add a blank line above this one

tests/ErrorHandlerTest.php
```diff
@@ -288,24 +190,30 @@ public function testHandleException(): void
...
     {
+        $listenerCalled = false;
         $exception = new \Exception('foo bar');
 
-        $this->callbackMock->expects($this->once())
-            ->method('__invoke')
-            ->with($this->identicalTo($exception));
+        $listener = function (\Throwable $throwable) use ($exception, &$listenerCalled): void {
+            $listenerCalled = true;
+            $this->assertSame($exception, $throwable);
```
ste93cry: Please add a blank line above this one

tests/ErrorHandlerTest.php
```diff
@@ -264,19 +161,24 @@ public function testHandleFatalErrorWithNonFatalErrorDoesNothing(): void
...
     {
+        $listenerCalled = false;
         $exception = new \Exception('foo bar');
-
-        $this->callbackMock->expects($this->once())
-            ->method('__invoke')
-            ->with($this->identicalTo($exception));
+        $listener = function (\Throwable $throwable) use ($exception, &$listenerCalled): void {
+            $listenerCalled = true;
+            $this->assertSame($exception, $throwable);
```
ste93cry: Please add a blank line above this one

tests/ErrorHandlerTest.php
```diff
@@ -358,6 +278,50 @@ public function testHandleExceptionWithThrowingPreviousExceptionHandler(): void
...
+    /**
+     * @dataProvider callableProvider
+     */
+    public function testAddListener(callable $listener): void
+    {
+        $exception = new \Exception();
+
+        try {
+            ErrorHandler::addExceptionListener($listener);
+
```
ste93cry: Please remove this blank line

tests/ErrorHandlerTest.php
```diff
@@ -358,6 +278,50 @@ public function testHandleExceptionWithThrowingPreviousExceptionHandler(): void
...
+            $this->assertSame(2, $listenerCalled);
+            $this->assertSame($exception1, $captured1);
+            $this->assertSame($exception2, $captured2);
         }
     }
+
+    /**
+     * @dataProvider callableProvider
+     */
+    public function testAddListener(callable $listener): void
```
ste93cry: This test doesn't seems solid, if for some reason the exception is not thrown from `handleException` the test will probably just say that no assertion has been made

tests/ErrorHandlerTest.php
```diff
@@ -358,6 +278,50 @@ public function testHandleExceptionWithThrowingPreviousExceptionHandler(): void
...
+            restore_error_handler();
+            restore_exception_handler();
+        }
+    }
+
+    public function callableProvider(): array
+    {
+        $stubErrorListener = new StubErrorListener();
+
+        return [
```
ste93cry: Can we format this array into multiple lines to improve readability?

tests/ErrorHandlerTest.php
```diff
@@ -358,6 +278,50 @@ public function testHandleExceptionWithThrowingPreviousExceptionHandler(): void
...
+            ErrorHandler::registerOnce()->handleException($exception);
+        } catch (\Throwable $rethrownException) {
+            $this->assertSame($exception, $rethrownException);
+        } finally {
+            restore_error_handler();
+            restore_exception_handler();
+        }
+    }
+
+    public function callableProvider(): array
```
ste93cry: I'm not sure it's worth to have and maintain this test, I would just use `call_user_func` all over the codebase and we're sure that if something that's an incorrect `callable` is passed to the functions it's user's fault and not our

tests/phpt/error_handler_refuses_negative_value.phpt
```diff
@@ -0,0 +1,27 @@
+--TEST--
+Test that the error handler is called regardless of the current `error_reporting` setting
```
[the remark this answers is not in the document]
Jean85: It's not, since the handler is a singleton, so it would be already registered by previous tests. The PHPT is run in isolation so there no such issue.

src/Integration/ExceptionListenerIntegration.php
Jean85: That would break the interface, and limit the extension possibilities for the end user.

tests/ErrorHandlerTest.php
Jean85: Good catch!

src/Integration/ExceptionListenerIntegration.php
ste93cry: Sorry I was not clear enough, I mean the lambda function. You can declare it as `static function (\Throwable $throwable): void` if you don't need to access `$this` inside it: http://php.net/manual/en/functions.anonymous.php#functions.anonymous-functions.static

tests/phpt/error_handler_refuses_negative_value.phpt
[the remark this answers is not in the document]
ste93cry: Then I believe that every single test of the error handler that we have in PHPUnit should be moved to PHPT or we should somehow (e.g. with reflection) unset the singleton variable or we will leak state between multiple tests and this is not good. Since we made the handler singleton here it's the correct PR to do it, regardless of the solution we choose

tests/ErrorHandlerTest.php
Jean85: I can understand the maintenance cost objection. I've changed the test to check for exceptions from listeners.

RobinHoutevelts:
Was just about to complain that `\Sentry\ErrorHandler::handleError` also called the previous**Exception**Handler with the handmade \ErrorException.

I'm not sure this PR also fixes that but I'll check it out when I find the time.

Either way this is a nice improvement 👍

Jean85:
We completely changed the approach there, splitting in two separate listeners, so that should be fixed!

ste93cry:
It's not, @RobinHoutevelts is right in saying that we're still calling the `$previousExceptionHandler` in the `handleException` method regardless of the type of exception. We should investigate if this is an unintended behavior and in case fix it

Jean85:
But that's no longer a problem, because we're no longer invoking the `handleException` from `handleError` nor `handleFatalError` methods, thanks to the listeners.

I broke that (not ideal) chain with this PR.

ste93cry:
You're right, sorry for the mistake then

PR #763, opened 2019-02-04
Fixes #761

### LINKED ISSUE #761: Stacktrace and FatalErrorException problem
Symfony and Laravel, and probably many other framework uses a Symfony\Component\Debug\Exception\FatalErrorException in its own Fatal error handling process before passing it to Sentry.
The problem is it doesn't return the stacktrace  in a standard debug_backtrace format, but replaces the arguments array indexes to the name of the variable if using xdebug.
`
[args] => Array
(
[0] => true
)
`
becomes
`
[args] => Array
(
[variableName] => true
)
`
This causes Sentry\StackTrace::getFrameArgumentsValues function to emit a Warning: A non-numeric value encountered while processing the above Exception. This whole thing makes debugging much more difficult.
If you think this is not edge case enough to warrant a fix, i would be happy to start working on a PR.

sztyup:


src/Stacktrace.php
```diff
@@ -301,14 +301,18 @@ protected function stripPrefixFromFilePath(string $filePath): string
      */
     protected function getFrameArgumentsValues(array $frame): array
     {
-        if (!isset($frame['args'])) {
+        if (!isset($frame['args']) || empty($frame['args'])) {
```
stayallive:
```suggestion
        if (empty($frame['args'])) {
```

The `empty` does also check for `isset` so it can be combined.

sztyup: Indeed :)

src/Stacktrace.php
```diff
@@ -301,14 +301,18 @@ protected function stripPrefixFromFilePath(string $filePath): string
...
-        if (!isset($frame['args'])) {
+        if (empty($frame['args'])) {
             return [];
         }
 
         $result = [];
 
-        foreach ($frame['args'] as $index => $argument) {
-            $result['param' . ($index + 1)] = $this->serializeArgument($argument);
+        if (\is_string(array_keys($frame['args'])[0])) {
```
ste93cry: Instead of this `if`/`else` statement with a combination of `foreach` and `array_map` couldn't we just loop as before and compute the key of the `$result` item according if the `$index` is a string or a number. POC:

```php
$index = 0;

foreach ($frame['args'] as $argumentName => $argumentValue) {
    $argumentName = is_string($argumentName) ?: 'param' . (++$index);

    $result[$argumentName] = $this->serializeArgument($argumentValue);
}
```

tests/StacktraceTest.php
```diff
@@ -297,6 +297,25 @@ public function testGetFrameArgumentsDoesNotModifyCapturedArgs(): void
...
+            'file' => __DIR__ . '/resources/a.php',
+            'line' => 9,
+            'args' => [
+                'foo' => 'bar',
+                'alice' => 'bob',
+            ],
+            'function' => 'a_test',
+        ];
+
+        $stacktrace = new Stacktrace(new Options(['max_value_length' => 5]), $this->serializer, $this->representationSerializer);
```
ste93cry: Any reason you set the `max_value_length` option?

tests/StacktraceTest.php
```diff
@@ -297,6 +297,25 @@ public function testGetFrameArgumentsDoesNotModifyCapturedArgs(): void
         $this->assertEquals($result['param2']['key'], 'xxxxx');
     }
 
+    public function testPreserveXdebugFrameArgumentNames(): void
```
ste93cry: Would it be possible to refactor this test to be similar to the others that use a JSON fixture and the `assertFrameEquals` method?

sztyup: assertFrameEquals does not check frame arguments :/

tests/StacktraceTest.php
sztyup: Fixed

PR #768, opened 2019-02-07
Remove link to unmaintained Nette community integration

WyskyNet:
Salamek/raven-nette is abandoned package and can be replaced by kdyby/monolog

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ The following integrations are fully supported and maintained by the Sentry team
 
 The following integrations are available and maintained by members of the Sentry community.
 
-- [Nette](https://github.com/Salamek/raven-nette)
+- [Nette](https://github.com/Kdyby/Monolog)
 - [ZendFramework](https://github.com/facile-it/sentry-module)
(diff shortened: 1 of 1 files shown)


ste93cry:
I agree, please just remove the entry from the list

Jean85:
It's still a community-supported integration, maybe we can just rename the link?

ste93cry:
The original package is not supported anymore, and the suggested one is not sentry-specific as said by @stayallive. If there is not a replacement for the package we should just remove it instead of linking a deprecated package

stayallive:
@Jean85 the replacement only adds monolog not even a note on how to add Sentry to that in the documentation (which is not wrong or a bad thing), so we should not link from Sentry that that is the way to go for Nette without it being either Sentry specific or have a note about how to integrate Sentry, no point in it.

Nothing wrong with the Nette Monolog package! And I don't expect them to add a piece of Sentry documentation or for that matter any of the other thousands of Monolog plugins.

@WyskyNet thanks for bringing to our attention the Nette package is no longer supported. Much appreciated!

PR #770, opened 2019-02-07
Update gitattributes

rvitaliy:
- remove useless files for production from package
- configure git to automatic line endings normalized

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,5 @@
+* text=auto
+
 /tests export-ignore
 /.appveyor.yml export-ignore
 /.craft.yml export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,15 @@
-* text=auto
+# Auto-detect text files, ensure they use LF.
+*       text=auto eol=lf
 
+# These files are always considered text and should use LF.
+# See core.whitespace @ http://git-scm.com/docs/git-config for whitespace flags.
+*.php   text eol=lf whitespace=blank-at-eol,blank-at-eof,space-before-tab,tab-in-indent,tabwidth=4 diff=php
+*.json  text eol=lf whitespace=blank-at-eol,blank-at-eof,space-before-tab,tab-in-indent,tabwidth=4
+*.yml   text eol=lf whitespace=blank-at-eol,blank-at-eof,space-before-tab,tab-in-indent,tabwidth=4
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1,3 +1,14 @@
-/examples export-ignore
-/docs export-ignore
+* text=auto
```
ste93cry: I don't think we should add this since it changes the behavior of how files are committed in the repository regardless of the operating system you're working on. According to the documentation:

>Setting the text attribute on a path enables end-of-line normalization and marks the path as a text file. End-of-line conversion takes place without guessing the content type.
>
>If the text attribute is unspecified, Git uses the core.autocrlf configuration variable to determine if the file should be converted.

If I don't recall wrong we have some files that must be treated as binary, and some others whose line endings should not change since we're relying on them being LF in the tests. Indeed normalization of things is always a good thing imho, but in this case I believe we should not force this setting to everyone. If line endings change for some reason it will be reported anyway in the diff

rvitaliy: https://help.github.com/articles/dealing-with-line-endings/#per-repository-settings
`* text=auto` = `Git will handle the files in whatever way it thinks is best. This is a good default option.`
it's not affect binary or unknown file types, but prevent bad configuration of `core.autocrlf`

Jean85: If it doesn't affect binaries, we should be good.

rvitaliy: it should not impact binary or unknown files

ste93cry: After having a look at how other repositories use this feature and at the Git documentation please add the following content too, then LGTM!

```gitattributes
*.php text whitespace=blank-at-eol,blank-at-eof,space-before-tab,tab-in-indent,tabwidth=4 diff=php
*.json text whitespace=blank-at-eol,blank-at-eof,space-before-tab,tab-in-indent,tabwidth=4
*.yml text whitespace=blank-at-eol,blank-at-eof,space-before-tab,tab-in-indent,tabwidth=4
*.md text whitespace=blank-at-eol,blank-at-eof
*.rst text whitespace=blank-at-eol,blank-at-eof
```

rvitaliy: i got inspiration from https://github.com/composer/composer/blob/master/.gitattributes and add `eol=lf` to force always to use `LF` for this files

ste93cry:
Thank you for your contribution, very much appreciated :smile:

PR #772, opened 2019-02-12
feat: Change readme to guide to meta package installation

HazAT:
Also ref: https://github.com/getsentry/sentry-docs/pull/768


HazAT:
I want to merge this as soon as we drop the `beta` tag.

PR #775, opened 2019-02-22
feat: Add http proxy option for curl client

HazAT:


src/Options.php
```diff
@@ -619,6 +619,28 @@ public function setMaxValueLength(int $maxValueLength): void
         $this->options = $this->resolver->resolve($options);
     }
 
+    /**
+     * Gets the http proxy setting.
+     *
+     * @return string
```
ste93cry: I don't know why PHPStan doesn't report this, but the docblock is incorrect since the return value can be `null`

src/Options.php
```diff
@@ -619,6 +619,28 @@ public function setMaxValueLength(int $maxValueLength): void
...
+     */
+    public function getHttpProxy(): ?string
+    {
+        return $this->options['http_proxy'];
+    }
+
+    /**
+     * Sets the http proxy. Be aware this option only works when curl client is used.
+     *
+     * @param string $httpProxy The http proxy
```
ste93cry: I don't know why PHPStan doesn't report this, but the docblock is incorrect since the parameter is `nullable`

src/ClientBuilder.php
```diff
@@ -314,6 +314,20 @@ private function createTransportInstance(): TransportInterface
 
         $this->messageFactory = $this->messageFactory ?? MessageFactoryDiscovery::find();
         $this->uriFactory = $this->uriFactory ?? UriFactoryDiscovery::find();
+
+        if ($this->options->getHttpProxy()) {
+            try {
+                $curlClientClass = 'Http\Client\Curl\Client';
+                if (class_exists($curlClientClass)) {
```
ste93cry: Please add a blank line above this one

src/ClientBuilder.php
```diff
@@ -314,6 +314,20 @@ private function createTransportInstance(): TransportInterface
 
         $this->messageFactory = $this->messageFactory ?? MessageFactoryDiscovery::find();
         $this->uriFactory = $this->uriFactory ?? UriFactoryDiscovery::find();
+
+        if (null === $this->httpClient && null !== $this->options->getHttpProxy()) {
+            try {
```
ste93cry: This `try`/`catch` block is misleading, since it's catching any exception to rethrow with a generic message that says that the cURL client must be installed while the problem may be something else entirely. What about removing it and just let a possible error to be thrown?

ste93cry:
I'm entirely :-1: on this change, there have been a lot of discussions already about why the SDK client should not know the HTTP client implementation (this is why we have the concept of transports). I obviously see the convenience of having such option available and its value, but I strongly believe that if someone wants to set it then he has to set its own client. It's not that hard after all and we're devs, so it should not be a problem for us to understand how something we're using works since it's our job...

src/ClientBuilder.php
HazAT: Restructured the code.

src/ClientBuilder.php
```diff
@@ -314,6 +315,17 @@ private function createTransportInstance(): TransportInterface
 
         $this->messageFactory = $this->messageFactory ?? MessageFactoryDiscovery::find();
         $this->uriFactory = $this->uriFactory ?? UriFactoryDiscovery::find();
+
+        if (null === $this->httpClient && null !== $this->options->getHttpProxy()) {
```
Jean85: We should make this fail hard if a client is already in, or the built client will be silently without proxy.

tests/ClientBuilderTest.php
```diff
@@ -290,6 +290,22 @@ public function testCreateWithNoOptionsIsTheSameAsDefaultOptions(): void
...
+        $options = new Options([
+            'dsn' => 'http://public:secret@example.com/sentry/1',
+            'http_proxy' => 'some-proxy',
+        ]);
+
+        $clientBuilder = new ClientBuilder($options);
+        $clientBuilder->setHttpClient($this->createMock(HttpAsyncClient::class));
+
+        $this->expectException(\RuntimeException::class);
+        $this->expectExceptionMessage('http_proxy');
```
ste93cry: I don't know how this test can pass since the message is different from the one in the codebase. I believe that the reason is that this method just asserts that the string is contained in the message, but since it's not clear please use the full message

PR #776, opened 2019-02-22
ref: Make getScope private

HazAT:


tests/State/HubTest.php
```diff
@@ -292,4 +292,13 @@ public function testCaptureEvent(): void
 
         $this->assertEquals('2b867534eead412cbdb882fd5d441690', $hub->captureEvent(['message' => 'test']));
     }
+
+    private static function getScope($hub): Scope
```
Jean85: Why `static`?

Also:
```suggestion
    private static function getScope(Hub $hub): Scope
```

ste93cry: Even better, `HubInterface`

tests/State/HubTest.php
```diff
@@ -292,4 +293,13 @@ public function testCaptureEvent(): void
 
         $this->assertEquals('2b867534eead412cbdb882fd5d441690', $hub->captureEvent(['message' => 'test']));
     }
+
+    protected function getScope(HubInterface $hub): Scope
```
ste93cry: Since this class is `final` it makes no sense to have protected methods

tests/State/HubTest.php
```diff
@@ -292,4 +293,13 @@ public function testCaptureEvent(): void
...
+    {
+        $class = new \ReflectionClass($hub);
```
ste93cry: We can use `ReflectionMethod` directly instead of passing for the class metadata

tests/SdkTest.php
```diff
@@ -94,7 +94,13 @@ public function testAddBreadcrumb(): void
 
         addBreadcrumb($breadcrumb);
 
-        $this->assertSame([$breadcrumb], Hub::getCurrent()->getScope()->getBreadcrumbs());
+        $hub = Hub::getCurrent();
+
+        $class = new \ReflectionClass($hub);
```
ste93cry: Same as above, we can use `ReflectionMethod` directly

tests/ClientTest.php
```diff
@@ -274,8 +274,14 @@ public function testHandlingExceptionThrowingAnException(): void
 
         $client = new Client(new Options(), $transport, $this->createEventFactory());
 
-        Hub::getCurrent()->bindClient($client);
-        $client->captureException($this->createCarelessExceptionWithStacktrace(), Hub::getCurrent()->getScope());
+        $hub = Hub::getCurrent();
+        $hub->bindClient($client);
+
+        $class = new \ReflectionClass($hub);
```
ste93cry: We can use `ReflectionMethod` directly

ste93cry:
I don't really like that we have to use reflection to test the internal behavior of an object, however at the moment I don't see any other way around so LGTM

PR #777, opened 2019-02-22
Support HTTPlug 2 / PSR-18

Jean85:
With this PR, I allow the usage of HTTPlug 2, that is PSR-18 compatible.

Unfortunately, the Curl client that uses HTTPlug 2 will be a new major (2.0), which hasn't released yet, so I've added a dedicated CI job to test it explicitly, replacing it with the Guzzle adapter.


tests/phpt/out_of_memory.phpt
```diff
@@ -20,23 +21,18 @@ while (!file_exists($vendor . '/vendor')) {
 require $vendor . '/vendor/autoload.php';
 
 init([
-    'dsn' => 'http://public:secret@local.host/1',
-    'send_attempts' => 1,
+    'before_send' => function (Event $event): ?Event {
+        Assert::assertTrue($event->getLevel()->isEqualTo(Severity::fatal()));
+        $error = $event->getExceptions()[0];
```
ste93cry: Please add an assert to check that the array is not empty. Also, I would move all the remaining asserts below this line

tests/phpt/out_of_memory.phpt
```diff
@@ -20,23 +21,18 @@ while (!file_exists($vendor . '/vendor')) {
...
+        Assert::assertContains('Allowed memory size', $error['value']);
+        echo 'Sending event...';
```
ste93cry: Please add a blank line before this one and remove the ending dots

tests/MockClientStrategy.php
```diff
@@ -0,0 +1,32 @@
...
+
+use Http\Client\HttpAsyncClient;
+use Http\Client\HttpClient;
+use Http\Discovery\Strategy\DiscoveryStrategy;
+use Http\Mock\Client as Mock;
+
+/**
+ * This class is ported from Http\Discovery\Strategy\MockClientStrategy.
+ *
+ * This versions provides the Mock client when an asycn client is required too.
```
ste93cry: `This version` (without the final `s`). Also it would make sense to open an issue or PR in HTTPlug repository and link it here so that we know we can remove this class when upstream will fix the problem

tests/MockClientStrategy.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\Tests;
+
+use Http\Client\HttpAsyncClient;
+use Http\Client\HttpClient;
+use Http\Discovery\Strategy\DiscoveryStrategy;
+use Http\Mock\Client as Mock;
```
ste93cry: Please alias the class to `HttpMockClient`, otherwise it's not clear what it is

composer.json
```diff
@@ -32,7 +32,6 @@
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.13",
         "monolog/monolog": "^1.3",
-        "php-http/curl-client": "^1.7.1",
         "php-http/message": "^1.5",
```
ste93cry: This should be a required dependency and not just a development one

src/ClientBuilder.php
```diff
@@ -321,13 +321,13 @@ private function createTransportInstance(): TransportInterface
                 throw new \RuntimeException('The `http_proxy` option does not work together with a custom client.');
             }
 
-            if (HttpAsyncClientDiscovery::safeClassExists(HttpCurlClient::class)) {
-                $this->httpClient = new HttpCurlClient(null, null, [
-                    CURLOPT_PROXY => $this->options->getHttpProxy(),
-                ]);
-            } else {
+            if (class_exists(HttpCurlClient::class)) {
```
HazAT: Shouldn't this be `!` 🤔 ?
```suggestion
            if (!class_exists(HttpCurlClient::class)) {
```

Jean85: Good catch!

PR #780, opened 2019-02-24
Stop using deprecated assertAttribute* methods of PHPUnit

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

PHPUnit is going to deprecate and drop the `assertAttribute*` methods in sebastianbergmann/phpunit#3338 and we use them in a few tests. This PR aims at refactoring such tests in some other way so that we are ready for a future upgrade of version of our test framework


tests/ClientBuilderTest.php
```diff
@@ -34,127 +35,142 @@ public function testCreate(): void
...
+
+        /** @var Plugin|MockObject $plugin2 */
+        $plugin2 = $this->createMock(Plugin::class);
+        $plugin2->expects($this->never())
+            ->method('handleRequest');
+
+        $client = ClientBuilder::create(['dsn' => 'http://public:secret@example.com/sentry/1'])
+            ->addHttpClientPlugin($plugin1)
+            ->addHttpClientPlugin($plugin1)
+            ->addHttpClientPlugin($plugin2)
```
Jean85: Could you move this add in the middle of the other two? Just to be sure that we're not popping a generic plugin, but instead removing the requested one.

Jean85:
If we need to fix the Travis builds triggers, we need to look at https://docs.travis-ci.com/user/pull-requests#double-builds-on-pull-requests

ste93cry:
I believe that the problem with Travis CI is that it's not running for draft PRs (new feature of GitHub), while AppVeyor does (for some reason, maybe different webhooks used)

stayallive:
This makes sense 👍 mark it as reviewable @ste93cry? Or is it not done yet?

ste93cry:
It is not finished yet, there are some points to think about how not to use the deprecated PHPUnit methods but keep testing those cases. Since this does not block any release, I will leave it in draft until I find a way to complete it

tests/phpt/spool_drain.phpt
```diff
@@ -1,68 +0,0 @@
...
-    private $spool;
-
-    public function __construct(SpoolInterface $spool)
-    {
-        $this->spool = $spool;
-    }
-
-    public function create(Options $options): TransportInterface
-    {
-        return new SpoolTransport($this->spool);
```
Jean85, line 42: Is this test not replaceable? Maybe we can extend/decorate the spool transport?

tests/ClientBuilderTest.php
```diff
@@ -96,14 +103,17 @@ public function testSetMessageFactory(): void
...
 
-        $clientBuilder = ClientBuilder::create(['dsn' => 'http://public:secret@example.com/sentry/1']);
-        $clientBuilder->setTransport($transport);
+        $client = ClientBuilder::create(['dsn' => 'http://public@example.com/sentry/1'])
+            ->setTransport($transport)
+            ->getClient();
 
-        $this->assertAttributeSame($transport, 'transport', $clientBuilder);
-        $this->assertAttributeSame($transport, 'transport', $clientBuilder->getClient());
+        $client->captureMessage('foo');
```
Jean85, line 116: Whenever possible, I would add an assert to check the mocked value, like here with the event ID.

tests/phpt/spool_drain.phpt
ste93cry: I deleted this entire test because there was no real point in keeping it: both the transport and the spool are already tested elsewhere, and this PHPT test was just combining the two by asserting that everything worked into a shutdown function which in no way influence the final result

PR #783, opened 2019-02-27
feat: Update readme and Upgrade2.0

HazAT:


UPGRADE-2.0.md
```diff
@@ -1,5 +1,38 @@
...
 
+Version `2.x` is a complete rewrite of the existing code base. The public API has been trimmed down to a minimum.
+The preferred way of using the SDK is through our static API.
+
+Here is a simple example to get started:
+
+```php
+\Sentry\init(['dsn' => '___PUBLIC_DSN___' ]);
+
+try {
```
stayallive: I would move the try block below the configure scope and breadcrumb to make it more clear those should become before exceptions are being thrown.

UPGRADE-2.0.md
```diff
@@ -1,5 +1,38 @@
 # Upgrade from 1.10 to 2.0
 
+Version `2.x` is a complete rewrite of the existing code base. The public API has been trimmed down to a minimum.
+The preferred way of using the SDK is through our static API.
```
stayallive: Not really `static API` but more like `(global) helpers` (better name possibly!? also their not really global...) as in the example below you using no `::` (except for some "enum" cases) which would indicate anything static method calls.

UPGRADE-2.0.md
```diff
@@ -1,5 +1,38 @@
...
+     $scope->setLevel(\Sentry\Severity::warning());
+     $scope->setExtra('character_name', 'Mighty Fighter');
+});
+
+\Sentry\addBreadcrumb(new Breadcrumb(Breadcrumb::LEVEL_ERROR, Breadcrumb::TYPE_ERROR, 'error_reporting', 'Message'));
+```
+
+The call to `\Sentry\init()` sets up global exception/error handlers and uncaught errors will be sent to Sentry.
+
+Version `>= 2.0` conforms the [Unified SDK API](https://docs.sentry.io/development/sdk-dev/unified-api/).
```
Jean85:
```suggestion
Version `>= 2.0` conforms to the [Unified SDK API](https://docs.sentry.io/development/sdk-dev/unified-api/).
```

UPGRADE-2.0.md
```diff
@@ -1,5 +1,38 @@
...
+     $scope->setTag('page_locale', 'de-at');
+     $scope->setUser(['email' => 'john.doe@example.com']);
+     $scope->setLevel(\Sentry\Severity::warning());
+     $scope->setExtra('character_name', 'Mighty Fighter');
+});
+
+\Sentry\addBreadcrumb(new Breadcrumb(Breadcrumb::LEVEL_ERROR, Breadcrumb::TYPE_ERROR, 'error_reporting', 'Message'));
+```
+
+The call to `\Sentry\init()` sets up global exception/error handlers and uncaught errors will be sent to Sentry.
```
ste93cry: I I would say `and any uncaught error will be`

UPGRADE-2.0.md
```diff
@@ -1,5 +1,38 @@
...
+
+Version `>= 2.0` conforms the [Unified SDK API](https://docs.sentry.io/development/sdk-dev/unified-api/).
+It has fundamentally different concepts, it's no longer recommended to just use a `Client` unless you really know what you are doing.
```
ste93cry: I would not start a new phrase here, but just say ``Version `2.0` conforms [...] and it has a foundamentally different concept since it's no longer recommended [...]`` (note also `concept` instead of `concepts`)

UPGRADE-2.0.md
```diff
@@ -1,5 +1,38 @@
...
+
+If you are still want to use Dependecy Injection you should inject and use `HubInterface` which comes closest to `Raven_Client` you knew before.
```
ste93cry: I would remove this phrase entirely since it provides no real value to the user

UPGRADE-2.0.md
```diff
@@ -91,7 +126,7 @@
 
 ### Client
 
-- The constructor of the `Raven_Client` class has changed its signature and
+- The constructor of the `Client` before (`Raven_Client`) class has changed its signature and
```
ste93cry: The word `before` should go inside the round brackets

UPGRADE-2.0.md
```diff
@@ -112,6 +147,8 @@
       // ...
   }
   ```
+  
+  In general if you need to create your own instance of a `Client` we advise you to use the `ClientBuilder` instead since it does a lot for you internally.
```
ste93cry: There is a space at the start of the line that should not exists. Also I would change the sentence to `The suggested way to create your own instance of the client is to use the provided builder that will take care of instantiating a few dependencies like the PSR-7 factories and the HTTP client`

UPGRADE-2.0.md
```diff
@@ -294,7 +331,7 @@
 
 - The `Raven_Client::getLastEventID` method has been removed. The ID of the
   last event that was captured is now returned by each of the `Client::capture*`
-  methods.
+  methods. Also you can use `Hub::getCurrent()->getLastEventId()`.
```
ste93cry: I believe that saying `You can also use [...]` is  better from a grammatical point of view

UPGRADE-2.0.md
HazAT: Updated the phrasing there, I want to keep the "Static API" since it's described like that on the docs page, linked later.

UPGRADE-2.0.md
HazAT: Well it would point them in the right direction to get started easily without having to look through the whole codebase what to use now.

ste93cry: But the sentence is misleading since it's not true that `HubInterface` comes closest to the old `Raven_Client` (we have the `Client` class now). Also dependency injection is a topic that is out of scope for this library since it's something users have to implement themselves, so talking about it is useless.

Issue #784, opened 2019-02-28
Checking for "@" in handleError on the 2.0 release


Jasonoro:
This issue is _very_ similar to #332. Currently Sentry automatically logs any error that’s suppressed by the “@” operator in PHP. There is however no easy way to disable this. You could do it by creating your own ErrorHandler with associated Integration but that’s a lot of work and a bunch of duplicate code due to those classes being final. Is there a way we can get a config option (or maybe this to be the default) so our Sentry issues don’t get spammed with a bunch of ignored 'errors'?

stayallive:
Hey @Jasonoro, we are already working on bringing it back how it was and make it opt-in via a configuration option.

ste93cry:
This has been resolved in `2.0.1`, please give it a try and feel free to reopen this issue if something is still wrong

PR #785, opened 2019-02-28
Handle silenced errors gracefully

Jean85:
This should solve #784.

I'm basically capturing silenced errors as `SilencedErrorException`, so we can handle them gracefully in the listener. I will add a `capture_silenced_error` option too, so the capture will be opt-in.

### DIFF
--- a/src/Exception/SilencedErrorException.php
+++ b/src/Exception/SilencedErrorException.php
@@ -1,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\Exception;
+
+class SilencedErrorException extends \ErrorException
+{
+}
--- a/src/Exception/SilencedErrorException.php
+++ b/src/Exception/SilencedErrorException.php
@@ -4,6 +4,10 @@ declare(strict_types=1);
 
 namespace Sentry\Exception;
 
-class SilencedErrorException extends \ErrorException
+/**
+ * Error that has been captured by the {@see ErrorHandler}, but that was silenced
+ * with `@` in the source code.
+ */
+final class SilencedErrorException extends \ErrorException
 {
 }
--- a/src/Options.php
+++ b/src/Options.php
@@ -642,10 +642,10 @@ final class Options
     }
 
     /**
-     * Return the capture_silenced_errors option
-     * 
-     * @return bool If true, errors silenced through the @ operator will be reported,
-     * ignored otherwise.
+     * Return the capture_silenced_errors option.
+     *
+     * @return bool if true, errors silenced through the @ operator will be reported,
+     *              ignored otherwise
      */
     public function shouldCaptureSilencedErrors(): bool
     {
@@ -654,9 +654,9 @@ final class Options
 
     /**
      * Sets the capture_silenced_errors option.
-     * 
-     * @param bool $shouldCapture If set to true, errors silenced through the @
-     * operator will be reported, ignored otherwise.
+     *
+     * @param bool $shouldCapture if set to true, errors silenced through the @
+     *                            operator will be reported, ignored otherwise
      */
     public function setCaptureSilencedErrors(bool $shouldCapture): void
     {
--- a/src/Options.php
+++ b/src/Options.php
@@ -642,6 +642,30 @@ final class Options
     }
 
     /**
+     * Return the capture_silenced_errors option
+     * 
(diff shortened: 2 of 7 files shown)


tests/phpt/silenced_errors.phpt
```diff
@@ -0,0 +1,30 @@
...
+$vendor = __DIR__;
+
+while (!file_exists($vendor . '/vendor')) {
+    $vendor = \dirname($vendor);
+}
+
+require $vendor . '/vendor/autoload.php';
+
+init([
+    'before_send' => function () {
```
ste93cry: This can be a `static` lambda

tests/phpt/silenced_errors.phpt
```diff
@@ -0,0 +1,30 @@
...
+        echo 'Event captured' . PHP_EOL;
+    }    
+]);
+
+echo 'Triggering silenced error' . PHP_EOL;
+@$a['missing'];
+echo 'End'
+?>
+--EXPECTF--
```
ste93cry: Since we are comparing the resulting output against a fixed message we can use `EXPECT`

src/Exception/SilencedErrorException.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\Exception;
+
+class SilencedErrorException extends \ErrorException
```
ste93cry: Missing the DocBlock describing what this class does. Also does it makes sense to make it `final`?

src/Options.php
```diff
@@ -641,6 +641,30 @@ public function setHttpProxy(?string $httpProxy): void
         $this->options = $this->resolver->resolve($options);
     }
 
+    /**
+     * Return the capture_silenced_errors option.
```
ste93cry: This sentence should describe what the option does rather than just say what it returns. I would change it to something like `Gets whether the silenced errors should be captured or not`

src/Options.php
```diff
@@ -641,6 +641,30 @@ public function setHttpProxy(?string $httpProxy): void
...
+     *
+     * @return bool if true, errors silenced through the @ operator will be reported,
```
ste93cry: The sentence should start with a capitalized letter

src/Options.php
```diff
@@ -641,6 +641,30 @@ public function setHttpProxy(?string $httpProxy): void
...
+     */
+    public function shouldCaptureSilencedErrors(): bool
+    {
+        return $this->options['capture_silenced_errors'];
+    }
+
+    /**
+     * Sets the capture_silenced_errors option.
+     *
+     * @param bool $shouldCapture if set to true, errors silenced through the @
```
ste93cry: The sentence should start with a capitalized letter

src/Options.php
```diff
@@ -641,6 +641,30 @@ public function setHttpProxy(?string $httpProxy): void
...
+     * @return bool if true, errors silenced through the @ operator will be reported,
+     *              ignored otherwise
+     */
+    public function shouldCaptureSilencedErrors(): bool
+    {
+        return $this->options['capture_silenced_errors'];
+    }
+
+    /**
+     * Sets the capture_silenced_errors option.
```
ste93cry: This sentence should describe what the option does rather than just say that it sets the option named X. I would change it to something like `Sets whether the silenced errors should be captured or not`

tests/phpt/silenced_errors.phpt
```diff
@@ -0,0 +1,41 @@
...
+init([
+    'before_send' => function () {
+        echo 'Event captured' . PHP_EOL;
+    }    
+]);
+
+echo 'Triggering silenced error' . PHP_EOL;
+@$a['missing'];
+echo 'End first test' . PHP_EOL;
+Hub::getCurrent()
```
ste93cry: Please add a blank line above this line and after the last call before the `echo`

tests/phpt/silenced_errors.phpt
```diff
@@ -0,0 +1,41 @@
...
+
+init([
+    'before_send' => function () {
+        echo 'Event captured' . PHP_EOL;
+    }    
+]);
+
+echo 'Triggering silenced error' . PHP_EOL;
+@$a['missing'];
+echo 'End first test' . PHP_EOL;
```
ste93cry: This `echo` statement is useless, whether an error is thrown or the execution continues the expected result will be different

PR #786, opened 2019-03-05
fix: Mark Sentry internal frames for attach_stacktrace to in_app false

HazAT:
This marks Sentry internal frames when using `attach_stacktrace` to `in_app` `false` to have better grouping and stack trace.

Before:
image

After:
image

### DIFF
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -424,6 +424,8 @@ class ClientTest extends TestCase
 
                 $this->assertInstanceOf(Stacktrace::class, $result);
                 $this->assertNotEmpty($result->getFrames());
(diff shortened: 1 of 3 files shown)


tests/ClientTest.php
```diff
@@ -424,6 +424,8 @@ public function testAttachStacktrace(): void
 
                 $this->assertInstanceOf(Stacktrace::class, $result);
                 $this->assertNotEmpty($result->getFrames());
+                $this->assertTrue($result->getFrames()[0]->isInApp());
```
ste93cry: Can we save the result of the call to `getFrames()` into a variable that we reuse later for the other assertions? This way we avoid calling that method multiple times

PR #791, opened 2019-03-15
improve breadcrumb coverage

AlessandroMinoccheri:


tests/BreadcrumbTest.php
```diff
@@ -102,4 +102,19 @@ public function testWithoutMetadata(): void
         $this->assertSame(['foo' => 'bar'], $breadcrumb->getMetadata());
         $this->assertArrayNotHasKey('foo', $newBreadcrumb->getMetadata());
     }
+
+    public function testBreadcrumbSerialized(): void
+    {
+        $type = Breadcrumb::TYPE_USER;
+        $category = 'foo';
+        $data = ['baz' => 'bar'];
+        $breadcrumb = new Breadcrumb(Breadcrumb::LEVEL_INFO, $type, $category, null, $data);
```
ste93cry: We should also assert that the level and the message of the breadcrumb are correctly serialized

tests/BreadcrumbTest.php
```diff
@@ -102,4 +102,19 @@ public function testWithoutMetadata(): void
...
+        $breadcrumbSerialized = $breadcrumb->jsonSerialize();
+
+        $this->assertEquals($type, $breadcrumbSerialized['type']);
+        $this->assertEquals($category, $breadcrumbSerialized['category']);
+        $this->assertNull($breadcrumbSerialized['message']);
+        $this->assertEquals($data, $breadcrumbSerialized['data']);
+        $this->assertArrayHasKey('timestamp', $breadcrumbSerialized);
```
ste93cry: Since this test is marked as 'time-sensitive' we should be able to test against the real value of timestamp. Also probably it makes more sense to compare against the expected result of `jsonSerialize` instead of comparing each field separately

tests/BreadcrumbTest.php
```diff
@@ -102,4 +102,19 @@ public function testWithoutMetadata(): void
         $this->assertSame(['foo' => 'bar'], $breadcrumb->getMetadata());
         $this->assertArrayNotHasKey('foo', $newBreadcrumb->getMetadata());
     }
+
+    public function testBreadcrumbSerialized(): void
```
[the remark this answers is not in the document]
AlessandroMinoccheri: Fixed the name of the test

tests/BreadcrumbTest.php
AlessandroMinoccheri: added missing parameters

tests/BreadcrumbTest.php
AlessandroMinoccheri: Now I compare an expected array with the result of jsonSerialize method

tests/BreadcrumbTest.php
```diff
@@ -102,4 +102,26 @@ public function testWithoutMetadata(): void
...
+        $expected = [
+            'type' => $type,
+            'category' => $category,
+            'message' => $message,
+            'level' => $level,
+            'timestamp' => microtime(true),
+            'data' => $data
+        ];
+        
+        $this->assertEquals($expected, $breadcrumbSerialized);
```
ste93cry: Since we don't use the `$breadcrumbSerialized` variable in more places we can inline the call to the `jsonSerialize` function here. Also build is failing due to CS issues

AlessandroMinoccheri:
Pr updated, it needs other improvements @ste93cry ?

PR #798, opened 2019-04-01
Add Breadcrumb::fromArray constructional helper function

peter-hoa:
If a Breadcrumb object is passed, do as it does today.

If an associative array is passed, use specific array items to create a new Breadcrumb (applying some defaults if needed).

Also added TYPE_DEFAULT to the Breadcrumb object, which appeared to be missing.

Added a test to SdkTest.php to include creation using new array argument option.

[Link to forum item which provides some background on this](https://forum.sentry.io/t/php-sdk-2-0-breadcrumb-creation-improvements/6350)

### DIFF
--- a/src/Breadcrumb.php
+++ b/src/Breadcrumb.php
@@ -14,6 +14,11 @@ use Sentry\Exception\InvalidArgumentException;
 final class Breadcrumb implements \JsonSerializable
 {
     /**
+     * This constant defines the default breadcrumb type.
+     */
+    public const TYPE_DEFAULT = 'default';
+
+    /**
      * This constant defines the http breadcrumb type.
      */
     public const TYPE_HTTP = 'http';
@@ -355,6 +360,22 @@ final class Breadcrumb implements \JsonSerializable
             'timestamp' => $this->timestamp,
             'data' => $this->metadata,
         ];
+    }
+
+    /**
+     * Helper method to create an instance of this class from an array of data.
+     *
+     * @param array $data Data used to populate the breadcrumb
+     */
+    public static function fromArray(array $data): self
+    {
+        return new self(
+            $data['level'],
+            $data['type'] ?? self::TYPE_DEFAULT,
+            $data['category'],
+            $data['message'] ?? null,
+            $data['data'] ?? []
+        );
     }
 
     /**
--- a/tests/BreadcrumbTest.php
+++ b/tests/BreadcrumbTest.php
@@ -43,6 +43,52 @@ final class BreadcrumbTest extends TestCase
         $this->assertEquals(microtime(true), $breadcrumb->getTimestamp());
     }
 
+    /**
+     * @dataProvider fromArrayDataProvider
+     */
+    public function testFromArray(array $requestData, array $expectedResult): void
+    {
+        $expectedResult['timestamp'] = microtime(true);
+        $breadcrumb = Breadcrumb::fromArray($requestData);
+
+        $this->assertEquals($expectedResult, $breadcrumb->toArray());
+    }
+
+    public function fromArrayDataProvider(): array
+    {
+        return [
+            [
+                [
+                    'level' => Breadcrumb::LEVEL_INFO,
+                    'type' => Breadcrumb::TYPE_USER,
+                    'category' => 'foo',
+                    'message' => 'foo bar',
+                    'data' => ['baz'],
+                ],
+                [
+                    'level' => Breadcrumb::LEVEL_INFO,
+                    'type' => Breadcrumb::TYPE_USER,
+                    'category' => 'foo',
+                    'message' => 'foo bar',
+                    'data' => ['baz'],
+                ],
+            ],
+            [
+                [
+                    'level' => Breadcrumb::LEVEL_INFO,
+                    'category' => 'foo',
+                ],
+                [
+                    'level' => Breadcrumb::LEVEL_INFO,
+                    'type' => Breadcrumb::TYPE_DEFAULT,
+                    'category' => 'foo',
+                    'message' => null,
+                    'data' => [],
+                ],
+            ],
+        ];
+    }
+
     public function testWithCategory(): void
     {
         $breadcrumb = new Breadcrumb(Breadcrumb::LEVEL_INFO, Breadcrumb::TYPE_USER, 'foo');


ste93cry:
First of all thank you for your contribution. Unfortunately removing the typehint from a function is a BC break, so as is now this PR cannot be accepted for a `2.x` release. Also, from my point of view I don't see the value in allowing this and I would like to ask you if you can provide any example of where this may be useful: a method with arguments is way better for a lot of reasons, e.g. it can be statically analyzed by IDEs to provide autocompletion. And since nearly all parameters required by the current constructor of the `Breadcrumb` class are mandatory to create a valid object I can't really see how accepting an array would help because you would have anyway to pass those parameters in the array with the added complexity of having to know which keys to use

src/Breadcrumb.php
```diff
@@ -13,6 +13,11 @@
  */
 final class Breadcrumb implements \JsonSerializable
 {
+    /**
+     * This constant defines the default breadcrumb type.
+     */
+    public const TYPE_DEFAULT = 'default';
```
Jean85: Apart from the rest of the PR, this addition should be retained. Thanks for spotting this.

peter-hoa:
Hey folks,

I have updated the description with a link to the forum topic that is the driver behind this PR.

Essentially, the main reason is to bring back some of the easy-reading programming style support that the 1.x API offered.  The coding style we use here favors clarity of the code vs what the 2.x API is offering currently.

Also, regarding the TYPE_DEFAULT - in the SDK documentation, it reads like if you don't provide this parameter, then a default value (which happens to be the literal 'default') will be applied for you.  But, the way the PHP API has been written, this parameter is *required*.  It's not optional.  Using the array of arguments allows for defaults to be defined in a very flexible way.

So, please bear the above in mind.  This improvement is providing a useful alternative to having to instantiate the Breadcrumb with unnamed arguments in the constructor.  It's also removing the need to construct the Breadcrumb object in client code at all, by embodying this within the Sdk method, which will lead to cleaner looking client code.

You might also want to offer the ability to pass the discrete constructor arguments to the Sdk->addBreadcrumb() rather than forcing the construction of the Breadcrumb object first.  This alone will help make client code cleaner and easier to read too.

ie:

```
Sentry->addBreadcrumb(new Sentry\Breadcrumb(Sentry\Breadcrumb::TYPE_DEFAULT, Sentry\Breadcrumb::LEVEL_USER, 'my category', 'my message'));
```

vs

```
Sentry->addBreadcrumb(Sentry\Breadcrumb::TYPE_DEFAULT, Sentry\Breadcrumb::LEVEL_USER, 'my category', 'my message');
```

and even readable: (using a type default)

```
Sentry->addBreadcrumb([
   'level' => 'user',  // or: Sentry\Breadcrumb::LEVEL_USER
   'category' => 'my category',
   'message' => 'my message',
]);
```

You already have support in the Breadcrumb for using a 'fluent' method of arriving at the object you really want to have.

This is just another layer of flexibility so that it will make the API very usable for people that have a different preferred coding style.

Jean85:
> Also, regarding the TYPE_DEFAULT - in the SDK documentation, it reads like if you don't provide this parameter, then a default value (which happens to be the literal 'default') will be applied for you. But, the way the PHP API has been written, this parameter is required. It's not optional. Using the array of arguments allows for defaults to be defined in a very flexible way.

As said on my previous comment, this is missing on our side; happy to merge this part, even in a separate PR if you want.

> This is just another layer of flexibility so that it will make the API very usable for people that have a different preferred coding style.

This instead is debatable. This is not just about "different coding styles": allowing a generic array as a way to construct breadcrumbs, or setters, may allow constructing Breadcrumbs objects in invalid states, forcing us to add a layer of validation (or fallbacks) before storing/sending them.

peter-hoa:
> As said on my previous comment, this is missing on our side; happy to merge this part, even in a separate PR if you want. [EDIT] maybe we can put the `DEFAULT` const as a default value for that parameter.

ummmm.... the parameter you're talking about has other no-default arguments after it, so I think you'd have to also make the category nullable too, which I suspect would not be desirable?

peter-hoa:
> This instead is debatable. This is not just about "different coding styles": allowing a generic array as a way to construct breadcrumbs, or setters, may allow constructing Breadcrumbs objects in invalid states, forcing us to add a layer of validation (or fallbacks) before storing/sending them.

Not sure what the issue here is.  You already have the fluent way of creating a breadcrumb.  And, by the way, the change I'm wanting to introduce is not really touching the Breadcrumb (aside from adding the missing TYPE_DEFAULT).  It's adding to the Sdk an option that we (and probably others) that are migrating from SDK 1.x would appreciate having.  The Breadcrumb object still has everything it had, and nothing less than it had before - including the ability to reject construction if the supplied arguments are not legal.

HazAT, in  [the forum conversation](https://forum.sentry.io/t/php-sdk-2-0-breadcrumb-creation-improvements/6350/3) was welcoming of the idea, as I understood it.

Jean85:
I've re-reviewed the code, and you're right, it seems to be likely to be merged. 

My personal preference though would be to move that code in a named constructor, something like `Breadcrumb::fromArray(): self` which would still be helpful, but available even outside of the global helper functions; those should remain a simple proxies, not have internal logic.

ste93cry:
>so I think you'd have to also make the category nullable too, which I suspect would not be desirable?

Looking at the [SDK reference](https://docs.sentry.io/development/sdk-dev/interfaces/breadcrumbs/) I see that nearly all parameters are optional, but this makes no sense to me because if everything can miss then I can't imagine what should be displayed in the UI... @HazAT can you give us help on this?

> And, by the way, the change I'm wanting to introduce is not really touching the Breadcrumb

Effectively looking again at the source code this change is non-breaking since it's changing a global function and not a method, thus there are no problems with people that could have overridden the method.

>This improvement is providing a useful alternative to having to instantiate the Breadcrumb with unnamed arguments in the constructor

I'm against this change too because I don't see a real value since the same motivation you are giving to justify it could be applied to nearly all methods of a library/application, but you don't for a reason: imagine if all of them would accept an array instead of parameters and the mess that would result in it... Plus, this motivation has a drawback: you have to know the keys that can be used in the array, so between choosing a method with typehinted parameters and IDE support for autocompletion and a plain array with no help at all from the IDE I would prefer the first (my opinion of course)

HazAT:
> I've re-reviewed the code, and you're right, it seems to be likely to be merged.
> 
> My personal preference though would be to move that code in a named constructor, something like `Breadcrumb::fromArray(): self` which would still be helpful, but available even outside of the global helper functions; those should remain a simple proxies, not have internal logic.

Thanks, @peter-hoa for the PR!
I like the idea to have `Breadcrumb::fromArray(): self` 👍 

I generally agree with what @Jean85 and @ste93cry said/reviewed.

@ste93cry 
> Looking at the SDK reference I see that nearly all parameters are optional, but this makes no sense to me because if everything can miss then I can't imagine what should be displayed in the UI... @HazAT can you give us help on this?

So the docs are correct here, a timestamp is the only required field, we will render an empty breadcrumb then.

ste93cry:
> I like the idea to have `Breadcrumb::fromArray(): self` 👍

It's fine for me too, I like this idea more than allowing removing the typehint from the global function

> So the docs are correct here, a timestamp is the only required field, we will render an empty breadcrumb then.

I don't see the value in doing this, so maybe you could discuss to change this with the other Sentry's dev

peter-hoa:
> I like the idea to have `Breadcrumb::fromArray(): self`

Hmmmmm.  One of the objectives I had with this was to make the code not be bloated with extra stuff that's really needless, to improve readability (yes, I know - at the expense of stating the name of the arguments each time - but that is a + for readability)

Lets take a look at what the suggestion will mean in use:

```
\Sentry\addBreadcrumb(\Sentry\Breadcrumb::fromArray([
   'level' => 'debug',
   'type' => 'error',
   'category' => 'my category',
   'message' => 'my message',
]));
```

versus what I contributed:

```
\Sentry\addBreadcrumb([
   'level' => 'debug',
   'type' => 'error',
   'category' => 'my category',
   'message' => 'my message',
]);
```

I have to say, I really don't like having to refer to the Breadcrumb::fromArray all the time.

I understand that making the addBreadcrumb be varargs friendly, and having the wiring that's in the Sdk is not attractive.

How about this:

```
\Sentry\addBreadcrumbFromArray([
   'level' => 'debug',
   'type' => 'error',
   'category' => 'my category',
   'message' => 'my message',
]);
```

and have Sdk's new addBreadcrumbFromArray() do the minimal work of instantiating the Breadcrumb via the static method on the Breadcrumb?

Sdk will be minimally involved.  Logic to marshal from the array to the existing constructor of the Breadcrumb will be put inside the Breadcrumb::fromArray(), and it will still use the existing constructor.

This way, you'll keep the strict type checking and hinting in IDE's that you are keen on, and those of us that want to use arrays will not be having to always explicitly refer to the Breadcrumb::fromArray.

?

Jean85:
I still don't like it. Try to understand that, as maintainers, we need to reduce the public API surface as much as possibile, to avoid bloating the SDK and have to maintain more stuff for no practical use.

This is a "readability" concern, which is pretty subjective, so anyone could ask for a different entrypoint for any feature.

If you're really worried about readability in your projects, you can still write your own wrapper, which would be super concise, up to your liking.

peter-hoa:
> I still don't like it. Try to understand that, as maintainers, we need to reduce the public API surface as much as possibile, to avoid bloating the SDK and have to maintain more stuff for no practical use.
> 
> This is a "readability" concern, which is pretty subjective, so anyone could ask for a different entrypoint for any feature.
> 
> If you're really worried about readability in your projects, you can still write your own wrapper, which would be super concise, up to your liking.


While it's certainly possible to create a wrapper, I believe that people migrating (as I am) from API 1.x would appreciate not having to do that, and have a nice migration path.

As maintainers of the API, isn't one of your goals that it be as successful and usable as possible?

I'm curious: how many of the maintainers are using the fluent style of Breadcrumb creation?

peter-hoa:
> I still don't like it. Try to understand that, as maintainers, we need to reduce the public API surface as much as possibile, to avoid bloating the SDK and have to maintain more stuff for no practical use.

You know, the Breadcrumb already has 'toArray'.  Adding 'fromArray' will make this be symmetrical.  ie: you can now construct a breadcrumb using the details obtained from 'toArray'.

peter-hoa:
Ok, I've backed off on pushing for the Sdk change, so now, it consists of adding the fromArray static method to the Breadcrumb object, the addition of the Breadcrumb::TEST_DEFAULT, and a test method to BreadcrumbTest.

usage will look like this:

```
\Sentry\addBreadcrumb(\Sentry\Breadcrumb::fromArray([
   'level' => 'debug',
   'type' => 'error',
   'category' => 'my category',
   'message' => 'my message',
]));
```

Please take another look.  Hopefully this will be accepted.

tests/BreadcrumbTest.php
```diff
@@ -43,6 +43,24 @@ public function testConstructor(): void
         $this->assertEquals(microtime(true), $breadcrumb->getTimestamp());
     }
 
+    public function testConstructFromArray(): void
```
ste93cry: As a naming convention for our tests we always use `test` + the name of the method being tested. In this case the correct name would be `testFromArray`

src/Breadcrumb.php
```diff
@@ -357,6 +362,22 @@ public function toArray(): array
         ];
     }
 
+    /**
+     * Constructon helper.
```
ste93cry: Please change this description to something like `Helper method to create an instance of this class from an array of data`. Also, as a convention we try to not write comments longer than 80 characters (it's a soft limit anyway, nothing enforced strictly) but if this is the case please consider splitting it into multiple lines to improve readability

src/Breadcrumb.php
```diff
@@ -357,6 +362,22 @@ public function toArray(): array
...
+     *
+     * @param array $array The set of construction key/value pairs
```
ste93cry: Please change this description to `Data used to populate the breadcrumb`

src/Breadcrumb.php
```diff
@@ -357,6 +362,22 @@ public function toArray(): array
...
+     */
+    public static function fromArray(array $array): self
```
ste93cry: Please rename the parameter to `$data`

src/Breadcrumb.php
```diff
@@ -357,6 +362,22 @@ public function toArray(): array
...
+    {
+        return new self(
+            $array['level'],                      // level is required
+            $array['type'] ?? self::TYPE_DEFAULT, // use default if not supplied
+            $array['category'],                   // category is required
+            $array['message'] ?? (string) null,   // use null cast to string if no value supplied
```
ste93cry: This casting is not necessary since the argument is `nullable`

src/Breadcrumb.php
```diff
@@ -357,6 +362,22 @@ public function toArray(): array
...
+            $array['data'] ?? []                  // supply empty array if none was supplied
```
ste93cry: Please remove such comments since it provides no real value (it's obvious from the usage of the `??` statement that we are doing fallback to a default)

src/Breadcrumb.php
```diff
@@ -357,6 +362,22 @@ public function toArray(): array
...
+     * Constructon helper.
+     *
+     * @param array $array The set of construction key/value pairs
+     */
+    public static function fromArray(array $array): self
+    {
+        return new self(
+            $array['level'],                      // level is required
+            $array['type'] ?? self::TYPE_DEFAULT, // use default if not supplied
+            $array['category'],                   // category is required
```
ste93cry: Please remove such comments since it provides no real value

src/Breadcrumb.php
```diff
@@ -357,6 +362,22 @@ public function toArray(): array
...
 
+    /**
+     * Constructon helper.
+     *
+     * @param array $array The set of construction key/value pairs
+     */
+    public static function fromArray(array $array): self
+    {
+        return new self(
+            $array['level'],                      // level is required
```
ste93cry: Please remove such comments since it provides no real value

src/Breadcrumb.php
```diff
@@ -357,6 +362,22 @@ public function toArray(): array
...
+            $array['type'] ?? self::TYPE_DEFAULT, // use default if not supplied
```
ste93cry: Please remove such comments since it provides no real value (it's obvious from the usage of the `?? `statement that we are doing fallback to a default)

ste93cry:
>You know, the Breadcrumb already has 'toArray'. Adding 'fromArray' will make this be symmetrical. ie: you can now construct a breadcrumb using the details obtained from 'toArray'.

ste93cry:
>You know, the Breadcrumb already has 'toArray'. Adding 'fromArray' will make this be symmetrical. ie: you can now construct a breadcrumb using the details obtained from 'toArray'.

I don't see how this could be a problem. However I believe that the `toArray` method (in this and other classes as well) should have never been exposed since it probably doesn't provide a real value for the public API and it's mainly used to serialize the data in the event, but this can be done anyway since we're implementing the `JsonSerializable` interface

ste93cry:
>You know, the Breadcrumb already has 'toArray'. Adding 'fromArray' will make this be symmetrical. ie: you can now construct a breadcrumb using the details obtained from 'toArray'.

I don't see how this could be a problem. However I believe that the `toArray` method (in this and other classes as well) should have never been exposed since it probably doesn't provide a real value for the public API and it's mainly used to serialize the data in the event, but this can be done anyway since we're implementing the `JsonSerializable` interface

>Regardless of the coding style preferences, adding what I'm asking for will make migrating from API 1.x such a simple thing.

You may be right on this, but since the `2.0` version of this library is almost a full rewrite imho it doesn't really matter because you will have anyway to change a lot of other things too

>I'm curious: how many of the maintainers are using the fluent style of Breadcrumb creation?

At my job we use a lot of value objects and indeed we have some classes that can be constructed from a payload of data (e.g. a model representing a REST payload) but in almost all cases where it's possible we prefer to use constructor or static methods to create objects since it helps IDE, static analysis and imho it improves readability of the code. It may or may not be more nicer to read and I agree on this, but from my experience it also depends on how the rest of the code is formatted/inlined, etc so I would definitely say that it's subjective

tests/BreadcrumbTest.php
```diff
@@ -43,6 +43,39 @@ public function testConstructor(): void
...
+
+        $this->assertEquals('foo', $breadcrumb->getCategory());
+        $this->assertEquals(Breadcrumb::LEVEL_INFO, $breadcrumb->getLevel());
+        $this->assertEquals('foo bar', $breadcrumb->getMessage());
+        $this->assertEquals(Breadcrumb::TYPE_USER, $breadcrumb->getType());
+        $this->assertEquals(['baz'], $breadcrumb->getMetadata());
+        $this->assertEquals(microtime(true), $breadcrumb->getTimestamp());
+    }
+
+    public function testFromArrayDefaults(): void
```
ste93cry: Instad of splitting the test into two different methods, simply use the one above with a data provider that provides the input `$data` parameter to pass to the `fromArray` method and an `$expectedResult` that can be compared to what the `toArray` method returns

ste93cry:
@peter-hoa thank you for your patience and hard work in getting this merged, however I think that there have been a misunderstanding in the last requested change: what I meant was to use a data provider so that the same test can be repeated multiple times, each time with different input values and expected result. There should be a `fromArrayDataProvider` method that returns the test data that is then used by the `testFromArray` method. Take a look at other tests to see how we did it

peter-hoa:
Ok folks.  So, I was asked to use the dataprovider approach to creating input and output arrays.

(I am referring to `BreadcrumbTest::testFromArray()`)

The first time, I had (in the dataprovider):

```
'timestamp' => microtime(true),
```

and the test result complained that the values did not match.  They were off by a very small amount.

So, I removed the timestamp element from the 'expects' array entirely, wondering if now, the timestamp element would be ignored, but no, it's still being evaluated for, and complaining that one array has it, and the other does not.

so, what, prey tell, is the magic sauce to make this work?

peter-hoa:
Ok, so, I changed the `fromArrayTest()` to jam the timestamp in to the expectsArray immediately before obtaining the result array, and it seems to have 'passed'.  But, I really don't like this - it seems very dependent on whether the processing happens within a very small window of time.  If the processing bounds adjacent times, won't this test fail for no apparent good reason on a seemingly random basis?

tests/BreadcrumbTest.php
peter-hoa: ok.  I think I've done as you ask now.  please review.

peter-hoa:
@HazAT , @ste93cry , @Jean85 

please review

ste93cry:
>so, what, prey tell, is the magic sauce to make this work?

There is no magic sauce, the data provider is still the right solution when you have a test method that needs to test different cases (this is why this feature exists in the first place), however due to how PHPUnit loads the test suite the data provider is executed before all the tests are ran, so mocking the time doesn't work inside it 😞 The only workaround until this is fixed in PHPUnit itself is to not use a data provider or workaround this limitation like you did

BTW, apart from those two little fixes LGTM. Can you also please squash the commits?

tests/BreadcrumbTest.php
```diff
@@ -43,6 +43,51 @@ public function testConstructor(): void
...
+     * @dataProvider testFromArrayDataProvider
+     */
+    public function testFromArray(array $requestData, array $expectedResult): void
+    {
+        $expectedResult['timestamp'] = microtime(true);
+        $breadcrumb = Breadcrumb::fromArray($requestData);
+        $this->assertEquals($expectedResult, $breadcrumb->toArray());
+    }
+
+    public function testFromArrayDataProvider(): array
```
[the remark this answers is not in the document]
peter-hoa: also changed the function name in the @dataprovider at line 47 so it's consistent

peter-hoa:
@ste93cry, changes you requested, and squashing of commits has been done.  please review and approve if you're happy with it.

ste93cry:
Thank you for your contribution, I know this is not really what you wanted in the beginning but it's a step towards it in case in the future we change our mind

peter-hoa:
@HazAT , @ste93cry , @Jean85

Forgive me, but, who / when will this be merged?  There's nothing more I need to do, right?

I suppose it might be good to change the title, since it's no longer accurate.  But, that might be more confusing :) ?

ste93cry:
>Forgive me, but, who / when will this be merged? There's nothing more I need to do, right?

This feature will be in the next minor version `2.1` which is under development. We will merge it asap, once other features that we want to ship too are ready

>I suppose it might be good to change the title, since it's no longer accurate. But, that might be more confusing :) ?

Feel free to change the title of the PR as you prefer

peter-hoa:
Thank you for the information, and all the reviewers, for persevering with me as this rather simple (as it turns out) change is added.  :)

It does still include the TYPE_DEFAULT being declared too, but the main focus was the helper.

PR #800, opened 2019-04-03
Fix code example

Horttcore:


ste93cry:
Since I'm evaluating in #788 to stop register implicitly the error handler in the `ErrorHandler::add*Listener` methods (and so make them non-static) to be future-proof I would change the example to something similar to

```php
$errorHandler = ErrorHandler::registerOnce();
$errorHandler->addErrorListener(...);
$errorHandler->addFatalErrorListener(...);
$errorHandler->addExceptionListener(...);
```

ste93cry:
The PR I was talking about has been merged, but the changes will be available only from version `2.1`. Since the `UPGRADE-2.0.md` is targeting of course the `2.0` version it seems to me that the changes here are fine as they are, so LGTM

ste93cry:
The PR I was talking about has been merged, but the changes will be available only from version `2.1`. Since the `UPGRADE-2.0.md` is targeting of course the `2.0` version it seems to me that the changes here are fine as they are except for the fact that as said by @Jean85 we should document that at least a listener should be added to report the errors. @Horttcore are you willing to address the requested changes so that we can merge this PR?

stayallive:
I am closing this PR because of inactivity, I even believe this problem was eventually addressed.

PR #806, opened 2019-04-19
Fix #805: Default type must be an array

### LINKED ISSUE #805: Default Type of Event::$exceptions must be an array
In Envent class, the default value for `private $exceptions;` is implicit `null`
https://github.com/getsentry/sentry-php/blob/master/src/Event.php#L127
But `public function getExceptions(): array` return only array
https://github.com/getsentry/sentry-php/blob/master/src/Event.php#L538

This cause a failure if I call `getExceptions` inside callback `before_send`  and don't have exceptions, for example, calling `captureMessage()`

tonybolzan:
In `Event` class, the default value for `private $exceptions;` is implicit `null` must be an `array`

### DIFF
--- a/src/Event.php
+++ b/src/Event.php
@@ -124,7 +124,7 @@ final class Event implements \JsonSerializable
     /**
      * @var array The exceptions
      */
-    private $exceptions;
+    private $exceptions = [];
 
     /**
      * @var Stacktrace|null The stacktrace that generated this event
@@ -640,7 +640,7 @@ final class Event implements \JsonSerializable
             $data['breadcrumbs']['values'] = $this->breadcrumbs;
         }
 
-        if (null !== $this->exceptions) {
+        if (!empty($this->exceptions)) {
             $reversedException = array_reverse($this->exceptions);
 
             foreach ($reversedException as $exception) {
--- a/src/Event.php
+++ b/src/Event.php
@@ -640,23 +640,19 @@ final class Event implements \JsonSerializable
             $data['breadcrumbs']['values'] = $this->breadcrumbs;
         }
 
-        if (!empty($this->exceptions)) {
-            $reversedException = array_reverse($this->exceptions);
+        foreach (array_reverse($this->exceptions) as $exception) {
+            $exceptionData = [
+                'type' => $exception['type'],
+                'value' => $exception['value'],
+            ];
 
-            foreach ($reversedException as $exception) {
-                $exceptionData = [
-                    'type' => $exception['type'],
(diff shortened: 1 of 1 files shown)


src/Event.php
```diff
@@ -640,7 +640,7 @@ public function toArray(): array
             $data['breadcrumbs']['values'] = $this->breadcrumbs;
         }
 
-        if (null !== $this->exceptions) {
+        if (!empty($this->exceptions)) {
```
ste93cry: I would say that this statement can just be removed since we would just loop over an empty array and do nothing, so it's useless.

tonybolzan: I have followed the standard of `fingerprint` and` breadcrumbs`
Do you want me to discard this PR and create another?

ste93cry: >I have followed the standard of `fingerprint` and `breadcrumbs`

Nope, it's different because in this case we would loop over the array and append each item to the `$data['exception']['values']` variable (implicitly creating the `$data['exception']` key), but since it would be empty it's a noop. In the other cases we're not looping but directly assigning the values to the `$data` array, so both keys would be set even if there is no need to

>Do you want me to discard this PR and create another?

Nope, no reason to do it :smiley:

tonybolzan: yes of course, but I did not want to make too many changes to the code, even because I'm using the github editor.
But I'll be more careful next time.

ste93cry: I saw you marked the issue as resolved, but it's not. Please update the PR with the requested changes so that we can merge it 😉

tonybolzan: oops... I don't use github pr, sorry, how can I append/merge a commit in existing pr ?

ste93cry: Just checkout the branch of this PR, edit the file, commit and push like always. Or you can edit the file again from the GitHub interface like I suppose you did, just be sure to edit the one in the branch [`patch-1`](https://github.com/tonybolzan/sentry-php/tree/patch-1)

tonybolzan: ok... Can I combine `array_reverse` inside of `foreach`, create a new variable `$reversedException` for this don't make sense for me.

ste93cry: Sure, feel free to make as much changes as you like if they makes sense and do not deviate too much from the purpose of the issue. Also remember to re-request the review when done, otherwise the reviewers may miss that more changes have been pushed if they receive a lot of notifications

tonybolzan:
Thanks @ste93cry for your mentoring on GitHub

Jean85:
Cross ref: this fixes #805

Issue #803, opened 2019-04-15
Request body not sent with 2.0


nesl247:
In 1.x, the request body was sent. From what I can see, it was sent in the `data` field of the `request` field. In 2.0, the `RequestIntegration` does not handle this at all, and there is no other location that I can see that even modifies the `request` field.

ste93cry:
This feature is not available out-of-the-box because the [SDK specs](https://docs.sentry.io/development/sdk-dev/interfaces/http/) says that it should not be sent by default since it can contain very large data and it makes sense. You can however make your own integration that will get the data and append it to the event

nesl247:
Was this in the upgrade guide in some way? I did not see it. Also, would it not make more sense to offer this as an option that is defaulted to false rather than to require every project to implement code to have the same functionality as version 1?

ste93cry:
>Was this in the upgrade guide in some way

I'm sorry for this, I've checked and we missed to document it. The good news is that we are going to implement the feature, but it will be disabled by default and it will work differently from before. To have a preview you can take a look at how the [`request_bodies`](https://docs.sentry.io/error-reporting/configuration/?platform=python#request-bodies) option in the Python SDK works

nesl247:
That's exactly what I'm looking for. Do you have an idea when this will be implemented? We're starting to rollout Sentry to all of our micro-services, and I'd love to have this already rather than doing this custom for now.

ste93cry:
I cannot provide an ETA because we don't know when we will release the next version due to other changes that we would like to do for that release too, but I can definitely say that this feature will be in `2.1`

Jean85:
@nesl247 feel free to submit a PR following the Python behavior if you want, it would be greatly appreciated. That way you'll be able to use your fork in the meantime, and be sure to have the same code in 2.1.

ste93cry:
Note: I'm already working on implementing such feature, be aware of this so that we don't do the work twice :wink:

nesl247:
I'll just wait on 2.1.

ste93cry:
Closing this since this has been fixed with the related PR

PR #807, opened 2019-04-20
Capture the body of the request and do a best effort to decode it

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.1
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

This is both a new feature and an improvement of one that already exists (the `RequestIntegration`). Infact, this is a porting of the same feature available in the Python SDK that solves #803. The integration does a best effort to deserialize the data, be it a submitted form or a JSON content. If it fails the RAW data is returned. There is also an option to configure the limit up to which the request data is captured.

### DIFF
--- a/src/Integration/RequestIntegration.php
+++ b/src/Integration/RequestIntegration.php
@@ -24,15 +24,15 @@ final class RequestIntegration implements IntegrationInterface
 {
     /**
      * This constant represents the size limit in bytes beyond which the body
-     * of the request is not captured when the `request_bodies` option is set
-     * to `small`.
+     * of the request is not captured when the `max_request_body_size` option
+     * is set to `small`.
      */
     private const REQUEST_BODY_SMALL_MAX_CONTENT_LENGTH = 10 ** 3;
 
     /**
      * This constant represents the size limit in bytes beyond which the body
-     * of the request is not captured when the `request_bodies` option is set
-     * to `medium`.
+     * of the request is not captured when the `max_request_body_size` option
+     * is set to `medium`.
      */
     private const REQUEST_BODY_MEDIUM_MAX_CONTENT_LENGTH = 10 ** 4;
 
@@ -153,10 +153,14 @@ final class RequestIntegration implements IntegrationInterface
      */
     private function captureRequestBody(ServerRequestInterface $serverRequest)
     {
-        $maxRequestBodySize = $this->options->getRequestBodies();
+        $maxRequestBodySize = $this->options->getMaxRequestBodySize();
         $requestBody = $serverRequest->getBody();
 
-        if ('none' === $maxRequestBodySize || ('small' === $maxRequestBodySize && $requestBody->getSize() > self::REQUEST_BODY_SMALL_MAX_CONTENT_LENGTH) || ('medium' === $maxRequestBodySize && $requestBody->getSize() > self::REQUEST_BODY_MEDIUM_MAX_CONTENT_LENGTH)) {
+        if (
+            'none' === $maxRequestBodySize ||
+            ('small' === $maxRequestBodySize && $requestBody->getSize() > self::REQUEST_BODY_SMALL_MAX_CONTENT_LENGTH) ||
+            ('medium' === $maxRequestBodySize && $requestBody->getSize() > self::REQUEST_BODY_MEDIUM_MAX_CONTENT_LENGTH)
+        ) {
             return null;
         }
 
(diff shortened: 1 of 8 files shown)


src/Integration/RequestIntegration.php
```diff
@@ -109,8 +132,64 @@ private function removePiiFromHeaders(array $headers): array
...
+     * @param ServerRequestInterface $serverRequest The server request
+     *
+     * @return mixed
+     */
+    private function captureRequestBody(ServerRequestInterface $serverRequest)
+    {
+        $maxRequestBodySize = $this->options->getRequestBodies();
+        $requestBody = $serverRequest->getBody();
+
+        if ('none' === $maxRequestBodySize || ('small' === $maxRequestBodySize && $requestBody->getSize() > self::REQUEST_BODY_SMALL_MAX_CONTENT_LENGTH) || ('medium' === $maxRequestBodySize && $requestBody->getSize() > self::REQUEST_BODY_MEDIUM_MAX_CONTENT_LENGTH)) {
```
Jean85: Please split this on multiple lines for better readability:

```suggestion
        if (
        'none' === $maxRequestBodySize 
        || ('small' === $maxRequestBodySize && $requestBody->getSize() > self::REQUEST_BODY_SMALL_MAX_CONTENT_LENGTH) 
        || ('medium' === $maxRequestBodySize && $requestBody->getSize() > self::REQUEST_BODY_MEDIUM_MAX_CONTENT_LENGTH)
        ) {
```

Jean85: Or even better, we could split this in a private method with a switch of the enum.

nesl247:
@HazAT I disagree on the rename. The other SDKs use this as the name, and it should be kept consistent IMO.

https://docs.sentry.io/error-reporting/configuration/?platform=python#request-bodies

The only difference between SDKs is the casing. Some use `request-bodies`, some `request_bodies`, others `requestBodies`, but it's always the same.

ste93cry:
I have to say that the name of the option as is means nothing to me and what @HazAT suggested is a nicer and clearer name. I would suggest to change it in the other SDKs instead

nesl247:
I agree that the name itself doesn't mean anything to me, but you guys are striving to maintain semver no? If so, it would be a BC break to change it in other SDKs, and therefore it would make more sense to keep `requestBodies` here until it's planned for all SDKs.

ste93cry:
Indeed it is a BC for other SDKs but not for our since the option was never implemented. Anyway they should just maintain both the old and the new option until the next major version, but please note that not all SDKs are following semver afaik and I don’t know what’s the way to suggest such a change. Usually this kind of change cames from upstream where the core developers of Sentry are more active (e.g. in the Python or JS SDK)

HazAT:
We did not have `requestBodies` in the new version, so the breaking change for this already happened. We will update the other SDKs to also use `max_request_body_size` instead. 
`max_request_body_size` is the way forward.

src/Options.php
```diff
@@ -705,6 +741,7 @@ private function configureOptions(OptionsResolver $resolver): void
             'max_value_length' => 1024,
             'http_proxy' => null,
             'capture_silenced_errors' => false,
+            'max_request_body_size' => 'none',
```
HazAT:
```suggestion
            'max_request_body_size' => 'medium',
```

ste93cry:
>aside from, that `medium` should be the default

According to the [documentation](https://docs.sentry.io/development/sdk-dev/interfaces/http/) the data should not be sent by default (even if there is the limit now I still agree on this). Are you sure we want to change this behavior?

HazAT:
@ste93cry We have to update the docs here, thanks. `medium` should be default so we want to send it.

untitaker:
@ste93cry Good catch! I have no idea why that is in there. Git blame doesn't give me much info, seems like it's copypasted from somewhere else.

Perhaps this was the result of miscommunication. Raw bodies should not be sent (there is not even an option to turn it on)

ste93cry:
>Perhaps this was the result of miscommunication. Raw bodies should not be sent (there is not even an option to turn it on)

Mmm, what I was referring to is the fact that in the documentation (that we found out to be wrong btw) there was written that the request data, in any form, should not be sent by default. Raw data will be sent if the SDK can't deserialize it (because it fails to parse the body as JSON or whatever other reason cause a failure) and this behavior cames from the Python SDK where this feature has been implemented first. Is it wrong?

Bjornar97:
Hi. In the documentation for laravel configuration it still says "request_bodies" with options: "never", "small", "medium", "always,

https://docs.sentry.io/platforms/php/guides/laravel/configuration/options/#request-bodies

PR #808, opened 2019-04-22
Add Monolog handler

### LINKED ISSUE #790: [2.0] Monolog
In sentry client v1 we were able to use the `Raven_Breadcrumbs_MonologHandler` handler. Is there a similar functionality in v2? There doesn't appear to be any documentation on this and it isn't listed as a removed feature either.

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.1
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

This PR implements a Monolog handler and closes #790, which is a highly requested feature. Code is a stripped down version of the one submitted by @HazAT in the official Monolog repository. There are a few questions to discuss before merging:

- Like in the original PR, the handler needs an hub instance to work, but since the current hub is something that can change during the lifecycle of an application (e.g. when the `Hub::setCurrent()` method is called) it's not clear how this should affect the handler itself: e.g. if the handler is binded to an old or different hub than the current one it will report errors to the wrong client (if it's different between hubs)
- Since Monolog is a logger library a message is likely to always be present along with the context that may include the exception and imho we want to log it regardless. Calling the `captureMessage` method provides no way to pass the exception while calling the `captureException` has no way to pass the message. To solve the issue, instead of calling those methods like in the original PR, I'm building the raw payload of data to build the event from and I'm using the `captureEvent` method directly. Since the two methods mentioned before have a slightly different behavior regarding the stacktrace capture (`attach_stacktrace` option), the same exact behavior cannot be accomplished with such way

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,12 @@
 ## Unreleased
 
 - Mark Sentry internal frames when using `attach_stacktrace` as `in_app` `false` (#786)
-- Increase default severity of `E_RECOVERABLE_ERROR` to `Severity::ERROR`, instead of warning (#792)
-- feat: Add Breadcrumb::fromArray (#798)
-- feat: Add support for SENTRY_ENVRIONMENT and SENTRY_RELEASE (#810)
-- fix: The default value of the $exceptions property of the Event class (#806)
+- Increase default severity of `E_RECOVERABLE_ERROR` to `Severity::ERROR`, instead
+  of `Severity::WARNING` (#792)
+- Add a static factory method to create a breadcrumb from an array of data (#798)
+- Add support for `SENTRY_ENVRIONMENT` and `SENTRY_RELEASE` environment variables (#810)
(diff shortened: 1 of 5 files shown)


Jean85:
We're missing just one small thing! The changelog!

ste93cry:
Done, I always forget to update it!

JanMikes:
Amazing! Thank you 👍

linaori:
Trying to update but am being blocked by this PR not being released. What's the ETA? :smile:

Jean85:
No ETA, but this will be released with 2.1. Check out the milestone for progress: https://github.com/getsentry/sentry-php/milestone/2

linaori:
Alright, thanks for the reply! I'll check back in a while and stay at sentry/sentry-symfony 2.x for now as it seems to not cause issues with Symfony 4.x

Issue #316, opened 2016-07-12
More flexible serialization object


MeSko:
Please add possibility to send more data about object in frame context.
At https://github.com/getsentry/sentry-php/blob/master/lib/Raven/Serializer.php#L54

```
if(is_object($value) || gettype($value) == 'object') {
    return 'Object '.get_class($value);
}
```

It will be excellent if you add interface with method serialize, when object implement you will run it:

```
if (is_object($value) || gettype($value) == 'object') {
    if ($value instanceof Raven_Custom_Serializer) {
        return $value->serialize();
    } else {
        return 'Object '.get_class($value);
    }
}
```

stayallive:
It would be even cooler if you could supply a callable to serialize custom classes, for classes you cannot control. This could word something like:

```php
(new Raven_Client(...))->addSerializerForType(Illuminate\Support\Collection::class, function ($value) {
    return $value->toArray();
});
```

This would add a lot of value IMHO, the `Raven_Custom_Serializer` is also a great idea!

Would love to hear what the Sentry folks think about this, would be happy to write a PR.

dcramer:
I would be down with making serialization more flexible (and effectively a type registry).

In Python we created a list of instances, and each instance has two methods: ``canSerialize`` and ``serialize``. This makes it easy to do things like "serialize all types of values this way".

mcuelenaere:
How about checking whether the [__debugInfo()](http://php.net/manual/en/language.oop5.magic.php#object.debuginfo) method is defined and using its output instead of returning `Object foo`?

This way the "serialization" (which in fact it's not, it's merely dumping the variable to human readable form) is controlled by the end-user and Sentry itself does not need any infrastructure for tracking custom serializers.

SeBsZ:
I came here looking to see if there is a way to view the contents of objects in Sentry. Right now all I see is "Object ReleaseObject". This is an object I have created. Is there a way to send some fields along? Has anything been implemented or is this just the way it is right now? Can I use something like "JsonSerializable" on this class? Thanks!

archon810:
Why not just use print_r or var_dump? Shouldn't that just work implicitly? http://stackoverflow.com/questions/3034530/php-print-all-properties-of-an-object

dcramer:
@archon810 it will create enormous objects in non-native types. Sure we could do it, but then you're going to by default sacrifice other data.

I'm all for allowing some simply callback for the serialization to allow you to manage it yourself, though I'm not sure the speed cost.

[...]

PR #809, opened 2019-04-24
Implement class serializers and interface to customize how objects are serialized

stayallive:
This allows the user (or framework library) to implement their own serialized values for any object instead of the `Object SomeObject` string that Sentry now shows.

Usage with interface (for object of your own):

```php
use Sentry\Serializer\Serializable;

class ExampleObject implements Serializable
{
	private $state = false;

    public function toSentry(): array
    {
        return [
            'internal_state' => $this->state,
        ];
    }
}
```

Usage with serializer "helper" for objects you don't control, register via an option `class_serializers`.

```php
use Sentry\ClientBuilder;
use Sentry\Options;

$serializers = [
    ExampleObject::class => static function (ExampleObject $object): array {
        return [
            'internal_state' => $object->getState(),
        ];
    },
];

$client = new ClientBuilder(new Options([
    'class_serializers' => $serializers,
]));
```

---

If a custom serializer or `toSentry()` generates an exception the value is discarded and the serializers serializes it as normal (resulting in a `Object ExampleClass` string). This is to prevent any issues caused by bad serializers or problems that occur there that might mask the actual exception that we are trying to send/serialize.

The `class_serializers` are resolved using `instanceof` to allow the user to add a custom serializer for a parent class or interface.

Currently it creates an array with the data nested to make it clear which class it actually is, this might be a concern because of the max depth but probably is not.

```php
[
    // get_class($object) but without the 'Object' prefix
    'class' => 'ExampleObject',

    // Whatever the __toSentry() or custom serializer returned:
    'data' => [
        'internal_state' => false,
    ],
]
```

---

/ref #443 #316


src/Serializer/Serializable.php
```diff
@@ -0,0 +1,15 @@
...
+namespace Sentry\Serializer;
+
+interface Serializable
+{
+    /**
+     * Return an array representation of the object for Sentry.
+     *
+     * @return array
+     */
+    public function __toSentry(): array;
```
Jean85: I'm against the use of the double underscore for the name, since it make it seems to say that this is a "magic method", which is not.

IMHO this is fine:
```suggestion
    public function toSentry(): array;
```

src/Options.php
```diff
@@ -705,6 +727,7 @@ private function configureOptions(OptionsResolver $resolver): void
             'max_value_length' => 1024,
             'http_proxy' => null,
             'capture_silenced_errors' => false,
+            'type_serializers' => [],
```
Jean85: Since all of this seems to be targeting just objects, I would prefer renaming it `class_serializers`

src/Serializer/Serializable.php
stayallive: Agreed! 👍

src/Options.php
stayallive: That is a better word for it indeed!

tests/Serializer/SerializerTest.php
```diff
@@ -96,6 +99,47 @@ public function testNull(bool $serializeAllObjects): void
...
+            'class' => \get_class($object),
+            'data' => $customSerializerCallback($object),
+        ], $this->invokeSerialization($serializer, $object));
+    }
+
+    public function testSerializableObject(): void
+    {
+        $serializer = $this->createSerializer();
+
+        $object = new StubSerializableInterfaceObject();
```
ste93cry: Instead of a real object cannot we simply mock the interface and its methods as always?

tests/Serializer/SerializerTest.php
```diff
@@ -96,6 +99,47 @@ public function testNull(bool $serializeAllObjects): void
...
+            'class_serializers' => [
+                StubObject::class => $customSerializerCallback = static function (StubObject $object): array {
+                    return [
+                        'purpose' => $object->getPurpose(),
+                    ];
+                },
+            ],
+        ]));
+
+        $object = new StubObject();
```
ste93cry: Instead of a real object cannot we simply mock the interface and its methods as always?

tests/Serializer/SerializerTest.php
```diff
@@ -96,6 +99,47 @@ public function testNull(bool $serializeAllObjects): void
...
+            'class' => \get_class($object),
+            'data' => $object->toSentry(),
+        ], $this->invokeSerialization($serializer, $object));
+    }
+
+    public function testSerializableObjectThatThrowsAnException(): void
+    {
+        $serializer = $this->createSerializer();
+
+        $object = new StubSerializableInterfaceObjectThrowingException();
```
ste93cry: Instead of a real object cannot we simply mock the interface and its methods as always?

src/Options.php
```diff
@@ -901,4 +926,22 @@ private function validateMaxBreadcrumbsOptions(int $value): bool
     {
         return $value >= 0 && $value <= self::DEFAULT_MAX_BREADCRUMBS;
     }
+
+    /**
+     * Validates all the class serializers are callables indexed by string keys.
+     *
+     * @param array $serializers
```
ste93cry: Missing description of this parameter

src/Options.php
```diff
@@ -901,4 +926,22 @@ private function validateMaxBreadcrumbsOptions(int $value): bool
...
+     *
+     * @return bool
+     */
+    private function validateClassSerializers(array $serializers): bool
+    {
+        foreach ($serializers as $key => $serializer) {
```
ste93cry: Please rename the `$key` variable to `$className`

src/Serializer/AbstractSerializer.php
```diff
@@ -122,14 +122,51 @@ protected function serializeRecursively($value, int $_depth = 0)
         }
 
         if (\is_object($value)) {
+            $classSerializer = $this->resolveClassSerializer($value);
+
+            if (null !== $classSerializer) {
+                try {
+                    if (\is_array($serializedObjectData = $classSerializer($value))) {
```
ste93cry: Please split the assignment statement from the conditional statement

ste93cry:
I see a few drawbacks in this implementation:

- if the user wants to serialize something based not only on the type of the object but also on some other conditions he can't since resolving the serializer is based on the FQCN
- if the user implements its own serializer and he's using the new option suddenly it won't work anymore without him having to reimplement all the logic. The same problem was unfortunately also caused forcibly by the introdction of the `http_proxy` option which only works with a certain transport and means that we are leaking parts of some features in the options where they should not stay

Said this I obviously see the value of such feature, but I'm not convinced that this is the way to go

stayallive:
> if the user wants to serialize something based not only on the type of the object but also on some other conditions he can't since resolving the serializer is based on the FQCN

The user get's the object passed so we could allow a null return value to skip the serializer and default back to the regular serialization to fix this?

> if the user implements its own serializer and he's using the new option suddenly it won't work anymore without him having to reimplement all the logic. The same problem was unfortunately also caused forcibly by the introdction of the http_proxy option which only works with a certain transport and means that we are leaking parts of some features in the options where they should not stay

Their own serializer will work just fine... it won't magically replace their serializer. If they want to switch they can, if they don't they don't... 

> Said this I obviously see the value of such feature, but I'm not convinced that this is the way to go

Please convince me about a better way forward than :wink:

tests/Serializer/SerializerTest.php
stayallive: This has no interface, it needs a real object to be able to register the class as a serializable, the StubObject class is just a class with a public method.

stayallive:
I did make a change which allows the user to register a `class_serializer` to possibly overwrite the value of a `SerializableInterface` object.

Also allowing to return `null` from a serializer allows you to write multiple serializer for more specific classes if you wanted to or overwrite the `SerializableInterface` return for a certain object with a `class_serializer`.

The only "issue" here is sorting, because the order you register serializers in the `class_serializer` matters, if you first have `Exception` and next have a `RuntimeException` serializer (as an example) and the first serializer always returns something the `RuntimeException` serializer will never be called, switching them around would fix that. However as long as it's documented I don't think it's an issue worth fixing if it were fixable at all without serious performance penalties.

ste93cry:
Thinking more about this feature, if someone wants to customize how objects are serialized or represented then the best way is to write their own serializer, which is why the `SerializerInterface` interface exists in the first place: to allow users to replace the implementation with something else and have full control over it. This option may be useful indeed because, but imho we are trying to ease too much the life of the developer when it's his job to make such customizations with something that is not flexible enough in all conditions and that feel "incomplete" for this reason

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
   and change the type of the reported exception to `\Sentry\Exception\FatalErrorException` (#788)
 - Add a static factory method to create a breadcrumb from an array of data (#798)
 - Add support for `SENTRY_ENVRIONMENT` and `SENTRY_RELEASE` environment variables (#810)
+- Implement class serializers and interface to customize how objects are serialized (#809)
```
ste93cry: I think a better description would be `Add the `class_serializers` option to make it possible to customize how objects are serialized in the event payload`

src/Options.php
```diff
@@ -701,6 +701,28 @@ public function setMaxRequestBodySize(string $maxRequestBodySize): void
...
+     */
+    public function getClassSerializers(): array
+    {
+        return $this->options['class_serializers'];
+    }
+
+    /**
+     * Sets the class serializers.
+     *
+     * @param array $serializers
```
ste93cry: Change the description to `The list of serializer callbacks` and the typehint to `array<string, callable>`

src/Options.php
```diff
@@ -701,6 +701,28 @@ public function setMaxRequestBodySize(string $maxRequestBodySize): void
...
+     *
+     * @return array
+     */
+    public function getClassSerializers(): array
+    {
+        return $this->options['class_serializers'];
+    }
+
+    /**
+     * Sets the class serializers.
```
ste93cry: Maybe a better description would be `Sets a list of callables that will be called to customize how objects are serialized in the event's payload. The list must be a map of FQCN/callable pairs.`

src/Options.php
```diff
@@ -701,6 +701,28 @@ public function setMaxRequestBodySize(string $maxRequestBodySize): void
         $this->options = $this->resolver->resolve($options);
     }
 
+    /**
+     * Gets the type serializers.
+     *
+     * @return array
```
ste93cry: Please change the return type to `array<string, callable>`

src/Options.php
```diff
@@ -701,6 +701,28 @@ public function setMaxRequestBodySize(string $maxRequestBodySize): void
         $this->options = $this->resolver->resolve($options);
     }
 
+    /**
+     * Gets the type serializers.
```
ste93cry: Please change the description to: `Gets the callbacks used to customize how objects are serialized in the payload of the event.`

src/Options.php
```diff
@@ -940,4 +965,22 @@ private function validateMaxBreadcrumbsOptions(int $value): bool
     {
         return $value >= 0 && $value <= self::DEFAULT_MAX_BREADCRUMBS;
     }
+
+    /**
+     * Validates all the class serializers are callables indexed by class names.
```
ste93cry: Change the description to `Validate that the value passed to the `class_serializers` option is valid.`

src/Options.php
```diff
@@ -940,4 +965,22 @@ private function validateMaxBreadcrumbsOptions(int $value): bool
...
+     *
+     * @param array $serializers The value to validate
+     *
+     * @return bool
+     */
+    private function validateClassSerializers(array $serializers): bool
```
ste93cry: Rename the function to `validateClassSerializersOption`

src/Serializer/SerializableInterface.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\Serializer;
+
+interface SerializableInterface
```
ste93cry: Add a description to this interface: `This interface can be used to customize how an object is serialized in the payload of an event.`

tests/Serializer/SerializerTest.php
ste93cry: WDYT about using an anonymous class?

tests/Serializer/SerializerTest.php
```diff
@@ -96,6 +98,63 @@ public function testNull(bool $serializeAllObjects): void
...
+
+        $this->assertEquals('Object ' . \get_class($object), $this->invokeSerialization($serializer, $object));
+    }
+
+    public function testSerializableObjectThatThrowsAnException(): void
+    {
+        $serializer = $this->createSerializer();
+
+        $object = $this->createMock(SerializableInterface::class);
+        $object->method('toSentry')->willThrowException(new \Exception('Doesn\'t matter what the exception is.'));
```
ste93cry: Since this is a fluent interface, please split each call to a new line to ease review, debugging and readability

tests/Serializer/SerializerTest.php
```diff
@@ -96,6 +98,63 @@ public function testNull(bool $serializeAllObjects): void
...
+            'data' => $serializedValue,
+        ], $this->invokeSerialization($serializer, $object));
+    }
+
+    public function testSerializableThatReturnsNull(): void
+    {
+        $serializer = $this->createSerializer();
+
+        $object = $this->createMock(SerializableInterface::class);
+        $object->method('toSentry')->willReturn(null);
```
ste93cry: Since this is a fluent interface, please split each call to a new line to ease review, debugging and readability

tests/Serializer/SerializerTest.php
```diff
@@ -96,6 +98,63 @@ public function testNull(bool $serializeAllObjects): void
...
+    public function testSerializableObject(): void
+    {
+        $serializer = $this->createSerializer();
+
+        $serializedValue = [
+            'testing' => 'value',
+        ];
+
+        $object = $this->createMock(SerializableInterface::class);
+        $object->method('toSentry')->willReturn($serializedValue);
```
ste93cry: Since this is a fluent interface, please split each call to a new line to ease review, debugging and readability

tests/Serializer/SerializerTest.php
stayallive: Thought this was more understandable, and thought it wouldn't work. It does.

Changed.

stayallive:
@ste93cry I think I got all your suggestions in, good ones!

Jean85:
Build failing due to CS.

src/Options.php
```diff
@@ -701,6 +701,31 @@ public function setMaxRequestBodySize(string $maxRequestBodySize): void
...
+     * @return array<string, callable>
+     */
+    public function getClassSerializers(): array
+    {
+        return $this->options['class_serializers'];
+    }
+
+    /**
+     * Sets a list of callables that will be called to customize how objects are
+     * serialized in the event's payload. * The list must be a map of FQCN/callable
```
ste93cry: Typo between the "payload" word and the rest of the description

src/Options.php
```diff
@@ -940,4 +968,22 @@ private function validateMaxBreadcrumbsOptions(int $value): bool
     {
         return $value >= 0 && $value <= self::DEFAULT_MAX_BREADCRUMBS;
     }
+
+    /**
+     * Validate that the value passed to the `class_serializers` option is valid.
```
ste93cry: Typo: `Validates that the values [...] are valid`

src/Serializer/AbstractSerializer.php
```diff
@@ -122,14 +122,61 @@ protected function serializeRecursively($value, int $_depth = 0)
...
+                try {
+                    $serializedObjectData = $classSerializer($value);
+
+                    if (\is_array($serializedObjectData)) {
+                        return [
+                            'class' => \get_class($value),
+                            'data' => $this->serializeRecursively($serializedObjectData, $_depth + 1),
+                        ];
+                    }
+                } catch (\Exception $e) {
```
ste93cry: It probably makes more sense to catch `\Throwable` instead of `\Exception`

src/Serializer/SerializableInterface.php
```diff
@@ -0,0 +1,19 @@
...
+namespace Sentry\Serializer;
+
+/**
+ * This interface can be used to customize how an object is serialized in the
+ * payload of an event.
+ */
+interface SerializableInterface
+{
+    /**
+     * Return an array representation of the object for Sentry.
```
ste93cry: Typo: `Returns`

tests/Serializer/SerializerTest.php
```diff
@@ -96,6 +97,73 @@ public function testNull(bool $serializeAllObjects): void
         $this->assertNull($result);
     }
 
+    public function testRegisteredObjectSerializers(): void
+    {
+        $object = new class() {
```
ste93cry: Since this class does not require any argument in the constructor we can leave out the brackets

tests/Serializer/SerializerTest.php
```diff
@@ -96,6 +97,73 @@ public function testNull(bool $serializeAllObjects): void
...
+                    return [
+                        'purpose' => $object->getPurpose(),
+                    ];
+                },
+            ],
+        ]));
+
+        $this->assertEquals([
+            'class' => $objectClass,
+            'data' => $customSerializerCallback($object),
```
ste93cry: Please define statically the expected result. Calling the same code used in the real scenario does not makes sense because it it's bugged then the test is bugged too

tests/Serializer/SerializerTest.php
```diff
@@ -96,6 +97,73 @@ public function testNull(bool $serializeAllObjects): void
...
+        ], $this->invokeSerialization($serializer, $object));
+    }
+
+    public function testSerializableThatReturnsNull(): void
+    {
+        $serializer = $this->createSerializer();
+
+        $object = $this->createMock(SerializableInterface::class);
+        $object->method('toSentry')
+               ->willReturn(null);
```
ste93cry: Please align correctly the fluent interface methods (always indent of 4 spaces)

tests/Serializer/SerializerTest.php
```diff
@@ -96,6 +97,73 @@ public function testNull(bool $serializeAllObjects): void
...
+    {
+        $serializer = $this->createSerializer();
+
+        $serializedValue = [
+            'testing' => 'value',
+        ];
+
+        $object = $this->createMock(SerializableInterface::class);
+        $object->method('toSentry')
+               ->willReturn($serializedValue);
```
ste93cry: Please align correctly the fluent interface methods (always indent of 4 spaces)

tests/Serializer/SerializerTest.php
```diff
@@ -96,6 +97,73 @@ public function testNull(bool $serializeAllObjects): void
...
+        $this->assertEquals('Object ' . \get_class($object), $this->invokeSerialization($serializer, $object));
+    }
+
+    public function testSerializableObjectThatThrowsAnException(): void
+    {
+        $serializer = $this->createSerializer();
+
+        $object = $this->createMock(SerializableInterface::class);
+        $object->method('toSentry')
+               ->willThrowException(new \Exception('Doesn\'t matter what the exception is.'));
```
ste93cry: Please align correctly the fluent interface methods (always indent of 4 spaces)

tests/Serializer/SerializerTest.php
Jean85: :-1: with this, it would be less readable, and also maybe against the current CS.

stayallive: I don't think that's even possible and is an actual syntax error, maybe commented on the wrong line?

ste93cry: It's not a syntax error, it's perfectly a valid syntax to write `new class {}`. I wouldn't say that it's less readable, but if you both agree on this feel free leave it as is

stayallive: Ah figured it out, for the anonymous class, yep, removed them.

stayallive: The line for me was wrong, however it's against CS, so I cannot change it and keep the build green, so I kept it in the end. I agree it could be left out, but phpcs says no.

mfn:
I find it confusing that you *have* to return an array from the custom class serializer.

My first attempt was:
```php
    public static function serializeCarbon(CarbonInterface $carbon): string
    {
        return get_class($carbon) . ': ' . $carbon->toIso8601String();
    }
```
But I never received this value because of the following in `\Sentry\Serializer\AbstractSerializer::serializeRecursively`:
```php
…
                foreach ($classSerializers as $classSerializer) {
                    try {
                        $serializedObjectData = $classSerializer($value);

                        if (\is_array($serializedObjectData)) {
                            return [
                            'class' => \get_class($value),
                            'data' => $this->serializeRecursively($serializedObjectData, $_depth + 1),
                        ];
                        }
                    } catch (\Throwable $e) {
                        // Ignore any exceptions generated by a class serializer
                    }
                }
…
```
Note the ` if (\is_array($serializedObjectData)) {`.

It you don't return an array, it's silently discard.

Is that really the intention 🤔

Simply doing this works:
```php
    public static function serializeCarbon(CarbonInterface $carbon): array
    {
        return [$carbon->toIso8601String()];
    }
```

And turns out this way on Sentry:
image

It's the same but slightly differently.

I just… I found a very unexpected restriction 🤷‍♀️

Now I understand I can switch the _whole_ serialize but… IMHO it would make sense to me to support simply returning strings here?

stayallive:
It was a choice made. In hindsight might have been better to allow for strings to. Design was for serializing objects (key-value) in mind mostly. 

I think we can broaden it to allow for strings or other scalar types without BC but not 100% sure.

PR #810, opened 2019-04-24
Add support for SENTRY_ENVRIONMENT and SENTRY_RELEASE

### LINKED ISSUE #804: SENTRY_RELEASE and SENTRY_ENVIRONMENT
It looks like these two (`SENTRY_RELEASE` and `SENTRY_ENVIRONMENT`) are missing support in at least v2. I haven't checked v1.

See https://docs.sentry.io/error-reporting/configuration/?platform=php#release

I see support for it in `sentry-python` but not here.

nesl247:
This fixes #804.

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -684,10 +684,10 @@ final class Options
             'attach_stacktrace' => false,
             'context_lines' => 5,
             'enable_compression' => true,
-            'environment' => null,
+            'environment' => $_SERVER['SENTRY_ENVIRONMENT'] ?? null,
             'project_root' => null,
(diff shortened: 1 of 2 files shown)


nesl247:
`SENTRY_DSN` is already there. I simply copied what was done, hence why I didn't add any tests.

mfb:
Filed https://github.com/getsentry/raven-ruby/issues/901 as it's a bit confusing that some SDKs use SENTRY_CURRENT_ENV and some use SENTRY_ENVIRONMENT

PR #814, opened 2019-05-05
Remove final keyword from all classes.

leo108:
We should allow developers to extend from existing classes.

### DIFF
--- a/src/Breadcrumb.php
+++ b/src/Breadcrumb.php
@@ -11,7 +11,7 @@ use Sentry\Exception\InvalidArgumentException;
  *
  * @author Stefano Arlandini <sarlandini@alice.it>
  */
-final class Breadcrumb implements \JsonSerializable
+class Breadcrumb implements \JsonSerializable
 {
     /**
      * This constant defines the default breadcrumb type.
--- a/src/Client.php
+++ b/src/Client.php
@@ -14,7 +14,7 @@ use Sentry\Transport\TransportInterface;
  *
  * @author Stefano Arlandini <sarlandini@alice.it>
  */
-final class Client implements ClientInterface
+class Client implements ClientInterface
 {
     /**
      * The version of the protocol to communicate with the Sentry server.
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -39,7 +39,7 @@ use Sentry\Transport\TransportInterface;
  *
  * @author Stefano Arlandini <sarlandini@alice.it>
  */
-final class ClientBuilder implements ClientBuilderInterface
+class ClientBuilder implements ClientBuilderInterface
 {
     /**
      * @var Options The client options
--- a/src/Context/UserContext.php
+++ b/src/Context/UserContext.php
@@ -8,7 +8,7 @@ namespace Sentry\Context;
  * This class is a specialized version of the base `Context` adapted to work
  * for the user context.
  */
-final class UserContext extends Context
+class UserContext extends Context
 {
     /**
      * Gets the ID of the user.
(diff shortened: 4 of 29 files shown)


Jean85:
Also, I would suggest to read this blog post from Marco Pivetta about this subject: https://ocramius.github.io/blog/when-to-declare-classes-final/

ste93cry:
I agree with @Jean85, `final` is there for a reason and removing the keyword indistinctly is not good. If you have a use case where `final` doesn't allow you to do some thing we are keen to discuss about it, but we won't blindly accept such PR

wi1dcard:
However it's a good trick which makes maintainers freer to do breaking changes. But `sentry-php` is a library, not a project. Users who want to use and extend the sdk have to fork and release a new package. It’s already very difficult to migrate from v1.0 to v2.0. Please stop blocking us users.

lx1036:
Agree with @wi1dcard

ste93cry:
>But `sentry-php` is a library, not a project

Like all the other libraries out of here that use `final` or are migrating to using it (Symfony, Doctrine, Guzzle, etc)

>Users who want to use and extend the sdk have to fork and release a new package

I honestly don't understand why you say this. Everything is typehinted against the interface where it makes sense to do it, so you should be able to just replace the dependency and as long as your new implementation satisfies the contract it will work. If you could share what you think you can't do it would be great

Jean85:
> Everything is typehinted against the interface where it makes sense to do it, so you should be able to just replace the dependency and as long as your new implementation satisfies the contract it will work.

Exactly. Also, since all of this is OSS, no one is stopping you from even copy-pasting our implementations and hot-fixing them where you need it, so we're not "blocking" anyone.

leo108:
We want to send a few extra data from some certain exceptions, e.g., extract request and response info from `GuzzleHttp\Exception\RequestException`.

After some research, I found that we can use the global scope to reach this, **But** we have to instantiate and pass a `Scope` object to method `captureException` every time to make global scopes working due to code lines at <https://github.com/getsentry/sentry-php/blob/master/src/Client.php#L158-L160>.

For now, we have an expedient solution is to create a new class or helper function like `MySentry::captureException` which will new a `Scope` object if a `null` was provided in the argument, that's not good enough.

If we can inherit the `Sentry\Client` class, we can override the `captureException` method to create a new `Scope` object if it's `null`, easily and beautifully.

Jean85:
You don't need to do those kind of extension there, in the client: that is the main difference with 1.x, where the client was a god-class. Now everything is neat and separated.

You can use `Hub::configureScope` or `Hub::withScope` to change what is attached to the event, or use the `before_send` callback to change the event at the last moment, just before is being sent to the HTTP transport.

wi1dcard:
> Exactly. Also, since all of this is OSS, no one is stopping you from even copy-pasting our implementations and hot-fixing them where you need it, so we're not "blocking" anyone.

Copy-pasting makes me feel like backing to 10 years ago without GitHub and Packagist.

Jean85:
> Copy-pasting makes me feel like backing to 10 years ago without GitHub and Packagist.

Still, you can't ask us maintainers to do more job on your behalf. We need to protect the integrity of the lib for the sake of all users, or otherwise we risk breaking stuff, or being cornered in doing ugly stuff (like a new major version) just to fix something that requires a BC to fix it.

Also, as demonstrated by this (now closed) issue, we designed this new version with numerous extension points, that allow customization without breaking the encapsulation of our design. Feel free to ask for help if you have specific need that you feel is not addressed by the current design.

wi1dcard:
So that users need to rely on this "extension" system or copy-paste the whole class to extend it. Why not use private functions instead of final classes to protect it against BC?

HazAT:
Hey all, @Jean85 @ste93cry made our position clear this conversation is not going anywhere productive that's why I am locking this thread. 

I have not a lot to add to what they have said, this is the direction where all Sentry SDKs are going. 

Please open a new issue if you think we need to expose more "hooks" or entry points for you guys to extend, but making everything public was the way it was before and we are not going back.

I hope you understand our decision, thanks.

PR #816, opened 2019-05-08
A non-numeric value encountered in Stacktrace.php

Deifinger:
I get string $index and observe "A non-numeric value encountered" error

7abf0c17e8

Please, take a look at my solution. I'm not sure that this is correct fix and the root of the problem

### DIFF
--- a/src/Stacktrace.php
+++ b/src/Stacktrace.php
@@ -311,7 +311,7 @@ class Stacktrace implements \JsonSerializable
             $result = array_map([$this, 'serializeArgument'], $frame['args']);
         } else {
             foreach (array_values($frame['args']) as $index => $argument) {
-                $result['param' . ($index + 1)] = $this->serializeArgument($argument);
+                $result['param' . ((int)$index + 1)] = $this->serializeArgument($argument);
             }
         }
 


ste93cry:
Are you using XDebug or some other debug extension? If so, can you try disabling it and retrying?

Deifinger:
@ste93cry Yeah, I have XDebug in my project. When I disabled it, the error disappeared.

src/Stacktrace.php
```diff
@@ -311,7 +311,7 @@ protected function getFrameArgumentsValues(array $frame): array
             $result = array_map([$this, 'serializeArgument'], $frame['args']);
         } else {
             foreach (array_values($frame['args']) as $index => $argument) {
-                $result['param' . ($index + 1)] = $this->serializeArgument($argument);
+                $result['param' . ((int)$index + 1)] = $this->serializeArgument($argument);
```
ste93cry: So, the culprit of the problem as we found out is XDebug. The error occurs because it changes how the arguments of a stacktrace frame are reported by using the argument name as key of each value of the `$args` variable instead of its 0-based position in the function signature. Blindly casting `$index` to an `integer` then would result in some indesiderate results (e.g. `foo` would be converted to `0`). The best fix we can do is to check whether the `$index` variable is a `string` or an `integer` and act accordingly. However, this exact same issue should have been fixed in #761 which is the reason why there is an `if`/`else` statement wrapping the loop, so I don't get why it happens and it would be cool if you could debug a bit more in depth to see what's happening

sztyup:
The screenshot you provided does not contain the fix by #763, could you make sure you are using the latest version of the library and see if the issue is still there?

ste93cry:
>The screenshot you provided does not contain the fix by #763

Good catch, I didn’t even noticed it!

Deifinger:
I'm going to update dependency on Monday and check the result

Deifinger:
@ste93cry Sorry for delay. I updated dependencies and now all works fine. I have the next versions:
sentry/sdk                                2.0.2              
sentry/sentry                             2.0.1             

Ticket going to be closed.

PR #818, opened 2019-05-16
Catch serialization errors

Jean85:
This bug emerged in getsentry/sentry-symfony#63. It's pretty nasty, due to the fact that it happens when we serialize a callable that refers to a file that contains a parse error; but my solution should catch a lot of possible issues anyway.

This PR stems from 2.0.1, so we can merge this on a 2.0.x branch if we want to release a 2.0.2; otherwise, if it's nearly ready, we can pack it up in 2.1.

### DIFF
--- a/tests/phpt/serialize_broken_class.phpt
+++ b/tests/phpt/serialize_broken_class.phpt
@@ -1,0 +1,37 @@
+--TEST--
+Test that requiring a broken class (with a parser error) will not explode during serialization
+--FILE--
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\Tests;
+
+use Sentry\ClientBuilder;
+use Sentry\Event;
+use Sentry\Integration\FatalErrorListenerIntegration;
+use Sentry\Options;
+use Sentry\Serializer\Serializer;
+use Sentry\State\Hub;
+use Sentry\Tests\resources\BrokenClass;
+use Sentry\Transport\TransportInterface;
+
+$vendor = __DIR__;
+
+while (!file_exists($vendor . '/vendor')) {
+    $vendor = dirname($vendor);
+}
+
+require $vendor . '/vendor/autoload.php';
+
+function testSerialization($value) {
+    $serializer = new Serializer(new Options());
+
+    echo $serializer->serialize($value);
+}
+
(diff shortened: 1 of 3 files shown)


tests/phpt/serialize_broken_class.phpt
```diff
@@ -0,0 +1,40 @@
...
+
+$vendor = __DIR__;
+
+while (!file_exists($vendor . '/vendor')) {
+    $vendor = dirname($vendor);
+}
+
+require $vendor . '/vendor/autoload.php';
+
+function testSerialization($value) {
```
ste93cry: Any reason for this method to exists instead of directly instantiating the serializer and calling it?

tests/resources/BrokenClass.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Sentry\Tests\resources;
+
+class BrokenClass
```
ste93cry: What about putting this file into `Fixtures/code` instead of in the current location? I would like to drop the `resources` folder someday since it's just a sort of duplicate of the first

tests/resources/BrokenClass.php
```diff
@@ -0,0 +1,11 @@
+<?php
```
ste93cry: What about putting this file into `Fixtures/code` instead of in the current location? I would like to drop the `resources` folder someday since it's just a sort of duplicate of the first

tests/phpt/serialize_broken_class.phpt
Jean85: Yes, it's needed to reproduce the issue, which is triggered by the backtrace serialization.

ste93cry: Can you please add a comment explaining why it's needed and/or link to the issue? Otherwise it's not clear and in the future someone may think that it's useless like I did

tests/resources/BrokenClass.php
Jean85: I did it to leverage the PHPStan exclusion rule; moved and ignored it anyway.

ste93cry:
Missing update of the `CHANGELOG`

PR #821, opened 2019-05-23
Handle string callable in case of non-static method

Jean85:
This is an improvement over #815. Thanks to @dmitry-pro for spotting the bug to start with.


Jean85:
Umh, there's something strange... I cannot reproduce the bug outside of Sentry's tests: https://3v4l.org/1CEnC

What's really triggering the issue in here?

dmitry-pro:
@Jean85 don't have much time for testing, but it is 
https://3v4l.org/XMiVr
Strange that it doesn't trigger in Your (previous) example - when I caught this error, it was found in public dynamic method, not function.
I'll test it more when I free.

Jean85:
I've found the difference: the issue is present with functions, not with class methods: https://3v4l.org/4ibt7

Not sure why there's a difference, the bug seems to be very subtle.

ste93cry:
Unfortunately we cannot remove the typehint or we will have a BC

Jean85:
> Unfortunately we cannot remove the typehint or we will have a BC

Crap that's true... maybe we can hotfix it in the previous method, but it would still be broken for other usages

chekalsky:
@Jean85 When are you planning to release it? Pretty annoying bug :-)

src/Serializer/AbstractSerializer.php
```diff
@@ -261,17 +261,44 @@ protected function serializeValue($value)
...
+
+        if (\is_string($callable) && !\function_exists($callable)) {
+            return $callable;
+        }
+
+        return $this->serializeCallable($callable);
+    }
+
+    /**
+     * @internal Use serializeCallableWithoutTypeHint instead (no type in argument)
```
ste93cry: We cannot mark an already existing method as `@internal` or we may be tempted in the future to make changes without worrying about BC when we must instead since it was part of the public API in the past. Just leave the note instead with a link to the issue and remove the comment from the parameter instead

src/Serializer/AbstractSerializer.php
```diff
@@ -261,17 +261,44 @@ protected function serializeValue($value)
...
+     * since using the callable type raises a deprecation in some cases.
+     *
+     * @param callable $callable
+     *
+     * @return string
+     */
+    protected function serializeCallableWithoutTypeHint($callable): string
+    {
+        if (!\is_callable($callable)) {
+            throw new \InvalidArgumentException('Expecting callable, got ' . \gettype($callable));
```
ste93cry: Please use `sprintf` and handle the fact that `$callable` may be either an object or a scalar

src/Serializer/AbstractSerializer.php
```diff
@@ -261,17 +261,44 @@ protected function serializeValue($value)
...
+     *
      * @param callable $callable
      *
      * @return string
      */
-    protected function serializeCallable(callable $callable): string
+    protected function serializeCallable(/* callable type to be removed in 3.0, see #821 => */callable $callable): string
     {
+        if (!\is_callable($callable)) {
+            throw new \InvalidArgumentException('Expecting callable, got ' . \gettype($callable));
```
ste93cry: Please use `sprintf` and handle the fact that `$callable` may be either an object or a scalar

src/Serializer/AbstractSerializer.php
```diff
@@ -357,4 +383,21 @@ public function getSerializeAllObjects(): bool
     {
         return $this->serializeAllObjects;
     }
+
+    /**
+     * @param mixed $callable
+     *
+     * @throws InvalidArgumentException
+     */
+    private function assertIsValidCallable($callable): void
```
ste93cry: I don't see any good reason to have (in this case) a `private` method that just checks whether the argument is a callable and throws if it's not. Please inline the code where it belongs to

src/Serializer/AbstractSerializer.php
```diff
@@ -261,17 +262,42 @@ protected function serializeValue($value)
...
+     * Use serializeCallableWithoutTypeHint instead (no type in argument).
+     *
+     * @see https://github.com/getsentry/sentry-php/pull/821
+     *
      * @param callable $callable
      *
      * @return string
      */
-    protected function serializeCallable(callable $callable): string
+    protected function serializeCallable(/* callable type to be removed in 3.0, see #821 => */callable $callable): string
```
ste93cry: Instead of putting this comment here use the description of the method

composer.json
```diff
@@ -74,7 +74,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-develop": "2.2-dev"
+            "dev-develop": "2.3-dev"
```
ste93cry: Please revert this change as this PR is targeting `master`

emircanerkul:
Hi @ste93cry @Jean85 

I face the same issue now while using sentry/sentry 3.22.1.

Is there any regression?

```
Deprecated: Use of "static" in callables is deprecated in /app/vendor/sentry/sentry/src/Serializer/AbstractSerializer.php on line 265 Deprecated: Use of "static" in callables is deprecated in /app/vendor/sentry/sentry/src/Serializer/AbstractSerializer.php on line 265 Deprecated: Use of "static" in callables is deprecated in /app/vendor/sentry/sentry/src/Serializer/AbstractSerializer.php on line 265 Deprecated: Use of "static" in callables is deprecated in /app/vendor/sentry/sentry/src/Serializer/AbstractSerializer.php on line 265 Deprecated: Use of "static" in callables is deprecated in /app/vendor/sentry/sentry/src/Serializer/AbstractSerializer.php on line 265 Deprecated: Use of "static" in callables is deprecated in /app/vendor/sentry/sentry/src/Serializer/AbstractSerializer.php on line 265
```

ste93cry:
It may be, even though it's the first time it is reported since years... I suggest to try version `4.x` of the SDK, and if the issue persist open a new issue.

PR #822, opened 2019-05-23
Change the behavior of excluded_exceptions option

### LINKED ISSUE #820: Event without exceptions and an empty message
Hi, I use the last version of [sentry-symfony](https://github.com/getsentry/sentry-symfony) integration and I have issues with the `excluded_exceptions` option.

If an exception that match `excluded_exceptions` is throw then an `<unlabeled event>` event is created in Sentry:

Capture d’écran 2019-05-22 à 07 29 43

Is seems the problem came from this: https://github.com/getsentry/sentry-php/blob/master/src/EventFactory.php#L148 -> the event has no message and no exception as the exception are ignored.

Jean85:
This PR stems/fixes #820 and the related getsentry/sentry-symfony#216.

The current behavior of the `excluded_exceptions` option is broken, since it just excludes exceptions from the `exception` key in the event payload, which results in broken or empty exception chains reported.

Instead, the correct behavior should be ignoring the automatic capturing of a set of exceptions, like excluding 404 exceptions from frameworks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Unreleased
+
+- Fix the behavior of the `excluded_exceptions` option: now it's used to skip capture of exceptions, not to purge the `exception` data of the event, which resulted in broken or empty chains of exceptions in reported events (#822) 
+
 ## 2.1.0 (2019-05-22)
 
 - Mark Sentry internal frames when using `attach_stacktrace` as `in_app` `false` (#786)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Unreleased
 
-- Fix the behavior of the `excluded_exceptions` option: now it's used to skip capture of exceptions, not to purge the `exception` data of the event, which resulted in broken or empty chains of exceptions in reported events (#822) 
+- Fix the behavior of the `excluded_exceptions` option: now it's used to skip capture of exceptions, not to purge the 
+`exception` data of the event, which resulted in broken or empty chains of exceptions in reported events (#822) 
 
 ## 2.1.0 (2019-05-22)
 
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -133,18 +133,18 @@ class ClientTest extends TestCase
     {
         return [
             [
-                true, 
-                \Exception::class, 
+                true,
+                \Exception::class,
                 new \Error(),
             ],
             [
-                false, 
-                \Exception::class, 
+                false,
+                \Exception::class,
                 new \LogicException(),
             ],
             [
-                false, 
-                \Throwable::class, 
+                false,
+                \Throwable::class,
                 new \Error(),
             ],
         ];
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -107,6 +107,42 @@ class ClientTest extends TestCase
         $client->captureException($exception);
     }
 
+    /**
+     * @dataProvider captureExceptionWithExcludedExceptionsDataProvider
+     */
+    public function testCaptureExceptionWithExcludedExceptions(bool $shouldCapture, string $excluded, \Throwable $thrown): void
+    {
+        $transport = $this->createMock(TransportInterface::class);
+
+        $transport->expects($shouldCapture ? $this->once() : $this->never())
+            ->method('send')
+            ->with($this->callback(function (Event $event) use ($thrown): bool {
+                $this->assertCount(1, $event->getExceptions());
+
+                $exceptionData = $event->getExceptions()[0];
+
+                $this->assertSame(\get_class($thrown), $exceptionData['type']);
+                $this->assertSame($thrown->getMessage(), $exceptionData['value']);
+
+                return true;
+            }));
+
+        $client = ClientBuilder::create(['excluded_exceptions' => [$excluded]])
+            ->setTransport($transport)
+            ->getClient();
+
+        $client->captureException($thrown);
+    }
+
+    public function captureExceptionWithExcludedExceptionsDataProvider(): array
+    {
+        return [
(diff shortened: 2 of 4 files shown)


tests/ClientTest.php
```diff
@@ -107,6 +107,42 @@ public function testCaptureException(): void
         $client->captureException($exception);
     }
 
+    /**
+     * @dataProvider captureExceptionWithExcludedExceptionsDataProvider
+     */
+    public function testCaptureExceptionWithExcludedExceptions(bool $shouldCapture, string $excluded, \Throwable $thrown): void
```
ste93cry: Please rename this method to `testCaptureExceptionDoesNothingIfExcludedExceptionsOptionMatches` (so that it's a bit more descriptive). Also rename the parameter `$excluded` to `$excludedExceptions` and `$thrown` to `$exception`

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Unreleased
+
+- Fix the behavior of the `excluded_exceptions` option: now it's used to skip capture of exceptions, not to purge the `exception` data of the event, which resulted in broken or empty chains of exceptions in reported events (#822)
```
ste93cry: When possible please wrap the lines after ~ 80 characters to keep it consistent with the rest of the changelog

tests/ClientTest.php
```diff
@@ -107,6 +107,42 @@ public function testCaptureException(): void
...
+    {
+        $transport = $this->createMock(TransportInterface::class);
+
+        $transport->expects($shouldCapture ? $this->once() : $this->never())
+            ->method('send')
+            ->with($this->callback(function (Event $event) use ($thrown): bool {
+                $this->assertCount(1, $event->getExceptions());
```
ste93cry: We're not really interested in how many exceptions there are but just that there is at least one, so we can use `assertNotEmpty`

tests/ClientTest.php
```diff
@@ -107,6 +107,42 @@ public function testCaptureException(): void
...
+            }));
+
+        $client = ClientBuilder::create(['excluded_exceptions' => [$excluded]])
+            ->setTransport($transport)
+            ->getClient();
+
+        $client->captureException($thrown);
+    }
+
+    public function captureExceptionWithExcludedExceptionsDataProvider(): array
```
ste93cry: Please rename this method to `testCaptureExceptionDoesNothingIfExcludedExceptionsOptionMatchesDataProvider`

tests/ClientTest.php
```diff
@@ -107,6 +107,42 @@ public function testCaptureException(): void
...
+    {
+        return [
+            [true, \Exception::class, new \Error()],
```
ste93cry: Can we split this array into multiple lines (one for each item)? Also apply this fix to the rest of the items of the parent array

ste93cry:
I was thinking if it makes sense to do the check of the option in the `captureEvent` method instead of in `captureException` (which is just a "proxy" method). The reason is that if someone uses the first one the option will not work, which may not be consistent with the expected behavior

Jean85:
I considered that approach, but it's less clear: it would have to inspect the `exception` key of the payload, and it doesn't have a way to discern if that is the origin of the event, or if it's just a collateral.

IMHO the `captureException` conveys the correct requested behavior, and it's more controllable. If anyone wants to apply a blanket ignore on the `exception` key of the payload, then `before_send` is the correct way to go, and it's not something that we should touch ourselves.

ste93cry:
>it doesn't have a way to discern if that is the origin of the event, or if it's just a collateral

I suppose we should not care if the exception is originating the event or is just additional data of the payload as long as the option works coherently

>If anyone wants to apply a blanket ignore on the `exception` key of the payload, then `before_send` is the correct way to go, and it's not something that we should touch ourselves

Setting the `exception` key of the payload is the official way to set an exception on the event object, so applying the option on it would just provide coherent behavior across all methods. I'm really torn on what to do here, maybe an opinion from @HazAT or @stayallive could help us take a decision. Apart from this point, LGTM

stayallive:
Hmm this is a tricky one, but I would say `captureException` makes more sense. Since that deals with capturing an exception specifically. 

If I were to call `captureEvent` myself I would not expect it to drop the event because I included an ignored exception.

ste93cry:
>If I were to call `captureEvent` myself I would not expect it to drop the event because I included an ignored exception

Wouldn't the same reasoning apply to the `captureException` method?

stayallive:
Yes. Yes it would. But since its called excluded_exceptions this makes more sense IMO. 

Best case would be to only exclude “auto” captured exceptions. But we have no way of identifying them currently so IMO this would be the best approach.

ste93cry:
I'm not really happy with some options (this one being one of them) that apply only in certain situations since they make the behavior less predictable or not predictable at all. However, LGTM, let's see what happens and if someone else complains we will act accordingly 👍

stayallive:
Probably a good thing to actually add this option to the documentation and be explicit about how it works there. Than at least it’s documented.

ste93cry:
Yup, if someone can take care of opening a PR in the [getsentry/sentry-docs](https://github.com/getsentry/sentry-docs) repository it would be great!

Jean85:
Done in getsentry/sentry-docs#1019. To be fair, the docs already described the behavior that this PR implements, I only added the `captureEvent` clarification.

pgrimaud:
Hi,

A new release with this fix is ​​expected soon? 😄
I hate having package as dev-master to my composer.json files.

Thanks

rvanlaak:
Could there be a regression introduced with the v2.2.x minor? https://github.com/getsentry/sentry-symfony/issues/216#issuecomment-549875053

Jean85:
Nope, that's related to the Symfony bundle. Also, this MR is present and released with d407aa1000ece8fd265ef536aef90b0bd7a7ded5 since 2.1.1.

Issue #303, opened 2016-06-01
[1.x] PHP Fatal error:  Uncaught ErrorException: Undefined index: function in /lib/Raven/Stacktrace.php:89


TPETb:
Guess this is PHP7 problem

problem is with the code
`'function' => $nextframe['function'],`

`$nextframe` doesn't necessary contain `function` index, it can be for example

```
array (
  'file' => '/vagrant/dev/admin/vendor/composer/ClassLoader.php',
  'line' => 412,
  'include_filename' => '/vagrant/dev/admin/app/SaaS/Admin/Models/User.php',
  'args' => 
  array (
  ),
)
```

Addressing undefined index leads to error posted in subject.

dcramer:
I dont know the greatest way we could write a test for this (which I'd love to have, to make sure we don't fuck it up in the future), but the attached commit should resolve it and is safe with our upstream API.

I'm wondering if this is only for use statements, or if its also include/require functions, and if we should mock a function call as well.

pelmered:
I know it's an old issue, but I have the same issue on this line: https://github.com/getsentry/sentry-php/blob/master/src/Stacktrace.php#L119

Jean85:
Feel free to write a PR to fix this issue too. Also, I would like to remind you that we have a 2.0 beta release which is near its stable release, you should try it.

pelmered:
Sure, I'll submit a PR. 
I'm already using 2.0.0-beta1. Is the master branch for 2.0 development?

ste93cry:
Yes it is. I'm wondering in which case the function name isn't available in the frame though...

Jean85:
Maybe if it's a closure/lambda?

ste93cry:
I don't think so because the `function` key at the line that is crashing is used only if the `class` is available too... By the way there is for sure some strange case, but the fix should be as simple as adding an `isset($frame['function`])` here:

https://github.com/getsentry/sentry-php/blob/dee22302443ccccbdec085ae3de1b57ca827618e/src/Stacktrace.php#L118

Jean85:
@pelmered are you able to create a reproducing case? I would like to port the resulting test to the master branch in case.

adnedelcu:
@Jean85 it happened to me too. I managed to reproduce it by using the Laravel framework version of the library, which in turn uses this, and it takes too long to return a response.

A case where there is a class index but not a function index is when the class for which the frame is generated is not executing a function but including / requiring a file.

[...]

PR #823, opened 2019-05-27
Fix undefined index generated by missing function in class

adnedelcu:
Fixing issue #303 for version 2.x.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -69,8 +69,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.0.x-dev",
-            "dev-fix-undefined-index": "2.0.x-dev"
+            "dev-master": "2.0.x-dev"
         }
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -69,7 +69,8 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.0.x-dev"
+            "dev-master": "2.0.x-dev",
+            "dev-fix-undefined-index": "2.0.x-dev"
         }
     }
 }
--- a/src/Stacktrace.php
+++ b/src/Stacktrace.php
@@ -117,8 +117,6 @@ class Stacktrace implements \JsonSerializable
 
         if (isset($backtraceFrame['class']) && isset($backtraceFrame['function'])) {
             $functionName = sprintf('%s::%s', $backtraceFrame['class'], $backtraceFrame['function']);
-        } elseif (isset($backtraceFrame['class'])) {
-            $functionName = $backtraceFrame['class'];
         } elseif (isset($backtraceFrame['function'])) {
             $functionName = $backtraceFrame['function'];
         } else {
--- a/src/Stacktrace.php
+++ b/src/Stacktrace.php
@@ -115,8 +115,10 @@ class Stacktrace implements \JsonSerializable
             $line = (int) $matches[2];
         }
 
-        if (isset($backtraceFrame['class'])) {
+        if (isset($backtraceFrame['class']) && isset($backtraceFrame['function'])) {
             $functionName = sprintf('%s::%s', $backtraceFrame['class'], $backtraceFrame['function']);
+        } elseif (isset($backtraceFrame['class'])) {
+            $functionName = $backtraceFrame['class'];
         } elseif (isset($backtraceFrame['function'])) {
             $functionName = $backtraceFrame['function'];
         } else {
--- a/tests/Fixtures/backtraces/anonymous_frame.json
+++ b/tests/Fixtures/backtraces/anonymous_frame.json
@@ -7,9 +7,6 @@
             "class": "TestClass"
         },
         {
-            "class": "CrashyClass"
-        },
-        {
             "file": "path/to/file",
             "line": 7,
             "function": "call_user_func"
--- a/tests/Fixtures/backtraces/anonymous_frame.json
+++ b/tests/Fixtures/backtraces/anonymous_frame.json
@@ -7,6 +7,9 @@
             "class": "TestClass"
         },
         {
+            "class": "CrashyClass"
+        },
+        {
             "file": "path/to/file",
             "line": 7,
             "function": "call_user_func"
--- a/tests/Fixtures/backtraces/exception.json
+++ b/tests/Fixtures/backtraces/exception.json
@@ -4,6 +4,11 @@
     "backtrace": [
         {
             "file": "path/to/file",
+            "line": 30,
+            "class": "CrashyClass"
+        },
+        {
+            "file": "path/to/file",
             "function": "triggerError",
             "line": 7,
             "class": "TestClass"
--- a/tests/Fixtures/backtraces/exception.json
+++ b/tests/Fixtures/backtraces/exception.json
@@ -4,11 +4,6 @@
     "backtrace": [
         {
             "file": "path/to/file",
-            "line": 30,
-            "class": "CrashyClass"
-        },
-        {
-            "file": "path/to/file",
             "function": "triggerError",
             "line": 7,
             "class": "TestClass"
--- a/tests/Fixtures/frames/class.json
+++ b/tests/Fixtures/frames/class.json
@@ -1,0 +1,5 @@
+{
+    "file": "path/to/file",
+    "line": 12,
+    "class": "CrashyClass"
+}
--- a/tests/StacktraceTest.php
+++ b/tests/StacktraceTest.php
@@ -39,9 +39,9 @@ final class StacktraceTest extends TestCase
     {
         $stacktrace = new Stacktrace($this->options, $this->serializer, $this->representationSerializer);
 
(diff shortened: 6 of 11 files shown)


composer.json
```diff
@@ -69,7 +69,8 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.0.x-dev"
+            "dev-master": "2.0.x-dev",
+            "dev-fix-undefined-index": "2.0.x-dev"
```
Jean85: This doubles the alias, it only leads to errors, please revert it.

src/Stacktrace.php
```diff
@@ -115,8 +115,10 @@ public function addFrame(string $file, int $line, array $backtraceFrame): void
             $line = (int) $matches[2];
         }
 
-        if (isset($backtraceFrame['class'])) {
+        if (isset($backtraceFrame['class']) && isset($backtraceFrame['function'])) {
             $functionName = sprintf('%s::%s', $backtraceFrame['class'], $backtraceFrame['function']);
+        } elseif (isset($backtraceFrame['class'])) {
```
ste93cry: In which case there can be the `class` key but not the `function` key? If there is no real reason for this `elseif` statement I would simply leave the following cases:

- `isset($backtraceFrame['class']) && isset($backtraceFrame['function'])`
- `isset($backtraceFrame['function'])`
- everything else

adnedelcu: You can check issue #303, for which this PR was created. The issue described there doesn't affect only version 1.x, but also 2.x. you can also see me commenting a case when the key function does not exist.

ste93cry: It I’m not missing something I see no reason the `class` key should be present but not the `function` key. This is what I meant

adnedelcu: Yes, I know, but yet there are some cases where it happens. One of the cases is when a class includes another file that has errors in it, as per my example in the issue.

composer.json
adnedelcu: Fixed

src/Stacktrace.php
adnedelcu: You can check [this comment](https://github.com/getsentry/sentry-php/issues/303#issuecomment-470053718) for more details of a case when it happened to me.

tests/Fixtures/frames/class.json
```diff
@@ -0,0 +1,5 @@
+{
+    "file": "path/to/file",
+    "line": 12,
+    "class": "CrashyClass"
```
ste93cry: It really changes nothing, but to be coherent with the rest of the fixtures please use `TestClass` as class name

tests/StacktraceTest.php
```diff
@@ -56,6 +58,7 @@ public function testStacktraceJsonSerialization(): void
 
         $stacktrace->addFrame('path/to/file', 1, ['file' => 'path/to/file', 'line' => 1, 'function' => 'test_function']);
         $stacktrace->addFrame('path/to/file', 2, ['file' => 'path/to/file', 'line' => 2, 'function' => 'test_function', 'class' => 'TestClass']);
+        $stacktrace->addFrame('path/to/file', 3, ['file' => 'path/to/file', 'line' => 2, 'class' => 'CrashyClass']);
```
ste93cry: To be coherent with the rest of the code please use `TestClass` as class name

tests/StacktraceTest.php
```diff
@@ -80,10 +84,11 @@ public function testAddFrame(): void
 
         $frames = $stacktrace->getFrames();
 
-        $this->assertCount(3, $frames);
-        $this->assertFrameEquals($frames[0], 'TestClass::test_function', 'path/to/file', 12);
-        $this->assertFrameEquals($frames[1], 'test_function', 'path/to/file', 12);
+        $this->assertCount(4, $frames);
+        $this->assertFrameEquals($frames[0], 'CrashyClass', 'path/to/file', 12);
```
ste93cry: Please use `TestClass` as class name to be coherent with the rest of the code

tests/Fixtures/backtraces/anonymous_frame.json
```diff
@@ -6,10 +6,13 @@
             "function": "triggerError",
             "class": "TestClass"
         },
+        {
```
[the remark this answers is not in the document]
adnedelcu: Reverted.

tests/Fixtures/backtraces/exception.json
```diff
@@ -2,6 +2,11 @@
     "file": "path/to/file",
     "line": 12,
     "backtrace": [
+        {
```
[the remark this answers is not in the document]
adnedelcu: Reverted.

tests/StacktraceTest.php
```diff
@@ -72,6 +75,7 @@ public function testAddFrame(): void
             $this->getJsonFixture('frames/eval.json'),
             $this->getJsonFixture('frames/runtime_created.json'),
             $this->getJsonFixture('frames/function.json'),
+            $this->getJsonFixture('frames/class.json'),
```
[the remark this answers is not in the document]
adnedelcu: Renamed.

tests/Fixtures/frames/class.json
adnedelcu: Renamed.

tests/StacktraceTest.php
adnedelcu: Renamed.

tests/StacktraceTest.php
adnedelcu: Renamed.

src/Stacktrace.php
ste93cry: Sentry treats a string without the `::` as a function name and in this case instead we would (incorrectly) report the name of the class as the name of the function. Please remove this `elseif` statement.

tests/StacktraceTest.php
```diff
@@ -39,15 +39,17 @@ public function testGetFramesAndToArray(): void
     {
         $stacktrace = new Stacktrace($this->options, $this->serializer, $this->representationSerializer);
 
-        $stacktrace->addFrame('path/to/file', 1, ['file' => 'path/to/file', 'line' => 1, 'function' => 'test_function']);
-        $stacktrace->addFrame('path/to/file', 2, ['file' => 'path/to/file', 'line' => 2, 'function' => 'test_function', 'class' => 'TestClass']);
+        $stacktrace->addFrame('path/to/file', 1, ['file' => 'path/to/file', 'line' => 2, 'class' => 'CrashyClass']);
```
ste93cry: It doesn't really matter given that this is a test, but for consistency the value of the line item in the 3rd argument should equal the value of the 2nd parameter as they represents the same thing which is the line at which the error originated

tests/StacktraceTest.php
```diff
@@ -39,15 +39,17 @@ public function testGetFramesAndToArray(): void
...
+        $stacktrace->addFrame('path/to/file', 2, ['file' => 'path/to/file', 'line' => 1, 'function' => 'test_function']);
```
ste93cry: It doesn't really matter given that this is a test, but for consistency the value of the line item in the 3rd argument should equal the value of the 2nd parameter as they represents the same thing which is the line at which the error originated

tests/StacktraceTest.php
```diff
@@ -39,15 +39,17 @@ public function testGetFramesAndToArray(): void
...
+        $stacktrace->addFrame('path/to/file', 3, ['file' => 'path/to/file', 'line' => 2, 'function' => 'test_function', 'class' => 'TestClass']);
```
ste93cry: It doesn't really matter given that this is a test, but for consistency the value of the line item in the 3rd argument should equal the value of the 2nd parameter as they represents the same thing which is the line at which the error originated

src/Stacktrace.php
adnedelcu: If I remove the `elseif` then the bug will return.

ste93cry: No, it will execute the next `elseif` that won't match too and then it will stop in the `else` statement. To clarify, the code should look like:

```php
if (isset($backtraceFrame['class']) && isset($backtraceFrame['function'])) {
    $functionName = sprintf('%s::%s', $backtraceFrame['class'], $backtraceFrame['function']);
} elseif (isset($backtraceFrame['function'])) {
    $functionName = $backtraceFrame['function'];
} else {
    $functionName = null;
}
```

In case the `class` key is present in `$backtraceFrame` but not the `function` key we should not report as function name the class name

ste93cry:
I though we already merged this but I was wrong. @adnedelcu are you willing to finish the PR?

adnedelcu:
@ste93cry Yes, I am willing but I have been busy relocating. I will try to do the changes tomorrow.

ste93cry:
Thank you for the great patience and for your contribution of course :tada:

adnedelcu:
So sorry it took so long.

PR #831, opened 2019-06-11
Read REMOTE_ADDR from server parameters

alemax-xyz:
Currently server IP is reported in a _User IP_ field because the library is checking for `REMOTE_ADDR` HTTP header.

`REMOTE_ADDR` is not an HTTP header, but rather a server variable.

This PR reads IP from the right place.

### DIFF
--- a/src/Integration/RequestIntegration.php
+++ b/src/Integration/RequestIntegration.php
@@ -96,8 +96,9 @@ final class RequestIntegration implements IntegrationInterface
         }
 
         if ($self->options->shouldSendDefaultPii()) {
-            if ($request->hasHeader('REMOTE_ADDR')) {
-                $requestData['env']['REMOTE_ADDR'] = $request->getHeaderLine('REMOTE_ADDR');
+            $server = $request->getServerParams();
+            if (isset($server['REMOTE_ADDR'])) {
+                $requestData['env']['REMOTE_ADDR'] = $server['REMOTE_ADDR'];
             }
 
             $requestData['cookies'] = $request->getCookieParams();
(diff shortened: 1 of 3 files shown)


src/Integration/RequestIntegration.php
```diff
@@ -96,17 +96,18 @@ public static function applyToEvent(self $self, Event $event, ?ServerRequestInte
         }
 
         if ($self->options->shouldSendDefaultPii()) {
-            if ($request->hasHeader('REMOTE_ADDR')) {
-                $requestData['env']['REMOTE_ADDR'] = $request->getHeaderLine('REMOTE_ADDR');
+            $server = $request->getServerParams();
+            if (isset($server['REMOTE_ADDR'])) {
```
ste93cry: Please add a blank line above this one to improve readability

src/Integration/RequestIntegration.php
```diff
@@ -96,17 +96,18 @@ public static function applyToEvent(self $self, Event $event, ?ServerRequestInte
         }
 
         if ($self->options->shouldSendDefaultPii()) {
-            if ($request->hasHeader('REMOTE_ADDR')) {
-                $requestData['env']['REMOTE_ADDR'] = $request->getHeaderLine('REMOTE_ADDR');
+            $server = $request->getServerParams();
```
ste93cry: `$serverParams` would be a better name for this variable imo

ste93cry:
Good catch, thank you for your contribution!

src/Integration/RequestIntegration.php
alemax-xyz: Done

src/Integration/RequestIntegration.php
alemax-xyz: Done

Issue #833, opened 2019-06-13
Monolog handler fails when the extra data has numeric array keys


pontus-mp:
If $key is an integer in the following line, the function call will fail with `Uncaught TypeError: Argument 1 passed to Sentry\State\Scope::setExtra() must be of the type string, integer given`.

https://github.com/getsentry/sentry-php/blob/50f58938ef857c398e188b79c2fc5036f1adbeb2/src/Monolog/Handler.php#L62

I suggest adding a cast to string there, or possibly making the first argument to setExtra more relaxed.

ste93cry:
Good catch! I don't have an example in mind of why the `$key` should not be a `string`, so I would say that casting is the best solution here. Are you willing to open a PR to fix it?

pontus-mp:
Sure, done!

PR #834, opened 2019-06-14
Fix exception in Monolog handler when extra data has numeric array keys

pontus-mp:
This fixes issue #833. A TypeError is raised if $key is numeric (either if the array key was an integer, or if the string was convertible to integer).

### DIFF
--- a/tests/Monolog/HandlerTest.php
+++ b/tests/Monolog/HandlerTest.php
@@ -281,5 +281,30 @@ final class HandlerTest extends TestCase
                 'bar.tag' => 'bar tag value',
             ],
         ];
+
+        yield [
+            [
+                'message' => 'foo bar',
+                'level' => Logger::INFO,
+                'level_name' => Logger::getLevelName(Logger::INFO),
+                'channel' => 'channel.foo',
+                'context' => [
+                    'extra' => [
+                        1 => 'numeric key',
+                    ],
+                ],
+                'extra' => [],
+            ],
+            [
+                'level' => Severity::info(),
+                'message' => 'foo bar',
+            ],
+            [
+                'monolog.channel' => 'channel.foo',
+                'monolog.level' => Logger::getLevelName(Logger::INFO),
+                1 => 'numeric key',
+            ],
+            [],
+        ];
     }
 }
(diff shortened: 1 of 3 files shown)


pontus-mp:
Fixed, and added!

tests/Monolog/HandlerTest.php
```diff
@@ -281,5 +281,30 @@ public function handleDataProvider(): \Generator
...
+                'extra' => [],
+            ],
+            [
+                'level' => Severity::info(),
+                'message' => 'foo bar',
+            ],
+            [
+                'monolog.channel' => 'channel.foo',
+                'monolog.level' => Logger::getLevelName(Logger::INFO),
+                1 => 'numeric key',
```
ste93cry: I would expect this to be a `string` and not an `integer`. Test is passing because the wonderful PHP casts strings containing numbers without decimals to integers when they are used as key of the arrays and so the comparison succeeds :man_facepalming:

```php
$a = [
    1 => 'foo',
];

$b = [
    '1' => 'foo',
];

var_dump($a == $b); // true
var_dump($a === $b); // true

$a = [
    1.2 => 'foo',
];

$b = [
    '1.2' => 'foo',
];

var_dump($a == $b); // false
var_dump($a === $b); // false
```

Please change the test case so that a float is used, this way the test should fail if something is wrong in the comparison of the expected output with the input data

pontus-mp: Attempting to use floats as array keys causes them to get truncated to integers, so that won't help.
```php
$a = [
    1.2 => 'foo',
];

$b = [
    '1' => 'foo',
];

var_dump($a == $b); // true
var_dump($a === $b); // true
```
The test doesn't pass without the cast (due to the exception) though, so from that point of view it's functional.

Where in the changelog should I add the notice, under 2.1.1?

ste93cry: >Where in the changelog should I add the notice, under 2.1.1?

Under the `Unreleased` section :+1:

pontus-mp: Sounds good!

ste93cry:
Thank you for your first contribution to the project :tada:

PR #839, opened 2019-06-25
Add support for contexts

williamdes:
https://github.com/getsentry/sentry-javascript/blob/faf038bef8d3812982f258e8630d7d3521a9ff02/packages/integrations/src/vue.ts#L93

https://github.com/getsentry/sentry-javascript/blob/c3b794026591ad40325f38250b852070f21c2982/packages/browser/test/index.test.ts#L45

`setContext` is missing
Contexts are very useful, the javascript sdk implements it correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix `Client::captureEvent` not considering the `attach_stacktrace` option (#940)
+
 ## 2.2.6 (2019-12-18)
 
 - Fix remaining PHP 7.4 deprecations (#930)
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -35,8 +35,6 @@ matrix:
 cache:
     - composer.phar
     - '%LOCALAPPDATA%\Composer\files'
-    - '%PROGRAMDATA%\chocolatey\bin -> .appveyor.yml'
-    - '%PROGRAMDATA%\chocolatey\lib -> .appveyor.yml'
     - C:\php -> .appveyor.yml
 
 init:
--- a/src/Client.php
+++ b/src/Client.php
@@ -108,7 +108,7 @@ final class Client implements FlushableClientInterface
      */
     public function captureEvent(array $payload, ?Scope $scope = null): ?string
     {
-        $event = $this->prepareEvent($payload, $scope);
+        $event = $this->prepareEvent($payload, $scope, $this->options->shouldAttachStacktrace());
 
         if (null === $event) {
             return null;
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -111,9 +111,9 @@ class ClientTest extends TestCase
         ];
     }
 
-    public function testCapture(): void
+    public function testCaptureEvent(): void
     {
-        /** @var TransportInterface|MockObject $transport */
+        /** @var TransportInterface&MockObject $transport */
         $transport = $this->createMock(TransportInterface::class);
         $transport->expects($this->once())
             ->method('send')
@@ -133,6 +133,43 @@ class ClientTest extends TestCase
         ];
 
         $this->assertEquals('500a339f3ab2450b96dee542adf36ba7', $client->captureEvent($inputData));
+    }
+
+    /**
+     * @dataProvider captureEventAttachesStacktraceAccordingToAttachStacktraceOptionDataProvider
+     */
+    public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOption(bool $shouldAttachStacktrace): void
+    {
+        /** @var TransportInterface&MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(static function (Event $event) use ($shouldAttachStacktrace): bool {
+                if ($shouldAttachStacktrace && null === $event->getStacktrace()) {
+                    return false;
+                }
+
+                if (!$shouldAttachStacktrace && null !== $event->getStacktrace()) {
+                    return false;
+                }
+
+                return true;
+            }))
+            ->willReturn('500a339f3ab2450b96dee542adf36ba7');
+
+        $client = ClientBuilder::create(['attach_stacktrace' => $shouldAttachStacktrace])
+            ->setTransport($transport)
+            ->getClient();
+
+        $this->assertEquals('500a339f3ab2450b96dee542adf36ba7', $client->captureEvent([]));
+    }
+
+    public function captureEventAttachesStacktraceAccordingToAttachStacktraceOptionDataProvider(): array
+    {
+        return [
+            [true],
+            [false],
+        ];
     }
 
     public function testCaptureLastError(): void
@@ -372,17 +409,7 @@ class ClientTest extends TestCase
         $transport->expects($this->once())
             ->method('send')
             ->with($this->callback(function (Event $event): bool {
-                $result = $event->getStacktrace();
-
-                $this->assertInstanceOf(Stacktrace::class, $result);
-
-                $frames = $result->getFrames();
-
-                $this->assertNotEmpty($frames);
-                $this->assertTrue($frames[0]->isInApp());
-                $this->assertFalse($frames[\count($frames) - 1]->isInApp());
-
-                return true;
+                return null !== $event->getStacktrace();
             }));
 
         $client = ClientBuilder::create(['attach_stacktrace' => true])
@@ -390,25 +417,6 @@ class ClientTest extends TestCase
             ->getClient();
 
         $client->captureMessage('test');
-    }
-
-    public function testAttachStacktraceShouldNotWorkWithCaptureEvent(): void
-    {
-        /** @var TransportInterface|MockObject $transport */
-        $transport = $this->createMock(TransportInterface::class);
-        $transport->expects($this->once())
-            ->method('send')
-            ->with($this->callback(function (Event $event): bool {
-                $this->assertNull($event->getStacktrace());
-
-                return true;
-            }));
-
-        $client = ClientBuilder::create(['attach_stacktrace' => true])
-            ->setTransport($transport)
-            ->getClient();
-
-        $client->captureEvent([]);
     }
 
     /**


williamdes:
Maybe my works is not perfect, I can improve it though.

williamdes:
image
```json
"contexts":{"akey":{"0":"avalue !!","type":"akey"},"keyscope":{"0":"efniurnuenrgergn","type":"keyscope"},"os":{"type":"os","name":"Linux"},"akey2":{"3":"ok","type":"akey2","4":{"rf0":"ok","key6":{"a":"b"},"key5":["ok"],"key4":"value"}}
```

src/State/Scope.php
```diff
@@ -96,6 +102,33 @@ public function getTags(): array
...
+     * Sets a new tag in the contexts context.
+     *
+     * @param string $key   The key that uniquely identifies the context
+     * @param array  $value The value
+     *
+     * @return $this
+     */
+    public function setContext(string $key, array $value): self
+    {
+        $this->contexts[$key] = (object) $value;
```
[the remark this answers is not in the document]
williamdes: You can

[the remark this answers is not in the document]
williamdes: https://www.php.net/manual/en/language.types.object.php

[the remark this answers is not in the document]
williamdes:
```php
<?php
$obj = (object) 'ciao';
echo $obj->scalar;  // outputs 'ciao'
?>
```

:)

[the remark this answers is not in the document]
Jean85: I didn't say you can't, I said you shouldn't. That representation doesn't benefit of a lot of performance improvements due to how arrays are handled; so the cast is an overhead, with no benefit, it's not idiomatic and it may suffer further performance issues.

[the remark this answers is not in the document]
williamdes: Oh okay, sorry
I need to handle this another way since Sentry only accepts objects :)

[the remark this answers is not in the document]
williamdes: @Jean85 I fixed it :)

williamdes:
@HazAT @Jean85 Can you please review ?

src/State/Scope.php
```diff
@@ -96,6 +102,38 @@ public function getTags(): array
...
 
+    /**
+     * Sets a new tag in the contexts context.
+     *
+     * @param string $key   The key that uniquely identifies the context
+     * @param array  $value The value
+     *
+     * @return $this
+     */
+    public function setContext(string $key, array $value): self
```
HazAT: Please rename `$key` to `$name`

src/Event.php
```diff
@@ -636,6 +652,10 @@ public function toArray(): array
             $data['contexts']['runtime'] = $this->runtimeContext->toArray();
         }
 
+        if (!$this->contexts->isEmpty()) {
```
HazAT: This has to be changed.
We want to merge with existing `$data['contexts']` in the event but the one set from the user should be stronger.
Please also provide tests for it since this is a common use case.

HazAT:
Thanks for the PR, good job, we definitely want to have `setContext`. I left a few notes.

src/State/Scope.php
```diff
@@ -96,6 +102,38 @@ public function getTags(): array
         return $this->tags->toArray();
     }
 
+    /**
+     * Sets a new tag in the contexts context.
```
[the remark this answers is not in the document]
williamdes: Thanks, done

src/Event.php
williamdes: Done :)

williamdes:
@HazAT I did the changes you suggested, please review one last time :)

williamdes:
@HazAT While reviewing I found out that I had forgot to add some lines (b60017bb8d07f6ca8cbe530d8ed3fc8720b9c3f7)

Changelog added by ac56ad207d5e09be22ac6218600f2f83ccbb5887

src/State/Scope.php
```diff
@@ -96,6 +102,38 @@ public function getTags(): array
...
+            $this->contexts[$name]->{$arrayKey} = $arrayValue;
+        }
+
+        return $this;
+    }
+
+    /**
+     * Gets the tags contained in the contexts context.
+     *
+     * @return array<string, array>
```
ste93cry: The typehint doesn't looks correct to me, I would expect something like `array<string, \stdClass>`

src/State/Scope.php
```diff
@@ -96,6 +102,38 @@ public function getTags(): array
...
+     * @param array  $value The value
+     *
+     * @return self
+     */
+    public function setContext(string $name, array $value): self
+    {
+        if (!isset($this->contexts[$name])) {
+            $this->contexts[$name] = new \stdClass();
+        }
+        foreach ($value as $arrayKey => $arrayValue) {
```
ste93cry: Please add a line above this one to improve readability

src/State/Scope.php
```diff
@@ -96,6 +102,38 @@ public function getTags(): array
...
+     *
+     * @param string $name  The name that uniquely identifies the context
+     * @param array  $value The value
+     *
+     * @return self
+     */
+    public function setContext(string $name, array $value): self
+    {
+        if (!isset($this->contexts[$name])) {
+            $this->contexts[$name] = new \stdClass();
```
ste93cry: I don't really like to have to use `stdClass`, what's the point of doing it instead of using `$value` directly?

williamdes: Sentry only accepts objects for `contexts`, any other value will make an error..

src/Context/Context.php
```diff
@@ -12,6 +12,11 @@
  */
 class Context implements \ArrayAccess, \JsonSerializable, \IteratorAggregate
 {
+    /**
+     * This constant defines the alias used for the contexts.
+     */
+    public const CONTEXT_CONTEXTS = 'contexts';
```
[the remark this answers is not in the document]
williamdes: True, removed.

src/State/Scope.php
```diff
@@ -64,6 +69,7 @@ final class Scope
      */
     public function __construct()
     {
+        $this->contexts = new Context();
```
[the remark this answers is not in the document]
williamdes: Done

tests/State/ScopeTest.php
```diff
@@ -172,6 +186,10 @@ public function testApplyToEvent(): void
         $this->assertSame([$breadcrumb], $event->getBreadcrumbs());
         $this->assertEquals(['foo' => 'bar'], $event->getTagsContext()->toArray());
         $this->assertEquals(['bar' => 'foo'], $event->getExtraContext()->toArray());
+        $this->assertEquals([
```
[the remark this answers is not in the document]
williamdes: Done ^^

src/State/Scope.php
```diff
@@ -96,6 +102,38 @@ public function getTags(): array
...
+
+        return $this;
+    }
+
+    /**
+     * Gets the tags contained in the contexts context.
+     *
+     * @return array<string, array>
+     *
+     * @internal
```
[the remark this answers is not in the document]
williamdes: No idea, copy pasta
fixed.

src/State/Scope.php
williamdes: Correct, done

williamdes:
@ste93cry I made the changes you requested :)

src/State/Scope.php
Jean85: This is reasonable only if the user is using numeric-only keys, so `json_encode` will mistake it for an array.

ste93cry: @wedamija you're right, but there is only one case that cames  to my mind that could cause PHP to JSON-encode in the wrong format the array: when it represents a pure list (so 0-based index). In all other cases it should work, and I expect people to use the context more as a map than a list. Of course this cannot be strictly enforced, we could add a check like in the `TagsContext` that the key is a string until PHP will have true generics that will allow typehinting the arrays

williamdes: @ste93cry What should I do ?

src/State/Scope.php
```diff
@@ -96,6 +102,38 @@ public function getTags(): array
...
+        }
+
+        return $this;
+    }
+
+    /**
+     * Gets the tags contained in the contexts context.
+     *
+     * @return array<string, \stdClass>
+     *
```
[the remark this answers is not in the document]
williamdes: Done

src/Context/Context.php
```diff
@@ -12,6 +12,7 @@
  */
 class Context implements \ArrayAccess, \JsonSerializable, \IteratorAggregate
 {
+
```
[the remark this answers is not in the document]
williamdes: Done

CHANGELOG.md
```diff
@@ -3,11 +3,12 @@
 ## Unreleased
 
 - Fix `TypeError` in `Sentry\Monolog\Handler` when the extra data array has numeric keys (#833).
+- Added `setContext` to `Scope` (#839).
```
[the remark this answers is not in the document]
williamdes: I used `- Allow setting custom contexts (`setContext`) in the `Scope` (#839).`

src/Event.php
```diff
@@ -636,6 +652,15 @@ public function toArray(): array
...
+            if (isset($data['contexts'])) {
+                // Merge the user defined contexts into existing ones
```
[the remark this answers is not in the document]
williamdes: Done

williamdes:
@ste93cry I made more changes as you requested.

----
> one thing that doesn't convince me of this feature is that it will allow people to just override whatever is in the contexts key, even the "standard" ones like server OS or request that cames from the interfaces.

Seems okay for me, you control the data you send to Sentry, no ?

> Also I find really confusing that there are multiple ways to configure the same thing (the particular contexts and the contexts as a whole thing)

Same thing ?
They are not in the same place in the Sentry GUI and in the request
Did I misunderstand ?

ste93cry:
> They are not in the same place in the Sentry GUI and in the request

What I meant is that some data like the information on the OS of the server or the HTTP request is stored under the `$contexts` key, which with this feature can be completely overridden by the user without any sort of validation. This means that before the data sent to Sentry was always complete filled correctly and strictly validated (if the server did not expect a `string` you could not send a `string`) while now we are going to lose all of these stuffs because we are just going to blindly replace all data with something else given by the user, and since some of the default contexts are named with common words (`server`, `request`) etc it will be really easy to lose useful data by mistake

williamdes:
> default contexts are named with common words (server, request, etc) it will be really easy to lose useful data by mistake

I agree

ste93cry:
To clarify some more things, when I talked about strict validation I meant that the data of the `contexts` key of the event's payload are handled by specialized classes (`ServerOsContext`, `RuntimeContext`, etc) that validate the data (both types and allowed values). By just replacing its value with random data from the user we're losing such feature. Also, supposing you're not passing all required attributes for those special default contexts or you are passing more attributes than what the server expects we are gonna break the logging of events without the user be ever notified of this because we don't log the errors during the sending. In this regards, the classes were introduced to have both a nice API to manage each context and to prevent such issues by validating and restricting what the user could do on them. This is why I undoubtedly see the value of having such feature but I also see that it's going to introduce a lot of pitfalls that won't be easily solved once we merge it as is because we won't be able to rollback without breaking the behavior, be it user or technical.  If there was some restriction, for example in disallowing/ignoring the override of the default contexts, I would be totally fine. Of course, if everyone agree anyway with this PR I'm ok to merge it

williamdes:
>  If there was some restriction, for example in disallowing/ignoring the override of the default contexts, I would be totally fine. Of course, if everyone agree anyway with this PR I'm ok to merge it

Okay, I someone else requests disallowing the override, I will invert the array merge (https://github.com/getsentry/sentry-php/pull/839/files#diff-a8ad00caade65c3a6d9d157e7e4b8708R656).

williamdes:
rebased.

williamdes:
rebased another time..

@HazAT @ste93cry @Jean85 Can you merge this feature or tell me what to change please ?

2 months later ..

ste93cry:
This feature will be scheduled for `2.3` since there are some more general changes regarding how we manage contexts that may affect this feature and how it’s coded. I’m sorry if this is taking more time than you expected, in the meanwhile you can use the `beforeSend` option to edit the raw event data before it’s sent

williamdes:
@ste93cry how can I add contexts like I do in this PR using `beforeSend` ?
I tried a long time ago and did not succeed
Please help me

williamdes:
Can someone help me fix the 2 psalm errors ?

src/State/Scope.php
helhum: > we could add a check like in the TagsContext that the key is a string

I'd be in favor of doing so as well.

williamdes: @ste93cry do I need to make a change here ?

as I said "Sentry only accepts objects for contexts, any other value will make an error."

ste93cry: >do I need to make a change here ?

Yes please, let's start with `array` and let's see what happens and how the situation evolves in the future. Also please update the code to use a plain array for the `contexts` variable and add the `setContexts` method to the `Event` class.

>"Sentry only accepts objects for contexts, any other value will make an error."

You're right, however it's unlikely that someone will put a list as value of a context and it's unlikely that someone will use `string` keys containing numbers knowing that they will be casted

williamdes: Okay, made the required changes (rebase, amend fix)
See 2419b39aa95b328af9e9a8bf12ba5387bc330618

williamdes:
I have no idea why psalm complains about, I did not change this file..
```
ERROR: InvalidOperand - src/Stacktrace.php:314:36 - Cannot perform a numeric operation with a non-numeric type array-key
                $result['param' . ($index + 1)] = $this->serializeArgument($argument);
```

williamdes:
@ste93cry I checked and the error shows on `develop` branch so I will not fix it.

src/State/Scope.php
```diff
@@ -98,6 +104,31 @@ public function setTags(array $tags): self
         return $this;
     }
 
+    /**
+     * Sets context data with the given name.
+     *
+     * @param string               $name  The name that uniquely identifies the context
+     * @param array<string, mixed> $value The value
```
ste93cry: Shouldn't this be `array<string, array<string, mixed>>`?

src/State/Scope.php
```diff
@@ -27,6 +27,11 @@ final class Scope
      */
     private $user;
 
+    /**
+     * @var Context The list of contexts associated to this scope
+     */
+    private $contexts;
```
ste93cry: For simplicity let's use an `array<string, array<string, mixed>>` too here. I don't think that there is any good reason to use the `Context` object which we may remove in `3.0` given that `contexts is able to override any data of the interfaces, thus effectively bypassing the validation provided by these objects

src/Event.php
```diff
@@ -421,6 +427,31 @@ public function setRequest(array $request): void
...
+    public function getContexts(): Context
+    {
+        return $this->contexts;
+    }
+
+    /**
+     * Sets context data with the given name.
+     *
+     * @param string               $name  The name that uniquely identifies the context
+     * @param array<string, mixed> $value The value
```
ste93cry: The typehint should be `array<string, array<string, mixed>>`

src/Event.php
```diff
@@ -101,6 +101,11 @@ final class Event implements \JsonSerializable
      */
     private $userContext;
 
+    /**
+     * @var Context An arbitrary mapping of additional contexts associated to this event
+     */
+    private $contexts;
```
ste93cry: Please switch to `array<string, array<string, mixed>>`, see similar comment below for the explanation of why I'm requesting this change

tests/EventTest.php
```diff
@@ -274,6 +362,25 @@ public function testGetExtraContext(): void
         $this->assertInstanceOf(Context::class, $event->getExtraContext());
     }
 
+    public function testGetContexts(): void
```
ste93cry: I don't think that this test is useful because using typehints we are guaranteed that the returned value is of the expected type

tests/EventTest.php
```diff
@@ -119,6 +119,94 @@ public function testToArray(): void
         $this->assertEquals($expected, $event->toArray());
     }
 
+    public function testToArrayMergeUserContexts(): void
```
ste93cry: What about merging this test and the one below using a data provider?

src/State/Scope.php
williamdes: I will try another time but I did have to revert this because psalm says it looks like an object

williamdes:
@ste93cry did my work get merged?

ste93cry:
Not yet, I wanted to rebase and change some things before requesting a final review and merging but somehow I messed your branch or whatever it happened (I honestly don't know) and GitHub closed the PR because it though there were no more commits... I will open a new PR as followup of this one with all your changes plus mines merged together. I feel very sorry for how I personally handled your contribution, how long it took to get my attention and how long it's taking to get merged

williamdes:
Thank you for your nice reply, keep me updated on the progress of this PR

>  I feel very sorry for how I personally handled your contribution, how long it took to get my attention and how long it's taking to get merged

I totally understand, I have the same challenges on the phpMyAdmin project with all the contributions and issues

PR #840, opened 2019-06-26
[Monolog] Add channel to payload as logger

### LINKED ISSUE #838: [Monolog] Make use of Event::$logger
Hello, this is more likely question and maybe proposal to do so - why are we not using `Sentry\Event::$logger` and instead using it as custom extra key: `$scope->setExtra('monolog.channel', $record['channel']);`

Imho saving channel name to the logger instead would be better for the UX.

JanMikes:
Closes #838

### DIFF
--- a/src/Monolog/Handler.php
+++ b/src/Monolog/Handler.php
@@ -47,7 +47,7 @@ final class Handler extends AbstractProcessingHandler
         $payload = [
             'level' => $this->getSeverityFromLevel($record['level']),
             'message' => $record['message'],
-            'logger' => sprintf('monolog.%s', $record['channel']),
+            'logger' => 'monolog.' . $record['channel'],
         ];
 
         if (isset($record['context']['exception']) && $record['context']['exception'] instanceof \Throwable) {
--- a/src/Monolog/Handler.php
+++ b/src/Monolog/Handler.php
@@ -47,6 +47,7 @@ final class Handler extends AbstractProcessingHandler
         $payload = [
             'level' => $this->getSeverityFromLevel($record['level']),
             'message' => $record['message'],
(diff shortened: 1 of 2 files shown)


src/Monolog/Handler.php
```diff
@@ -47,6 +47,7 @@ protected function write(array $record): void
         $payload = [
             'level' => $this->getSeverityFromLevel($record['level']),
             'message' => $record['message'],
+            'logger' => sprintf('monolog.%s', $record['channel']),
```
JanMikes: I was thinking about adding just channel name, but decided to add it this way - `monolog.$channel` imho it is better, feedback is appreciated.

Jean85: This seems sensible to me :+1:

ste93cry: I fail to see a good reason to force the prefix on a user-submitted value. If he wants to distinguish between an error logged from a "plain" client or from Monolog then he can just change the `logger` property (in this case by setting the Monolog channel)

JanMikes: Since the property is named `logger` and not `channel` i though that knowing it was logged via monolog + combining with channel's name might be really helpful - at least for me it has added value.

If we take away this prefix, most people (it is the most common scenario that people uses monolog in some framework, etc Symfony, and channel name is `default`) will see only `default` vs `monolog.default`. I would be thinking about what the hell does the `default` mean, when i see `monolog.default` i know that it is monolog channel on first sight.

ste93cry: Got it, just not really happy as I was saying to force something on user's behalf. By the way let's merge it as is and let's see how it goes

PR #843, opened 2019-06-26
Added support for setting transaction name via monolog extra context

### LINKED ISSUE #842: [Monolog] Possibility to add transaction to record
**Problem:** Currently, there is no way how to set `transaction` on event sending to Sentry only via monolog - i found out that `before_send` callable can be used on the client, but i am sure many people would love to keep using monolog only with the monolog recommended way of adding extra information - processors.

My proposed way is to check if there is `$record['extra']['transaction']` if yes, set it to the payload from which the event is created.

JanMikes:
Closes #842

### DIFF
--- a/src/Monolog/Handler.php
+++ b/src/Monolog/Handler.php
@@ -53,6 +53,10 @@ final class Handler extends AbstractProcessingHandler
             $payload['exception'] = $record['context']['exception'];
         }
 
+        if (isset($record['extra']['transaction'])) {
+            $payload['transaction'] = $record['extra']['transaction'];
+        }
+
         $this->hub->withScope(function (Scope $scope) use ($record, $payload): void {
             $scope->setExtra('monolog.channel', $record['channel']);
             $scope->setExtra('monolog.level', $record['level_name']);
--- a/tests/Monolog/HandlerTest.php
+++ b/tests/Monolog/HandlerTest.php
@@ -306,5 +306,28 @@ final class HandlerTest extends TestCase
             ],
             [],
         ];
+
+        yield [
+            [
+                'message' => 'foo bar',
+                'level' => Logger::DEBUG,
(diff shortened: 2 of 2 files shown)


ste93cry:
Change looks good and clean, really good work. Just wondering if we should prefix all these `extra` keys to not clash with other libraries using the same names for something totally different. This is one of the main reasons I didn't support such feature when I coded the handler: I though that it was best to leave to the user the choice of how to implement it and from where to retrieve such information

JanMikes:
> Just wondering if we should prefix all these extra keys to not clash with other libraries using the same names for something totally different. 

I see what you mean and totally agree. My personal opinion is that it is not necessary as it is possible to register processors for specific handlers or either exclude them, so i think this might not be issue. 

> I though that it was best to leave to the user the choice of how to implement it and from where to retrieve such information

I agree with this as well, i am the kind of person who prefers sending data to sentry via monolog and configure it via monolog processors - i missed this options so i am adding them, if you are okay with that, few more PRs are following soon 😄

ste93cry:
>it is not necessary as it is possible to register processors for specific handlers or either exclude them, so i think this might not be issue.

It may be an issue as long as I want to use both processors at the same time. If that's the case, then I have to choose between reimplementing one of the two, not use any of them or use only one of them. 

>if you are okay with that, few more PRs are following soon

We would be glad to get some more contributions and feedback from you and all the community of course :smile: I'm really undecided about the prefixed keys but I would say that since those worked fine in version `1.x` let's keep using them here and hope for the best

JanMikes:
> It may be an issue as long as I want to use both processors at the same time. If that's the case, then I have to choose between reimplementing one of the two, not use any of them or use only one of them.

Yeah, sometimes i have problem with that i am trying to solve problems that might exist only in my head and other developers are not facing them 😄 let's see and wait for some feedback.

Issue #845, opened 2019-07-01
Remove static method from HubInterface


B-Galati:
Hello,

I don't see the point of having these 2 methods in HubInterface:

- `public static function getCurrent(): self`
- `public static function setCurrent(self $hub): self`

As we cannot know which implementation is the right one to use. 

IMHO, it's great that the SDK proposed a static access to the Hub but it should be opt-in and not part of the initial contract.

An alternative would be to type hint the interface with `HubInterface` instead of self, WDYT?

FYI I got the issue because I created a decorator to deal with the Spool Transport that is hard to setup.

ste93cry:
There seems to be a lot of questions here, so I will try to address all of them :wink: 

>An alternative would be to type hint the interface with HubInterface instead of self, WDYT?

That is indeed a wrong typehint that should have never been that, but we cannot change it anymore without breaking BC so it must stay as that. For the same reason, no methods can be added or removed from the `HubInterface` interface.

>it's great that the SDK proposed a static access to the Hub but it should be opt-in and not part of the initial contract

The hub holds in its state the current hub instance (there can be only one active at once) and those methods are the way to get and set it. I agree that it's not the purpose of this class to hold such state, the alternative would be to have a singleton object that holds such instance. To clarify, in the JS SDK this is represented by the `Carrier` concept while in the Python SDK this should be the `HubManager`. Like before, we cannot change the interface anyway without breaking BC so those methods can became a proxy to something else but have to stay until the next major version which is unlikely to came soon

B-Galati:
> There seems to be a lot of questions here, so I will try to address all of them wink
> 
> > An alternative would be to type hint the interface with HubInterface instead of self, WDYT?
> 
> That is indeed a wrong typehint that should have never been that, but we cannot change it anymore without breaking BC so it must stay as that. For the same reason, no methods can be added or removed from the `HubInterface` interface.

It would not be a dramatic BC break to change self by `HubInterface` as it stays relatively compatible.

Thank you for your answers :+1:

Jean85:
I would like to agree with B-Galati, changing to interface widens the return, so it shouldn't be a BC, but unfortunately it is, PHP doesn't like widen interfaces: https://3v4l.org/R78b3

Proper contravariance will be present in PHP 7.4: https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters

ste93cry:
[...]

PR #847, opened 2019-07-09
Deprecate the methods on the HubInterface to get and set the current hub

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.2 (develop)
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| License       | MIT

As discussed in #845, putting the `getCurrentHub` and `setCurrentHub` on the `HubInterface` was an error. ~Based on what has been done in other SDKs (I used Python and JS as reference) I decided to deprecate them and move the handling of the current hub in a `HubManager` class that can be used as a service in a proper DI-based environment or through the global functions for those that prefers to use them. PR is not ready yet, but I would like an opinion on a few implementation details. In particular, I'm undecided on how the global function `getHubManager` should be written. The main thing I would like to support is the retrieval of the instance from a service container, so I had to pass a callback around that will be used as a sort of factory method to instantiate the singleton variable the first time it's accessed. Another thing I had to solve is that since we (especially in an application based on a framework) cannot be sure that the function won't be accessed before the service container is built and initialized I needed a way to force replacing the instance even if it was already set. I don't really like this because of all the bad things that could occur when replacing an instance when you shouldn't, but I couldn't came up with a better solution.~ Any idea, comment or feedback is greatly appreciated.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
     },
     "autoload": {
         "files": [
-            "src/Sdk.php"
+            "src/functions.php"
         ],
         "psr-4" : {
             "Sentry\\" : "src/"
@@ -59,11 +59,8 @@
         "tests": [
             "vendor/bin/phpunit --verbose"
         ],
-        "tests-report": [
-            "vendor/bin/phpunit --verbose --configuration phpunit.xml.dist --coverage-html tests/html-report"
-        ],
         "phpcs": [
-            "vendor/bin/php-cs-fixer fix --config=.php_cs --verbose --diff --dry-run"
+            "vendor/bin/php-cs-fixer fix --verbose --diff --dry-run"
         ],
         "phpstan": [
             "vendor/bin/phpstan analyse"
--- a/src/Integration/ErrorListenerIntegration.php
+++ b/src/Integration/ErrorListenerIntegration.php
@@ -8,7 +8,7 @@ use Sentry\ErrorHandler;
 use Sentry\Exception\FatalErrorException;
 use Sentry\Exception\SilencedErrorException;
 use Sentry\Options;
-use Sentry\State\Hub;
+use Sentry\SentrySdk;
 
 /**
  * This integration hooks into the global error handlers and emits events to
@@ -58,9 +58,8 @@ final class ErrorListenerIntegration implements IntegrationInterface
                 return;
             }
 
-            $currentHub = Hub::getCurrent();
-            $integration = $currentHub->getIntegration(self::class);
-            $client = $currentHub->getClient();
+            $integration = SentrySdk::getIntegration(self::class);
+            $client = SentrySdk::getClient();
 
             // The client bound to the current hub, if any, could not have this
             // integration enabled. If this is the case, bail out
@@ -78,7 +77,7 @@ final class ErrorListenerIntegration implements IntegrationInterface
                 return;
             }
 
-            $currentHub->captureException($exception);
+            SentrySdk::captureException($exception);
         });
     }
 }
--- a/src/Integration/ErrorListenerIntegration.php
+++ b/src/Integration/ErrorListenerIntegration.php
@@ -58,8 +58,9 @@ final class ErrorListenerIntegration implements IntegrationInterface
                 return;
             }
 
-            $integration = SentrySdk::getIntegration(self::class);
-            $client = SentrySdk::getClient();
+            $currentHub = SentrySdk::getCurrentHub();
+            $integration = $currentHub->getIntegration(self::class);
+            $client = $currentHub->getClient();
 
             // The client bound to the current hub, if any, could not have this
             // integration enabled. If this is the case, bail out
@@ -77,7 +78,7 @@ final class ErrorListenerIntegration implements IntegrationInterface
                 return;
             }
 
-            SentrySdk::captureException($exception);
+            $currentHub->captureException($exception);
         });
     }
 }
--- a/src/Integration/ExceptionListenerIntegration.php
+++ b/src/Integration/ExceptionListenerIntegration.php
@@ -5,7 +5,7 @@ declare(strict_types=1);
 namespace Sentry\Integration;
 
 use Sentry\ErrorHandler;
-use Sentry\State\Hub;
+use Sentry\SentrySdk;
 
 /**
  * This integration hooks into the global error handlers and emits events to
@@ -21,8 +21,7 @@ final class ExceptionListenerIntegration implements IntegrationInterface
         /** @psalm-suppress DeprecatedMethod */
         $errorHandler = ErrorHandler::registerOnce(ErrorHandler::DEFAULT_RESERVED_MEMORY_SIZE, false);
         $errorHandler->addExceptionHandlerListener(static function (\Throwable $exception): void {
-            $currentHub = Hub::getCurrent();
-            $integration = $currentHub->getIntegration(self::class);
+            $integration = SentrySdk::getIntegration(self::class);
 
             // The client bound to the current hub, if any, could not have this
             // integration enabled. If this is the case, bail out
@@ -30,7 +29,7 @@ final class ExceptionListenerIntegration implements IntegrationInterface
                 return;
             }
 
-            $currentHub->captureException($exception);
+            SentrySdk::captureException($exception);
         });
     }
 }
--- a/src/Integration/ExceptionListenerIntegration.php
+++ b/src/Integration/ExceptionListenerIntegration.php
@@ -21,7 +21,8 @@ final class ExceptionListenerIntegration implements IntegrationInterface
         /** @psalm-suppress DeprecatedMethod */
         $errorHandler = ErrorHandler::registerOnce(ErrorHandler::DEFAULT_RESERVED_MEMORY_SIZE, false);
         $errorHandler->addExceptionHandlerListener(static function (\Throwable $exception): void {
-            $integration = SentrySdk::getIntegration(self::class);
+            $currentHub = SentrySdk::getCurrentHub();
+            $integration = $currentHub->getIntegration(self::class);
 
             // The client bound to the current hub, if any, could not have this
             // integration enabled. If this is the case, bail out
@@ -29,7 +30,7 @@ final class ExceptionListenerIntegration implements IntegrationInterface
                 return;
             }
 
-            SentrySdk::captureException($exception);
+            $currentHub->captureException($exception);
         });
     }
 }
--- a/src/Integration/FatalErrorListenerIntegration.php
+++ b/src/Integration/FatalErrorListenerIntegration.php
@@ -42,8 +42,9 @@ final class FatalErrorListenerIntegration implements IntegrationInterface
     {
         $errorHandler = ErrorHandler::registerOnceFatalErrorHandler();
         $errorHandler->addFatalErrorHandlerListener(function (FatalErrorException $exception): void {
-            $integration = SentrySdk::getIntegration(self::class);
-            $client = SentrySdk::getClient();
+            $currentHub = SentrySdk::getCurrentHub();
+            $integration = $currentHub->getIntegration(self::class);
+            $client = $currentHub->getClient();
 
             // The client bound to the current hub, if any, could not have this
             // integration enabled. If this is the case, bail out
@@ -57,7 +58,7 @@ final class FatalErrorListenerIntegration implements IntegrationInterface
                 return;
             }
 
-            SentrySdk::captureException($exception);
+            $currentHub->captureException($exception);
         });
     }
 }
--- a/src/Integration/FatalErrorListenerIntegration.php
+++ b/src/Integration/FatalErrorListenerIntegration.php
@@ -7,7 +7,7 @@ namespace Sentry\Integration;
 use Sentry\ErrorHandler;
 use Sentry\Exception\FatalErrorException;
 use Sentry\Options;
-use Sentry\State\Hub;
+use Sentry\SentrySdk;
 
 /**
  * This integration hooks into the error handler and captures fatal errors.
@@ -42,9 +42,8 @@ final class FatalErrorListenerIntegration implements IntegrationInterface
     {
         $errorHandler = ErrorHandler::registerOnceFatalErrorHandler();
         $errorHandler->addFatalErrorHandlerListener(function (FatalErrorException $exception): void {
-            $currentHub = Hub::getCurrent();
-            $integration = $currentHub->getIntegration(self::class);
-            $client = $currentHub->getClient();
+            $integration = SentrySdk::getIntegration(self::class);
+            $client = SentrySdk::getClient();
 
             // The client bound to the current hub, if any, could not have this
             // integration enabled. If this is the case, bail out
@@ -58,7 +57,7 @@ final class FatalErrorListenerIntegration implements IntegrationInterface
                 return;
             }
 
-            $currentHub->captureException($exception);
+            SentrySdk::captureException($exception);
         });
     }
 }
--- a/src/Integration/ModulesIntegration.php
+++ b/src/Integration/ModulesIntegration.php
@@ -7,7 +7,7 @@ namespace Sentry\Integration;
 use Jean85\PrettyVersions;
 use PackageVersions\Versions;
 use Sentry\Event;
-use Sentry\State\Hub;
+use Sentry\SentrySdk;
 use Sentry\State\Scope;
 
 /**
@@ -27,7 +27,7 @@ final class ModulesIntegration implements IntegrationInterface
     public function setupOnce(): void
     {
         Scope::addGlobalEventProcessor(function (Event $event) {
-            $integration = Hub::getCurrent()->getIntegration(self::class);
+            $integration = SentrySdk::getIntegration(self::class);
 
             // The integration could be bound to a client that is not the one
             // attached to the current hub. If this is the case, bail out
--- a/src/Integration/ModulesIntegration.php
+++ b/src/Integration/ModulesIntegration.php
@@ -27,7 +27,7 @@ final class ModulesIntegration implements IntegrationInterface
     public function setupOnce(): void
     {
         Scope::addGlobalEventProcessor(function (Event $event) {
-            $integration = SentrySdk::getIntegration(self::class);
+            $integration = SentrySdk::getCurrentHub()->getIntegration(self::class);
 
             // The integration could be bound to a client that is not the one
             // attached to the current hub. If this is the case, bail out
--- a/src/SentrySdk.php
+++ b/src/SentrySdk.php
@@ -1,0 +1,312 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry;
+
+use Sentry\Integration\IntegrationInterface;
+use Sentry\State\Hub;
+use Sentry\State\HubInterface;
+use Sentry\State\Scope;
+
+/**
+ * This class is the main entry point for all the most common SDK features.
+ *
+ * @author Stefano Arlandini <sarlandini@alice.it>
+ */
+final class SentrySdk
+{
+    /**
+     * @var HubInterface|null The current hub
+     */
+    private static $currentHub;
+
+    /**
+     * Constructor.
+     */
+    private function __construct()
+    {
+    }
+
+    /**
+     * Initializes the SDK by creating a new hub instance each time this method
+     * gets called.
+     *
+     * @return HubInterface
+     */
+    public static function init(): HubInterface
+    {
+        self::$currentHub = new Hub();
+
+        return self::$currentHub;
+    }
+
+    /**
+     * Gets the current hub. If it's not initialized then creates a new instance
+     * and sets it as current hub.
+     *
+     * @param bool $throwDeprecation Whether to throw a deprecation message when
+     *                               using this method. Internal only!
+     *
+     * @return HubInterface
+     *
+     * @deprecated since version 2.2, to be removed in 3.0
+     *
+     * @internal
+     */
+    public static function getCurrentHub(bool $throwDeprecation = true): HubInterface
+    {
+        if ($throwDeprecation) {
+            @trigger_error(sprintf('The %s() method is deprecated since version 2.2 and will be removed in 3.0.', __METHOD__), E_USER_DEPRECATED);
+        }
+
+        if (null === self::$currentHub) {
+            self::$currentHub = new Hub();
+        }
+
+        return self::$currentHub;
+    }
+
+    /**
+     * Sets the current hub.
+     *
+     * @param HubInterface $hub              The hub to set
+     * @param bool         $throwDeprecation Whether to throw a deprecation message when
+     *                                       using this method. Internal only!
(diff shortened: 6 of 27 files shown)


src/HubManager.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    private static $currentHub;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getCurrentHub(): HubInterface
+    {
+        if (null === self::$currentHub) {
+            self::$currentHub = new Hub();
```
B-Galati: Why returning a new Hub if it has not been defined yet?

One solution I see to avoid the pitfall of global state like this is to allow calling `setCurrentHub` only one time. Following calls would throw an exception.

`getCurrentHub` could then: throw an exception if no Hub is defined or return null. I would prefer throwing an exception but it's not a big deal.

src/HubManagerInterface.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry;
+
+use Sentry\State\HubInterface;
+
+interface HubManagerInterface
```
B-Galati: Out of curiosity, the hub manager is here to act as a Singleton right? So why defining an interface for it?

His role is to allow access to the global instance of the HubInterface and that's all no?

src/HubManager.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /**
+     * @var HubInterface|null The current hub
+     */
+    private static $currentHub;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getCurrentHub(): HubInterface
```
B-Galati: If it acts as a singleton I guess the class can be completely static.

So only static method and mark the constructor as private.

src/HubManagerInterface.php
ste93cry: Actually it's a singleton because it's a leftover of some tries I did to find a solution that would satisfy the requirements I would like to meet for this PR to be ready for review and that I wrote in the first comment. In particular I would like to be able to support getting the hub manager from a service container, so it should not be a singleton and this is the reason I "migrated" to hold the instance in the global function `getHubManager`. However, this solution has drawbacks too:

- Once the variable is initialized you have to have a way to replace it in case someone called the function too early and you don't have control on this (this is the purpose of the `$replace` parameter)
- Having a singleton (this doesn't really depends on where the singleton is placed, but of course it's even worse if it's in a global function) causes problems in the unit tests where the state of the "application" is shared between them
- Singletons are generally speaking a bad idea in all cases, from testing to usage

Unfortunately supporting API based on global functions means that we need the singleton, no way around this. Another solution would probably be having a factory that holds the singleton so that neither `getHubManager` nor the hub manager itself holds the instance, but I didn't investigate if this idea can work and/or makes sense. WDYT?

src/HubManager.php
ste93cry: >Why returning a new Hub if it has not been defined yet?

We cannot throw an exception or return `null` because the behavior must be the same as calling `Hub::getCurrentHub()` with the current code. Also, creating the instance here means that the calling code does not need to handle creating the hub each time it needs it and this happens a lot of time (e.g. when calling the global functions). Finally, creating the instance outside this class means that the implementation would be hardcoded in places where the user cannot change it, instead here he can just (ideally) switch the hub manager implementation and do whatever he likes

>One solution I see to avoid the pitfall of global state like this is to allow calling `setCurrentHub` only one time

I don't have any valid reason about why you should not be able to call it more than once. Probably it makes more sense in other SDKs where the programming language needs more than one hub (e.g. one per thread) but the current API allows this and not permitting it anymore would be a BC :sob:

src/HubManagerInterface.php
B-Galati: IMHO, only one and one instance could exist at a time and this instance should be created by calling `Sentry\init()` explicitly.

Trying to magically instanciate the Hub is pointless as the SDK cannot guess the whole configuration wanted by the user. Trying to get some of them from the environment variable is a bad idea as it is a global state that you don't want to deal with in the library. 

Static method are ok to be used in service container so that:

- It stays ok to be used standalone without FW / service container
- It can ensure only once instance exist at a time
- It acts as the only way to configure / get a Hub

ste93cry: I've changed the design to be more similar to what the C# SDK does, which basically is having a class that acts as main entry point and that holds the instance of the hub, and if someone wants to use the DI pattern then he can register in the container the `HubAdapter` which acts as a sort of proxy between the user and the class `SentrySdk` so that you can typehint against the `HubInterface` instead of using directly that singleton. I've also deprecated the global functions, I'm not sure this is something we want to do but it would simplify the whole code a bit given that they are now just duplicates of the methods of the `SentrySdk` class

ste93cry:
After looking at the implementation of several other SDKs I decided to follow the way of the .NET SDK and implement a singleton class with static methods only that will be the main entry point for all the SDK common features like capturing messages, errors, etc. To support getting the hub instance from a service container the users will need to register the instance of the `HubAdapter` class and then they will be able to typehint the `HubInterface` to avoid relying on the static class. I decided to deprecate the global functions even though we may still provide them, however they are just duplicate code of what the new `SentrySdk` class already provides. The integrations deserves a separate discussion on how to get the current hub instance in a flexible way: the .NET SDK differs from the standard `setupOnce` method used in all the other implementations and calls a `Register` method to which both the hub instance and the options are passed as arguments. At the moment I'm getting the hub from the constructor but I'm not sure this is the best and recommended way as it implies that if the current hub gets changed hunder the hood then magically the extension will use the new one and this is something we may not want. @HazAT I would like your opinion on this PR as it's a pretty big change and you have more in-depth knowledge of the unified API and how it should work

composer.json
```diff
@@ -58,7 +60,7 @@
             "vendor/bin/phpunit --verbose --configuration phpunit.xml.dist --coverage-html tests/html-report"
         ],
         "phpcs": [
-            "vendor/bin/php-cs-fixer fix --config=.php_cs --verbose --diff --dry-run"
+            "vendor/bin/php-cs-fixer fix --config=.php_cs.dist --verbose --diff --dry-run"
```
Jean85: Maybe we could omit the `--config` altogether? IIRC it chooses the right one by default.

src/Integration/ErrorListenerIntegration.php
```diff
@@ -61,7 +69,7 @@ public function setupOnce(): void
                 return;
             }
 
-            Hub::getCurrent()->captureException($exception);
+            $this->hub->captureException($exception);
```
Jean85: This changes the behavior: the hub is now bound at construction and, if the main is changed afterwards, this integration will not acknowledge that. Is this intended?

src/Integration/ExceptionListenerIntegration.php
```diff
@@ -5,22 +5,38 @@
...
     /**
      * {@inheritdoc}
      */
     public function setupOnce(): void
     {
         $errorHandler = ErrorHandler::registerOnce(ErrorHandler::DEFAULT_RESERVED_MEMORY_SIZE, false);
-        $errorHandler->addExceptionHandlerListener(static function (\Throwable $exception): void {
-            Hub::getCurrent()->captureException($exception);
+        $errorHandler->addExceptionHandlerListener(function (\Throwable $exception): void {
+            $this->hub->captureException($exception);
```
Jean85: Here too, same behavioral change.

src/Integration/FatalErrorListenerIntegration.php
```diff
@@ -42,7 +50,7 @@ public function setupOnce(): void
                 return;
             }
 
-            Hub::getCurrent()->captureException($exception);
+            $this->hub->captureException($exception);
```
Jean85: Here too.

src/Sdk.php
```diff
@@ -1,100 +0,0 @@
...
-namespace Sentry;
-
-use Sentry\State\Hub;
-
-/**
- * Creates a new Client and Hub which will be set as current.
- *
- * @param array $options The client options
- */
-function init(array $options = []): void
```
Jean85: This removal is a huge BC.

src/Integration/ExceptionListenerIntegration.php
HazAT: This is a valid point and I just realized something else.
Apparently, no one reported this so far but:

We need to check if the integration is active on the client like we do in the ModulesIntegration for example.

https://github.com/getsentry/sentry-php/blob/62a48faf3fc4ae5b0564a0d87f7a91c47ea61ad9/src/Integration/RequestIntegration.php#L68-L72

The reason being is that if someone binds another client to the current hub, where this integration is not enabled we shouldn't report errors.

src/Integration/ModulesIntegration.php
```diff
@@ -21,13 +22,28 @@ final class ModulesIntegration implements IntegrationInterface
...
+    }
+
     /**
      * {@inheritdoc}
      */
     public function setupOnce(): void
     {
         Scope::addGlobalEventProcessor(function (Event $event) {
-            $self = Hub::getCurrent()->getIntegration(self::class);
+            $self = $this->hub->getIntegration(self::class);
```
HazAT: Checking if the integration is active on the current client, we need to ask for the current hub on the time when this is called. It's not valid to pass it into the Ctor and rely on this instance since this Hub can already be gone.

src/Integration/ExceptionListenerIntegration.php
HazAT: So to clarify what we need to do:
```
    $hub = $this->hub  ?? HubAdapter::getInstance();
    $self = $hub->getIntegration(self::class); 
  
    if (!$self instanceof self) { 
        return $event; 
    } 

    $hub->captureException($exception);
```

src/Sdk.php
ste93cry: I fu**ed up the rebase... Thanks for spotting this!

ste93cry: I fu**ed up the rebase, this file has only been renamed and nothing got deleted

src/Integration/ErrorListenerIntegration.php
ste93cry: >Is this intended?

No, this is an issue I didn't account for when coding. I will revert the change

src/Integration/ExceptionListenerIntegration.php
ste93cry: I reverted the change, however I will fix the bug in a separate PR targeting `2.1` so that we can release it earlier than the version this PR is targeting

ste93cry:
>I don't fully understand the motivation of this change, it was fine before 

The change originated because as reported by @B-Galati the singleton should be managed somewhere else: it's not up to the implementors of the `HubInterface` to handle it. This in turn was a problem because I would had to save the instance in a global variable or in a global function as local static variable, but in this case then it was not possible to set it without making a getter that could act as a setter too according to the number of parameters that it was invoked with. So, I took inspiration from the .NET SDK and just adapted what was done there deprecating the global functions and moving everything into a static class that acts as central point for the SDK. The `HubAdapter` is just a proxy to avoid depending on such class. The rest is just a refactoring to make everything work

>It changes a lot of internal behaviour which is now inconsitent at times.

Indeed I didn't account for some things, I reverted the changes and I will address the found bugs not strictly related to this change in a separate PR so that we can ship the fix as soon as possible

Jean85:
#861 seems a simpler approach to this issue...

ste93cry:
Nope, there is a lot of confusion here. The PR you referenced is fixing a bug with the integrations and the fact that they do not check if they should run according to the options of the client bound to the current hub while this PR solves an architectural problem and tries to move away from a wrong implementation that is breaking SRP and a lot of other things

src/Integration/ExceptionListenerIntegration.php
```diff
@@ -19,8 +19,8 @@ final class ExceptionListenerIntegration implements IntegrationInterface
     public function setupOnce(): void
     {
         $errorHandler = ErrorHandler::registerOnce(ErrorHandler::DEFAULT_RESERVED_MEMORY_SIZE, false);
-        $errorHandler->addExceptionHandlerListener(static function (\Throwable $exception): void {
-            Hub::getCurrent()->captureException($exception);
+        $errorHandler->addExceptionHandlerListener(function (\Throwable $exception): void {
```
Taluu: you can leave the static here

src/Integration/ModulesIntegration.php
```diff
@@ -27,7 +27,7 @@ final class ModulesIntegration implements IntegrationInterface
     public function setupOnce(): void
     {
         Scope::addGlobalEventProcessor(function (Event $event) {
-            $self = Hub::getCurrent()->getIntegration(self::class);
+            $self = SentrySdk::getIntegration(self::class);
 
             if ($self instanceof self) {
                 self::applyToEvent($self, $event);
```
Taluu: I know it's not really the scope here, but why bother with getting the `$self` from the sdk + checking, and not use `$this`... directly ? same in the `applyToEvent` method, which would transform itself from static to non-static

(Same for the `RequestIntegration`, I was about to make a PR for that...)

Taluu: With this, we can then remove the hard dependency on the sdk, and thus for monolog integration etc, we can be free of the hub / sdk / whatever and just apply whatever integration we want.

src/SentrySdk.php
```diff
@@ -0,0 +1,287 @@
...
+     * @var HubInterface|null The current hub
+     */
+    private static $currentHub;
+
+    /**
+     * Constructor.
+     */
+    private function __construct()
+    {
+    }
```
Taluu: do the same for the clone

src/SentrySdk.php
```diff
@@ -0,0 +1,287 @@
...
+        return self::$currentHub;
+    }
+
+    /**
+     * Gets the current hub. If it's not initialized then creates a new instance
+     * and sets it as current hub.
+     *
+     * @return HubInterface
+     *
+     * @deprecated
```
Taluu: don't really see the point of adding yet a new method that will be deprecated anyways...

src/SentrySdk.php
```diff
@@ -0,0 +1,287 @@
...
+     *
+     * @internal
+     */
+    public static function setCurrentHub(HubInterface $hub): HubInterface
+    {
+        @trigger_error(sprintf('The %s() method is deprecated since version 2.2 and will be removed in 3.0.', __METHOD__), E_USER_DEPRECATED);
```
Taluu: same thing here

src/Integration/ModulesIntegration.php
ste93cry: Explanation about why this is needed has been given in #868

src/SentrySdk.php
ste93cry: The reason is that this method is needed to make the (also) deprecated method `Hub::getCurrentHub` work

src/SentrySdk.php
ste93cry: The reason is that this method is needed to make the (also) deprecated method `Hub::setCurrentHub` work

src/State/HubAdapter.php
```diff
@@ -0,0 +1,170 @@
...
+
+        return SentrySdk::getCurrentHub();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function setCurrent(HubInterface $hub): HubInterface
+    {
+        @trigger_error(sprintf('The %s() method is deprecated since version 2.2 and will be removed in 3.0', __METHOD__), E_USER_DEPRECATED);
```
Jean85: Wouldn't this generate a double deprecation?

ste93cry: Yes it does, fixed!

B-Galati:
@HazAT In the Java implementation there is a Sentry class that is the entrypoint of the SDK -> https://github.com/getsentry/sentry-java/blob/master/sentry/src/main/java/io/sentry/Sentry.java.
This class manage the global state and is static.

And then there is a class called SentryClient https://github.com/getsentry/sentry-java/blob/master/sentry/src/main/java/io/sentry/SentryClient.java that is non static and is responsible to talk with Sentry Server.

It is as simple as that.

Taluu:
Having every integrations (at least, didn't bother to check further) aware of the hub is also a huge PITA, especially if you try to integrate it with a dependency injection container, rendering it useless in case of you need multiple hubs (that's the joke).

Each integrations should be built with their own hub, their own client and that's pretty much it. No need to access a global state to fetch those...

ste93cry:
I agree, having a class rather than a bunch of global functions provides a bit more of encapsulation and gives us a bit more flexibility in how we want to store the global state. If the problem is deprecating the global functions we can leave them there, although they are pretty much useless since they would just call the static class under the hood. About the hub, having `getCurrentHub` and `setCurrentHub` on the interface makes totally no sense as said by @B-Galati since the implementors should not have to worry about where to save the instance of the current hub.

>I also highly doubt one would ever create their own Hub if they do, they should be familiar with the internals and why we need a `current` hub.

First of all, since there is an interface then we must suppose that someone could implement it and we cannot hide behind the "I highly doubt someone would even create..." discussion. Second, having the interface in this case is a good thing because in an project using dependency injection you want to avoid at all costs to inject statics and singletons, which is why the guys that created the C# SDK and in this PR too there is an implementation of the interface that serves as proxy to the current hub and that can be injected to classes willing to call the hub methods. Finally, I always see that other SDKs differs from the official specs (Java SDK, C# SDK) so I don't really see any reason about why we cannot have the same freedom here. To me it seems that we are improving the design of the Unified API concept and probably the right question would be if the other SDKs should adapt to us in some cases like this one and not viceversa 😉

src/SentrySdk.php
```diff
@@ -0,0 +1,312 @@
...
+
+        return self::$currentHub->getIntegration($className);
+    }
+
+    /**
+     * @see https://www.php.net/manual/en/language.oop5.cloning.php#object.clone
+     */
+    public function __clone()
+    {
+        throw new \BadMethodCallException('Cloning is forbidden.');
```
stayallive: If you are forbidding something it would be nice to have a tiny explanation why (you know, I know but someone reading this maybe doesn't). There are some other places too.

ste93cry:
>people that have just ^2.0 in their composer.json will be bombarded with ErrorExceptions from their error tracking library.

This is part of how deprecation works, even in PHP... If you use something that is deprecated then unless you suppress the errors you will get warnings and this is good because you may not even notice a docblock that marks a function as deprecated. Even frameworks and other libraries out of here throws such warnings and we did the same in the past. This PR should not throw any deprecation as far as I coded it right , but if someone used deprecated functions then he will get them.

ste93cry:
@HazAT are you more willing to approve this PR if I don't deprecate the global functions? Even though I strongly disagree because we have duplicate ways to do the same things in this case users would still be able to use the same API as now as the change would be transparent for them, apart from `Hub::getCurrent` that would throw a deprecation (and there is no way around this as it's the subject of this PR). Otherwise I'm out as I don't have the desire to further discuss why this change is better than our current situation as it seems that the people to involve are not willing to change their mind

src/SentrySdk.php
ste93cry: I don't think that it adds value to specify that cloning a singleton is not allowed since ensuring that just one instance of the class exists during the execution of the program is the foundation of the concept of a singleton...

src/State/HubInterface.php
```diff
@@ -123,17 +123,21 @@ public function addBreadcrumb(Breadcrumb $breadcrumb): bool;
      * Returns the current global Hub.
      *
      * @return HubInterface
+     *
+     * @deprecated since version 2.2, to be removed in 3.0
```
HazAT: Can we do `@see SentrySdk`?

src/State/HubInterface.php
```diff
@@ -123,17 +123,21 @@ public function addBreadcrumb(Breadcrumb $breadcrumb): bool;
...
+    public static function getCurrent(): self;
 
     /**
      * Sets the Hub as the current.
      *
      * @param HubInterface $hub The Hub that will become the current one
      *
      * @return HubInterface
+     *
+     * @deprecated since version 2.2, to be removed in 3.0
```
HazAT: Can we do `@see SentrySdk`?

src/SentrySdk.php
ste93cry: I dropped the `__clone` and `__unserialize` methods as this class only have `static` methods and the constructor is `private`, thus there is no way to get an instance of this class (you could use reflection, but there is no reason to use it)

src/Integration/ErrorListenerIntegration.php
```diff
@@ -58,7 +58,7 @@ public function setupOnce(): void
                 return;
             }
 
-            $currentHub = Hub::getCurrent();
+            $currentHub = SentrySdk::getCurrentHub();
```
Taluu: this is still binding *any* integration to the "current" hub.

What should be done would be to inject the hub as a constructor argument (if null, then fetch the current hub from the sentry sdk as bc).

ste93cry: We could inject the `HubAdapter` in the constructor (it has been coded for cases as this one), however if I don't recall wrong binding an integration to a specific hub does not play well with the current architecture, or at least it doesn't work well with all the integrations provided out of the box. I'm sure @HazAT can give a better explanation of why it isn't a so easy task to make this change, so maybe it's better to discuss about this in a separate issue. The purpose of this PR is mainly to move away from the `Hub` class the storage of the instance of the current hub

src/State/HubAdapter.php
```diff
@@ -0,0 +1,186 @@
...
+use Sentry\Breadcrumb;
+use Sentry\ClientInterface;
+use Sentry\Integration\IntegrationInterface;
+use Sentry\SentrySdk;
+use Sentry\Severity;
+
+/**
+ * An implementation of {@see HubInterface} which forwards any call to {@see SentrySdk}.
+ * This allows testing classes which otherwise would need to depend on it by
+ * having them depend on the interface instead, which can be mocked.
```
HazAT: I would change the description to something like:

```
An implementation of {@see HubInterface} uses {@see SentrySdk::getCurrentHub()}
internally to manage active instance of the Hub.
```

Jean85:
@HazAT the update on `sentry-symfony` is doable, it would require a bump in the minimum version of the SDK. That would be useful for me because I was already thinking about doing a bump due to https://github.com/getsentry/sentry-symfony/issues/236.

Jean85:
@HazAT another question: the two integrations will need a bump but they rely on `sentry-sdk`... Should we tag the bump there or should we add the direct constraint in the integrations?

ste93cry:
You could also just use `class_exists` to decide whether to use the `Hub` or `SentrySdk` class to get and set the current hub so that there is no need to bump the minimum required version of the main SDK. Or am I missing something?

Jean85:
@ste93cry yes you could. @HazAT it's up to you then, you should decide which approach we should choose.

HazAT:
I am for bumping/code change if possible.

PR #849, opened 2019-07-10
 fix: parameter and return type of HubInterface

### LINKED ISSUE #845: Remove static method from HubInterface

B-Galati:
as per https://3v4l.org/ljZp6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Fix `TypeError` in `Sentry\Monolog\Handler` when the extra data array has numeric keys (#833).
+- Changed type hint for both parameter and return value of `HubInterface::getCurrentHub` and `HubInterface::setCurrentHub()` methods (#849)
 
 ## 2.1.1 (2019-06-13)
 
--- a/src/State/HubInterface.php
+++ b/src/State/HubInterface.php
@@ -122,18 +122,18 @@ interface HubInterface
     /**
      * Returns the current global Hub.
      *
-     * @return self
+     * @return HubInterface
      */
-    public static function getCurrent(): self;
+    public static function getCurrent(): HubInterface;
 
     /**
      * Sets the Hub as the current.
      *
-     * @param self $hub The Hub that will become the current one
+     * @param HubInterface $hub The Hub that will become the current one
      *
-     * @return self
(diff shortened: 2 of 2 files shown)


ste93cry:
I don't know why I convinced myself that this could not be done without breaking BC, but apparently it works because there was just one way to get it working right now and that way is the same as the change :rocket:

ste93cry:
There is one missing thing: can you please update the CHANGELOG?

Jean85:
@ste93cry that's probably because all the implementation that would be broken are possible only with PHP 7.4 where proper type widening (contravariance) is implemented: https://3v4l.org/EWG2S

So if we release this before PHP 7.4 we're good :smile:

B-Galati:
> @ste93cry that's probably because all the implementation that would be broken are possible only with PHP 7.4 where proper type widening (contravariance) is implemented: https://3v4l.org/EWG2S

It would be ok for the return type to use `self` as it is covariant but it's not for the parameter type to have `self` as it must be contravariant as per the RFC -> https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters.

B-Galati:
Changelog updated!

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Fix `TypeError` in `Sentry\Monolog\Handler` when the extra data array has numeric keys (#833).
+- Fix parameter and return type of HubInterface (#849)
```
ste93cry: I would like to be more specific in regards to what has been done here since it's a really small change, so a phrase like ``Changed the typehints for both the parameter and the return value of the `HubInterface::getCurrentHub` and `HubInterface::setCurrentHub()` methods`` makes more sense.

B-Galati: Done! Thx!

PR #852, opened 2019-07-18
Add functions to clear breadcrumbs and bulk set tags and extras in the scope

### LINKED ISSUE #797: QoL Changes
- Add more scope functions:
`setTags`, `setExtras`, `clearBreadcrumbs`
https://github.com/getsentry/sentry-javascript/blob/6852dde3467218583dd320c8ad95d1844ffa6ebe/packages/types/src/scope.ts#L25
https://github.com/getsentry/sentry-javascript/blob/6852dde3467218583dd320c8ad95d1844ffa6ebe/packages/types/src/scope.ts#L38
https://github.com/getsentry/sentry-javascript/blob/6852dde3467218583dd320c8ad95d1844ffa6ebe/packages/types/src/scope.ts#L72

- Add setter functions to breadcrumb to make it easier to create a breadcrumb step by step
ref: https://forum.sentry.io/t/php-sdk-2-0-breadcrumb-creation-improvements/6350

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.2 (develop)
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

This closes #797 by implementing the `setTags`, `setExtras` and `clearBreadcrumbs` methods on the `Scope` class. I also refactored the tests to stop relying on reflection to access the active scope of an hub. I wonder if it makes sense to deprecate all the getters on the scope as the other SDKs do not implement them

### DIFF
--- a/tests/State/HubTest.php
+++ b/tests/State/HubTest.php
@@ -7,25 +7,19 @@ namespace Sentry\Tests\State;
 use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
 use Sentry\Breadcrumb;
-use Sentry\ClientBuilder;
 use Sentry\ClientInterface;
+use Sentry\Event;
+use Sentry\Integration\IntegrationInterface;
+use Sentry\Options;
 use Sentry\Severity;
 use Sentry\State\Hub;
-use Sentry\State\HubInterface;
 use Sentry\State\Scope;
 
 final class HubTest extends TestCase
 {
-    public function testConstructorCreatesScopeAutomatically(): void
-    {
-        $hub = new Hub(null, null);
-
-        $this->assertNotNull($this->getScope($hub));
-    }
-
     public function testGetClient(): void
     {
-        /** @var ClientInterface|MockObject $client */
+        /** @var ClientInterface&MockObject $client */
         $client = $this->createMock(ClientInterface::class);
         $hub = new Hub($client);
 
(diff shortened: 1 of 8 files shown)


tests/State/HubTest.php
```diff
@@ -7,25 +7,19 @@
...
-    {
-        $hub = new Hub(null, null);
-
-        $this->assertNotNull($this->getScope($hub));
-    }
-
     public function testGetClient(): void
     {
-        /** @var ClientInterface|MockObject $client */
+        /** @var ClientInterface&MockObject $client */
```
Jean85: Isn't this PHPDoc superfluous with the `phpstan-phpunit` extension?

ste93cry: Yes it is, however it helps the autocompletion of the IDE that otherwise would not recognize that the type of the variable is a `ClientInterface` and not only a `MockObjec`. I think that this annotation is still wrong as it should probably be `MockObject<ClientInterface>`, but then it's useless again as PHPStorm (in my case) doesn't recognize the syntax for objects that are not `iterable`

Jean85: If that's the case, you should use the PHPUnit plugin for PHPStorm and drop the annotation: https://github.com/Haehnchen/idea-php-phpunit-plugin

(it's a fork of the previous one which was broken)

ste93cry:
@HazAT should we deprecate the getters on the `Scope` class and discourage their usage?

HazAT:
@ste93cry Yes, we could add `@deprecated` before we make them `protected` or something.

ste93cry:
Since they are marked as `@internal` we can remove them immediatly without breaking BC per PHPDoc-definition as they should not be used outside the application or library:

>The @internal tag is used to denote that associated Structural Elements are elements internal to this application or library

tests/State/ScopeTest.php
```diff
@@ -15,62 +15,127 @@ final class ScopeTest extends TestCase
...
         $scope->setExtra('bar', 'baz');
 
-        $this->assertEquals(['foo' => 'bar', 'bar' => 'baz'], $scope->getExtra());
+        $event = $scope->applyToEvent(new Event(), []);
+
+        $this->assertNotNull($event);
+        $this->assertSame(['foo' => 'bar', 'bar' => 'baz'], $event->getExtraContext()->toArray());
+    }
+
+    public function setExtras(): void
```
Jean85:
```suggestion
    public function testSetExtras(): void
```

PR #853, opened 2019-07-19
Revert extracting the transaction name from structured log data in Monolog handler

ste93cry:
This PR reverts #843 after the decision of not wanting to support extracting information from structured log data due to the following reasons:

- some key names (e.g. `transaction`) are too generic and may not represent the same concept intended by Sentry
- blindly setting anything from the structured log data into the event payload is not a good move as the data could be really big or it may contain unwanted things

/cc @JanMikes


ste93cry:
Actually the only reason we can simply revert this commit without breaking BC is that we didn't release yet the version in which it was merged :wink:

Issue #787, opened 2019-03-06
HttpTransport example does not work


cosminardeleanu:
Your example from https://docs.sentry.io/platforms/php/#httptransport does not work, or is incomplete.

Used version 2.0.1.

Please add a working version and/or fix the code.

Context: I am trying to send messages, via queues, so i store the `Event` and dispatch it later, by using the HttpTransport.

```php
use Http\Discovery\HttpAsyncClientDiscovery;
use Http\Discovery\MessageFactoryDiscovery;
use Sentry\ClientBuilder;
use Sentry\Transport\HttpTransport;

$options = ['dsn' => '...'];
$builder = ClientBuilder::create($options);
$options = $builder->getOptions();

$transport = new HttpTransport($options, HttpAsyncClientDiscovery::find(), MessageFactoryDiscovery::find());
$builder->setTransport($transport);
$client = $builder->getClient();

echo $client->captureMessage('test message'); // Does not work.

$event = ... // Gathered from queue, serialized object.
$transport->send($event); // Also does not work.
```

This example does not work.
Looking deep in the sentry code, i have seen that the http transport model, is being filled with various plugins, that actually makes the guzzle6 adapter to work: 

`new BaseUriPlugin($uriFactory->createUri($options->getDsn()))` and `new AuthenticationPlugin(new SentryAuthentication($options, $sdkIdentifier, $sdkVersion));`.

The only working around was to use the `PluginClient` and copy/paste the logic from `ClientBuilder`, `private function createHttpClientInstance(): PluginClient`.

Jean85:
As a workaround, you could use the Hub with no default integrations, and send the events from there. IT should result in a smaller code example that works.

javascriptlove:
Can't get it to work also, the example is broken and/or incomplete. Seems like we still have to use the old Raven

Jean85:
This issue could be related to the issues we're facing with async transport...

ste93cry:
I don't think so, I believe that as written by @cosminardeleanu when you set your transport then you lose all the automation done to set the plugins to the HTTP client needed to authenticate with the server. It's a pity and we can think about fixing it, but for now it's this way

B-Galati:
Just had the same issue, IMHO the spool transport is not usable right now from a DX perspective.

B-Galati:
My workaround for now is to use reflection: 

```
$reflection = new \ReflectionClass(ClientBuilder::class);
$method = $reflection->getMethod('createTransportInstance');
[...]

PR #855, opened 2019-07-19
Add wrapHttpClientWithPlugins method to ClientBuilder.

RageZBla:
This should ease the process of registering a custom transport.

Related to #787


src/ClientBuilder.php
```diff
@@ -354,4 +343,34 @@ private function createEventFactory(): EventFactoryInterface
...
     }
+
+    public function wrapHttpClientWithPlugins(HttpAsyncClient $client): PluginClient
+    {
+        return new PluginClient($client, $this->httpClientPlugins);
+    }
+
+    private function registerDefaultPlugins(): void
+    {
+        if (0 !== \count($this->httpClientPlugins)) {
```
Jean85: I would move this `if` at the start of the `wrapHttpClientWithPlugins` method and use it to call `registerDefaultPlugins`

RageZBla: @Jean85 thanks for reviewing the PR; We have applied your suggestion.

ste93cry:
I don’t like to expose such method from the builder as it seems we’re leaking some internal behavior of the builder. I didn’t investigate if it’s possible, but we should probably instead wrap the HTTP client internally

RageZBla:
@ste93cry I can see your point but in that case the documentation to register custom transport should have all the boilerplate to basically reproduce the `registerDefaultPlugin` behavior.

ste93cry:
The problem arises because if someone sets the transport then we don't wrap the HTTP client because the transport may not need it and so doing it would be useless

https://github.com/getsentry/sentry-php/blob/98a07e7a800b6d2fee52157ef6936cc3edda6bdd/src/ClientBuilder.php#L309-L313

What I was suggesting was to always wrap the client regardless of the transport set. A clearer solution imo would be to have a `setTransportFactory` method in place of the `setHttpClient`, `setTransport`, `setMessageFactory`, etc methods that you can use to set your own transport so that the builder does not need to care about instantiating the HTTP client based on what transport you want to use. This would not solve the problem that you would have to wrap your HTTP client with all the needed plugins, but I don't really get why if you decide to use your own transport then the SDK should be responsible to partially "configure" it

RageZBla:
@ste93cry thanks for the feedback. I'll try to update my pull request over the week end.

RageZBla:
@ste93cry may I ask for your opinion? 

This is not the complete thing, I still need to write the tests but before that I would like you to acknowledge if the solution looks good enough for you. 

Regarding the `testGetClientTogglesCompressionPluginInHttpClient`, we should probably move it to the `DefaultTransportFactory` test when I write it.

ste93cry:
I think that the direction may be good. Probably it makes sense to passo to the `create` method of the factory the options of the client. I would remove the setters from the factory, once it's initialized I don't see any reason to let users change its state.

RageZBla:
@ste93cry let me know if it looks like something that could be merged or if I need to work on it a little bit more.

ste93cry:
I did some changes to your PR, may I ask if I can push them to your branch? In that case please ensure that maintainers can edit your PR by checking the checkbox in the right sidebar

RageZBla:
@ste93cry the checkbox is ticked so you should be able to push. Thanks!

ste93cry:
I pushed the changes and there is still a lot of work that must be done, from documentation to unit tests, but this should be a good start to provide the flexibility you were looking for at the beginning of this PR to create a custom HTTP client without having to also think about the plugins needed to make it work. I didn't have the chance to try it with a real Sentry server yet, so it may be that everything doesn't work anymore. It's just a starting point, please let me know if you're ok with them and then we can interate over this if you agree and see what cames out. The only downside to all of these changes is that we are polluting the codebase with deprecated things and the complexity is increasing rapidly and this makes me wonder if this is a good way to solve the issue

RageZBla:
@ste93cry Thanks, I see you went all the way to create a dedicated factory for the `HttpClient`. 

Yes, it is covering my original use case :+1:

I'll try it out with real server tomorrow at work. Given that we moved code around, I think it should work. 

In my opinion, it looks good and add some modularity and good extension points.

Do you want me to fix the broken tests and PHPStan? On the other hand what kind of test you had in mind for the `HttpClientFactory`?

RageZBla:
@ste93cry tested it works :ballot_box_with_check: 

Moving on to fix the tests.

RageZBla:
@ste93cry I have added test sadly the deprecations are in the way. Let me know what you want to do about it?

Jean85:
@RageZBla when you run a test with an expected deprecation you can set it as `@legacy` and use an `@expectDeprecation` annotation. You can see it example of this in other places in the tests or in #847

RageZBla:
@Jean85 gotcha let me do that.

RageZBla:
@Jean85 build is green :raised_hands:

RageZBla:
@ste93cry any feedback on this?

ste93cry:
This PR needs to be rebased but since it's scheduled for 2.3 and there are still things left to be merged it may be that it will need to be done again. We should also find a way to still support adding custom plugins to the HTTP client to not break BC. Since this is new code we should finally avoid using the deprecated `assertAttribute` methods of PHPUnit and find a better way to test the things that use them at the time of writing. As said before, as a last step we should document all methods, classes and interfaces added

src/Transport/DefaultTransportFactory.php
```diff
@@ -0,0 +1,37 @@
...
+
+declare(strict_types=1);
+
+namespace Sentry\Transport;
+
+use Http\Message\MessageFactory as MessageFactoryInterface;
+use Sentry\HttpClient\HttpClientFactoryInterface;
+use Sentry\Options;
+
+final class DefaultTransportFactory implements TransportFactoryInterface
```
ste93cry: Please add the docblock with the description of what this class does, e.g. `This class is the default implementation of the {@see TransportFactoryInterface} interface.`

src/Transport/TransportFactoryInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\Transport;
+
+use Sentry\Options;
+
+interface TransportFactoryInterface
```
ste93cry: Please add a docblock with the description of the purpose of this interface, e.g. `This interface defines a contract for all classes willing to create instances of the transport to use with the Sentry client.`

RageZBla:
@ste93cry can you review and give us feedback? 

Regarding this: 

> It would be also good to rework all the tests to stop using assertAttribute which is deprecated by PHPUnit

I am fine with that but that would require to add getters only for testing, are you OK with that?

ste93cry:
>I am fine with that but that would require to add getters only for testing, are you OK with that?

I reworked some of the newly added tests to not use the reflection to test internal state of the objects and I also created an additional HTTP client factory to support wrapping a client with the plugins. I decided to go that way because in `3.0` I would like to stop supporting out-of-the-box this feature and leave to the user the choice of how to do it, so it made more sense to not tie the `HttpClientFactory::create` method with the list of plugins. I also fixed some code style issues and polished a few things. WDYT?

src/HttpClient/HttpClientFactory.php
```diff
@@ -0,0 +1,116 @@
...
+     */
+    public function create(Options $options): HttpAsyncClientInterface
+    {
+        if (null === $options->getDsn()) {
+            throw new \RuntimeException('Cannot create an HTTP client without the Sentry DSN set in the options.');
+        }
+
+        $httpClient = $this->httpClient;
+
+        if (null !== $httpClient && null !== $options->getHttpProxy()) {
```
Jean85: I would check if the client is a compatible one, instead of `null` altogether.

src/HttpClient/HttpClientFactory.php
```diff
@@ -0,0 +1,116 @@
...
+            throw new \RuntimeException('The "http_proxy" option does not work together with a custom HTTP client.');
+        }
+
+        if (null === $httpClient && null !== $options->getHttpProxy()) {
```
Jean85: Same here.

src/HttpClient/HttpClientFactory.php
```diff
@@ -0,0 +1,116 @@
...
+            if (!class_exists(HttpCurlClient::class)) {
+                throw new \RuntimeException('The "http_proxy" option requires the `php-http/curl-client` package to be installed.');
+            }
+
+            $httpClient = new HttpCurlClient(null, null, [
```
Jean85: This piece of code will conflict with #890.

src/HttpClient/HttpClientFactory.php
```diff
@@ -0,0 +1,116 @@
...
+                CURLOPT_PROXY => $options->getHttpProxy(),
+            ]);
+        }
+
+        if (null === $httpClient) {
+            $httpClient = HttpAsyncClientDiscovery::find();
+        }
+
+        $httpClientPlugins = [
```
Jean85: I would move this to a `getDefaultPlugins()` method.

src/HttpClient/PluggableHttpClientFactory.php
```diff
@@ -0,0 +1,52 @@
...
+use Sentry\Options;
+
+/**
+ * This factory can be used to decorate an HTTP client with a list of plugins.
+ *
+ * @author Stefano Arlandini <sarlandini@alice.it>
+ *
+ * @deprecated since version 2.3, to be removed in 3.0
+ */
+final class PluggableHttpClientFactory implements HttpClientFactoryInterface
```
Jean85: This class seems an overkill to me. IMHO we should just use the `getDefaultPlugins()` method that I suggested as public and add a `$plugins` optional argumento to `HttpClientFactoryInterface::create`

src/HttpClient/HttpClientFactory.php
ste93cry: Indeed, depending on which one will be merged first I will rebase the second PR

src/HttpClient/PluggableHttpClientFactory.php
ste93cry: It was done like this before, but since in `3.0` I don't want to support customizing the plugins of the HTTP client out-of-the-box (I don't see a real value in supporting this use case) I don't want to tie the `create` method to the list of plugins. This class will be used automatically to wrap the default factory only if the user added a custom plugin to the builder and we can remove it in the next major version without changing anything else besides the builder itself

Jean85: The value is in the fact that at least some of them are necessary to use the Sentry APIs correctly. Wrapping them there would mean that the transport factory would have the responsibility of knowing the default plugins, and instead the client factory should care about the HTTP client only.

ste93cry: I see a bit of confusion here, so I will try to explain how things are organized. There are two factories: the transport factory and the HTTP client factory. The first one has been introduced to encapsulate the logic to create the transport and move it out of the builder. The second factory instead is the most important and its purpose is to create a fully configured HTTP client ready to be used by Sentry to send events to the server. To do such thing is adds certain plugins to the client, and extracting such logic from the factory would make it useless as the configured client would not be working out-of-the-box without them. Additional plugins that the users wants to add instead are not part of the client configuration and it's not responsibility of the default factory to add them to the HTTP client

Jean85: Sure, and that's why I'm suggesting a different approach. With your current one, the TransportFactory is just an empty shell without any proper responsibility; I'm suggesting moving the responsibility of attaching the needed plugins there, so the HTTPClientFactory instead will focus on the pure HTTP client and its options, without worrying about what Sentry server needs.

In this way, the end user will be able to intervene onto those two different concerns (plugins and HTTP client) independently, with a simple substitution of one of those two factories.

Jean85: We discussed it at length via chat, I'm ok with the current implementation.

src/HttpClient/HttpClientFactory.php
ste93cry: We could, but the result would not be the same. The whole logic is:

- If a HTTP client is set and the `http_proxy` option is set then throw an error because we cannot modify the options of an already instantiated client at runtime
- If a HTTP client is not set and the `http_proxy` option is set then try to instantiate a Guzzle client or a Httplug Curl client in this order
- If a HTTP client is set and no `http_proxy` option is set then use it
- In all other cases try to use autodiscovery or throw if we cannot find a suitable HTTP client

As you can see the logic cannot be replaced with a simple `if` that compares whether the HTTP client is supported out-of-the-box or not as the conditions to check are multiple

tests/ClientTest.php
```diff
@@ -10,21 +10,19 @@
...
 use Sentry\Transport\HttpTransport;
+use Sentry\Transport\TransportFactoryInterface;
 use Sentry\Transport\TransportInterface;
 
 class ClientTest extends TestCase
 {
     public function testCaptureMessage(): void
     {
-        /** @var TransportInterface|MockObject $transport */
+        /** @var TransportInterface&MockObject $transport */
```
Jean85: With the proper PHPStan extension this annotation should be superfluous.

I will not repeat this comment on the other lines.

ste93cry: It is, but the IDE is not able to recognize the full return type for the `createMock` function and the annotation helps it

Jean85: If that's the case (and you're using PHPStorm) you can use this plugin to make it work: https://plugins.jetbrains.com/plugin/9674-phpunit-enhancement/

PR #857, opened 2019-07-24
Fix sending of GZIP-compressed requests when the enable_compression option is on

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.1
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

When the `enable_compression` option was enabled it was affecting only the responses received by the server while first of all the requests should be compressed with GZIP. This PR fixes the problem. Note that if Sentry is running in uWSGI mode and behind a NGINX server as described in the [documentation](https://docs.sentry.io/server/performance/#performance-web-server) and the client sends a chunked request (as in our case) then a bug in the latter makes the proxied request have both the `Transfer-Encoding` and `Content-Length` headers set, which per HTTP/1.1 specs is not valid. For this reason Sentry will be unable to read the request content and will silently discard the event logging an error. The solution is to unset the `Transfer-Encoding` header in the NGINX config.

```nginx
server {
  ...

  location / {
    include uwsgi_params;
    
    uwsgi_param HTTP_TRANSFER_ENCODING "";

    ...
  }
}
```

ping @HazAT can you please update the docs with the note about NGINX + uWSGI?

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -32,7 +32,6 @@ cache:
 
 init:
     - SET PATH=c:\php\%PHP_VERSION%;%PATH%
-    - SET SYMFONY_DEPRECATIONS_HELPER=strict
     - SET ANSICON=121x90 (121x90)
     - SET INSTALL_PHP=1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Fix `TypeError` in `Sentry\Monolog\Handler` when the extra data array has numeric keys (#833).
+- Fix sending of GZIP-compressed requests when the `enable_compression` option is `true` (#857)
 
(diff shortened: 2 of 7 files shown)


.appveyor.yml
```diff
@@ -32,7 +32,6 @@ cache:
 
 init:
     - SET PATH=c:\php\%PHP_VERSION%;%PATH%
-    - SET SYMFONY_DEPRECATIONS_HELPER=strict
```
Jean85: Do we need to take care of some deprecations?

ste93cry: We have some deprecations in the codebase, but the reason this got removed is that it was useless as we have a good default in the `phpunit.xml.dist` file. In fact, Travis CI didn't set this environment variable :wink:

PR #858, opened 2019-07-25
Allow numeric value in tags.

leo108:
It's a very common scene that we use a numeric value as a tag, e.g. user id. 

Developers are not always aware that should pass a real string value to Sentry's TagsContext, and  InvalidArgumentException thrown in Sentry may not be caught by framework's exception handler because Sentry is called in the exception handler in most cases, thus no error logs in framework nor reports in Sentry, which makes it hard to track error.

### DIFF
--- a/src/Context/TagsContext.php
+++ b/src/Context/TagsContext.php
@@ -21,13 +21,7 @@ class TagsContext extends Context
             throw new \InvalidArgumentException('The tags context does not allow recursive merging of its data.');
         }
 
-        foreach ($data as $value) {
-            if (!\is_string($value)) {
-                throw new \InvalidArgumentException('The $data argument must contains a simple array of string values.');
-            }
-        }
-
-        parent::merge($data);
+        parent::merge(self::sanitizeData($data));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Context/TagsContext.php
```diff
@@ -21,52 +21,53 @@ public function merge(array $data, bool $recursive = false): void
...
+        }
+
         if (!\is_string($value)) {
             throw new \InvalidArgumentException('The $value argument must be a string.');
         }
 
         parent::offsetSet($offset, $value);
     }
+
+    private static function sanitizeTags(array $tags): array
```
ste93cry: Missing the description for both the function and the parameter. Also, maybe to be consistent with the names of the rest of the methods we could call this function `sanitizeData`, the parameter `$data` and the value being checked `$item`

leo108:
@Jean85 Test added.

@ste93cry Method renamed.

ste93cry:
Still missing the documentation of the private function, apart from that LGTM :+1:

ste93cry:
Still missing the documentation of the private function, apart from that it seems fine to me :+1:

leo108:
@ste93cry Doc added, sorry for the poor English.

leo108:
Any chance to merge this PR?

ste93cry:
@leo108 I did some changes to your PR, please let me know if you're ok with them

leo108:
@ste93cry LGTM, Thanks.

PR #861, opened 2019-07-31
Skip integrations not bound to the current client and fetch their options from it

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.1
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| License       | MIT

As noticed by @HazAT in #847, some integrations were executed even though the client binded to the current hub was not using them. I also changed how we pass the options to the integrations by stopping to use the constructor and fetching them from the client binded to the current hub so that we can avoid having a circular reference between the two

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Fix `TypeError` in `Sentry\Monolog\Handler` when the extra data array has numeric keys (#833).
 - Fix sending of GZIP-compressed requests when the `enable_compression` option is `true` (#857)
 - Fix error thrown when trying to set the `transaction` attribute of the event in a CLI environment (#862)
+- Fix integrations that were not skipped if the client binded to the current hub was not using them (#861)
 
 ## 2.1.1 (2019-06-13)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 - Fix `TypeError` in `Sentry\Monolog\Handler` when the extra data array has numeric keys (#833).
 - Fix sending of GZIP-compressed requests when the `enable_compression` option is `true` (#857)
 - Fix error thrown when trying to set the `transaction` attribute of the event in a CLI environment (#862)
-- Fix integrations that were not skipped if the client binded to the current hub was not using them (#861)
+- Fix integrations that were not skipped if the client bound to the current hub was not using them (#861)
 
 ## 2.1.1 (2019-06-13)
 
--- a/src/Integration/ErrorListenerIntegration.php
+++ b/src/Integration/ErrorListenerIntegration.php
@@ -61,7 +61,7 @@ final class ErrorListenerIntegration implements IntegrationInterface
             $integration = $currentHub->getIntegration(self::class);
             $client = $currentHub->getClient();
 
-            // The client binded to the current hub, if any, could not have this
+            // The client bound to the current hub, if any, could not have this
             // integration enabled. If this is the case, bail out
             if (null === $integration || null === $client) {
                 return;
--- a/src/Integration/ErrorListenerIntegration.php
+++ b/src/Integration/ErrorListenerIntegration.php
@@ -58,19 +58,12 @@ final class ErrorListenerIntegration implements IntegrationInterface
             }
 
             $currentHub = Hub::getCurrent();
+            $integration = $currentHub->getIntegration(self::class);
             $client = $currentHub->getClient();
 
-            // The client could have been detached from the hub. If this is the
-            // case this integration should not run
-            if (null === $client) {
-                return;
-            }
-
-            $integration = $client->getIntegration(self::class);
-
-            // The integration could be binded to a client that is not the one
-            // attached to the current hub. If this is the case, bail out
-            if (null === $integration) {
+            // The client binded to the current hub, if any, could not have this
+            // integration enabled. If this is the case, bail out
+            if (null === $integration || null === $client) {
                 return;
             }
 
--- a/src/Integration/ErrorListenerIntegration.php
+++ b/src/Integration/ErrorListenerIntegration.php
@@ -17,7 +17,7 @@ use Sentry\State\Hub;
 final class ErrorListenerIntegration implements IntegrationInterface
 {
     /**
-     * @var Options The options, to know which error level to use
+     * @var Options|null The options, to know which error level to use
      */
     private $options;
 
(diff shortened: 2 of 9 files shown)


ste93cry:
The reason is just for backwards compatibility. I could simply stop using the options passed to the constructor and always use the client to fetch them, however people that implemented the integrations wrong (like we did!) and that (unlikely case, but possible so we must take it into account) change the hub during execution and have a different client bound to it will suddenly find out that their app breaks because the behavior changed and the options are now different from the ones they expected (even though they are now the ones that should be)

src/Integration/ErrorListenerIntegration.php
```diff
@@ -53,15 +57,34 @@ public function setupOnce(): void
...
+            $currentHub = Hub::getCurrent();
+            $client = $currentHub->getClient();
+
+            // The client could have been detached from the hub. If this is the
+            // case this integration should not run
+            if (null === $client) {
+                return;
+            }
+
+            $integration = $client->getIntegration(self::class);
```
Jean85: Sure, I understood that. But I see that here you're getting the integration from the client...

src/Integration/ExceptionListenerIntegration.php
```diff
@@ -20,7 +20,16 @@ public function setupOnce(): void
     {
         $errorHandler = ErrorHandler::registerOnce(ErrorHandler::DEFAULT_RESERVED_MEMORY_SIZE, false);
         $errorHandler->addExceptionHandlerListener(static function (\Throwable $exception): void {
-            Hub::getCurrent()->captureException($exception);
+            $currentHub = Hub::getCurrent();
+            $integration = $currentHub->getIntegration(self::class);
```
Jean85: ..and here from the hub.

src/Integration/ErrorListenerIntegration.php
HazAT: Yup this is how it should be done:
```
$currentHub = Hub::getCurrent();
$integration = $currentHub->getIntegration(self::class);
```

ste93cry: It's the same, since `Hub::getCurrent()->getIntegration(...)` is really just a shortcut for `Hub::getCurrent()->getClient()->getIntegration(...)`

Jean85: Then we should consistently use always the same path, I would say from the Hub for shortness.

ste93cry: It's the same, since `Hub::getCurrent()->getIntegration(...)` is really just a shortcut for `Hub::getCurrent()->getClient()->getIntegration(...)`. The reason I did it like this here and not below is that I don't always need the `$client` (e.g. for getting the options), so in the other places I can avoid writing some code

src/Integration/ErrorListenerIntegration.php
```diff
@@ -29,17 +29,21 @@ final class ErrorListenerIntegration implements IntegrationInterface
     /**
      * Constructor.
      *
-     * @param Options $options           The options to be used with this integration
-     * @param bool    $handleFatalErrors Whether to handle fatal errors or not
+     * @param Options|null $options           The options to be used with this integration
+     * @param bool         $handleFatalErrors Whether to handle fatal errors or not
      */
-    public function __construct(Options $options, bool $handleFatalErrors = true)
+    public function __construct(?Options $options = null, bool $handleFatalErrors = true)
```
HazAT: This is kind of an internal note to myself.
In JS for example we have specific options just for the integration that are not reflected in the global options.
In PHP we have all possible options in the global options object, so making the options deprecated here makes kinda sense. TBD if this should be homogenized between SDKs.

src/Integration/RequestIntegration.php
```diff
@@ -57,13 +61,17 @@ public function __construct(Options $options)
     public function setupOnce(): void
     {
         Scope::addGlobalEventProcessor(function (Event $event): Event {
-            $self = Hub::getCurrent()->getIntegration(self::class);
+            $currentHub = Hub::getCurrent();
+            $integration = $currentHub->getIntegration(self::class);
+            $client = $currentHub->getClient();
 
-            if (!$self instanceof self) {
+            // The client binded to the current hub, if any, could not have this
```
HazAT:
```suggestion
            // The client bound to the current hub, if any, could not have this
```

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Fix `TypeError` in `Sentry\Monolog\Handler` when the extra data array has numeric keys (#833).
 - Fix sending of GZIP-compressed requests when the `enable_compression` option is `true` (#857)
 - Fix error thrown when trying to set the `transaction` attribute of the event in a CLI environment (#862)
+- Fix integrations that were not skipped if the client binded to the current hub was not using them (#861)
```
HazAT:
```suggestion
- Fix integrations that were not skipped if the client bound to the current hub was not using them (#861)
```

src/Integration/ErrorListenerIntegration.php
```diff
@@ -53,15 +57,27 @@ public function setupOnce(): void
                 return;
             }
 
-            if ($exception instanceof SilencedErrorException && !$this->options->shouldCaptureSilencedErrors()) {
+            $currentHub = Hub::getCurrent();
+            $integration = $currentHub->getIntegration(self::class);
+            $client = $currentHub->getClient();
+
+            // The client binded to the current hub, if any, could not have this
```
HazAT:
```suggestion
            // The client bound to the current hub, if any, could not have this
```

src/Integration/ExceptionListenerIntegration.php
```diff
@@ -20,7 +20,16 @@ public function setupOnce(): void
...
+
+            // The client binded to the current hub, if any, could not have this
```
HazAT:
```suggestion
            // The client bound to the current hub, if any, could not have this
```

src/Integration/FatalErrorListenerIntegration.php
```diff
@@ -38,11 +42,23 @@ public function setupOnce(): void
     {
         $errorHandler = ErrorHandler::registerOnceFatalErrorHandler();
         $errorHandler->addFatalErrorHandlerListener(function (FatalErrorException $exception): void {
-            if (!($this->options->getErrorTypes() & $exception->getSeverity())) {
+            $currentHub = Hub::getCurrent();
+            $integration = $currentHub->getIntegration(self::class);
+            $client = $currentHub->getClient();
+
+            // The client binded to the current hub, if any, could not have this
```
HazAT:
```suggestion
            // The client bound to the current hub, if any, could not have this
```

src/Integration/ModulesIntegration.php
```diff
@@ -27,10 +27,12 @@ final class ModulesIntegration implements IntegrationInterface
     public function setupOnce(): void
     {
         Scope::addGlobalEventProcessor(function (Event $event) {
-            $self = Hub::getCurrent()->getIntegration(self::class);
+            $integration = Hub::getCurrent()->getIntegration(self::class);
 
-            if ($self instanceof self) {
-                self::applyToEvent($self, $event);
+            // The integration could be binded to a client that is not the one
```
HazAT:
```suggestion
            // The integration could be bound to a client that is not the one
```

src/Integration/ErrorListenerIntegration.php
ste93cry: I think that it makes sense to have per-integration options, definitely! Honestly I'm not sure about whether we should deprecate passing options to the constructor or just ignore them as this PR is basically fixing a bug. Anyway, I don't have in mind a way to move the existing options from the global configuration to each integration without breaking BC

HazAT: Maybe we just don't make it `optional` then the migration moving the options to the integration is a bit easier.

ste93cry: We have to make it `optional` because otherwise we cannot throw the deprecation saying that passing global options is discouraged as we don't have a way to know if they did or not

PR #865, opened 2019-08-09
Add an integration to set the transaction attribute of the event

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.2
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Stemming from #864, this PR adds a new integration similar to the one present in the JS SDK for NodeJS that sets the `transaction` attribute of the event. This is a much more cleaner approach than setting it into the event factory as it can be configured from the user so that if he doesn't want to report such information he won't.

### DIFF
--- a/src/Integration/TransactionIntegration.php
+++ b/src/Integration/TransactionIntegration.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\Integration;
+
+use Sentry\Event;
+use Sentry\State\Hub;
+use Sentry\State\Scope;
+
+/**
+ * This integration sets the `transaction` attribute of the event to the value
+ * found in the raw event payload or to the value of the `PATH_INFO` server var
+ * if present.
+ *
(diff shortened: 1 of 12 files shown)


src/Integration/TransactionIntegration.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+final class TransactionIntegration implements IntegrationInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function setupOnce(): void
+    {
+        Scope::addGlobalEventProcessor(static function (Event $event, array $payload): Event {
+            if (null !== $event->getTransaction()) {
```
HazAT: We also need the 
```
$currentHub = Hub::getCurrent();
$integration = $currentHub->getIntegration(self::class);
$client = $currentHub->getClient();
// The client bound to the current hub, if any, could not have this
// integration enabled. If this is the case, bail out
if (null === $integration || null === $client) {
    return event;
}
```
here.

PR #867, opened 2019-08-14
Make cleanupPendingRequests public

ruudk:
This allows for cleaning up pending requests in a long running process (queue consumer).

I want to create a [LongRunning Cleaner](https://github.com/LongRunning/LongRunning) that automatically calls this method when the cleaner is executed. For example after the execution of a RabbitMQ message in a consumer.

### DIFF
--- a/src/Transport/HttpTransport.php
+++ b/src/Transport/HttpTransport.php
@@ -102,7 +102,7 @@ final class HttpTransport implements TransportInterface
      * Cleanups the pending requests by forcing them to be sent. Any error that
      * occurs will be ignored.
      */
-    private function cleanupPendingRequests(): void
+    public function cleanupPendingRequests(): void
     {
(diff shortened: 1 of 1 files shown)


ste93cry:
Thank you for the contribution, this issue is going to be solved with the implementation of a public `flush` method on the `Client` class and subsequent refactoring of the transports to align them with the [Unified API specs](https://docs.sentry.io/development/sdk-dev/unified-api/). The update is going to be merged as part of `2.2` that will came hopefully soon. Making this method public now means that we cannot revert this change anymore because we would break BC. Right now the suggested workaround is to create your own transport and implement the logic you need to flush the events on-demand. Other solutions have been posted in #811. Closing this, you can check the referenced issue for updates

ruudk:
@ste93cry Thanks for the fast reply. I should have searched for the issue first. Glad to hear that you are working on this 💪

PR #868, opened 2019-08-20
Remote static integrations

Taluu:
As the classes are final, these are not bc breaks ; instead of using a "a la python" mecanism to apply to events on the request integration (getting the instance of the hub), we should use the current one and be done with it.

this allows us to make the request integration independent from the Hub, making it way easier to integrate it from 3rd party (such as Monolog).

### DIFF
--- a/src/Integration/ModulesIntegration.php
+++ b/src/Integration/ModulesIntegration.php
@@ -7,7 +7,6 @@ namespace Sentry\Integration;
 use Jean85\PrettyVersions;
 use PackageVersions\Versions;
 use Sentry\Event;
-use Sentry\State\Hub;
 use Sentry\State\Scope;
 
 /**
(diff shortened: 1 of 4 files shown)


Taluu:
Note : Not sure if I should I target master or develop though for this. Feel free to change it (if there are any conflicts, I'll take care of them).

Taluu:
Locking this PR, as the last work made on the request removes the static things (not all of them though).

This requires another idea I have to make it work without the "static" things in it, making it so much easier to properly integrate it.

HazAT:
The reason it's currently like this is that there could be multiple clients running at the same time and not all of them use all integrations. That's why an integration needs to ask the current hub if the active client has this integration enabled.

Taluu:
Yep, the solution / idea I had in mind was to inject the client when building the... well, the client.

So that this integration only works when using said client.

PR #872, opened 2019-08-24
Allow extras to be cleared

adam1010:
I am trying to send along data (extras) that is specific to a particular error. Currently when I set tags it attaches them to ALL errors that I send. After I send an error I want to be able to clear out that data so it doesn't pollute future calls. This is easier than trying to build up a full new scope each time

### DIFF
--- a/src/State/Scope.php
+++ b/src/State/Scope.php
@@ -274,6 +274,18 @@ final class Scope
     }
 
     /**
+     * Resets the extra data
+     *
+     * @return $this
+     */
+    public function clearExtra(): self
+    {
+        $this->extra->clear();
+
+        return $this;
+    }
+
+    /**
      * Applies the current context and fingerprint to the event. If the event has
      * already some breadcrumbs on it, the ones from this scope won't get merged.
      *


Jean85:
You're using the wrong method then. You can use `withScope` to momentarely set new tags and send events inside a callback, of take leverage of the Scope stack: Scopes are in a stack, so you can push and pop any time you want, and any new pushed Scope will retain all the data from the previous one; popping it later will restore easily the previous situation.

adam1010:
@Jean85 Thanks for explaining that! I'm reading about `withScope` and it looks like it will create a new scope for that event, so global scope I've setup previously (like the logged in user) wouldn't get included -- is that correct? I would want that part of the scope to be preserved, it's only the event-specific data (the `extra` data) that I'm wanting to set on a per-event basis.

Jean85:
`withScope` only pushes temporarely a new `Scope` for you before and pops it automatically after the callback execution, so no, same behavior applies: data from previous Scope is copied over.

adam1010:
So then how can I copy the user context into my new scope from the previous scope so that it gets included?  Seems like my workaround of just clearing the tags on the global scope after sending my event is still a feasible option

ste93cry:
To clarify, there are two functions of interest here: `configureScope` and `withScope`. The first one lets you set data on the current scope (breadcrumbs, extra, tags, user, etc) and it stays forever until you decide to overwrite or delete it. The second function instead pushes a new scope on the stack, meaning that the data from the previous scope is merged with the data of the newly created scope. So effectively `withScope` is just cloning the previous scope and any edit to the data is reverted once the callback function ends its execution. This is an example to show what to expect using those functions:

```php
configureScope(static function (Scope $scope): void {
    $scope->setExtra('foo', 'bar');
});

captureMessage('foo bar'); // event's extra will contain `foo => bar`

withScope(static function (Scope $scope): void {
    $scope->setExtra('bar', 'baz');

    captureMessage('bar baz'); // event's extra will contain `foo => bar` and `bar => baz`
});

captureMessage('foo bar baz'); // event's extra will contain `foo => bar`
```

stayallive:
As detailed, `withScope` should be the route to take to accomplish this. Since the "issue" is addressed and there is already a way to accomplish what this PR tries to implement I'm going to close this.

PR #874, opened 2019-08-28
Add 3rd party integrations

christophlehmann:


ste93cry:
Or even remove them as I already proposed some time ago. In the README of the version 2 of this library there is absolutely no reason to link to integrations that are not compatible with it. Of course the old branch for `1.x` should still link to them

ste93cry:
>Sorry to hijack your PR for this discussion but this would also mean this PR will not be accepted because the 2 clients you added support only the 1.x Sentry client.

Well, if we want to advertise the clients for `1.x` then we can just change the target branch of this PR to the correct one. Still, for `master` and `develop` I already expressed my opinion. @HazAT did you change your mind on this since the last time?

HazAT:
Can you please create a separate section for integrations that still use raven-php?
It could confuse users if that is not clear.

ste93cry:
@HazAT wouldn't be clearer if each branch has its own README where only the integrations that are compatible with the related version are listed? For example, the `1.x` branch would list all integrations that works with the old SDK while `master` and `develop` branches would list only the integrations compatible with the new SDK

HazAT:
@ste93cry No, please put it in the main readme with sections.
No one besides us (contributors) ever switches branches on Github. We need to build towards great out of the box experience for new users that visit Github.

christophlehmann:
The Client for Flow Framework now uses the new SDK. The other projects of us (TYPO3 & Neos) will follow soon

ste93cry:
So, the list of 3rd party integrations to move to a section dedicated to the SDK `1.x` are:

- Drupal
- ZendFramework
- OpenCart

mfb:
We can release the new Drupal integration when 2.2 is out

Jean85:
The ZendFramework one was updated too: https://github.com/facile-it/sentry-module/blob/d9d9e6dace1cbb447d88874eb7193aaf20068b1f/composer.json#L48

mfb:
> We can release the new Drupal integration when 2.2 is out

ok I'm a little delayed on this as I need to also make a custom HTTP client to support additional configurations like CA cert, timeout, etc.

ste93cry:
@christophlehmann are you willing to make the requested changes so that we can merge this PR?

christophlehmann:
TYPO3 and Neos follow in another PR.

ste93cry:
If the release of both packages is not too far in the future we may want to just wait to merge this. Let me know if you have any ETA

christophlehmann:
Sure, should not take too long

stayallive:
@christophlehmann thanks for taking the time for making and keeping up with us changing the PR to what it is now 👍 

Happy to accept a PR as soon as the other clients are up-to-date too!

PR #876, opened 2019-08-30
add option to disable `ssl_verify_peer` in curl client

tldev-de:
This pull request will allow php sentry client to work with self-signed certificates or behind a company proxy, that breaks ssl connections.


ste93cry:
The options of the HTTP client depends on the adapter you choose to use, so we don’t want to encourage adding global options to the Sentry client to configure such things as it’s not our responsibility. We aknowledge that customizing the HTTP client is not so easy right now due to the architecture of the client builder and we will make it simpler in release `2.3`. However, this won’t change the fact that you will still need to set your own client.

PR #877, opened 2019-09-06
GZIP-encode the request in memory instead of streaming

### LINKED ISSUE #871: Events receiving HTTP 400 in v2.1.2
Exactly what it says on the tin. After upgrading from 2.1.1 to 2.1.2 this morning, events sent to our on-prem Sentry would get back an HTTP 400 with the following response body: `{"error":"No JSON data was found"}`. Upon further investigation (hampered by the fact that there's no `debug` option for the PHP SDK), I discovered that the request being made did, in fact, have an empty body. This was seen by manually dumping out both the request and response bodies from the `catch` in `HttpTransport->cleanupPendingRequests()`.

My `init` takes only a DSN and environment - everything else is left to the defaults.

I then worked around the issue by reverting to 2.1.1.

stayallive:
This is a fix for a PHP bug that was discovered causing the `GzipEncodeStream` to output an empty string when `exit($code)` is used instead of the expected encoded string which causes requests to Sentry to fail because the event data was never sent.

/fixes #871
/closes getsentry/sentry-laravel#271

### DIFF
--- a/src/HttpClient/Plugin/GzipEncoderPlugin.php
+++ b/src/HttpClient/Plugin/GzipEncoderPlugin.php
@@ -41,7 +41,7 @@ final class GzipEncoderPlugin implements PluginInterface
 
         $encodedBody = gzcompress($requestBody->getContents(), -1, ZLIB_ENCODING_GZIP);
 
-        if ($encodedBody === false) {
+        if (false === $encodedBody) {
             throw new \RuntimeException('Failed to GZIP-encode the request body.');
         }
 
--- a/src/HttpClient/Plugin/GzipEncoderPlugin.php
+++ b/src/HttpClient/Plugin/GzipEncoderPlugin.php
@@ -5,7 +5,7 @@ declare(strict_types=1);
 namespace Sentry\HttpClient\Plugin;
 
 use Http\Client\Common\Plugin as PluginInterface;
-use Http\Message\StreamFactory;
+use Http\Message\StreamFactory as StreamFactoryInterface;
 use Http\Promise\Promise as PromiseInterface;
 use Psr\Http\Message\RequestInterface;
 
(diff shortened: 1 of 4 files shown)


src/HttpClient/Plugin/GzipEncoderPlugin.php
```diff
@@ -16,16 +16,25 @@
...
+
     /**
      * Constructor.
      *
+     * @param StreamFactoryInterface $streamFactory The stream factory
+     *
      * @throws \RuntimeException If the zlib extension is not enabled
      */
-    public function __construct()
+    public function __construct(StreamFactoryInterface $streamFactory)
```
Jean85: This is a BC; we could make that argument optional and use the `StreamFactoryDiscovery::find()` paired with a deprecation.

src/HttpClient/Plugin/GzipEncoderPlugin.php
```diff
@@ -39,8 +48,14 @@ public function handleRequest(RequestInterface $request, callable $next, callabl
             $requestBody->rewind();
         }
 
+        $encodedBody = gzcompress($requestBody->getContents(), -1, ZLIB_ENCODING_GZIP);
+
+        if (false === $encodedBody) {
+            throw new \RuntimeException('Failed to GZIP-encode the request body.');
```
Jean85: Wouldn't this make the app blow up?

stayallive: No because we catch any and all errors occurring in the transport and discard them to prevent the blowing up part.

src/HttpClient/Plugin/GzipEncoderPlugin.php
stayallive: You are correct. I've added that, should be all sorted.

src/HttpClient/Plugin/GzipEncoderPlugin.php
ste93cry: Wouldn't this make the app blow up?

It would, and since it's an error happening while sending the event I don't have a good reason we should not thrown an exception. In case users are using our `HttpTransport` the exception will be caught and silently discarded though...

PR #890, opened 2019-09-27
Add support for Guzzle as default HTTP client

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.3
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

With this PR I'm going to add the support out-of-the-box for Guzzle so that we can switch later the default HTTP client from Httplug Curl which is bugged and far less developed and maintained

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -15,6 +15,15 @@ parameters:
         -
             message: "/^Parameter #1 \\$function of function register_shutdown_function expects callable\\(\\): void, 'register_shutdown…' given\\.$/"
             path: src/Transport/HttpTransport.php
+        -
+            message: '/^Class Http\\Adapter\\Guzzle6\\Client not found\.$/'
+            path: src/ClientBuilder.php
+        -
+            message: '/^Call to static method createWithConfig\(\) on an unknown class Http\\Adapter\\Guzzle6\\Client\.$/'
+            path: src/ClientBuilder.php
+        -
+            message: '/^Access to constant PROXY on an unknown class GuzzleHttp\\RequestOptions\.$/'
+            path: src/ClientBuilder.php
     excludes_analyse:
         - tests/resources
         - tests/Fixtures
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Sentry;
 
+use GuzzleHttp\RequestOptions as GuzzleHttpClientOptions;
+use Http\Adapter\Guzzle6\Client as GuzzleHttpClient;
 use Http\Client\Common\Plugin as PluginInterface;
 use Http\Client\Common\Plugin\AuthenticationPlugin;
 use Http\Client\Common\Plugin\BaseUriPlugin;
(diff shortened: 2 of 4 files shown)


phpstan.neon
```diff
@@ -15,6 +15,15 @@ parameters:
...
             path: src/Transport/HttpTransport.php
+        -
+            message: '/^Class Http\\Adapter\\Guzzle6\\Client not found\.$/'
+            path: src/ClientBuilder.php
+        -
+            message: '/^Call to static method createWithConfig\(\) on an unknown class Http\\Adapter\\Guzzle6\\Client\.$/'
+            path: src/ClientBuilder.php
+        -
+            message: '/^Access to constant PROXY on an unknown class GuzzleHttp\\RequestOptions\.$/'
+            path: src/ClientBuilder.php
```
Jean85: Wouldn't be better to require-dev those packages? We would just have to add another CI job where we remove those to check that our code works even without them...

src/ClientBuilder.php
```diff
@@ -332,14 +334,19 @@ private function createTransportInstance(): TransportInterface
...
             }
 
-            if (!ClassDiscovery::safeClassExists(HttpCurlClient::class)) {
-                throw new \RuntimeException('The `http_proxy` option requires the `php-http/curl-client` package to be installed.');
+            if (ClassDiscovery::safeClassExists(CurlHttpClient::class)) {
+                /** @psalm-suppress InvalidPropertyAssignmentValue */
+                $this->httpClient = new CurlHttpClient(null, null, [
+                    CURLOPT_PROXY => $this->options->getHttpProxy(),
+                ]);
+            } elseif (ClassDiscovery::safeClassExists(GuzzleHttpClient::class)) {
```
Jean85: I would reverse the sequence of this `if`, so we would actually prefer Guzzle over Curl.

ste93cry: I though about it, but if someone installed the Httplug Curl client on purpose and relied on the default configuration instead of forcefully setting the instance through the builder, then he would suddenly find out that it isn't the default anymore. Instead, with the order of the statements as-is people installing the newer version of `getsentry/sentry-php-sdk` will use Guzzle if they didn't install the Curl client, otherwise they will keep using the old default. Does it makes sense for you?

phpstan.neon
ste93cry: Since this part is not even tested, adding more steps to the CI won't allow us to know if this is gonna work or not so it's kinda useless

src/ClientBuilder.php
Jean85: I agree but I don't see a problem in the silent switch, the code is trnasport agnostic.
@HazAT WDYT?

ste93cry: >I agree but I don't see a problem in the silent switch

The problem is that all versions of the `sentry-php-sdk` package are requiring the core SDK using the `^x.x` constraint. This in turn makes possible the following scenario:

- `getsentry/sentry-php` version `2.2.x` installed
- `getsentry/sentry-php-sdk` version `2.1.x` installed (requires SDK `^2.2`)

He then updates `getsentry/sentry-php-sdk` because a new version is installed, however the core SDK is not updated. In this case he will download the Guzzle client, but the SDK will not support it out-of-the-box and if he's using the `http_proxy` option it will not work anymore. To avoid this we may evaluate to start syncing the two packages using a more restrictive constraint like `2.x`, but for the past versions the problem will remain and we will have to be careful about changes like this one. Hope it's not clear what the problem is, if not let me know and I will try to explain it in another way

Jean85: I don't understand how the SDK package will have support issues. It doesn't have any code, how could that be an issue? If you DON'T update the SDK, you will be stuck with Curl, and your `if`, even if put last, will work, because you will not go inside the first one.

ste93cry: After more thinking you are right, however the metapackage must bump the minimum version or you may update (and you will install Guzzle) without updating the core (which will not support it)

PR #892, opened 2019-10-01
Fix #891 by adding the errcontext argument

### LINKED ISSUE #891: previousErrorHandler number of arguments
```
app/vendor/sentry/sentry/src/ErrorHandler.php on line 386 and exactly 5 expected in /app/vendor/shopware/shopware/engine/Shopware/Plugins/Default/Core/ErrorHandler/Bootstrap.php on line 183
```

Originally caused by the following error:
```
The Sentry\State\Hub::getCurrent() method is deprecated since version 2.2 and will be removed in 3.0. Use SentrySdk::getCurrentHub() instead.
```

The `previousErrorHandler` registered accepts 5 arguments. The fifth argument is the `$errcontext` according to https://www.php.net/manual/de/function.set-error-handler.php.

I've fixed the deprecated call for now, but the errorHandler should in general deal with the fifth optional argument.

Jean85:


src/ErrorHandler.php
```diff
@@ -357,18 +357,19 @@ public function addExceptionHandlerListener(callable $listener): void
...
-     *                        of the E_* constants
-     * @param string $message The error message
-     * @param string $file    The filename the error was raised in
-     * @param int    $line    The line number the error was raised at
+     * @param int    $level      The level of the error raised, represented by
+     *                           one of the E_* constants
+     * @param string $message    The error message
+     * @param string $file       The filename the error was raised in
+     * @param int    $line       The line number the error was raised at
+     * @param array  $errcontext The error context (deprecated in PHP 7.2)
```
ste93cry: I think that it's more correct to say `deprecated since ...` instead of `deprecated in`

src/ErrorHandler.php
```diff
@@ -357,18 +357,19 @@ public function addExceptionHandlerListener(callable $listener): void
...
      *
      * @return bool If the function returns `false` then the PHP native error
      *              handler will be called
      *
      * @throws \Throwable
      */
-    private function handleError(int $level, string $message, string $file, int $line): bool
+    private function handleError(int $level, string $message, string $file, int $line, array $errcontext): bool
```
ste93cry: According to the documentation the fifth parameter is optional, so add a default value for it (empty array)

PR #894, opened 2019-10-02
Fix failing build on PHP 7.4

Jean85:
With this we will be sure of being compatible with PHP 7.4. The build was failing just because a PHPT test was triggering a new notice present under 7.4.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,6 @@ env:
 
 matrix:
     fast_finish: true
-    allow_failures:
-        - php: 7.4snapshot
 
 cache:
     directories:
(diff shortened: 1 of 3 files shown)


Jean85:
I've fixed the issue with the test, but now the build is failing for another issue in the `--prefer-lowest` part: https://travis-ci.org/getsentry/sentry-php/jobs/592558205

This seems not reproducible locally, but I'm using 7.4RC1, so I suspect that is just CI related. Putting this on hold for now.

Jean85:
Rebased on top of develop, but I fear that I need to wait #915 to have a clean build.

Jean85:
Re-rebased after #915 got merged, now we have one last failure in PHP 7.4 with prefer-lowest due to a Diactoros bug... 

Should we test that config? If yes, we would probably bump the Diactoros minimum requirement... I'll have to investigate up to which version though.

ste93cry:
The bug has been fixed in version `1.7.1` and I'm fine to update the minimum required version to that (or even to the latest version of `1.x`)

PR #898, opened 2019-10-09
fix: Catch expections that are caused by transport

HazAT:
This fixes an issue if Sentry replies with `429` error code that we no longer throw an exception.

```
PHP Fatal error:  Uncaught Http\Client\Common\Exception\ClientErrorException: TOO MANY REQUESTS in /Users/haza/Projects/demo-projects/php/demo-php/vendor/php-http/client-common/src/Plugin/ErrorPlugin.php:83
Stack trace:
#0 /Users/haza/Projects/demo-projects/php/demo-php/vendor/php-http/client-common/src/Plugin/ErrorPlugin.php(65): Http\Client\Common\Plugin\ErrorPlugin->transformResponseToException(Object(GuzzleHttp\Psr7\Request), Object(Zend\Diactoros\Response))
#1 [internal function]: Http\Client\Common\Plugin\ErrorPlugin->Http\Client\Common\Plugin\{closure}(Object(Zend\Diactoros\Response))
#2 /Users/haza/Projects/demo-projects/php/demo-php/vendor/php-http/curl-client/src/PromiseCore.php(215): call_user_func(Object(Closure), Object(Zend\Diactoros\Response))
#3 /Users/haza/Projects/demo-projects/php/demo-php/vendor/php-http/curl-client/src/MultiRunner.php(99): Http\Client\Curl\PromiseCore->fulfill()
#4 /Users/haza/Projects/demo-projects/php/demo-php/vendor/php-http/curl-client/src/CurlPromise.php(99): Http\Client\Curl\MultiRun in /Users/haza/Projects/demo-projects/php/demo-php/vendor/php-http/client-common/src/Plugin/ErrorPlugin.php on line 83

Fatal error: Uncaught Http\Client\Common\Exception\ClientErrorException: TOO MANY REQUESTS in /Users/haza/Projects/demo-projects/php/demo-php/vendor/php-http/client-common/src/Plugin/ErrorPlugin.php on line 83

Http\Client\Common\Exception\ClientErrorException: TOO MANY REQUESTS in /Users/haza/Projects/demo-projects/php/demo-php/vendor/php-http/client-common/src/Plugin/ErrorPlugin.php on line 83

Call Stack:
    0.9151    3181944   1. Sentry\ErrorHandler->handleException() /Users/haza/Projects/demo-projects/php/demo-php/vendor/sentry/sentry/src/ErrorHandler.php:0
```

### DIFF
--- a/src/Transport/HttpTransport.php
+++ b/src/Transport/HttpTransport.php
@@ -110,7 +110,7 @@ final class HttpTransport implements TransportInterface
             } else {
                 $promise->wait(false);
             }
-            
+
             return $event->getId();
         } catch (ClientErrorException $e) {
             return null;
--- a/src/Transport/HttpTransport.php
+++ b/src/Transport/HttpTransport.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Sentry\Transport;
 
+use Http\Client\Common\Exception\ClientErrorException;
 use Http\Client\HttpAsyncClient as HttpAsyncClientInterface;
 use Http\Message\RequestFactory as RequestFactoryInterface;
 use Http\Promise\Promise as PromiseInterface;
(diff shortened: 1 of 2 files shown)


src/Transport/HttpTransport.php
```diff
@@ -88,28 +89,32 @@ public function __destruct()
      */
     public function send(Event $event): ?string
     {
-        $projectId = $this->config->getProjectId();
-
-        if (null === $projectId) {
-            throw new MissingProjectIdCredentialException();
+        try {
```
ste93cry: Instead of catching the entire code of the method we should just wrap the `wait` function call. The same should be done in the `cleanupPendingRequests` method. The strange thing is that when `false` is passed to the `wait` method no exceptions should be thrown but just returned (effectively doing the same as a `try`/`catch` block), so there is a bug somewhere in the vendor or the exception is not raised there (the other point where it could be raised is the `sendAsyncRequest` method, but since we saw that the request is not sent until `wait` is called I don't think it's the case)

PR #899, opened 2019-10-09
Fix/http error master

HazAT:
This fixes an issue if Sentry replies with `429` error code that we no longer throw an exception.

```
PHP Fatal error:  Uncaught Http\Client\Common\Exception\ClientErrorException: TOO MANY REQUESTS in /Users/haza/Projects/demo-projects/php/demo-php/vendor/php-http/client-common/src/Plugin/ErrorPlugin.php:83
Stack trace:
#0 /Users/haza/Projects/demo-projects/php/demo-php/vendor/php-http/client-common/src/Plugin/ErrorPlugin.php(65): Http\Client\Common\Plugin\ErrorPlugin->transformResponseToException(Object(GuzzleHttp\Psr7\Request), Object(Zend\Diactoros\Response))
#1 [internal function]: Http\Client\Common\Plugin\ErrorPlugin->Http\Client\Common\Plugin\{closure}(Object(Zend\Diactoros\Response))
#2 /Users/haza/Projects/demo-projects/php/demo-php/vendor/php-http/curl-client/src/PromiseCore.php(215): call_user_func(Object(Closure), Object(Zend\Diactoros\Response))
#3 /Users/haza/Projects/demo-projects/php/demo-php/vendor/php-http/curl-client/src/MultiRunner.php(99): Http\Client\Curl\PromiseCore->fulfill()
#4 /Users/haza/Projects/demo-projects/php/demo-php/vendor/php-http/curl-client/src/CurlPromise.php(99): Http\Client\Curl\MultiRun in /Users/haza/Projects/demo-projects/php/demo-php/vendor/php-http/client-common/src/Plugin/ErrorPlugin.php on line 83

Fatal error: Uncaught Http\Client\Common\Exception\ClientErrorException: TOO MANY REQUESTS in /Users/haza/Projects/demo-projects/php/demo-php/vendor/php-http/client-common/src/Plugin/ErrorPlugin.php on line 83

Http\Client\Common\Exception\ClientErrorException: TOO MANY REQUESTS in /Users/haza/Projects/demo-projects/php/demo-php/vendor/php-http/client-common/src/Plugin/ErrorPlugin.php on line 83

Call Stack:
    0.9151    3181944   1. Sentry\ErrorHandler->handleException() /Users/haza/Projects/demo-projects/php/demo-php/vendor/sentry/sentry/src/ErrorHandler.php:0
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ cs-dry-run:
 	vendor/bin/php-cs-fixer fix --config=.php_cs --verbose --diff --dry-run
 
 cs-fix:
-	vendor/bin/php-cs-fixer fix --config=.php_cs
+	vendor/bin/php-cs-fixer fix --config=.php_cs.dist
 
 phpstan:
 	vendor/bin/phpstan analyse
--- a/src/Transport/HttpTransport.php
+++ b/src/Transport/HttpTransport.php
@@ -110,7 +110,7 @@ final class HttpTransport implements TransportInterface
             } else {
                 $promise->wait(false);
             }
-            
+
             return $event->getId();
         } catch (ClientErrorException $e) {
             return null;
--- a/src/Transport/HttpTransport.php
+++ b/src/Transport/HttpTransport.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Sentry\Transport;
 
+use Http\Client\Common\Exception\ClientErrorException;
 use Http\Client\HttpAsyncClient as HttpAsyncClientInterface;
 use Http\Message\RequestFactory as RequestFactoryInterface;
 use Http\Promise\Promise as PromiseInterface;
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -15,7 +15,7 @@ cs-dry-run:
 	vendor/bin/php-cs-fixer fix --config=.php_cs --verbose --diff --dry-run
 
 cs-fix:
-	vendor/bin/php-cs-fixer fix --config=.php_cs
+	vendor/bin/php-cs-fixer fix --config=.php_cs.dist
```
Jean85: We could omit this entirely:

```suggestion
	vendor/bin/php-cs-fixer fix
```

and let the auto-select do its magic

src/Transport/HttpTransport.php
```diff
@@ -106,7 +107,11 @@ public function send(Event $event): ?string
         if ($this->delaySendingUntilShutdown) {
             $this->pendingRequests[] = $promise;
         } else {
-            $promise->wait(false);
+            try {
+                $promise->wait(false);
+            } catch (ClientErrorException $e) {
```
ste93cry: We use to name the variable `$exception`

src/Transport/HttpTransport.php
```diff
@@ -119,7 +124,10 @@ public function send(Event $event): ?string
     private function cleanupPendingRequests(): void
     {
         while ($promise = array_pop($this->pendingRequests)) {
-            $promise->wait(false);
+            try {
+                $promise->wait(false);
+            } catch (ClientErrorException $e) {
```
ste93cry: We use to name the variable `$exception`

src/Transport/HttpTransport.php
```diff
@@ -119,7 +124,10 @@ public function send(Event $event): ?string
     private function cleanupPendingRequests(): void
     {
         while ($promise = array_pop($this->pendingRequests)) {
-            $promise->wait(false);
+            try {
+                $promise->wait(false);
```
ste93cry: You should remove the `false` parameter

src/Transport/HttpTransport.php
```diff
@@ -106,7 +107,11 @@ public function send(Event $event): ?string
         if ($this->delaySendingUntilShutdown) {
             $this->pendingRequests[] = $promise;
         } else {
-            $promise->wait(false);
+            try {
+                $promise->wait(false);
```
ste93cry: You should remove the `false` parameter

helhum:
Hey folks. From my perspective of being an author of a custom sentry integration, this change is a breaking change.

In the integration I have code like this:

```php
try {
    $hub->captureEvent($payload);
} catch (\Throwable $e) {
    // Avoid hard failure in case connection to sentry failed
    if ($this->fallbackWriter) {
        $this->fallbackWriter->writeLog(
            new LogRecord(
                'Sentry.Writer',
                LogLevel::ERROR,
                'Failed to write to Sentry',
                ['exception' => $e],
                $record->getRequestId()
            )
        );
    }
}
```

It means that I rely on the exception being throw to be able to handle that case and use a fallback strategy to capture the event, to make sure it is not silently discarded.

With this change, there is no way to detect such failure, except of course, by implementing my own HttpTransport.

Any thoughts?

ste93cry:
Technically speaking I think that we can consider this a BC indeed... However, please note that before `2.2.0` we were even more strict about catching and this PR is just going to fix a regression since `wait(false)` should **never** throw according to the documentation, but for some reason out of our control it did... :no_good:

helhum:
@ste93cry You are right, thanks. I missed that my code actually never did what I expected, as requests have been sent on shutdown only before 2.2.0. So In fact, there is no change in behavior here and I'm now onto implementing my own HttpTransport to be able to have a fallback feature in my integration.

Thanks a gain for the quick response!

PR #902, opened 2019-10-15
#901: Replace critical with fatal status

### LINKED ISSUE #901: Invalid breadcrumb level
Based on your [documentation](https://docs.sentry.io/enriching-error-data/breadcrumbs/?platform=php) breadcrumbs support the level `debug`, `info`, `warning`, `error` and `fatal` (see screenshot).

image

But the Breadcrumb class has a constant with the value `critical`, which is part of the allowed levels arrey, which is also used for validation in the constructor ([Breadcrumb.php#L64](https://github.com/getsentry/sentry-php/blob/be044bfa83ff98c5dd0ea0a7ff243ff0eade721c/src/Breadcrumb.php#L64)).

This issue causes confusion and conflict when translating PSR3 log levels to sentry breadcrumb levels. In addition neither the level `critical` nor the level `fatal` have special highlighting on a sentry exception. Only `warning` and `error` status receive a meaningful styling.

Please clearify which is now correct, `critical` or `fatal`. Either way please update the documentation and/or the php sdk. If `fatal` is the correct one, I would suggest adding the correct constant and deprecating the critical one.

TheLevti:
Critical breadcrumb status is not supported. Replaced it with fatal.

Fixes #901

### DIFF
--- a/src/Breadcrumb.php
+++ b/src/Breadcrumb.php
@@ -60,8 +60,15 @@ final class Breadcrumb implements \JsonSerializable
 
     /**
      * This constant defines the critical level for a breadcrumb.
+     *
+     * @deprecated 2.2.2 Use fatal instead. Will be removed in a later version.
      */
     public const LEVEL_CRITICAL = 'critical';
+
+    /**
+     * This constant defines the fatal level for a breadcrumb.
+     */
+    public const LEVEL_FATAL = 'fatal';
 
     /**
      * This constant defines the list of values allowed to be set as severity
@@ -73,6 +80,7 @@ final class Breadcrumb implements \JsonSerializable
         self::LEVEL_WARNING,
         self::LEVEL_ERROR,
         self::LEVEL_CRITICAL,
+        self::LEVEL_FATAL,
     ];
 
     /**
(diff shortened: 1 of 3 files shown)


src/Breadcrumb.php
```diff
@@ -59,9 +59,9 @@ final class Breadcrumb implements \JsonSerializable
     public const LEVEL_ERROR = 'error';
 
     /**
-     * This constant defines the critical level for a breadcrumb.
+     * This constant defines the fatal level for a breadcrumb.
      */
-    public const LEVEL_CRITICAL = 'critical';
```
ste93cry: Thank you for the contribution. However, this is a breaking change as the constant is `public` and may be used by someone. Please add a new constant that defines the correct value and mark as deprecated this one

TheLevti: Thanks, will do that.

TheLevti: Should the deprecated constant be still in the allowed array?

ste93cry: Yes, if we don't allow the value anymore we would possibly still break existing applications

TheLevti: Done.

src/Breadcrumb.php
```diff
@@ -60,9 +60,16 @@ final class Breadcrumb implements \JsonSerializable
 
     /**
      * This constant defines the critical level for a breadcrumb.
+     *
+     * @deprecated 2.2.2 Use fatal instead. Will be removed in a later version.
```
ste93cry: Please change this description to `@deprecated since version 2.2.2, to be removed in 3.0` so that it's consistent with the rest of the code

Jean85:
```suggestion
     * @deprecated since version 2.2.2, to be removed in 3.0; use fatal instead.
```

Jean85:
Also, please add an entry to the changelog.

Thank you for spotting and fixing this!

PR #905, opened 2019-10-17
Fix regression on default sending behavior of the HttpTransport transport

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.2
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

This PR fixes two regressions introduced in #899 and #885. In particular, with the first PR when an error occurs during the sending of an event we returned `null` instead of the event ID while before we always returned the latter. The second PR instead accidentally changed the default behavior of the `HttpTransport` transport by sending events immediatly instead of waiting the shutdown of the application. This is a deprecated behavior, but the default must not change as it may have a huge impact on performances


src/Transport/HttpTransport.php
```diff
@@ -108,25 +119,26 @@ public function send(Event $event): ?string
         } else {
             try {
                 $promise->wait();
-            } catch (\Exception $exception) {
-                return null;
+            } catch (\Throwable $exception) {
+                return $event->getId();
```
Jean85:
```suggestion
                // ignore transport issues
```

In the future we could create a hook here to log issues.

helhum:
Perfect! That is exactly how I understood that it should be. Thanks a bunch for creating the PR and for helping me out understanding current and desired state!

ste93cry:
I'm not sure we actually want to consider returning `null` when an event fails to be sent a regression though. @HazAT do Unified API gives some hint on how this should be handled? Since the `send` method can return `null` values probably what you did is fine as-is. If that's the case I will revert the change

src/Transport/HttpTransport.php
helhum: See: #907

helhum: optionally the return can be removed and let it fall through to the final return

ste93cry: I reverted this change as thinking again about it it's not a regression since the method could already return `null` values and it makes a lot of sense if something fails to not return the event ID

src/Transport/HttpTransport.php
```diff
@@ -108,24 +119,41 @@ public function send(Event $event): ?string
         } else {
             try {
                 $promise->wait();
-            } catch (\Exception $exception) {
+            } catch (\Throwable $exception) {
                 return null;
```
HazAT: 👍 return `null` in case we are not able to send the event is what we also do in other SDKs.

Issue #886, opened 2019-09-23
Unable to know status of the event


JulienChavee:
I'm trying to migrate from V1 to V2.

In V1 of this SDK, we were able to know if an exception failed to be sent via some methods or by checking if eventID is null.

In V2, it seems impossible to know if the exception was correctly sent to Sentry.

```
$eventID = \Sentry\captureException($myException);
```
This code will always give me an EventID even if the exception wasn't correctly sent.

I checked a bit in the code and I saw that in Transport/HttpTransport::send() you are doing this :

```
 $promise->wait(false);
```
Passing false to this function means that you don't want to check the status the request : "bool $unwrap Whether to return resolved value / throw reason or not"

If I change this value to "true", I can now catch error when the exception wasn't correctly sent (as the method is throwing an Exception).

Now, I would like to know if it is possible (out of the box) to check if the exception is correctly sent? 
And furthermore, is it intended to not warn the user that the exception isn't correctly sent?

ste93cry:
>This code will always give me an EventID even if the exception wasn't correctly sent.

Even though I totally agree that it makes no sense, Unified API works that way and so this behavior is expected

>Now, I would like to know if it is possible (out of the box) to check if the exception is correctly sent?

Out-of-the-box it's not possible at the moment of writing to know if an exception was sent correctly or not, but you can build your own transport that instead of silently catching all errors will do whatever you think is best. See the next part of my reply for what you can expect in the next version to solve or at least improve the handling of this issue

>And furthermore, is it intended to not warn the user that the exception isn't correctly sent?

Complex topic here. We catch any exception that occurs during the sending of the event to avoid breaking the application, but it makes sense to log it somehow, be it with a simple logger or by capturing it and sending it as any other error. What I plan to do in `2.3` is to support a PSR-3 logger in some of the critical parts of the library so that we can log what is happening. Would that be enough?

Jean85:
In general, raising an exception while reporting an error to Sentry is a risky default behavior: it may happen that the error is a network failure, and that would create a cascade of exceptions, since even Sentry server wouldn't be reachable. 

IMHO the current behavior, as a "best effort", is a good one. Both approaches proposed by @ste93cry (custom transport and logger) are good :+1:

JulienChavee:
Many thanks for the answers.

We will probably write our custom transport (it the next idea if there is no solution out-of-the-box).

[...]

PR #907, opened 2019-10-17
Add PSR-3 error logger to ClientBuilder

helhum:
Q | A
-- | --
Branch? | 2.3
Bug fix? | no
New feature? | yes
BC breaks? | no
Deprecations? | no
License | MIT

To be able to log Sentry PHP framework incidents,
a logger is added to the option, which be default logs nothing.

Integrations can set any PSR-3 logger by setting
an instance to the "error_logger" option.

The logger is currently only used to log failures when sending
an event.

### DIFF
--- a/src/Transport/HttpTransport.php
+++ b/src/Transport/HttpTransport.php
@@ -8,9 +8,10 @@ use GuzzleHttp\Promise\FulfilledPromise;
 use GuzzleHttp\Promise\PromiseInterface;
 use Http\Client\HttpAsyncClient as HttpAsyncClientInterface;
 use Http\Message\RequestFactory as RequestFactoryInterface;
-use Http\Promise\Promise as HttpPromiseInterface;
+use Psr\Log\LoggerAwareInterface;
 use Sentry\Event;
 use Sentry\Exception\MissingProjectIdCredentialException;
+use Sentry\Log\OptionalLoggerAwareTrait;
 use Sentry\Options;
 use Sentry\Util\JSON;
 
@@ -20,8 +21,10 @@ use Sentry\Util\JSON;
  *
  * @author Stefano Arlandini <sarlandini@alice.it>
  */
-final class HttpTransport implements TransportInterface, ClosableTransportInterface
+final class HttpTransport implements TransportInterface, ClosableTransportInterface, LoggerAwareInterface
 {
+    use OptionalLoggerAwareTrait;
+
     /**
      * @var Options The Raven client configuration
      */
@@ -38,7 +41,7 @@ final class HttpTransport implements TransportInterface, ClosableTransportInterf
     private $requestFactory;
 
     /**
-     * @var HttpPromiseInterface[] The list of pending promises
+     * @var PendingRequest[] The list of pending requests
      */
     private $pendingRequests = [];
 
@@ -115,13 +118,9 @@ final class HttpTransport implements TransportInterface, ClosableTransportInterf
         $promise = $this->httpClient->sendAsyncRequest($request);
 
         if ($this->delaySendingUntilShutdown) {
-            $this->pendingRequests[] = $promise;
+            $this->pendingRequests[] = new PendingRequest($promise, $event);
         } else {
-            try {
-                $promise->wait();
-            } catch (\Throwable $exception) {
-                return null;
-            }
+            return $this->sendPendingRequest(new PendingRequest($promise, $event));
         }
 
         return $event->getId();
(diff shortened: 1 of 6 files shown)


src/Options.php
```diff
@@ -789,6 +814,7 @@ private function configureOptions(OptionsResolver $resolver): void
         $resolver->setAllowedTypes('in_app_exclude', 'array');
         $resolver->setAllowedTypes('project_root', ['null', 'string']);
         $resolver->setAllowedTypes('logger', 'string');
+        $resolver->setAllowedValues('error_logger', \Closure::fromCallable([$this, 'validateLoggerOption']));
```
Jean85: I think you should be able to use `setAllowedTypes('error_logger', LoggerInterface::class);` here

src/Transport/HttpTransport.php
```diff
@@ -122,10 +129,17 @@ public function send(Event $event): ?string
      */
     private function cleanupPendingRequests(): void
     {
-        while ($promise = array_pop($this->pendingRequests)) {
+        while ($pendingRequest = array_pop($this->pendingRequests)) {
             try {
```
Jean85: This whole `try...catch` block could be extracted into a private method.

tests/Transport/HttpTransportTest.php
```diff
@@ -62,9 +62,11 @@ public function testSendDoesNotDelayExecutionUntilShutdownWhenConfiguredToNotDoI
         $config = new Options(['dsn' => 'http://public@example.com/sentry/1']);
         $transport = new HttpTransport($config, $httpClient, MessageFactoryDiscovery::find(), false);
 
-        $transport->send(new Event());
+        $eventId = $transport->send(new Event());
 
         $this->assertAttributeEmpty('pendingRequests', $transport);
+        // null is returned if sending request failed
```
Jean85: Do we return anything else in the other case?

src/Options.php
helhum: Yes, this works. Thanks. Will update the PR

src/Transport/HttpTransport.php
helhum: True. Tried as little refactoring as possible, but can certainly extract this into a private method

tests/Transport/HttpTransportTest.php
helhum: Yes. Otherwise the event ID is returned. I added this, because of the confusion here: #905 and here: #886

helhum: shall I remove this drive by change and create a separate PR out of this?

ste93cry:
Review apart, I strongly disagree on having the logger as an option as it's a dependency of some classes and not a configuration. Also, it makes more difficult to map the configuration to other formats different from PHP (e.g. YAML/XML in framework's configs). Also, I expect the logger to be used in some other places, e.g. to log when an event gets discarded (and why, e.g. a processor returned `null` or the `beforeSend` callback returned `null`). I think that some analysis should be done before even continuing the code implementation

helhum:
> Review apart, I strongly disagree on having the logger as an option as it's a dependency of some classes and not a configuration
> I think that some analysis should be done before even continuing the code implementation

Thanks for your feedback. Indeed I chose a shortcut here in order to skip the analysis part. I agree with your concerns and am willing to go the long path, but would need some guidance in what direction I should go. Typically loggers are dependencies to classes, which could be resolved with a DI container, but I'm not sure whether this isn't too heavy for this purpose.

When no DI container should be used, then it would be possible to add methods to set and get a logger to `SentrySdk` class, but this also does not seem too nice to me.

Any other ideas?

helhum:
Anyway, this PR is obsolete then

ste93cry:
>Any other ideas?

I was thinking of a more simpler way to set the logger, which is set it through the client builder. A default `NullLogger` logger can be set so that by default it logs nothing. Then we should inject this into all classes that will need it: transport, maybe the client and the the hub

>Anyway, this PR is obsolete then

It's not. Logging what happens is still useful, for example when an event is not sent because the `sendBefore` or a processor returns `null`. Yes, it's possible to implement this in userland but we may use it also for debugging some critical parts of the library. If you still want to continue we can discuss further how to proceed

helhum:
> It's not
> If you still want to continue we can discuss further how to proceed

Alright thanks. I changed the subject of the PR and will look into this later.

helhum:
Here now my suggestion to use LoggerAwareInterface for the classes that intend to use a logger. I added a trait, to ease backwards compatibility as well and use this trait in ClientBuilder and HttpTransport. wdyt?

ste93cry:
I would prefer to have the logger injected in the constructor as it's a dependency. The argument can be `nullable` until `3.0` when we may re-evaluate if it makes sense to keep it like that or make it mandatory (that would not prevent us from using a `NullLogger` anyway). Also I don't really like the `setter` because I don't find a reason to leave open the possibility of changing the logger once the class has been instantiated

ste93cry:
@helhum are you willing to continue the work on this PR?

helhum:
@ste93cry yepp. I'm currently in preparation for holiday season, but in general I will update the code to use constructor injection for the loggers as soon as possible

ste93cry:
Superseded by #989

PR #909, opened 2019-10-22
Add option to whitelist specific paths as 'in app'

### LINKED ISSUE #595: More flexible app path inclusion / exclusion
Currently we have the `app_path` and the `excluded_app_paths` option to define if a stack frame is in the app context or outside.

How we do it in Laravel as a default:

```php
new Raven_Client([
    // ... other options
	'app_path' => '/var/www/app',
    'excluded_app_paths' => [
        '/var/www/app/vendor',
    ],
]);
```

However I would like to do mark some vendored folders as `in_app` for our own packages or packages that we maintain.

-----

I am not sure what the best way to go about this is but simply adding a `included_app_paths` seems silly... but is also the only way I can think of to keep backwards compat and allow the above to work.

```php
new Raven_Client([
    // ... other options
	'app_path' => '/var/www/app',
    'excluded_app_paths' => [
        '/var/www/app/vendor',
    ],
	'included_app_paths' => [
        '/var/www/app/vendor/foo/bar',
		'/var/www/app/vendor/bar/foo',
    ],
]);
```

Also I don't think adding all composer package folders you don't care about to `excluded_app_paths` is a no-go either since that makes a huge list that needs maintaining on every composer update operation or when adding/removing packages.

What are your thoughts?

mfrischbutter:
To bypass the default value for _in_app_exclude_ from getsentry/sentry-laravel, I added the _in_app_whitelist_ option.

The problem is that our own _vendor/packages_ do not appear in the 'in app' stackstrace.

With this option it is possible to whitelist subfolders if you don't want to exclude all folders individually. 

To use this option just add _in_app_whitelist_ to your sentry configuration and define some relative paths as an array.

Resolves #595

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Add `in_app_include` option to whitelist paths that should be marked as part of the app (#909)
 - Fix `Client::captureEvent` not considering the `attach_stacktrace` option (#940)
 - Replace `ramsey/uuid` dependency with `uuid_create` from the PECL [`uuid`](https://pecl.php.net/package/uuid) extension or [`symfony/polyfill-uuid`](https://github.com/symfony/polyfill-uuid) (#937)
 - Deprecate `Scope::setUser` behaviour of replacing user data. (#929)
--- a/src/Options.php
+++ b/src/Options.php
@@ -842,6 +842,8 @@ final class Options
                 return null;
             }
 
+            @trigger_error('The option "project_root" is deprecated. Please use the "in_app_include" option instead.', E_USER_DEPRECATED);
+
             return $this->normalizeAbsolutePath($value);
         });
 
--- a/src/Options.php
+++ b/src/Options.php
@@ -301,6 +301,28 @@ final class Options
     }
 
     /**
+     * Gets the list of paths which has to be identified as in_app.
+     *
+     * @return string[]
+     */
+    public function getInAppIncludedPaths(): array
+    {
+        return $this->options['in_app_include'];
+    }
+
+    /**
+     * Set the list of paths to include in in_app detection.
+     *
+     * @param string[] $paths The list of paths
+     */
+    public function setInAppIncludedPaths(array $paths): void
+    {
+        $options = array_merge($this->options, ['in_app_include' => $paths]);
+
+        $this->options = $this->resolver->resolve($options);
+    }
+
+    /**
      * Gets the project ID number to send to the Sentry server.
      */
     public function getProjectId(): ?string
@@ -769,6 +791,7 @@ final class Options
             },
             'excluded_exceptions' => [],
             'in_app_exclude' => [],
+            'in_app_include' => [],
             'send_default_pii' => false,
             'max_value_length' => 1024,
             'http_proxy' => null,
@@ -786,6 +809,7 @@ final class Options
         $resolver->setAllowedTypes('environment', ['null', 'string']);
         $resolver->setAllowedTypes('excluded_exceptions', 'array');
         $resolver->setAllowedTypes('in_app_exclude', 'array');
+        $resolver->setAllowedTypes('in_app_include', 'array');
         $resolver->setAllowedTypes('project_root', ['null', 'string']);
         $resolver->setAllowedTypes('logger', 'string');
         $resolver->setAllowedTypes('release', ['null', 'string']);
@@ -826,6 +850,10 @@ final class Options
         });
 
         $resolver->setNormalizer('in_app_exclude', function (SymfonyOptions $options, array $value) {
+            return array_map([$this, 'normalizeAbsolutePath'], $value);
+        });
+
+        $resolver->setNormalizer('in_app_include', function (SymfonyOptions $options, array $value) {
             return array_map([$this, 'normalizeAbsolutePath'], $value);
         });
     }
--- a/src/Stacktrace.php
+++ b/src/Stacktrace.php
@@ -148,23 +148,25 @@ class Stacktrace implements \JsonSerializable
             $includedAppPaths = $this->options->getInAppIncludedPaths();
             $absoluteFilePath = @realpath($file) ?: $file;
             $isApplicationFile = 0 === strpos($absoluteFilePath, $this->options->getProjectRoot());
+            $isInApp = false;
 
-            if (!$isApplicationFile) {
+            foreach ($includedAppPaths as $includedPath) {
+                if (0 === mb_strpos($absoluteFilePath, $includedPath)) {
+                    $isInApp = true;
+                    break;
+                }
+            }
+
+            if (!$isApplicationFile && !$isInApp) {
                 $frame->setIsInApp(false);
             } elseif (!empty($excludedAppPaths)) {
                 foreach ($excludedAppPaths as $excludedPath) {
                     if (0 === mb_strpos($absoluteFilePath, $excludedPath)) {
-                        foreach ($includedAppPaths as $includedPath) {
-                            if (false === mb_strpos($absoluteFilePath, $includedPath)) {
-                                $frame->setIsInApp(false);
-                            } elseif (0 === mb_strpos($absoluteFilePath, $includedPath)) {
-                                $frame->setIsInApp(true);
-                                break;
-                            }
-                        }
-
-                        if (empty($includedAppPaths)) {
+                        if ($isInApp) {
+                            $frame->setIsInApp(true);
+                        } elseif (!$isInApp) {
                             $frame->setIsInApp(false);
+                            break;
                         }
 
                         break;
--- a/src/Stacktrace.php
+++ b/src/Stacktrace.php
@@ -145,15 +145,27 @@ class Stacktrace implements \JsonSerializable
 
         if (null !== $this->options->getProjectRoot()) {
             $excludedAppPaths = $this->options->getInAppExcludedPaths();
+            $includedAppPaths = $this->options->getInAppIncludedPaths();
             $absoluteFilePath = @realpath($file) ?: $file;
             $isApplicationFile = 0 === strpos($absoluteFilePath, $this->options->getProjectRoot());
 
             if (!$isApplicationFile) {
                 $frame->setIsInApp(false);
             } elseif (!empty($excludedAppPaths)) {
-                foreach ($excludedAppPaths as $path) {
-                    if (0 === mb_strpos($absoluteFilePath, $path)) {
-                        $frame->setIsInApp(false);
+                foreach ($excludedAppPaths as $excludedPath) {
+                    if (0 === mb_strpos($absoluteFilePath, $excludedPath)) {
+                        foreach ($includedAppPaths as $includedPath) {
+                            if (false === mb_strpos($absoluteFilePath, $includedPath)) {
+                                $frame->setIsInApp(false);
+                            } elseif (0 === mb_strpos($absoluteFilePath, $includedPath)) {
+                                $frame->setIsInApp(true);
+                                break;
+                            }
+                        }
+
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Add in_app_include option
```
Jean85:
```suggestion
- Add `in_app_include` option (#909)
```

src/Stacktrace.php
```diff
@@ -145,14 +145,26 @@ public function addFrame(string $file, int $line, array $backtraceFrame): void
...
+            $isApplicationFile = 0 === strpos(
+                $absoluteFilePath, $this->options->getProjectRoot()
+            );
 
             if ($isApplicationFile && !empty($excludedAppPaths)) {
-                foreach ($excludedAppPaths as $path) {
-                    if (0 === mb_strpos($absoluteFilePath, $path)) {
-                        $frame->setIsInApp(false);
-
+                foreach ($excludedAppPaths as $excludedPath) {
```
Jean85: This loop seems a bit convoluted, we could take the occasion to refactor it and/or extract to a private method.
Also, it seems we're not handling the case of a in-app path that is not under the project root, am I right? @HazAT what's the expected behavior here?

CHANGELOG.md
ste93cry: Even better, I suggest ``Add `in_app_include` option to whitelist paths that should be marked as part of the app (#909)``

src/Options.php
```diff
@@ -287,6 +287,18 @@ public function setInAppExcludedPaths(array $paths): void
         $this->options = $this->resolver->resolve($options);
     }
 
+    public function getInAppIncludedPaths(): array
```
ste93cry: Missing the PHPDoc block

src/Options.php
```diff
@@ -287,6 +287,18 @@ public function setInAppExcludedPaths(array $paths): void
...
+    {
+        return $this->options['in_app_include'];
+    }
+
+    public function setInAppIncludedPaths(array $paths): void
```
ste93cry: Missing the PHPDoc block. Please document both what the function does and the argument with a description

src/Stacktrace.php
```diff
@@ -145,14 +145,26 @@ public function addFrame(string $file, int $line, array $backtraceFrame): void
 
         if (null !== $this->options->getProjectRoot()) {
             $excludedAppPaths = $this->options->getInAppExcludedPaths();
+            $includedAppPaths = $this->options->getInAppIncludedPaths();
             $absoluteFilePath = @realpath($file) ?: $file;
-            $isApplicationFile = 0 === strpos($absoluteFilePath, $this->options->getProjectRoot());
```
ste93cry: Please revert this change and leave everything on one line

src/Stacktrace.php
```diff
@@ -145,14 +145,26 @@ public function addFrame(string $file, int $line, array $backtraceFrame): void
...
+                    if (0 === mb_strpos($absoluteFilePath, $excludedPath)) {
+                        foreach ($includedAppPaths as $includedPath) {
+                            if (false === mb_strpos($absoluteFilePath, $includedPath)) {
+                                $frame->setIsInApp(false);
+                            } elseif (0 === mb_strpos($absoluteFilePath, $includedPath)) {
+                                $frame->setIsInApp(true);
+                                break;
+                            }
+                        }
+                        if (empty($includedAppPaths)) {
```
ste93cry: Please add a blank line above this one

src/Stacktrace.php
```diff
@@ -145,14 +145,26 @@ public function addFrame(string $file, int $line, array $backtraceFrame): void
...
+                            $frame->setIsInApp(false);
+                        }
```
ste93cry: Please add a blank line below this line

tests/OptionsTest.php
```diff
@@ -247,6 +248,29 @@ public function excludedPathProviders()
         ];
     }
 
+    /**
+     * @dataProvider includedPathProviders
+     *
+     * @param string $value
```
ste93cry: Please remove this line and use the typehint on the parameter

tests/OptionsTest.php
```diff
@@ -247,6 +248,29 @@ public function excludedPathProviders()
...
+     * @param string $expected
```
ste93cry: Please remove this line and use the typehint on the parameter

tests/OptionsTest.php
```diff
@@ -247,6 +248,29 @@ public function excludedPathProviders()
...
+     */
+    public function testIncludedAppPathsOverrideExcludedAppPaths($value, $expected)
+    {
+        $configuration = new Options(['in_app_include' => [$value]]);
+
+        $this->assertSame([$expected], $configuration->getInAppIncludedPaths());
+    }
+
+    public function includedPathProviders()
```
ste93cry: Please add the typehint for the return type

tests/StacktraceTest.php
```diff
@@ -143,6 +143,39 @@ public function testAddFrameMarksAsInApp(): void
...
+
+        $stacktrace = new Stacktrace($this->options, $this->serializer, $this->representationSerializer);
+
+        $stacktrace->addFrame('path/to/excluded/path/file.php', 12, ['function' => 'test_function']);
+        $stacktrace->addFrame('path/to/recursive/file.php', 12, ['function' => 'test_function']);
+        $stacktrace->addFrame('path/to/recursive/excluded/path/file.php', 12, ['function' => 'test_function']);
+        $stacktrace->addFrame('path/to/excluded/folder/file.php', 12, ['function' => 'test_function']);
+
+        $frames = $stacktrace->getFrames();
+        // var_dump($frames);
```
ste93cry: This should not be here :smile:

tests/StacktraceTest.php
```diff
@@ -143,6 +143,39 @@ public function testAddFrameMarksAsInApp(): void
...
+
+        $this->options->setInAppExcludedPaths(
+            [
+                'path/to/excluded/folder',
+                'path/to/excluded/path',
+                'path/to/recursive/excluded/path',
+            ]
+        );
+        $this->options->setInAppIncludedPaths(
+            [
```
ste93cry: Please format the code like so:

```php
$this->options->setInAppIncludedPaths([
    ...
])
```

tests/StacktraceTest.php
```diff
@@ -143,6 +143,39 @@ public function testAddFrameMarksAsInApp(): void
         $this->assertTrue($frames[1]->isInApp());
     }
 
+    public function testAddFrameMarksAsInAppOverrideExcludedPaths(): void
+    {
+        $this->options->setProjectRoot('path/to');
+
+        $this->options->setInAppExcludedPaths(
```
ste93cry: Please format the code like so:

```php
$this->options->setInAppExcludedPaths([
    ...
])
```

ste93cry:
Since this is a new feature it must target the `develop` branch, thus the PR needs a rebase

src/Options.php
ste93cry: To clarify, I would like the PHPDoc to document that the returned array is a list of `string`

src/Options.php
ste93cry: To clarify, I would like the PHPDoc to document that the `$paths` argument accepts a list of `string`

src/Options.php
```diff
@@ -273,6 +273,26 @@ public function setInAppExcludedPaths(array $paths): void
...
+     */
+    public function getInAppIncludedPaths(): array
+    {
+        return $this->options['in_app_include'];
+    }
+
+    /**
+     * Set the list of paths to include in in_app detection.
+     *
+     * @param array $paths The list of paths
```
ste93cry: The correct type to document for this parameter is `string[]`

src/Stacktrace.php
mfrischbutter: What am I supposed to do here?

tests/StacktraceTest.php
```diff
@@ -145,6 +145,34 @@ public function testAddFrameMarksAsInApp(): void
...
+                'path/to/excluded/path',
+                'path/to/recursive/', //recursive
+        ]);
+
+        $stacktrace = new Stacktrace($this->options, $this->serializer, $this->representationSerializer);
+
+        $stacktrace->addFrame('path/to/excluded/path/file.php', 12, ['function' => 'test_function']);
+        $stacktrace->addFrame('path/to/recursive/file.php', 12, ['function' => 'test_function']);
+        $stacktrace->addFrame('path/to/recursive/excluded/path/file.php', 12, ['function' => 'test_function']);
+        $stacktrace->addFrame('path/to/excluded/folder/file.php', 12, ['function' => 'test_function']);
```
Jean85: We have one missing edge case here: paths which are included and NOT under the project root.

src/Stacktrace.php
ste93cry: I might be wrong, but I cannot find in the documentation any hint to the `project_root` option, so it may be something that we only have in the PHP SDK and that we may decide to deprecate. Thinking that all the application source code is under just one folder may be a too strong assumption, e.g. in Symfony `2.x` the `AppKernel` class was in a folder different than the one that contained the rest of the source code but technically speaking is part of the application's source. With the `include_app_paths` and the `exclude_app_paths` the user has full flexibility in regards to defining what belongs or not to the application and doesn't need a third option just to restrict to which is the base path those options apply to

Jean85: Still, we cannot change this behavior without breaking BC. I would prefer to deprecate/change that in a different PR.

I would leave this piece like this for now, and refactor it later.

tests/StacktraceTest.php
mfrischbutter: What is the expected behavior here?
Because of the changes in (#911) all paths outside the project root getting excluded, even if i add them into _in_app_include_. 
If a path is specified in _in_app_include_, should it also get marked as _in_app_ if the path is outside of the project root?

ste93cry: According to the documentation `in_app_include` takes precedence over `in_app_excluded`. `in_app` should became an entry of the `in_app_include` option, so the code should be adjusted accordingly to work in the way described before

ste93cry:
@mfrischbutter I did some changes to your PR but it seems that I cannot push them because I don't have permissions for the branch of this PR. Would you be so kind to check if the "Allow edits from maintainers" flag is checked?

mfrischbutter:
@ste93cry 
Yeah, the checkbox was already checked. Maybe i should change anything inside my repository or is the checkbox enough?

Jean85:
I can edit the files from the web interface, so the permissions are ok.

mfrischbutter:
I just rebased onto develop.
Is there anything left to do?

ste93cry:
>Yeah, the checkbox was already checked. Maybe i should change anything inside my repository or is the checkbox enough?

It should be enough, maybe the fact that the PR is not a branch owned by you but by another user (@mogic-le) is the problem as I get 403 as soon as I try to push

>I can edit the files from the web interface, so the permissions are ok.

I just tried and yes, it opens the web editor although I didn't try to save any change

>Is there anything left to do?

There are my changes that I would like to merge with this PR if everything is ok for you. I pushed the commit to my own [fork](https://github.com/ste93cry/sentry-php/tree/add-in_app_whitelist-configuration), if you can cherry-pick it here and give your opinion it would be cool

ste93cry:
@mfrischbutter are you willing to rebase and merge the changes from my fork so that we can proceed with a new (and hopefully final) review of the code?

mfrischbutter:
Sorry it took a while, thanks for your patience.
I pushed your commit @ste93cry.  The code looks very good and all the test passed successfully on the first try. I have not made any changes.

PR #911, opened 2019-10-28
Fix frame inApp detection

grongor:
Paths outside the project root should be considered "not in app"


grongor:
Done. Yeah, I was quite confused by this and I look around for about an hour "what am I doing wrong", and then I traced the code to this place, which didn't really make any sense :D

grongor:
Do you mean `in_app_exclude`? That is covered here: https://github.com/getsentry/sentry-php/pull/911/files#diff-c2c964f09de947d90752fedd3917d9d5R138

PR #914, opened 2019-10-31
Don't autoload functions

ruudk:
To me, this is a real bad pattern. It scatters the global context with functions. Why not let people opt-in if they want this?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,9 +43,6 @@
         "raven/raven": "*"
     },
     "autoload": {
-        "files": [
-            "src/functions.php"
-        ],
         "psr-4" : {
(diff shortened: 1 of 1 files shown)


stayallive:
This does not scatter the global context with functions since they're namespaced under `Sentry`.

You'll need to do `\Sentry\init();` to call them.

So unless we missed something there should be no problem loading these under our namespace for convenience.

ruudk:
Hmm, weird, my PHPStorm just allowed them without. Maybe it automagically imported them.

stayallive:
That is what phpstorm does and is good at :)

PR #919, opened 2019-11-06
Make the `integrations` option accept a callable

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.3
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

This PR aligns the SDK behavior of the `integrations` option with the one present in the JS SDK. Basically, instead of accepting only a list of integrations, a `callable` can now be set that will receive the list of default integrations and will return a (customized) list of integrations. This allows users to pick only certain default integrations without having to maintain in their own project the list of default integrations that must be kept in sync with the one of our library

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -56,7 +56,7 @@ final class Options
     private $resolver;
 
     /**
-     * @var IntegrationInterface[] The list of default integrations
+     * @var IntegrationInterface[]|null The list of default integrations
      */
     private $defaultIntegrations;
 
@@ -68,13 +68,6 @@ final class Options
     public function __construct(array $options = [])
     {
         $this->resolver = new OptionsResolver();
-        $this->defaultIntegrations = [
-            new ExceptionListenerIntegration(),
-            new ErrorListenerIntegration(null, false),
-            new FatalErrorListenerIntegration(),
-            new RequestIntegration(),
-            new TransactionIntegration(),
-        ];
 
         $this->configureOptions($this->resolver);
 
@@ -541,7 +534,7 @@ final class Options
      */
     public function getIntegrations(): array
     {
-        $defaultIntegrations = $this->options['default_integrations'] ? $this->defaultIntegrations : [];
+        $defaultIntegrations = $this->getDefaultIntegrations();
         $userIntegrations = $this->options['integrations'];
         $integrations = [];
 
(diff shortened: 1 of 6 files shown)


src/Options.php
```diff
@@ -558,7 +577,40 @@ public function setIntegrations(array $integrations): void
      */
     public function getIntegrations(): array
     {
-        return $this->options['integrations'];
+        $defaultIntegrations = $this->options['default_integrations'] ? $this->defaultIntegrations : [];
+        $userIntegrations = $this->options['integrations'];
+        $integrations = [];
+
+        if (\is_array($userIntegrations)) {
```
Jean85: I would reverse the check, and do a 
```suggestion
        if (\is_callable($userIntegrations)) {
            return $userIntegrations($defaultIntegrations);
        }
```
to reduce the indentation level of the whole method.

src/Options.php
```diff
@@ -558,7 +577,40 @@ public function setIntegrations(array $integrations): void
...
+            foreach ($userIntegrations as $userIntegration) {
+                $userIntegrationClass = \get_class($userIntegration);
+
+                if (isset($pickedIntegrationsClasses[$userIntegrationClass])) {
+                    continue;
+                }
+
+                $pickedIntegrationsClasses[$userIntegrationClass] = true;
+                $integrations[] = $userIntegration;
+            }
```
Jean85: What is this filter about? I'm not sure I understand this logic, but it seems complicated and founded on some assumption that the end user could miss.

ste93cry: The logic is a bit complicated (I carried it over from the JS SDK) and what it does is:

- It ensures that just one instance of the same integratione exists in the list
- It adds all the missing default integrations if the `enable_integrations` option is set to `true`
- It calls the user-supplied callable by passing the list of default integrations if any

Jean85: I don't like that you're prioritizing the default integrations over the user supplied ones. An user could want to change the behavior of a default integration by supplying it's own, maybe changing just the constructor arguments, and in this way you're losing that.

Also, I think that you could simplify this logic by using an array with FQCN as keys, and do an array merge after.

src/Options.php
```diff
@@ -58,6 +68,13 @@ final class Options
     public function __construct(array $options = [])
     {
         $this->resolver = new OptionsResolver();
+        $this->defaultIntegrations = [
+            new ExceptionListenerIntegration(),
+            new ErrorListenerIntegration(null, false),
+            new FatalErrorListenerIntegration(),
+            new RequestIntegration(),
+            new TransactionIntegration(),
+        ];
```
stayallive: It's probably a micro optimization but shouldn't we move this to a private getter so they're only created (I know most constructors are not doing anything but still) when needed?

Jean85:
Can we please remember to do a PR to the docs once this is released?

PR #920, opened 2019-11-07
Fix for invalid encoding

### LINKED ISSUE #828: Could not encode value into JSON format. Error was: \"Malformed UTF-8 characters, possibly incorrectly encoded
```
{
"message": "Could not encode value into JSON format. Error was: \"Malformed UTF-8 characters, possibly incorrectly encoded\".",
"exception": "Sentry\\Exception\\JsonException",
"file": "/var/www/app/vendor/sentry/sentry/src/Util/JSON.php",
"line": 30,
"trace": [
{
"file": "/var/www/app/vendor/sentry/sentry/src/Transport/HttpTransport.php",
"line": 78,
"function": "encode",
"class": "Sentry\\Util\\JSON",
"type": "::"
},
{
"file": "/var/www/app/vendor/sentry/sentry/src/Client.php",
"line": 105,
"function": "send",
"class": "Sentry\\Transport\\HttpTransport",
"type": "->"
},
{
"file": "/var/www/app/vendor/sentry/sentry/src/Client.php",
"line": 96,
"function": "captureEvent",
"class": "Sentry\\Client",
"type": "->"
},
```

Jean85:
This is a crude attempt to fix #828. 

Using `JSON_INVALID_UTF8_IGNORE` seems the way to go, but it requires PHP 7.2; in 3.0 we could require PHP 7.3 so we could drop the `JSON` class altogether and use `JSON_THROW_ON_ERROR` too.

Under PHP 7.1, I'm using our own serializer, but it's a big messy approach.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,10 @@ cs-fix:
 phpstan:
 	vendor/bin/phpstan analyse
 
-test: cs-fix phpstan
+psalm:
+	vendor/bin/psalm --config=psalm.xml.dist
+
+test: cs-fix phpstan psalm
 	vendor/bin/phpunit --verbose
 
 setup-git:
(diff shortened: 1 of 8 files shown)


Jean85:
As @ste93cry tells me, this doesn't work because the serializer isn't able to serialize objects and returns `Object of type Sentry\Event` :disappointed:

Issue #922, opened 2019-11-16
Add support for Symfony 5


garak:
Symfony 5 is currently in beta and will be stable [in two weeks](https://symfony.com/releases).

ste93cry:
Would you mind opening a PR for this targeting the `master` branch?

garak:
I tried to, but unfortunately it's currently not possible until we get support from `friendsofphp/php-cs-fixer` and `php-http/client-common`.
Opened issues in respective repos too.

Jean85:
Can you cross-link them here?

Jean85:
First one already resolved: https://github.com/php-http/client-common/pull/175 now we need to wait for a tag.

The other one has a pending PR, which has a green build: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4608

pkruithof:
Both issues have been resolved/tagged. So I don't think this issue is blocked anymore. Can we proceed with it?

PR #925, opened 2019-11-22
Symfony 5 compatibility

rvanlaak:
fixes https://github.com/getsentry/sentry-php/issues/922

related to https://github.com/getsentry/sentry-symfony/pull/266

tested by running phpunit locally

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,15 @@
     "name": "sentry/sentry",
     "type": "library",
     "description": "A PHP SDK for Sentry (http://sentry.io)",
-    "keywords": ["sentry", "log", "logging", "error-monitoring", "error-handler", "crash-reporting", "crash-reports"],
+    "keywords": [
+        "sentry",
+        "log",
+        "logging",
+        "error-monitoring",
+        "error-handler",
+        "crash-reporting",
+        "crash-reports"
+    ],
     "homepage": "http://sentry.io",
     "license": "BSD-3-Clause",
     "authors": [
@@ -24,7 +32,7 @@
         "php-http/message": "^1.5",
         "psr/http-message-implementation": "^1.0",
         "ramsey/uuid": "^3.3",
-        "symfony/options-resolver": "^2.7|^3.0|^4.0",
+        "symfony/options-resolver": "^2.7|^3.0|^4.0|^5.0",
         "zendframework/zend-diactoros": "^1.4|^2.0"
     },
     "require-dev": {
@@ -35,7 +43,7 @@
         "phpstan/phpstan": "^0.11",
         "phpstan/phpstan-phpunit": "^0.11",
         "phpunit/phpunit": "^7.5",
-        "symfony/phpunit-bridge": "^4.3",
+        "symfony/phpunit-bridge": "^4.3|^5.0",
         "vimeo/psalm": "^3.4"
     },
     "suggest": {
@@ -49,8 +57,8 @@
         "files": [
             "src/functions.php"
         ],
-        "psr-4" : {
-            "Sentry\\" : "src/"
+        "psr-4": {
+            "Sentry\\": "src/"
         }
     },
     "autoload-dev": {
@@ -75,6 +83,7 @@
     "config": {
         "sort-packages": true
     },
+    "prefer-stable": true,
     "extra": {
         "branch-alias": {
             "dev-develop": "2.2-dev"


ste93cry:
This should close #922 and I don't expect any issue with Symfony 5 anyway, but let's wait until we can really test it in the CI because right now the `highest` jobs are using the components at version `4.4`

Jean85:
Symfony 5 got released (live on stage at SymfonyCon, I've seen it! :smile: )

I'll restart the CI job...

TomasVotruba:
@Jean85 That's not a blocker for this package. If I don't use php-cs-fixer, I could already use these

ste93cry:
The problem is not that it's blocking the installation for someone but that we cannot test the compatibility with Symfony 5 as the test suite cannot be executed to prove that it works, even though as I said I'm sure that it would

TomasVotruba:
Yes, those are 2 different problems.

TomasVotruba:
Ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4608

rvanlaak:
What about removing CS Fixer as dependency, and start using https://github.com/jakzal/phpqa or a Github action?

Big-Shark:
@Jean85 PHP-CS-Fixer updated, and now we can restart checkers?

ste93cry:
@rvanlaak can you please force-push? I can restart the Travis CI build but not AppVeyor's one

rvanlaak:
Force push ✔️ checks are still green

Jean85:
It seems that CS-Fixer has the compat on the 2.15.4, but Composer is preferring 2.16.0, which isn't compatible. Also, PHPStan is getting in the way too.

PHPStan will be released as a PHAR from 0.12 so it will auto-resolve, CS-Fixer needs just a tag (the compat is already merged in the 2.16 branch).

rvanlaak:
There is no need to share dependencies with an independent tool like CS-Fixer. What about removing CS-Fixer and running it via the phpqa docker image?

What do you think about merging this PR together with the Symfonyt 5 changes? https://github.com/scribbr/sentry-php/pull/1

ste93cry:
>What about removing CS-Fixer and running it via the phpqa docker image?

There is no reason to use Docker to run in the CI such tools as using the PHARs should be enough. PHPStan 0.12 will be released December 3rd, so let's wait until that date and then let's see what's the status of the dependencies and act accordingly

DavidBadura:
Yeah, tools like php-cs-fixer shouldn't install via composer as dev dependency. You can download it in the CI as phar archive and use that. Otherwise, you will always have the problem that the dependencies do not fit. Either the lib can not be tested with the latest version or the tools can not be updated.

ste93cry:
>Yeah, tools like php-cs-fixer shouldn't install via composer as dev dependency

Sometimes it's convenient to install it that way as you don't have to deal with downloading the PHARs, and sometimes it's necessary as you have to load extensions and it wouldn't be possible by using PHARs: one such example is PHPStan, that's way I'm curious of understanding how `0.12` solves the issue

rvanlaak:
For now it is CS Fixer that is blocking this PR 👍  ... that is why somehow removing it as dependency would be great.

```
$ composer why-not symfony/options-resolver:^5
friendsofphp/php-cs-fixer  v2.16.0     requires  symfony/options-resolver (^3.0 || ^4.0)
```
https://github.com/getsentry/sentry-php/pull/925#pullrequestreview-321716954

If the phar is the way to go, I will make another PR for that.

garak:
You can just require `"friendsofphp/php-cs-fixer": "^2.17@dev"`until a new stable tag will be released.
I tried locally and it works, installing Symfony 5 components (and with green tests)

ste93cry:
>If the phar is the way to go, I will make another PR for that.

No need for another PR, we can use this one :wink: but again, I want to wait PHPStan `0.12` so that we don't have to make one more PR to update how we fetch its PHAR

>You can just require "friendsofphp/php-cs-fixer": "^2.17@dev"until a new stable tag will be released.

It's not a great idea to use a non-stable dependency, isn't it?

garak:
> > You can just require "friendsofphp/php-cs-fixer": "^2.17@dev"until a new stable tag will be released.
> 
> It's not a great idea to use a non-stable dependency, isn't it?

You can do it only on your non-stable release, then everything matches

Jean85:
> You can do it only on your non-stable release, then everything matches

That's not a solution, we want to release a patch and declare compat with SF5. Surely making us indipendent from dev tooling is the best approach, but I agree that a Docker container is a bit overkill here; also, we need to take into account local DX, otherwise noone will fix CS locally and we will get a lot of errors in future PRs.

ste93cry:
PHP-CS-Fixer has been released, @rvanlaak please can you trigger the build one more time?

ste93cry:
LGTM then :+1:

rvanlaak:
Did force-push to rerun CI ✅

ste93cry:
Thank you, even though it was not needed as @Jean85 closed and reopened the PR to forcefully trigger the CI. Everything's fine, LGTM

PR #928, opened 2019-12-06
Add the IgnoreErrorsIntegration integration and deprecate the excluded_exceptions option

### LINKED ISSUE #924: Excluded exceptions are being ignored when monolog error handler is enabled
I've recently enabled the monolog error handler on a Symfony project and realized that the excluded exceptions I configured are being reported.

After doing some digging, I've found that `Sentry\Monolog\Handler` relies on the `captureEvent` method which does not check if the payload contains an exception and if that exception should be ignored. I was wondering if it is an intended beheavior? 

If not, I would be happy to submit a PR. The idea would be to use either the `captureException` or the `captureEvent` method whether the payload has an `exception` key set. 

```php
// Sentry\Monolog\Handler.php

{
    ....

        $payload['exception'] = $record['context']['exception'];
    }

    $this->hub->withScope(function (Scope $scope) use ($record, $payload): void {
        ...

        if (array_key_exists('exception', $payload)) {
            $this->hub->captureException($payload['exception']);
        } else {
            $this->hub->captureEvent($payload);
        }
    });
}
```

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.3
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| License       | MIT

Closes #924. I decided that I didn't want to use the `$payload` argument of the `capture*` methods as an hacky way to pass options that configured the behavior of the method. Instead, like in the [JS SDK](https://github.com/getsentry/sentry-javascript/blob/c565329fac4876856da106e36790b0d975029a68/packages/core/src/integrations/inboundfilters.ts) I added a new integration that can be further expanded in the future and that will centralize all the logic to decide whether an event should be captured or not according to a set of criteria that must match with the event data. Being an integration means that the configuration now applies to all those methods and not only to the `captureException` one, so we cannot enable it by default until `3.0`

### DIFF
--- a/src/Integration/InboundFiltersIntegration.php
+++ b/src/Integration/InboundFiltersIntegration.php
@@ -1,0 +1,103 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\Integration;
+
+use Sentry\Event;
+use Sentry\SentrySdk;
+use Sentry\State\Scope;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * This integration decides whether an event should not be captured according
+ * to a series of options that must match with its data.
+ *
+ * @author Stefano Arlandini <sarlandini@alice.it>
+ */
+final class InboundFiltersIntegration implements IntegrationInterface
+{
+    /**
+     * @var array<string, mixed> The options
+     */
+    private $options;
+
+    /**
+     * Creates a new instance of this integration and configures it with the
+     * given options.
+     *
+     * @param array<string, mixed> $options The options
+     *
+     * @psalm-param array{
+     *     ignore_exceptions?: bool
+     * } $options
+     */
+    public function __construct(array $options = [])
+    {
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults([
+            'ignore_exceptions' => [],
+        ]);
+
+        $resolver->setAllowedTypes('ignore_exceptions', ['array']);
+
+        $this->options = $resolver->resolve($options);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setupOnce(): void
+    {
+        Scope::addGlobalEventProcessor(function (Event $event): ?Event {
+            $currentHub = SentrySdk::getCurrentHub();
+            $integration = $currentHub->getIntegration(self::class);
+
+            if (null !== $integration && $integration->shouldDropEvent($event, $integration->options)) {
+                return null;
+            }
+
+            return $event;
+        });
+    }
+
+    /**
+     * Checks whether the given event should be dropped according to the options
+     * that configures the current instance of this integration.
+     *
+     * @param Event                $event   The event to check
+     * @param array<string, mixed> $options The options of the integration
+     */
+    private function shouldDropEvent(Event $event, array $options): bool
+    {
+        if ($this->isIgnoredException($event, $options)) {
+            return true;
+        }
+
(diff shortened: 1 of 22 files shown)


src/Integration/InboundFiltersIntegration.php
```diff
@@ -0,0 +1,103 @@
...
+    private function isIgnoredException(Event $event, array $options): bool
+    {
+        $exceptions = $event->getExceptions();
+
+        if (empty($exceptions) || !isset($exceptions[0]['type'])) {
+            return false;
+        }
+
+        foreach ($options['ignore_exceptions'] as $ignoredException) {
+            if (is_a($exceptions[0]['type'], $ignoredException, true)) {
```
Jean85: Why not use `instanceof`

ste93cry: Because `['type']` contains a string and not an object, so I cannot use the `instanceof` operator

ste93cry:
I see that the Go SDK has a similar extension named `IgnoreErrorsIntegration` and I wonder if I should use the same name as it seems more appropriate and also clearer. Any opinion?

Jean85:
:+1: for that name, I like it

VincentLanglet:
So we'll have to rename `sentry.options.excluded_exceptions` by `sentry.options.ignore_exceptions` ?

ste93cry:
No, you can leave your code as is and it will keep working until `3.0`. However, the `excluded_exceptions` option has been deprecated in favour of enabling the `IgnoreErrorsIntegration` that will be applied every time an event is captured, so it will work also when the Monolog handler is involved

VincentLanglet:
> so it will work also when the Monolog handler is involved

The `excluded_Exceptions` wasn't working for the Monolog handler (https://github.com/getsentry/sentry-symfony/issues/273)

As soon as the 2.3 version is released, which of the following sentences is true ?
- The excluded_exceptions will work also for the Monolog handler
- We'll have to enabling the IgnoreErrorsIntegration if we want to ignore exceptions for the Monolog handler (But how ? With a `ignore_exceptions` ?)
- Some development is necessary for sentry-symfony in order to handle the IgnoreErrorsIntegration. 

Thanks !

ste93cry:
>The `excluded_Exceptions` wasn't working for the Monolog handler

This is the reason of this PR, to make such option work in all cases regardless of who is calling the `Client::capture*` methods. To do it, an integration has been created and once enabled it will affect **all** those methods (instead the `excluded_exceptions` was not affecting the `Client::captureEvent` method that was used by the Monolog handler): this is the reason we cannot enable it out-of-the-box until `3.0` or we risk of breaking existing applications. So, at the end of the story:

- The `excluded_exceptions` option is deprecated as of version `2.3` and will be removed in `3.0`
- The `IgnoreErrorsIntegration` will take the place of the old option. To use it you will need to use the `integrations` option to instantiate the class and pass to its constructor the values that you did pass to the `excluded_exceptions`. You can then stop using the deprecated option and everything should work as before except that this new integration will affect all the `Client::capture*` methods (and not only the `captureError` and `captureException`) and consequently the Monolog handler

```diff
Sentry\init([
-    excluded_exceptions => [
-        Exception::class,
-    ],
+    integrations => [
+        new Sentry\Integration\IgnoreErrorsIntegration([
+            Exception::class,
+        ]),
+    ],
]);
```

PR #930, opened 2019-12-09
Fix PHP 7.4 deprecations

Jean85:
We had the deprecation handler on a too-low setting, so we missed deprecations under PHP 7.4: https://travis-ci.org/getsentry/sentry-php/jobs/622409178#L330-L345

This fixes them.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -6,6 +6,10 @@
          colors="true"
          bootstrap="tests/bootstrap.php"
 >
+    <php>
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0" />
+    </php>
+
     <testsuites>
         <testsuite name="Sentry for PHP">
             <directory>tests</directory>
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -6,10 +6,6 @@
          colors="true"
          bootstrap="tests/bootstrap.php"
 >
-    <php>
-        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0" />
-    </php>
-
     <testsuites>
         <testsuite name="Sentry for PHP">
             <directory>tests</directory>
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -3,10 +3,12 @@ build: false
 clone_depth: 2
 clone_folder: c:\projects\sentry-php
 skip_branch_with_pr: true
+image: Visual Studio 2019
 branches:
     only:
         - master
         - develop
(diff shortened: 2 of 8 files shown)


ste93cry:
Can you please also update the AppVeyor config file to use `Visual Studio 2019` as image and test against PHP `7.4`? furthermore please also update the Composer version to `1.9.1`

phpunit.xml.dist
```diff
@@ -6,10 +6,6 @@
          colors="true"
          bootstrap="tests/bootstrap.php"
 >
-    <php>
-        <env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=0" />
```
ste93cry: This is not a good idea as the build will break as soon as a third party library deprecates something triggered by our own code. This is why this line was added. Please leave it as is or check the [documentation](https://symfony.com/doc/current/components/phpunit_bridge.html#internal-deprecations) to see if there is a more appropriate configuration that we can use

Jean85: We can try `max[direct]=0`, but I have to check if it detects this kind of issues.

ste93cry: `max[self]` worked as far as I see from build logs, it was reporting 3 legacy deprecations, thus I don’t get what you are trying to fix by changing/removing this setting

Jean85: It didn't make the build fail, hence I had to do this late PR.

ste93cry: It did’t make the build fail on purpose, as doing it would move the burden of fixing the issue to the next person that will contribute a change, and if no one do it we will never know anyway that we are generating deprecation notices...

Jean85: IMHO we should let the build fail instead. We can always open a parallel PR to fix the deprecations, and ignore them as a last resort when merging.

ste93cry: As I wrote above until someone opens a PR and gets the failing build in the face we won’t know that something is generating deprecation notices, so I don’t really see how removing this line will help us to not ignore/miss such notices. We should periodically run a build just to check that in the meanwhile another library didn’t deprecate something we rely on, so this change is kinda useless in my opinion. The fact is that in the rush of releasing a version compatible with PHP 7.4 we forgot to check if any deprecation was there. To clarify, if no change was required to release such version we would have missed the deprecations too

Pnoexz:
Any estimated date for this to be released? These deprecation warnings are getting logged on Sentry and it's burning through my events.

ste93cry:
Don't take these words as written on stone, but probably tomorrow morning

PR #931, opened 2019-12-10
Allows merging user context on Scope::setUser()

### LINKED ISSUE #929: Allow merging user data on Scope
## Summary

Currently [the Scope only allows us to replace user data](https://github.com/getsentry/sentry-php/blob/24aa98be497ff089e2d58807ee3db0bb6a021fc2/src/State/Scope.php#L139) with the setUser() method, and because of this it's not possible to gradually include more specific user data. It would be good to merge that data.

## Motivation

We don't always get all the user info from the same place, so it would be useful to gradually populate that data. For instance: adding the IP before the user is authenticated, then adding username and email afterwards.

## Additional Context

The motivation for this request is the fact that the [RequestListener](https://github.com/getsentry/sentry-symfony/blob/0e12fee8e4232e7cc45cf177d8d7a662175fec24/src/EventListener/RequestListener.php) from [getsentry/sentry-symfony](https://github.com/getsentry/sentry-symfony) is overwriting my user data. As it does add some useful data, it would be best if it was possible to keep the user data it provided while also adding some data myself without replacing it.

brunowowk:
Closes #929

- Deprecate `Scope::setUser` behaviour of replacing user data. (#929)
- Add the $merge parameter on `Scope::setUser` to allow merging user context. (#929)
- Add the `Scope::clearUser` method to clear user context. (#929)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 - Deprecate `Scope::setUser` behaviour of replacing user data. (#929)
-- Add the $merge parameter on `Scope::setUser` to allow merging user context. (#929)
+- Add the `$merge` parameter on `Scope::setUser` to allow merging user context. (#929)
 - Add the `Scope::clearUser` method to clear user context. (#929)
 
 ## 2.2.4 (2019-11-04)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- Deprecate `Scope::setUser` behaviour of replacing user data. (#929)
+- Add the $merge parameter on `Scope::setUser` to allow merging user context. (#929)
+- Add the `Scope::clearUser` method to clear user context. (#929)
+
 ## 2.2.4 (2019-11-04)
 
 - Suggest installing Monolog to send log messages directly to Sentry (#908)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@
 
 - Deprecate `Scope::setUser` behaviour of replacing user data. (#929)
 - Add the `$merge` parameter on `Scope::setUser` to allow merging user context. (#929)
-- Add the `Scope::removeUser` method to remove specified data from user context. (#929)
 - Fix remaining PHP 7.4 deprecations (#930)
 - Make the `integrations` option accept a `callable` that will receive the list of default integrations and returns a customized list (#919)
 - Add the `IgnoreErrorsIntegration` integration to deprecate and replace the `exclude_exceptions` option (#928)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Deprecate `Scope::setUser` behaviour of replacing user data. (#929)
 - Add the `$merge` parameter on `Scope::setUser` to allow merging user context. (#929)
-- Add the `Scope::clearUser` method to clear user context. (#929)
+- Add the `Scope::removeUser` method to remove specified data from user context. (#929)
 - Fix remaining PHP 7.4 deprecations (#930)
 - Make the `integrations` option accept a `callable` that will receive the list of default integrations and returns a customized list (#919)
 - Add the `IgnoreErrorsIntegration` integration to deprecate and replace the `exclude_exceptions` option (#928)
--- a/src/State/Scope.php
+++ b/src/State/Scope.php
@@ -155,7 +155,7 @@ final class Scope
      *
      * @return $this
      */
-    public function clearUser(): self
+    public function removeUser(): self
     {
         $this->user->clear();
 
--- a/src/State/Scope.php
+++ b/src/State/Scope.php
@@ -131,12 +131,34 @@ final class Scope
      * Sets the given data in the user context.
      *
      * @param array $data The data
+     * @param bool $merge If true, $data will be merged into user context instead of replaciong it
      *
      * @return $this
      */
-    public function setUser(array $data): self
+    public function setUser(array $data, bool $merge = false): self
     {
+        if ($merge) {
+            $this->user->merge($data);
+
+            return $this;
+        }
+
+        @trigger_error('Replacing user context on setUser is deprecated since sentry-php 2.3 and will be changed to a merge on 3.0. '
+             . 'For now you can use $merge = true to merge instead of replacing it.', E_USER_DEPRECATED);
+
         $this->user->replaceData($data);
+
+        return $this;
+    }
+
+    /**
+     * Clears all user context data.
+     *
+     * @return $this
+     */
+    public function clearUser(): self
+    {
+        $this->user->clear();
 
         return $this;
     }
--- a/src/State/Scope.php
+++ b/src/State/Scope.php
@@ -131,7 +131,7 @@ final class Scope
      * Sets the given data in the user context.
      *
      * @param array $data The data
-     * @param bool $merge If true, $data will be merged into user context instead of replaciong it
+     * @param bool $merge If true, $data will be merged into user context instead of replacing it
      *
      * @return $this
      */
--- a/src/State/Scope.php
+++ b/src/State/Scope.php
@@ -151,20 +151,6 @@ final class Scope
     }
 
     /**
-     * Removes specified data from user context.
-     *
-     * @param string $key The key of the information to remove
-     *
-     * @return $this
-     */
-    public function removeUser(string $key): self
-    {
-        $this->user->offsetUnset($key);
-
-        return $this;
-    }
-
-    /**
      * Sets the list of strings used to dictate the deduplication of this event.
      *
      * @param string[] $fingerprint The fingerprint values
--- a/src/State/Scope.php
+++ b/src/State/Scope.php
@@ -143,8 +143,7 @@ final class Scope
             return $this;
         }
 
-        @trigger_error('Replacing user context on setUser is deprecated since sentry-php 2.3 and will be changed to a merge on 3.0. '
-             . 'For now you can use $merge = true to merge instead of replacing it.', E_USER_DEPRECATED);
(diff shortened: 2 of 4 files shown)


tests/EventTest.php
```diff
@@ -91,9 +91,11 @@ public function testEventIsGeneratedWithUniqueIdentifier(): void
 
     public function testToArray(): void
     {
+        $event = new Event();
+
         $expected = [
             'event_id' => str_replace('-', '', static::GENERATED_UUID[0]),
-            'timestamp' => gmdate('Y-m-d\TH:i:s\Z'),
+            'timestamp' => $event->getTimestamp(),
```
brunowowk: Changed to avoid this occasional failure: https://ci.appveyor.com/project/sentry/sentry-php/builds/29441367/job/hnjqwl6587bv6w4v

tests/State/ScopeTest.php
```diff
@@ -29,57 +29,68 @@ public function testSetTag(): void
         $this->assertSame(['foo' => 'bar', 'bar' => 'baz'], $event->getTagsContext()->toArray());
     }
 
-    public function setTags(): void
```
brunowowk: `setTags` was probably pasted as reference when writing the tests and accidentally committed. Wasn't being used, so I removed it.

src/State/Scope.php
```diff
@@ -130,17 +130,39 @@ public function setExtras(array $extras): self
...
-    public function setUser(array $data): self
+    public function setUser(array $data, bool $merge = false): self
     {
+        if ($merge) {
+            $this->user->merge($data);
+
+            return $this;
+        }
+
+        @trigger_error('Replacing user context on setUser is deprecated since sentry-php 2.3 and will be changed to a merge on 3.0. '
```
ste93cry: Please make this function call one-line and replace the text with ``Replacing the data is deprecated since version 2.3 and will stop working from version 3.0. Set the second argument to `true` to merge the data instead.``

src/State/Scope.php
```diff
@@ -130,17 +130,39 @@ public function setExtras(array $extras): self
...
 
         return $this;
     }
 
+    /**
+     * Clears all user context data.
+     *
+     * @return $this
+     */
+    public function clearUser(): self
```
ste93cry: Although having the `clear` word in the method's name is fine for me, Unified API specs talks about a `removeUser` method so we should stick with it. But if @HazAT is fine with changing such specs (I don't know if any of the other SDKs implement this method) we may leave it as is or rename it to `clearUserContext` (but this would be inconsistent with the rest of the methods, which I would have named with the `Context` suffix by the way)

tests/State/ScopeTest.php
```diff
@@ -29,20 +29,15 @@ public function testSetTag(): void
...
+    public function testSetTags(): void
     {
         $scope = new Scope();
         $scope->setTags(['foo' => 'bar']);
-
```
ste93cry: Please revert this and the subsequent changes on the blank line as they improve the readability and simplifies the code review

tests/State/ScopeTest.php
```diff
@@ -82,7 +77,12 @@ public function testSetExtras(): void
         $this->assertSame(['foo' => 'bar', 'bar' => 'baz'], $event->getExtraContext()->toArray());
     }
 
-    public function testSetUser(): void
+    /**
+     * @group legacy
+     *
+     * @expectedDeprecation Replacing user context on setUser is deprecated since sentry-php 2.3 and will be changed to a merge on 3.0. For now you can use $merge = true to merge instead of replacing it.
+     */
+    public function testDeprecatedSetUser(): void
```
ste93cry: Please rename this method to `testSetUserThrowsDeprecation`

tests/State/ScopeTest.php
```diff
@@ -237,7 +282,8 @@ public function testClear(): void
         $scope->setFingerprint(['foo']);
         $scope->setExtras(['foo' => 'bar']);
         $scope->setTags(['bar' => 'foo']);
-        $scope->setUser(['foobar' => 'barfoo']);
+        $merge = true;
```
ste93cry: Instead of using a variabile just pass `true` as value of the 2nd argument of the `setUser` function

tests/State/ScopeTest.php
```diff
@@ -273,7 +319,8 @@ public function testApplyToEvent(): void
         $scope->addBreadcrumb($breadcrumb);
         $scope->setTag('foo', 'bar');
         $scope->setExtra('bar', 'foo');
-        $scope->setUser(['foo' => 'baz']);
+        $merge = true;
```
ste93cry: Instead of using a variabile just pass `true` as value of the 2nd argument of the `setUser` function

tests/EventTest.php
ste93cry: I agree that it's a pity when it happens, but we should avoid using the getter because if it returns the date in the wrong format (it's just an example) we won't catch it as we are basically comparing that what the getter returns is what we expect. Yes, in a lot of other tests we do the same but in this specific case the format of the date is what we wanna test other than the value itself of course

src/State/Scope.php
brunowowk: 0f2e231

src/State/Scope.php
brunowowk: I named it that way to be consistent with `Scope::clear()` and `Scope::clearBreadcrumbs()`, but yeah, we should go with the unified spec. Renamed it to removeUser(): 304803c

(I find the behaviour of clearing the context on setUser() with an empty array to be pretty cryptic)

tests/State/ScopeTest.php
brunowowk: 8e26743

tests/State/ScopeTest.php
brunowowk: c3dfd42

brunowowk: c3dfd42

tests/State/ScopeTest.php
brunowowk: 04406b5

src/State/Scope.php
brunowowk: whoops 3482376

tests/EventTest.php
brunowowk: bf6d717

src/State/Scope.php
ste93cry: I had a look at the other SDKs and I read again the documentation and I think I misunderstood what was written there. Basically, the documentation suggests as a way to remove data from a context the `remove*` method. However this method should not clear the whole context, but just remove a given item from it, so can you please update the code to selectively unset a certain key of the context?

ste93cry:
I just noticed that this PR is pointing to the wrong branch. Since it’s not a bug fix it should target `develop`. @brunowowk would you mind to switch target branch and in case rebase?

src/State/Scope.php
brunowowk: Ref: https://docs.sentry.io/development/sdk-dev/unified-api/#scope

> scope.set_user(user): Shallow merges user configuration (email, username, …). Removing user data is SDK-defined, either with a remove_user function or by passing nothing as data.

Yeah, it's not absolutelly clear if they meant for it to remove all context or just part of the data, but look at this other spec:
> scope.set_extra(key, value): Sets the extra key to an arbitrary value, overwriting a potential previous value. Removing a key is SDK-defined, either with a remove_extra function or by passing nothing as data.

Comparing with this spec, what I get is that on methods which set a single key of context, remove* should remove a single key (worded: "Removing **a key** is SDK-defined". The specifically mention "**a key**"). On the other hand, when having methods that set the whole context at once, without a key, the remove* method should clear the whole context (worded: "Removing **user data** is SDK-defined", as opposed to "**a key**"). 

WDYT?

brunowowk: We could also open an issue on the docs asking them to clarify on that topic and go without a `removeUser` method for now. I can't actually see much use for it anyways, as the scope should be entirely cleared when changing context on the application.

ste93cry: I think that you are right in saying that generally speaking the specs should be more clear on how these methods should act, because in certain parts they are and in others they are not. I suggest to go with the `removeUser` method that unsets a given key and not the whole context, as it may turn useful for example if something out of the control of the user sets some information on a parent scope and without such method you would have no way to remove it because inner scopes just inherit all data. We can add a `clearUser` context if we see that people need it

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- Deprecate `Scope::setUser` behaviour of replacing user data. (#929)
+- Add the $merge parameter on `Scope::setUser` to allow merging user context. (#929)
```
stayallive:
```suggestion
- Add the `$merge` parameter on `Scope::setUser` to allow merging user context. (#929)
```

src/State/Scope.php
brunowowk: Makes sense, there you go: 744a040

brunowowk: We can revisit this after getting feedback on https://github.com/getsentry/sentry-docs/issues/1407 ✌

ste93cry: I'm sorry to change one more time my mind, but since the discussion in the referenced issue may last for a long time, let's drop the `removeUser` method so that we can merge this PR as-is and if something will change in the future regarding this topic we will update the SDK accordingly

brunowowk: there you go 031baaf

pilif:
> Add the Scope::clearUser method to clear user context. (#929)

this seems to not have been part of the PR - what is the recommended way now to set a user to specific value without merging?

ste93cry:
If you mean how you can remove an item from the root of the user context, there is no way to do it now. If you need to push some data and clear it after some operation finished the recommended way is to use a new scope with `\Sentry\withScope`

Cyrills:
So, how I need to set up user data when I try to initialize Sentry manualy in my project ?

ste93cry:
It depends on what you are trying to do. Please take a look at the [documentation](https://docs.sentry.io/platforms/php/), and if something is not clear feel free to open a new issue and/or write to the support

PR #934, opened 2019-12-15
Fix error thrown when serializing an event containing invalid UTF-8 strings

### LINKED ISSUE #828: Could not encode value into JSON format. Error was: \"Malformed UTF-8 characters, possibly incorrectly encoded

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.2
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Fixes #828. Originally I wanted to continue on @Jean85's PR but since he closed it and I force pushed to start over the work I could not, so here we go with a brand new. I abandoned the idea of using the serializer and instead hard-coded the sanitization of strings into the `JSON` util class. For PHP 7.2 or greater I don't expect to have to run the custom code because of the `JSON_INVALID_UTF8_SUBSTITUTE` option, so what I did is trying to serialize the data one first time and if it fails because of invalid UTF-8 characters I loop over the data where possible and try to convert all strings to UTF-8 before trying once more the encoding to JSON. If it fails again I throw the exception as it was done before. Due to this double serialization happening in case of PHP < 7.2 and an invalid UTF-8 string, there is a small performance penalty but I suppose it's acceptable

### DIFF
--- a/src/Util/JSON.php
+++ b/src/Util/JSON.php
@@ -18,13 +18,29 @@ final class JSON
      *
      * @param mixed $data The data to encode
      *
+     * @return mixed
+     *
      * @throws JsonException If the encoding failed
      */
-    public static function encode($data): string
+    public static function encode($data)
     {
-        $encodedData = json_encode($data, JSON_UNESCAPED_UNICODE);
+        $options = JSON_UNESCAPED_UNICODE;
 
-        if (JSON_ERROR_NONE !== json_last_error() || false === $encodedData) {
+        if (\PHP_VERSION_ID >= 70200) {
+            $options |= JSON_INVALID_UTF8_SUBSTITUTE;
+        }
+
+        $encodedData = json_encode($data, $options);
+
+        // This should never happen on PHP >= 7.2 as the substitution of invalid
+        // UTF-8 characters is done internally. On lower versions instead, we
+        // try to sanitize the data ourselves before retrying encoding. If it
+        // fails again we throw an exception as usual.
+        if (JSON_ERROR_UTF8 === json_last_error()) {
+            $encodedData = json_encode(self::sanitizeData($data, 256), $options);
+        }
+
+        if (JSON_ERROR_NONE !== json_last_error()) {
             throw new JsonException(sprintf('Could not encode value into JSON format. Error was: "%s".', json_last_error_msg()));
         }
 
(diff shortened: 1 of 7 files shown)


src/Util/JSON.php
```diff
@@ -50,4 +66,78 @@ public static function decode(string $data)
...
+     * @param int   $maxDepth The maximum depth to walk through `$data`
+     *
+     * @return mixed
+     *
+     * @throws JsonException
+     */
+    private static function sanitizeData($data, int $maxDepth)
+    {
+        if ($maxDepth < 0) {
+            throw new JsonException('Max depth limit of data to be sanitized reached.');
```
Jean85: I would not throw an exception here, but stop silently and truncate the serialization, like in our serializer.

src/Util/JSON.php
```diff
@@ -18,13 +18,29 @@ final class JSON
...
+
+        $encodedData = json_encode($data, $options);
 
-        if (JSON_ERROR_NONE !== json_last_error() || false === $encodedData) {
+        // This should never happen on PHP >= 7.2 as the substitution of invalid
+        // UTF-8 characters is done internally. On lower versions instead, we
+        // try to sanitize the data ourselves before retrying encoding. If it
+        // fails again we throw an exception as usual.
+        if (JSON_ERROR_UTF8 === json_last_error()) {
+            $encodedData = json_encode(self::sanitizeData($data, 256), $options);
```
Jean85: I would set a lower default limit as default depth: if we encounter a loop in the serialization it would make a mess!

src/Util/JSON.php
```diff
@@ -50,4 +66,78 @@ public static function decode(string $data)
...
+    {
+        $previousSubstituteCharacter = mb_substitute_character();
+        $encoding = mb_detect_encoding($value, mb_detect_order(), true);
+
+        mb_substitute_character(0xfffd);
+
+        if (false === $encoding) {
+            $value = mb_convert_encoding($value, 'UTF-8', 'UTF-8');
+        } else {
+            $value = mb_convert_encoding($value, 'UTF-8', $encoding);
```
Jean85: You could simplify this whole if with 
```suggestion
            $value = mb_convert_encoding($value, 'UTF-8', $encoding ?: 'UTF-8');
```

src/Util/JSON.php
ste93cry: I took this limit from the XDebug default configuration for the maximum level of nested function calls. We can lower it of course, but I'm not sure it's worth it because regardless of the value the result is always the same: once the limit is reached the serialization stops.

src/Util/JSON.php
ste93cry: I'm not convinced that we should remove this exception: `json_encode` will throw an error anyway if the max depth limit (3rd argument, set to `512` by default) is reached. Plus, this check also ensures that the function is always called with a valid argument

ste93cry:
As always I forget it, I will never learn :flushed:... Done, just one remaining conversation open though

PR #937, opened 2019-12-18
Use PECL uuid function

jderusse:
This PR replaces the RamserUUID object `uuid_create`.

Calls to the PECL library are faster and avoid storing complexe object for nothing.

### DIFF
--- a/tests/EventTest.php
+++ b/tests/EventTest.php
@@ -8,14 +8,11 @@ use Jean85\PrettyVersions;
 use PHPUnit\Framework\TestCase;
 use Sentry\Breadcrumb;
 use Sentry\Client;
-use Sentry\ClientBuilder;
-use Sentry\ClientInterface;
 use Sentry\Context\Context;
 use Sentry\Context\RuntimeContext;
 use Sentry\Context\ServerOsContext;
 use Sentry\Context\TagsContext;
 use Sentry\Event;
-use Sentry\Options;
 use Sentry\Severity;
 use Sentry\Util\PHPVersion;
 
@@ -24,22 +21,6 @@ use Sentry\Util\PHPVersion;
  */
 final class EventTest extends TestCase
 {
-    /**
-     * @var Options
-     */
-    protected $options;
-
-    /**
-     * @var ClientInterface
-     */
-    protected $client;
-
-    protected function setUp(): void
-    {
-        $this->client = ClientBuilder::create()->getClient();
-        $this->options = $this->client->getOptions();
-    }
-
     public function testEventIsGeneratedWithUniqueIdentifier(): void
     {
         $event1 = new Event();
--- a/tests/EventTest.php
+++ b/tests/EventTest.php
@@ -5,11 +5,7 @@ declare(strict_types=1);
 namespace Sentry\Tests;
 
 use Jean85\PrettyVersions;
-use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
-use Ramsey\Uuid\Uuid;
-use Ramsey\Uuid\UuidFactory;
-use Ramsey\Uuid\UuidFactoryInterface;
 use Sentry\Breadcrumb;
 use Sentry\Client;
 use Sentry\ClientBuilder;
@@ -28,21 +24,6 @@ use Sentry\Util\PHPVersion;
  */
 final class EventTest extends TestCase
 {
-    private const GENERATED_UUID = [
-        '4d310518-9e9d-463c-8161-bd46416f7817',
-        '431a2537-d1de-49da-80b6-b7861954c9cf',
-    ];
-
-    /**
-     * @var int
-     */
-    protected $uuidGeneratorInvokationCount;
-
-    /**
-     * @var UuidFactoryInterface
-     */
-    protected $originalUuidFactory;
-
     /**
      * @var Options
      */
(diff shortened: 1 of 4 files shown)


tests/EventTest.php
```diff
@@ -55,44 +37,26 @@ final class EventTest extends TestCase
...
+        $this->assertNotEquals($event1->getId(), $event2->getId());
     }
 
     public function testToArray(): void
     {
+        $event = new Event();
+
         $expected = [
-            'event_id' => str_replace('-', '', static::GENERATED_UUID[0]),
+            'event_id' => $event->getId(),
```
jderusse: That's the drawback. We can't predict the UUID, we just know that it's a random UUID (see `testEventIsGeneratedWithUniqueIdentifier`).

stayallive:
I'm fine with this, the leaner the better.

I am wondering why you have chosen for `UUID_TYPE_DCE` over `UUID_TYPE_RANDOM` what we used before (which should be `v4`). I know those 2 are the same in the polyfill but I can't find any documentation on the PECL extension and if it makes a difference, but I prefer `UUID_TYPE_RANDOM` better since it's more recognizable for the uninformed reader.

Also the build is failing because you left some `@throws` comments and unused imports (check the Travis logs) :)

tests/EventTest.php
stayallive: I personally think this is acceptable to prevent having to use the Ramsey dependency just for validating the ID is generated from there in testing.

Checking if there is an ID with an expected length and contents is fine for me.

tests/EventTest.php
```diff
@@ -55,44 +37,26 @@ final class EventTest extends TestCase
 
     protected function setUp(): void
     {
-        $this->uuidGeneratorInvokationCount = 0;
-        $this->originalUuidFactory = new UuidFactory();
         $this->client = ClientBuilder::create()->getClient();
         $this->options = $this->client->getOptions();
```
ste93cry: Change not strictly related to this PR, but it doesn't seems to me we're using neither of the two properties above in the class so we can drop them with the `setUp` method

stayallive:
That is interesting... it seems like `ramsey/uuid` might not be so bad after all and might be best to keep that since it looks like it is quite a bit faster than the PECL extension.

The polyfill itself is even faster but it looks like with the PECL extension you get a big downgrade in performance (comparing against `ramsey/uuid`) but without the PECL extension there is a bit of a win.

Overal it might be better to keep `ramsey/uuid` to be honest.

Great find on the announcement, did not even question the "extension is faster" statement: https://symfony.com/blog/introducing-the-new-symfony-uuid-polyfill

ste93cry:
Ramsey can use the PECL extension too, however by default its usage is disabled and must be manually enabled, even if it's available. Also I don't know how much this extension is installed, so I cannot comment on it. The downside of this change is that the polyfill automatically uses it if it finds that it's available and there is no way to prevent it

stayallive:
I agree, so if the consideration to make this change is speed I think this is going backwards when the PECL extension is actually installed and only slightly better if it's not :)

As far as I've gathered the PECL extension might be more secure/resistent to collisions since it uses a better random source, but that is not critically important in this case.

ste93cry:
Ramsey should use the `random_bytes` feature of PHP as far as I saw, but I didn't look at the full source code. One of the reasons I chose it instead of keeping the custom function like in `1.x` was because it's a battle-tested library with a strong and safe implementation. I really think that the only thing that should be taken in consideration here is if we want to depend on such 3rd party library just to generate a string (because we only do that) or if we are keen to slow down things for users having the PECL extension installed but make it faster, but probably less safe, for those who don't. One of the reasons the extension is slower is because it opens a file descriptor to `/dev/urandom` every time a UUID v4 is generated

https://jolicode.com/blog/uuid-generation-in-php

Given that this change is just an internal change and doesn't affect in any way the public API it makes sense in my opinion to merge it and accept the tradeoffs of slowing things down if the extension is enabled because I doubt that such extension is so commonly used

tests/EventTest.php
jderusse: Good catch. Fixed

jderusse:
Thank for your review and comments. Yes PECL, is 3 micro seconds slower than Ramsey/polyfill.
But I think that `Event` are not instanciated that often, and we can afford it.

The purpose of this PR was more about removing a dependency to a package (that already have 3 major version and a 4th in dev) and removing the burdon of maintaing/updating the dependency while keep Backward Compatibility with old version.

Feel free to close the PR if you think we won't affort such performance drawback, I wouldn't be offended :wink:

ste93cry:
>The purpose of this PR was more about removing a dependency to a package

As I already said, I'm more than happy to drop a dependency if we can and it makes sense and in this case for what we are doing which is just generating a UUID and then converting it to a `string` it totally makes sense, so you have my approval for the change. There is just that small change on the test class that I requested, then for me LGTM

jderusse:
tests are green. (sorry @stayallive i force-pushed you commit).

Looks like Appveyor is out

ste93cry:
>Looks like Appveyor is out

I'm looking at the issue, since PHP 7.1 is not maintained anymore the link to download the Windows version is not working anymore 🤦‍♂

stayallive:
@jderusse looks like everything is here so no worries :)

Just some AppVeyor issues, but the PR is solid, thanks!

ste93cry:
@jderusse can you please force push to retrigger the CI? Once all checks are green I will merge

jderusse:
@ste93cry Done. But this PR targets `develop` (changed by you 2 days ago) While #938 has been merged in master.

ste93cry:
Yes sorry my bad, I didn't notice it! I will rebase `master` on `develop` as soon as possible and then we should be fine here too

PR #938, opened 2019-12-18
Use Chocolatey to install PHP on AppVeyor to avoid broken download links

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.2
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

When a PHP version reaches end of life the download link to get its latest version does not work anymore and we should use the links pointing to the [Archive](https://windows.php.net/downloads/releases/archives/) to download a specific version down to patch level. To avoid the burden of needing to handle the download links, I decided to switch to installing PHP on AppVeyor using Cholocatey

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -12,21 +12,21 @@ branches:
 
 environment:
     matrix:
-        - PHP_VERSION: 7.1-Win32-VC14
+        - PHP_VERSION: 7.1
           DEPENDENCIES: lowest
-        - PHP_VERSION: 7.1-Win32-VC14
+        - PHP_VERSION: 7.1
           DEPENDENCIES: highest
-        - PHP_VERSION: 7.2-Win32-VC15
+        - PHP_VERSION: 7.2
           DEPENDENCIES: lowest
-        - PHP_VERSION: 7.2-Win32-VC15
+        - PHP_VERSION: 7.2
           DEPENDENCIES: highest
-        - PHP_VERSION: 7.3-Win32-VC15
+        - PHP_VERSION: 7.3
           DEPENDENCIES: lowest
-        - PHP_VERSION: 7.3-Win32-VC15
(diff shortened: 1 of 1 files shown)


.appveyor.yml
```diff
@@ -12,54 +11,52 @@ branches:
...
           DEPENDENCIES: highest
 
 matrix:
     fast_finish: true
 
 cache:
     - composer.phar
-    - '%LocalAppData%\Composer\files'
-    - 'c:\php -> .appveyor.yml'
+    - '%LOCALAPPDATA%\Composer\files -> composer.lock'
```
Jean85: IMHO we shouldn't thrash this cache completely only because the .lock changes. This is only Composer's cache, so we can normally partially reuse it. It would be updated at the end of the build anyway.
```suggestion
    - '%LOCALAPPDATA%\Composer\files'
```

Jean85: Also, are we caching `composer.phar`?

ste93cry: >we shouldn't thrash this cache completely only because the .lock changes

We don't commit the `composer.lock` file, so the cache would not be wiped out anyway, thus I reverted the change

>Also, are we caching `composer.phar`?

We are, but it never worked because the path is wrong. By the way, it's a really small file so imho it makes sense to always download it to be sure we always get the latest version

Jean85: I suggest you to use `composer self-update` when the PHAR is present.

ste93cry:
Build is finally green. For the records, we're affected by chocolatey/choco#1843 and the suggested workaround for the moment is to downgrade Chocolatey itself. We should keep an eye on when version `0.16` will be released because it should fix the problem

stayallive:
Nicely done @ste93cry props for figuring this out so quickly!

PR #940, opened 2019-12-19
Fix Client::captureEvent not considering the attach_stacktrace option

### LINKED ISSUE #939: Adding stack traces via captureEvent
Hi,

seems `Sentry\Client#captureEvent` doesn't apply the `attach_stacktrace` option as `captureMessage` does, and there seems to be no easy way to send stack traces along with events, unless it is an exception.

Can this tiny but important feature be added, please?

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.2
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Fixes #939: the `Client::captureEvent` method should consider the `attach_stacktrace`, instead it doesn't (it was done on purpose for some reason because there was a test for this, but I cannot remember/understand why we chose like that)

### DIFF
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -111,9 +111,9 @@ class ClientTest extends TestCase
         ];
     }
 
-    public function testCapture(): void
+    public function testCaptureEvent(): void
     {
-        /** @var TransportInterface|MockObject $transport */
+        /** @var TransportInterface&MockObject $transport */
         $transport = $this->createMock(TransportInterface::class);
         $transport->expects($this->once())
             ->method('send')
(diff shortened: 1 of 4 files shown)


tests/ClientTest.php
```diff
@@ -135,6 +135,43 @@ public function testCapture(): void
...
+     * @dataProvider captureEventAttachesStacktraceAccordingToAttachStacktraceOptionDataProvider
+     */
+    public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOption(bool $shouldAttachStacktrace): void
+    {
+        /** @var TransportInterface&MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(static function (Event $event) use ($shouldAttachStacktrace): bool {
+                if ($shouldAttachStacktrace && null === $event->getStacktrace()) {
```
Jean85: Couldn't you simplify this with a `return $shouldAttachStacktrace && null !== $event->getStacktrace();`?

ste93cry: No, when `$shouldAttachStacktrace` is `false` then the stacktrace must not be present, and that `if` statement would compute as `false && false => false` when instead it should return `true`

pilif:
> it was done on purpose for some reason because there was a test for this, but I cannot remember/understand why we chose like that

because if you don't do it, the sentry.io web client won't display the backtrace of the captured exception but instead the backtrace to where you called `captureException`.

See #957

PR #942, opened 2019-12-22
Add TYPO3 SDK integration package to the list of 3rd party vendors supporting SDK 2.x

christophlehmann:
The TYPO3 client now uses sentry/sdk


ste93cry:
I'm wondering if it's worth keeping listing the package also in the "3rd party integrations using old SDK 1.x" section for users still looking for the old SDK...

Jean85:
@ste93cry we can discuss that in a separate PR. Anyway, since the readme on master is the first document that a user reaches here, I would prefer to leave it.

PR #945, opened 2019-12-31
Replace zendframework/zend-diactoros with guzzlehttp/psr7

DeyV:
Move from zendframework/zend-diactoros to guzzlehttp/psr7

### DIFF
--- a/tests/Fixtures/code/BrokenClass.php
+++ b/tests/Fixtures/code/BrokenClass.php
@@ -1,4 +1,4 @@
-<?php /** @noinspection SyntaxError */
+<?php
 
 namespace Sentry\Tests\Fixtures\code;
 
--- a/tests/Fixtures/code/BrokenClass.php
+++ b/tests/Fixtures/code/BrokenClass.php
@@ -1,4 +1,4 @@
-<?php
+<?php /** @noinspection SyntaxError */
 
 namespace Sentry\Tests\Fixtures\code;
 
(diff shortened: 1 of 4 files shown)


tests/Fixtures/code/BrokenClass.php
```diff
@@ -1,4 +1,4 @@
-<?php
+<?php /** @noinspection SyntaxError */
```
ste93cry: I would like to avoid as much as possible IDE-specific annotations, so please remove it

tests/Integration/RequestIntegrationTest.php
```diff
@@ -39,7 +39,7 @@ public function testInvokeWithRequestHavingIpAddress(bool $shouldSendPii, array
         $event = new Event();
         $event->getUserContext()->setData(['foo' => 'bar']);
 
-        $request = new ServerRequest(['REMOTE_ADDR' => '127.0.0.1']);
+        $request = new ServerRequest('GET', new Uri(''), [], null, '1.1', ['REMOTE_ADDR' => '127.0.0.1']);
```
ste93cry: For coherence with the rest of the code, can you please create a valid `Uri` object?

DeyV:
> Thank you for the hard work. Just one note though for the future: there is no need to close the old PR and open an entirely new one to target a different branch, just use the **Edit** button at the top of the page and then rebase on top of the new branch

thx- I forgot about this possibility

PR #946, opened 2020-01-02
replace zend-diactoros with laminas-diactoros

hebbet:
"Package zendframework/zend-diactoros is abandoned, you should avoid using it. Use laminas/laminas-diactoros instead."

https://packagist.org/packages/zendframework/zend-diactoros


ste93cry:
Closing this, we already switched to `guzzlehttp/psr7` in #945

PR #951, opened 2020-01-09
Fix client creation with proxy

marmichalski:
The exception has always been thrown, no matter if vendors were present or not.

### DIFF
--- a/src/HttpClient/HttpClientFactory.php
+++ b/src/HttpClient/HttpClientFactory.php
@@ -111,9 +111,9 @@ final class HttpClientFactory implements HttpClientFactoryInterface
                 $this->httpClient = new CurlHttpClient($this->responseFactory, $this->streamFactory, [
                     CURLOPT_PROXY => $options->getHttpProxy(),
                 ]);
+            } else {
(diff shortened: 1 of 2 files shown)


src/HttpClient/HttpClientFactory.php
```diff
@@ -113,7 +113,9 @@ public function create(Options $options): HttpAsyncClientInterface
                 ]);
             }
 
-            throw new \RuntimeException('The "http_proxy" option requires either the "php-http/curl-client" or the "php-http/guzzle6-adapter" package to be installed.');
+            if (null === $this->httpClient) {
```
ste93cry: What about making this the `else` statement instead of an additional `if`?

marmichalski: Right. Don't know how I have missed that. Done, thanks.

ste93cry:
Ops, I just noticed that CHANGELOG entry is missing. Can you please add it?

marmichalski:
Sure, changelog updated.

marmichalski:
Thanks! Could this be released anytime soon?

ste93cry:
It's an important bug so definitely yes, but I would like to wait a few days to see if some other bug appears to avoid releasing too many times in a short period of time. In the meanwhile you can require the commit hash or the `dev-master` version

Issue #953, opened 2020-01-13
Changed logic for in_app_include && in_app_exclude


tpdrs:
Before #909 we deducted from project root all excludes. Now logic changed and we add to excludes all includes.

### Example before #909
- project_root = /app
- in_app_exclude = /app/vendor

And if comes frame with path /app/vendor/xxx it **not appears** in sentry.

### Example after #909 
- in_app_include = /app
- in_app_exclude = /app/vendor

And if comes frame with path /app/vendor/xxx it **appears** in sentry because we check in_app_include in the end of function.

ste93cry:
You're right, the `in_app_include` option should take precedence over the `in_app_exclude` option. I will fix it asap

PR #958, opened 2020-01-16
Fixed check for exclude

tpdrs:
#953

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Allow unsetting the stack trace on an `Event` by calling `Event::setStacktrace(null)` (#961)
 - Fix sending of both `event.stacktrace` and `event.exceptions` when `attach_stacktrace = true` (#960)
 - Fix issues with memory addresses in anonymous class stack traces (#956)
+- By default all paths are included in stacktrace (#958)
 
 ## 2.3.0 (2020-01-08)
 
--- a/src/Stacktrace.php
+++ b/src/Stacktrace.php
@@ -435,7 +435,7 @@ class Stacktrace implements \JsonSerializable
         $excludedAppPaths = $this->options->getInAppExcludedPaths();
         $includedAppPaths = $this->options->getInAppIncludedPaths();
         $absoluteFilePath = @realpath($file) ?: $file;
-        $isInApp = false;
+        $isInApp = true;
 
         if (null !== $projectRoot) {
             $isInApp = 0 === mb_strpos($absoluteFilePath, $projectRoot);
--- a/tests/StacktraceTest.php
+++ b/tests/StacktraceTest.php
@@ -206,7 +206,18 @@ final class StacktraceTest extends TestCase
             ],
             'path/to/file',
             'test_function',
-            false,
+            true,
+        ];
+
+        yield 'in_app_include not specified && file path not matching' => [
+            [
+                'project_root' => null,
+                'in_app_exclude' => [],
+                'in_app_include' => [],
+            ],
+            'path/to/file',
+            'test_function',
+            true,
         ];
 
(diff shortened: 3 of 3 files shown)


ste93cry:
Thank you for the contribution, however this as-is cannot be merged because Unified API docs clearly says that `in_app_include` overrides `in_app_exclude`, so the behavior you're trying to fix is expected and is not a bug, although I think that we all agree that this depending on the needs of each user may or may not be a good thing

>This option can be overridden using in-app-include.
https://docs.sentry.io/error-reporting/configuration/?platform=php#in-app-exclude

ste93cry:
>This PR restores the behavior described in the unified APIs

I would like to clarify this sentence because I feel that it's not: the behavior described in the Unified API is that the `in_app_exclude` option overrides the `in_app_include`, so what has been implemented in `2.3` is the correct behavior and thus there is technically nothing to change. You should not blindly migrate from `project_root` to `in_app_include`: instead you should make each of its entries more specific than the ones in `in_app_exclude` and it should work

tpdrs:
So before I set project_root = /app and excluded /app/vendor, now I have to mention all folders to include (/app/src, /app/bin, /app/tests and don't use app_exclude)

tpdrs:
Let's look to the code

```
    /**
     * Checks whether a certain frame should be marked as "in app" or not.
     *
     * @param string      $file         The file to check
     * @param string|null $functionName The name of the function
     */
    private function isFrameInApp(string $file, ?string $functionName): bool
    {
        if (self::INTERNAL_FRAME_FILENAME === $file) {
            return false;
        }

        if (null !== $functionName && 0 === strpos($functionName, 'Sentry\\')) {
            return false;
        }

        $projectRoot = $this->options->getProjectRoot();
        $excludedAppPaths = $this->options->getInAppExcludedPaths();
        $includedAppPaths = $this->options->getInAppIncludedPaths();
        $absoluteFilePath = @realpath($file) ?: $file;
        $isInApp = false;

        if (null !== $projectRoot) {
            $isInApp = 0 === mb_strpos($absoluteFilePath, $projectRoot);
        }

        foreach ($excludedAppPaths as $excludedAppPath) {
            if (0 === mb_strpos($absoluteFilePath, $excludedAppPath)) {
                $isInApp = false;

                break;
            }
        }

        foreach ($includedAppPaths as $includedAppPath) {
            if (0 === mb_strpos($absoluteFilePath, $includedAppPath)) {
                $isInApp = true;

                break;
            }
        }

        return $isInApp;
    }
```

If you say $projectRoot is deprecated then `$excludedAppPaths` is useless because `$isInApp` by default is `false` and you can remove foreach with check excluded.

tpdrs:
@ste93cry any comments?

ste93cry:
So, there is indeed one bug that should be addressed here: `$isInApp` should be `true` by default. This would make the things working as follow:

- Everything is included by default (regardless of any option)
- `in_app_exclude` sets `$isInApp` to `false` if any entry matches
- `in_app_include` sets `$isInApp` to `true` if any entry matches

With this behavior it's possible to include everything by default, exclude some folders (e.g. `app/vendor`) and still include some paths like `app/vendor/xxx` if needed. Does it makes more sense to you? If so, it would be really cool if you could adjust this PR to fit with the above specs

tpdrs:
@ste93cry I fixed default value for isInApp if in_app_include is not specified.

ste93cry:
As I said above, the default value of `$isInApp` must be true **regardless** of any option, otherwise we will report all files as not-in-app if you didn't configure `$projectRoot` nor `$includedAppPaths` nor `$excludedAppPaths`.

tpdrs:
Fixed

tpdrs:
Yep. This one solves my issues.

tpdrs:
Updated changelog

PR #960, opened 2020-01-17
do not auto-attach a backtrace when logging an exception

### LINKED ISSUE #957: 2.3.0 with manual integration produces useless backtrace
We have configured Sentry with a manual integration in order to do "stuff" with certain throwables before sending them off to sentry. 

Since upgrading to 2.3.0, the backtrace reported on sentry.io is completely useless because it contains only one frame, the one where our custom handler calls `captureException`.

Here's what we're doing:

```php
$options = new Options([
    'default_integrations' => false,
    'dsn' => $dsn,
    'error_types' => error_reporting(),
    'environment' => '' /* some code to find the env */,
    'release' => '' /* some code to find the release */,
    'attach_stacktrace' => true,
    'prefixes' => [APP_ROOT],
    'project_root' => APP_ROOT,
    'in_app_exclude' => [APP_ROOT . '/vendor/sentry'],
]);
$options->setIntegrations([
    new ModulesIntegration(),
    new RequestIntegration($options),
    new class () implements IntegrationInterface{

        {
            $errorHandler = ErrorHandler::registerOnceExceptionHandler();
            $errorHandler->addExceptionHandlerListener(static function (\Throwable $throwable): void {
                if ($throwable instanceof SomeClass) {
                    handle_this_one_ourselves($throwable);
                    return;
                }
                SentrySdk::getCurrentHub()->captureException($throwable);
            });
            $fhh = ErrorHandler::registerOnceFatalErrorHandler();
            $fhh->addFatalErrorHandlerListener(function (FatalErrorException $exception): void {
                SentrySdk::getCurrentHub()->captureException($exception);
            });
        }
    }
]);
$builder = new ClientBuilder($options);
SentrySdk::getCurrentHub()->bindClient($builder->getClient());
```

in 2.0, 2.1, and 2.2 up to and including 2.2.6, this is fine and on sentry.io it reports unhandled exceptions with a backtrace pointing to the exact spot there the exception happened.

In 2.3.0 however, the exception has an empty backtrace that just contains the one frame where we do `captureException` above

Screen Shot 2020-01-16 at 13 12 59

I did check the release notes for 2.3.0 but I could not see anything that pointed to this having changed.

pilif:
this fixes #957.

The sentry server prefers the event's stacktrace over an attached exception's stacktrace, so if a stacktrace is attached to an event containing an exception, the web interface would show the backtrace to where the exception was captured rather than the exception's backtrace.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fix client creation with proxy (#951)
 - Allow unsetting the stack trace on an `Event` by calling `Event::setStacktrace(null)` (#961)
+- Fix sending of both `event.stacktrace` and `event.exceptions` when `attach_stacktrace = true` (#960)
 
 ## 2.3.0 (2020-01-08)
 
--- a/src/Client.php
+++ b/src/Client.php
@@ -85,7 +85,7 @@ final class Client implements FlushableClientInterface
             'level' => $level,
         ];
 
-        $event = $this->prepareEvent($payload, $scope, $this->options->shouldAttachStacktrace());
+        $event = $this->prepareEvent($payload, $scope);
 
         if (null === $event) {
             return null;
@@ -111,7 +111,7 @@ final class Client implements FlushableClientInterface
      */
     public function captureEvent(array $payload, ?Scope $scope = null): ?string
     {
-        $event = $this->prepareEvent($payload, $scope, $this->options->shouldAttachStacktrace());
+        $event = $this->prepareEvent($payload, $scope);
 
         if (null === $event) {
             return null;
@@ -162,13 +162,12 @@ final class Client implements FlushableClientInterface
     /**
      * Assembles an event and prepares it to be sent of to Sentry.
      *
-     * @param array      $payload        the payload that will be converted to an Event
-     * @param Scope|null $scope          optional scope which enriches the Event
-     * @param bool       $withStacktrace True if the event should have and attached stacktrace
+     * @param array      $payload the payload that will be converted to an Event
+     * @param Scope|null $scope   optional scope which enriches the Event
      *
      * @return Event|null returns ready to send Event, however depending on options it can be discarded
      */
-    private function prepareEvent(array $payload, ?Scope $scope = null, bool $withStacktrace = false): ?Event
+    private function prepareEvent(array $payload, ?Scope $scope = null): ?Event
     {
         $sampleRate = $this->getOptions()->getSampleRate();
 
@@ -176,12 +175,7 @@ final class Client implements FlushableClientInterface
             return null;
         }
 
-        // when 'exception' is set, no backtrace should be auto-attached (#957)
-        if (
-            $withStacktrace
-            && !isset($payload['exception'])
-            && !isset($payload['stacktrace'])
-        ) {
+        if ($this->getOptions()->shouldAttachStacktrace() && !isset($payload['exception']) && !isset($payload['stacktrace'])) {
             $event = $this->eventFactory->createWithStacktrace($payload);
         } else {
             $event = $this->eventFactory->create($payload);
--- a/src/Client.php
+++ b/src/Client.php
@@ -85,13 +85,7 @@ final class Client implements FlushableClientInterface
             'level' => $level,
         ];
 
-        $event = $this->prepareEvent($payload, $scope);
-
-        if (null === $event) {
-            return null;
-        }
-
-        return $this->transport->send($event);
+        return $this->captureEvent($payload, $scope);
     }
 
     /**
--- a/src/Client.php
+++ b/src/Client.php
@@ -176,7 +176,12 @@ final class Client implements FlushableClientInterface
             return null;
         }
 
-        if ($withStacktrace) {
+        // when 'exception' is set, no backtrace should be auto-attached (#957)
+        if (
+            $withStacktrace
+            && !isset($payload['exception'])
+            && !isset($payload['stacktrace'])
+        ) {
             $event = $this->eventFactory->createWithStacktrace($payload);
         } else {
             $event = $this->eventFactory->create($payload);
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -197,6 +197,27 @@ class ClientTest extends TestCase
         ];
     }
 
+    public function testCaptureEventPrefersExplicitStacktrace(): void
+    {
+        $explicitStacktrace = $this->createMock(Stacktrace::class);
+        $payload = ['stacktrace' => $explicitStacktrace];
+
+        /** @var TransportInterface&MockObject $transport */
+        $transport = $this->createMock(TransportInterface::class);
+        $transport->expects($this->once())
+            ->method('send')
+            ->with($this->callback(static function (Event $event) use ($explicitStacktrace): bool {
+                return $explicitStacktrace === $event->getStacktrace();
+            }))
+            ->willReturn('500a339f3ab2450b96dee542adf36ba7');
+
+        $client = ClientBuilder::create(['attach_stacktrace' => true])
+            ->setTransportFactory($this->createTransportFactory($transport))
+            ->getClient();
+
+        $this->assertEquals('500a339f3ab2450b96dee542adf36ba7', $client->captureEvent($payload));
+    }
+
     public function testCaptureLastError(): void
     {
         /** @var TransportInterface&MockObject $transport */
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -166,13 +166,34 @@ class ClientTest extends TestCase
         $this->assertEquals('500a339f3ab2450b96dee542adf36ba7', $client->captureEvent($payload));
     }
 
-    public function captureEventAttachesStacktraceAccordingToAttachStacktraceOptionDataProvider(): array
+    public function captureEventAttachesStacktraceAccordingToAttachStacktraceOptionDataProvider(): \Generator
     {
-        return [
-            [true, [], true],
-            [false, [], false],
-            [true, ['exception' => new \Exception()], false],
-            [false, ['exception' => new \Exception()], false],
+        yield 'Stacktrace attached when attach_stacktrace = true and both payload.exception and payload.stacktrace are unset' => [
+            true,
+            [],
+            true,
+        ];
+
+        yield 'No stacktrace attached when attach_stacktrace = false' => [
+            false,
+            [],
+            false,
+        ];
+
+        yield 'No stacktrace attached when attach_stacktrace = true and payload.exception is set' => [
+            true,
+            [
+                'exception' => new \Exception(),
+            ],
+            false,
+        ];
+
+        yield 'No stacktrace attached when attach_stacktrace = false and payload.exception is set' => [
+            true,
+            [
+                'exception' => new \Exception(),
+            ],
+            false,
         ];
     }
 
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -140,7 +140,7 @@ class ClientTest extends TestCase
     /**
      * @dataProvider captureEventAttachesStacktraceAccordingToAttachStacktraceOptionDataProvider
      */
-    public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOption(bool $shouldAttachStacktrace): void
+    public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOption(bool $attachStacktraceOption, array $payload, bool $shouldAttachStacktrace): void
     {
         /** @var TransportInterface&MockObject $transport */
         $transport = $this->createMock(TransportInterface::class);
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Fix client creation with proxy (#951)
+- Make `attach_stacktrace` not attach a stack trace when capturing an Exception in order to prevent sentry.io from showing a useless backtrace (#940, #957)
```
ste93cry: Since this issue is not related to [sentry.io](https://sentry.io) but to the server in general, can you please update the description to ``Fix sending of both `event.stacktrace` and `event.exceptions` when `attach_stacktrace = true``. Also please drop the reference to the issue and leave only the PR as that is enough

src/Client.php
```diff
@@ -176,7 +176,12 @@ private function prepareEvent(array $payload, ?Scope $scope = null, bool $withSt
             return null;
         }
 
-        if ($withStacktrace) {
+        // when 'exception' is set, no backtrace should be auto-attached (#957)
```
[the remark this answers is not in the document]
pilif: I can do that. The reason I put the comment there is because back in #940 where an explicit test for this was removed, there was a comment posted where people didn't remember what the check was for.

As this behavior is unexpected and has been broken once even though it was there previously, I felt like it warranted the comment.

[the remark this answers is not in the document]
ste93cry: We should be covered by documenting each test case of this piece of code: in the `captureEventAttachesStacktraceAccordingToAttachStacktraceOptionDataProvider` method we should return a custom description for each test case similarly to what has been done in `addFrameSetsInAppFlagCorrectlyDataProvider` so that if for some reason we change this code and we break an existing test case it's clear why we did that thing in the past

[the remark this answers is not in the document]
pilif: fixed.

src/Client.php
```diff
@@ -176,7 +176,12 @@ private function prepareEvent(array $payload, ?Scope $scope = null, bool $withSt
...
+        if (
```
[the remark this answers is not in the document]
pilif: fixed

tests/ClientTest.php
```diff
@@ -159,18 +159,21 @@ public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOpt
...
-        $client = ClientBuilder::create(['attach_stacktrace' => $shouldAttachStacktrace])
+        $client = ClientBuilder::create(['attach_stacktrace' => $attachConfigSetting])
             ->setTransportFactory($this->createTransportFactory($transport))
             ->getClient();
 
-        $this->assertEquals('500a339f3ab2450b96dee542adf36ba7', $client->captureEvent([]));
+        $this->assertEquals('500a339f3ab2450b96dee542adf36ba7', $client->captureEvent($payload));
     }
 
     public function captureEventAttachesStacktraceAccordingToAttachStacktraceOptionDataProvider(): array
```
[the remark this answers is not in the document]
pilif: fixed

tests/ClientTest.php
```diff
@@ -159,18 +159,21 @@ public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOpt
...
     {
         return [
-            [true],
-            [false],
+            [true, [], true],
+            [false, [], false],
+            // when 'exception' is set, no backtrace should be auto-attached (#957)
```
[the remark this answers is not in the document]
pilif: fixed

tests/ClientTest.php
```diff
@@ -159,18 +159,21 @@ public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOpt
...
             ->setTransportFactory($this->createTransportFactory($transport))
             ->getClient();
 
-        $this->assertEquals('500a339f3ab2450b96dee542adf36ba7', $client->captureEvent([]));
+        $this->assertEquals('500a339f3ab2450b96dee542adf36ba7', $client->captureEvent($payload));
     }
 
     public function captureEventAttachesStacktraceAccordingToAttachStacktraceOptionDataProvider(): array
     {
         return [
```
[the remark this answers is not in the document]
pilif: added

CHANGELOG.md
pilif: fixed

pilif:
haha - creating conflicts with myself - and on CHANGELOG.md no less.

rebasing 😊

src/Client.php
```diff
@@ -85,7 +85,7 @@ public function captureMessage(string $message, ?Severity $level = null, ?Scope
             'level' => $level,
         ];
 
-        $event = $this->prepareEvent($payload, $scope, $this->options->shouldAttachStacktrace());
+        $event = $this->prepareEvent($payload, $scope);
```
ste93cry: Instead of calling `prepareEvent`, we can now simply do `return captureEvent`

src/Client.php
```diff
@@ -162,21 +162,21 @@ public function flush(?int $timeout = null): PromiseInterface
...
+     * @param array      $payload the payload that will be converted to an Event
+     * @param Scope|null $scope   optional scope which enriches the Event
      *
      * @return Event|null returns ready to send Event, however depending on options it can be discarded
      */
-    private function prepareEvent(array $payload, ?Scope $scope = null, bool $withStacktrace = false): ?Event
+    private function prepareEvent(array $payload, ?Scope $scope = null): ?Event
     {
         $sampleRate = $this->getOptions()->getSampleRate();
+        $withStacktrace = $this->getOptions()->shouldAttachStacktrace();
```
ste93cry: Can you inline this method call in the `if` statement?

tests/ClientTest.php
```diff
@@ -159,19 +159,57 @@ public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOpt
...
     }
 
-    public function captureEventAttachesStacktraceAccordingToAttachStacktraceOptionDataProvider(): array
+    public function captureEventAttachesStacktraceAccordingToAttachStacktraceOptionDataProvider(): \Generator
     {
-        return [
-            [true],
-            [false],
+        yield 'attaches a stacktrace when attach_stacktrace=true but none is provided' => [
+            true, [], true,
```
ste93cry: Please split each entry into its own line

tests/ClientTest.php
```diff
@@ -159,19 +159,57 @@ public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOpt
...
+        ];
+
+        yield 'attaches no stacktrace when attach_stacktrace=false and none is provided' => [
+            false, [], false,
+        ];
+
+        yield "attaches no stacktrace even when attach_stacktrace=true but an Exception is present (server prefers attached stacktrace to an Exception's trace)" => [
+            true, [
```
ste93cry: Please split each entry into its own line

tests/ClientTest.php
```diff
@@ -159,19 +159,57 @@ public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOpt
...
     {
-        return [
-            [true],
-            [false],
+        yield 'attaches a stacktrace when attach_stacktrace=true but none is provided' => [
+            true, [], true,
+        ];
+
+        yield 'attaches no stacktrace when attach_stacktrace=false and none is provided' => [
+            false, [], false,
```
ste93cry: Please split each entry into its own line

tests/ClientTest.php
```diff
@@ -159,19 +159,57 @@ public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOpt
...
+
+        yield "attaches no stacktrace even when attach_stacktrace=true but an Exception is present (server prefers attached stacktrace to an Exception's trace)" => [
+            true, [
+                'exception' => new \Exception(),
+            ],
+            false,
         ];
+
+        yield 'attaches no stacktrace even when attach_stacktrace=false and Exception is present' => [
+            true, [
```
ste93cry: Please split each entry into its own line

tests/ClientTest.php
```diff
@@ -140,7 +140,7 @@ public function testCaptureEvent(): void
     /**
      * @dataProvider captureEventAttachesStacktraceAccordingToAttachStacktraceOptionDataProvider
      */
-    public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOption(bool $shouldAttachStacktrace): void
+    public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOption(bool $attachConfigSetting, array $payload, bool $shouldAttachStacktrace): void
```
ste93cry: Please rename the `$attachConfigSetting` variable to `$attachStacktraceOption`

tests/ClientTest.php
```diff
@@ -159,19 +159,57 @@ public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOpt
...
+        ];
+
+        yield "attaches no stacktrace even when attach_stacktrace=true but an Exception is present (server prefers attached stacktrace to an Exception's trace)" => [
+            true, [
+                'exception' => new \Exception(),
+            ],
+            false,
         ];
+
+        yield 'attaches no stacktrace even when attach_stacktrace=false and Exception is present' => [
```
ste93cry: Please change this description to `No stacktrace attached when attach_stacktrace = false and payload.exception is set`

tests/ClientTest.php
```diff
@@ -159,19 +159,57 @@ public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOpt
...
-            [false],
+        yield 'attaches a stacktrace when attach_stacktrace=true but none is provided' => [
+            true, [], true,
+        ];
+
+        yield 'attaches no stacktrace when attach_stacktrace=false and none is provided' => [
+            false, [], false,
+        ];
+
+        yield "attaches no stacktrace even when attach_stacktrace=true but an Exception is present (server prefers attached stacktrace to an Exception's trace)" => [
```
ste93cry: Please change this description to `No stacktrace attached when attach_stacktrace = true and payload.exception is set`

tests/ClientTest.php
```diff
@@ -159,19 +159,57 @@ public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOpt
...
+    public function captureEventAttachesStacktraceAccordingToAttachStacktraceOptionDataProvider(): \Generator
     {
-        return [
-            [true],
-            [false],
+        yield 'attaches a stacktrace when attach_stacktrace=true but none is provided' => [
+            true, [], true,
+        ];
+
+        yield 'attaches no stacktrace when attach_stacktrace=false and none is provided' => [
```
ste93cry: Please change this description to `No stacktrace attached when attach_stacktrace = false`

tests/ClientTest.php
```diff
@@ -159,19 +159,57 @@ public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOpt
...
+        $this->assertEquals('500a339f3ab2450b96dee542adf36ba7', $client->captureEvent($payload));
     }
 
-    public function captureEventAttachesStacktraceAccordingToAttachStacktraceOptionDataProvider(): array
+    public function captureEventAttachesStacktraceAccordingToAttachStacktraceOptionDataProvider(): \Generator
     {
-        return [
-            [true],
-            [false],
+        yield 'attaches a stacktrace when attach_stacktrace=true but none is provided' => [
```
ste93cry: Please change this description to `Stacktrace attached when attach_stacktrace = true and both payload.exception and payload.stacktrace are set`

tests/ClientTest.php
```diff
@@ -159,19 +159,57 @@ public function testCaptureEventAttachesStacktraceAccordingToAttachStacktraceOpt
...
+            true, [
+                'exception' => new \Exception(),
+            ],
+            false,
+        ];
+    }
+
+    public function testCaptureEventPrefersExplicitStacktrace(): void
+    {
+        $explicit_stacktrace = $this->createMock(Stacktrace::class);
```
ste93cry: Please use the camelCase format

src/Client.php
pilif: I see you don't seem to have an issue with long lines 😆 - this will grow to 128 characters when I inline it.

tests/ClientTest.php
pilif: shouldn't this be "… and both payload.exception and payload.stacktrace are **unset**"?

src/Client.php
ste93cry: I'm more inclined to keep the code on one line, it's just a matter of preference of course and sometimes I myself find that for really long lines of code (e.g. functions with a long list of arguments) it indeed makes sense to use multiline, however most of the codebase is written this way so I prefer to stick with it as much as possible. Also our code style is implicitly based on the one of Symfony which in turn prefers one-line style :smile: By the way, the real reason I asked to inline the statement here was to avoid a variable as it's used only once

tests/ClientTest.php
ste93cry: Yes, nice catch!

src/Client.php
pilif: yeah - the reason I used the variable was to keep the line short.

in the end it's totally a matter of taste and I gladly adopt to the existing style.

src/Client.php
pilif: fixed

src/Client.php
pilif: fixed

tests/ClientTest.php
pilif: fixed

tests/ClientTest.php
pilif: fixed

tests/ClientTest.php
pilif: fixed

tests/ClientTest.php
pilif: fixed

tests/ClientTest.php
pilif: fixed

tests/ClientTest.php
pilif: fixed

tests/ClientTest.php
pilif: fixed

tests/ClientTest.php
pilif: fixed

PR #961, opened 2020-01-17
allow the stacktrace to be set to null

pilif:
as part of fixing #957 application-side in a before_send hook, it turend out that even though the $stacktrace field is nullable, the setter didn't allow setting it to `null` because of an erroneous type hint

see [this comment](https://github.com/getsentry/sentry-php/issues/957#issuecomment-575385598) by @ste93cry

### DIFF
--- a/tests/EventTest.php
+++ b/tests/EventTest.php
@@ -10,6 +10,7 @@ use Sentry\Breadcrumb;
 use Sentry\Client;
 use Sentry\Event;
 use Sentry\Severity;
+use Sentry\Stacktrace;
 use Sentry\Util\PHPVersion;
 
 /**
@@ -251,6 +252,20 @@ final class EventTest extends TestCase
             ['fingerprint', ['foo', 'bar'], ['fingerprint' => ['foo', 'bar']]],
             ['environment', 'foo', ['environment' => 'foo']],
         ];
+    }
+
+    public function testSetStacktrace(): void
+    {
+        $stacktrace = $this->createMock(Stacktrace::class);
+
+        $event = new Event();
+        $event->setStacktrace($stacktrace);
+
+        $this->assertSame($stacktrace, $event->getStacktrace());
+
+        $event->setStacktrace(null);
+
+        $this->assertNull($event->getStacktrace());
     }
 
     public function testEventJsonSerialization(): void
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Fix client creation with proxy (#951)
+- Allow unsetting the stack trace on an `Event` by calling `Event::setStacktrace(null)` (#961)
 
 ## 2.3.0 (2020-01-08)
 
--- a/src/Event.php
+++ b/src/Event.php
@@ -552,9 +552,9 @@ final class Event implements \JsonSerializable
     /**
      * Sets the stacktrace that generated this event.
      *
-     * @param Stacktrace $stacktrace The stacktrace instance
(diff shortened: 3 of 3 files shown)


pilif:
yup - no problem. will do.

I wasn't convinced this would actually require a CHANGELOG entry as it's very close to the internals. And for the test - I mean - it's a setter for a field. what could possibly go wrong? But you're right - better safe than sorry.

pilif:
I have amended the commit to also contain the test (the generic `testGettersAndSetters` test wasn't useful to prove the setter allows unsetting a set backtrace), so I wrote a dedicated method) and added a second commit with the changelog update.

ste93cry:
>as it's very close to the internals

Not really as the method is part of the public API and so it makes sense to document the change even though probably few people will care about it

>I mean - it's a setter for a field. what could possibly go wrong

I'm also in doubt of whether testing setters and getters is worth, and I don't have a strong opinion on this even though somehow I agree that having a test that just does the following is kinda pointless and instead the methods should be tested as part of a feature that uses them

```php
$object->setValue(...);

$this->assertSame(..., $object->getValue());
```

However, let's suppose that the setter does something like skipping to set the variable if the argument is `null`. In that case, without testing that both setting a value other than `null` and setting `null` works would prevent us to see a potential bug. Does it makes sense for you?

```php
function setValue(?string $value)
{
    if (null === $value) {
        return;
    }

    $this->value = $value;
}

$object->setValue('foo');

$this->assertSame('foo', $object->getValue());

$object->setValue(null);

$this->assertNull($object->getValue()); // Without this line we would not notice that the value is not being set to `null`, thus leading to a bug
```

pilif:
>  In that case, without testing that both setting a value other than null and setting null works would prevent us to see a potential bug. Does it makes sense for you?

I see what you mean. yep. setting it to `null` could also be interpreted as "don't change what's currently stored".

tests/EventTest.php
```diff
@@ -253,6 +254,18 @@ public function gettersAndSettersDataProvider(): array
         ];
     }
 
+    public function testBacktraceSetter(): void
```
ste93cry: As a convention we are used to name the test methods `test[name of the method]` or `test[name of the method][what is testing]`. Can you kindly change it to `testSetStacktrace`?

tests/EventTest.php
```diff
@@ -253,6 +254,18 @@ public function gettersAndSettersDataProvider(): array
...
+    {
+        $stacktrace = $this->createMock(Stacktrace::class);
+        $event = new Event();
+        $event->setStacktrace($stacktrace);
+
+        $this->assertSame($stacktrace, $event->getStacktrace());
+
+        $event->setStacktrace(null);
+        $this->assertNull($event->getStacktrace());
```
ste93cry: Please add a blank line above this one to improve readability

tests/EventTest.php
```diff
@@ -253,6 +254,18 @@ public function gettersAndSettersDataProvider(): array
         ];
     }
 
+    public function testBacktraceSetter(): void
+    {
+        $stacktrace = $this->createMock(Stacktrace::class);
+        $event = new Event();
```
ste93cry: Please add a blank line above this one to improve readability

pilif: done and amended

tests/EventTest.php
pilif: done and amended

tests/EventTest.php
pilif: fixed.

ste93cry:
Thank you for your hard work on contributing this bugfix, LGTM!

PR #967, opened 2020-01-23
Add timeout option to limit the request duration

stremlau:
Added the timeout option, similar to https://docs.sentry.io/clients/java/config/#timeout-advanced

Let me know if something is missing. I implemented tests like the ones for the http_proxy option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Add `timeout` option to limit the request duration
+
 ## 2.3.1 (2020-01-23)
 
 - Allow unsetting the stack trace on an `Event` by calling `Event::setStacktrace(null)` (#961)
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -35,6 +35,9 @@ parameters:
             message: '/^Access to constant PROXY on an unknown class GuzzleHttp\\RequestOptions\.$/'
             path: src/HttpClient/HttpClientFactory.php
         -
+            message: '/^Access to constant TIMEOUT on an unknown class GuzzleHttp\\RequestOptions\.$/'
+            path: src/HttpClient/HttpClientFactory.php
+        -
             message: '/^Property Sentry\\HttpClient\\HttpClientFactory::\$httpClient \(Http\\Client\\HttpAsyncClient\|null\) does not accept Http\\Client\\Curl\\Client.$/'
             path: src/HttpClient/HttpClientFactory.php
         -
--- a/src/HttpClient/HttpClientFactory.php
+++ b/src/HttpClient/HttpClientFactory.php
@@ -109,14 +109,15 @@ final class HttpClientFactory implements HttpClientFactoryInterface
                 $guzzleConfig = [];
 
                 if (null !== $options->getHttpProxy()) {
+                    /** @psalm-suppress UndefinedClass */
                     $guzzleConfig[GuzzleHttpClientOptions::PROXY] = $options->getHttpProxy();
                 }
 
                 if (null !== $options->getTimeout()) {
+                    /** @psalm-suppress UndefinedClass */
                     $guzzleConfig[GuzzleHttpClientOptions::TIMEOUT] = $options->getTimeout() / 1000;
                 }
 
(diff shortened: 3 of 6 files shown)


ste93cry:
I have to admit that the work is amazing, however I'm against adding this option as I was when we were forced to add the `http_proxy` one: the reason is that it's an hell to maintain as for every HTTP client that we want to support out-of-the-box we have to remap the option to work with them, plus it ceases to work as soon as you want to switch the client from the default one, so it's usefulness in my opinion is 0 and it seems that something is broken. Also, this option is clearly something that does not belongs to the SDK because only a certain type of transport needs it (the HTTP transport). The proper way to set the timeout in my opinion is to instantiate your own HTTP client (with `2.3` it's far more easier to do it) and configure it however you like

stremlau:
Thank you for your fast reply!

The timeout option should be used by everyone who uses sentry in production. Without it, the application uptime depends on the uptime of sentry, which is not suitable for a monitoring tool. We use the self-hosted version of sentry in a not high available setup and every time when sentry is down, all our http nodes are getting blocked by long running requests to sentry. (After reducing the amount of events we plan to switch to the hosted version and hope that we will have fewer downtimes, but it is also not guaranteed, that the hosted version is 100% up).

I assume, that the majority of the users of this library don't need and want to switch the default client and therefore this option is a benefit. For example it is very hacky to use a custom client in sentry-laravel. When you need to switch the client, it is still possible and the exception will guide you to disable the timeout option.

Is there a chance to get this merged? Otherwise please close it and I will use a custom http client :-|

ste93cry:
>The timeout option should be used by everyone who uses sentry in production

I agree and it makes sense at all

>For example it is very hacky to use a custom client in sentry-laravel

I'm sure that we can improve the DX and in this regards any contribution is more than welcome

>Is there a chance to get this merged?

After my first comment I had a look again at the docs and Unified API doesn't really talk about such option, so it's one more reason I'm against the merging. This option is specific of the Java SDK and in fact is marked as "Advanced" in the documentation. Using a custom HTTP client to configure such settings is the right way, also in the light of the statement I wrote above:

>this option is clearly something that does not belongs to the SDK because only a certain type of transport needs it (the HTTP transport)

ste93cry:
After discussing this PR with @stayallive, we came to the conclusion that some defaults should be changed to be more tolerant in case the connection with the server cannot be established or the transfer takes too much time, so if you are willing to make the needed changes we would gladly accept them. What should be done is:

- Set the cURL `CURLOPT_CONNECTTIMEOUT` option to `2` (this was the default value in version `1.x`)
- Set the cURL `CURLOPT_TIMEOUT` option to `5` (in version `1.x` the default was `2`)

Of course this must be done for both HTTP clients that we support out-of-the-box. We think that these are good default values, if you still want to customize them then my advice of instantiating your own HTTP client is still valid and for the moment I prefer to avoid adding options to customize such things as it's pretty clear that there is the potential for having dozens of them. Please also target the `develop` branch because this is not a bugfix

stremlau:
We use multiple different APIs in our application, so we have the need to set the timeouts individual e.g. a working sentry setup resonds in a few hundred milliseconds and it is ok to lose some events vs. our invoice service takes up to a few seconds and we should wait for it.

I still think that this option is useful for this library. For now there is just the http transport and other options are referring to it (http_proxy). Since we cannot wait to use the timeout, I have to find a solution anyways, so feel free to close this PR.

ste93cry:
>I still think that this option is useful for this library. For now there is just the http transport and other options are referring to it (http_proxy).

Setting the timeout is indeed useful, and maybe the documentation should provide an example of how to do it given that it doesn't seems to be so straightforward. I already explained the reason I'm against adding more options that are useful only in the context of the HTTP transport and that will cease to work as soon as you stop using the default HTTP client provided out-of-the-box from this library

>I have to find a solution anyways, so feel free to close this PR.

I already gave you the solution: you can use the `HttpClientFactoryInterface` (or one of its implementations if it's enough for you) to create your own instance of the HTTP client, configure it however you like and then set it using the transport factory. I'm going to close this issue and I will open a new PR to fix the default values for the timeouts, which are wrong in any case

Issue #973, opened 2020-02-07
Bug in error handler?


thomasvargiu:
I don't know why in these lines of code the error handler throws the exception again.

This is causing the exception to be converted in an `E_ERROR`, then catched again by the shutdown function. This behaviour causes the error to be logger twice, triggering other shutdown functions, and logging it (the `E_ERROR`) with a stack trace (because when converted, the message contains the stack trace) with possible secrets in parameters (it's logging my database secrets when my app fails to connect).

https://github.com/getsentry/sentry-php/blob/a74999536b9119257cb1a4b1aa038e4a08439f67/src/ErrorHandler.php#L444-L452

The previous handler is already called, I think we should just `return`, but maybe I'm not thinking to other behaviours.

ste93cry:
Can you please provide a simple test case that reproduces the issue? Being the first time I see this it would help a lot to understand what's going on

thomasvargiu:
Hi, IMHO the problem is that even if the exception is handled, it is always thrown.
The exception handler is to handle an exception, so it's not necessary to throw it again.
I saw that the code is very similar to the symfony error handler, but it throws an exception only when there is another exception while handling it:

https://github.com/symfony/error-handler/blob/33ea3274749b635b6af9a5dbb2cdf53dabc55536/ErrorHandler.php#L597-L603

So I think the following assertion is wrong:

https://github.com/getsentry/sentry-php/blob/7acdcdd1d507e29a38b4c282527a2ae3c9fca130/tests/phpt/error_handler_captures_exception.phpt#L44-L46

ste93cry:
>even if the exception is handled, it is always thrown

This was meant to be that way because the error handler of Sentry is not meant to catch exceptions and hide them but just to log them. If you want to catch exceptions and e.g. render an error page then it's up to you to set another exception handler. However having that `return` statement probably makes definitely sense and I would gladly accept a contribution to fix it, but before doing it having a small example that reproduces the issue and what you expect would make things a lot easier if we have to reason about it

>I saw that the code is very similar to the symfony error handler

It is because in effect the code is largely based on that component

ste93cry:
>Hi, IMHO the problem is that even if the exception is handled, it is always thrown.
>The exception handler is to handle an exception, so it's not necessary to throw it again.

Having that `return` statement probably makes definitely sense and I would gladly accept a contribution to fix it, but before doing it having a small example that reproduces the issue and what you expect would make things a lot easier if we have to reason about it

[...]

PR #974, opened 2020-02-10
Try to make the exception handler more transparent

Jean85:
This is a first attempt to solve #973. I'll comment the code to explain my attempt, ~which is incomplete for now~.

Build is now green, PR is ready for review.


src/ErrorHandler.php
```diff
@@ -434,6 +434,8 @@ private function handleException(\Throwable $exception): void
         try {
             if (null !== $previousExceptionHandler) {
                 $previousExceptionHandler($exception);
+
+                return;
```
Jean85: This is the hearth of the change: if we call down the chain of registered error handler and we don't get any exception thrown, we shouldn't do anything else, so the behavior with and without Sentry should be the same.

src/ErrorHandler.php
```diff
@@ -451,7 +453,9 @@ private function handleException(\Throwable $exception): void
             throw $exception;
         }
 
-        $this->handleException($previousExceptionHandlerException);
+        if ($exception !== $previousExceptionHandlerException) {
+            $this->handleException($previousExceptionHandlerException);
+        }
```
Jean85: This is another helpful change: let's not go into recursion if we have the same exception we were reporting on.

tests/phpt/error_handler_captures_exception.phpt
```diff
@@ -41,6 +41,3 @@ throw new \Exception('foo bar');
 --EXPECTF--
 Exception listener called
 Custom exception handler called
-Fatal error: Uncaught Exception: foo bar in %s:%d
-Stack trace:
-%a
```
Jean85: I changed this expectation because it's what is happening if I remove completely Sentry from the test.

tests/phpt/error_handler_captures_exception_thrown_from_next_handler.phpt
```diff
@@ -0,0 +1,77 @@
...
+throw new \Exception('foo bar');
+?>
+--EXPECTF--
+Sentry error listener called: User Deprecated%s
+Custom exception handler called
+Calling Sentry exception handler...
+Sentry exception listener called: foo bar
+Sentry handler rethrowed
+Throwing a new exception from the custom exception handler
+Sentry fatal error listener called: Secondary exception thrown
```
Jean85: This line is currently making the test fail: if another exception handler registered after Sentry throws an exception (causing so a fatal), Sentry is not reporting it, we will receive only the first exception that was causing the entire chain to trigger.

tests/phpt/error_handler_captures_rethrown_exception_once.phpt
```diff
@@ -51,8 +51,10 @@ SentrySdk::getCurrentHub()->bindClient($client);
...
 --EXPECTF--
-Transport called
+Transport called: foo bar
 Custom exception handler called
+
 Fatal error: Uncaught Exception: foo bar in %s:%d
 Stack trace:
-%a
+#0 {main}
+  thrown in %s on line %d
```
Jean85: This doesn't change, I've just made the expectation stricter.

src/ErrorHandler.php
ste93cry: It's not needed because of the `if` statement right above these lines of code that throws if the exception thrown by the previous handler is the same as the one we're trying to handle. To say it simple, this case was already covered

src/ErrorHandler.php
```diff
@@ -402,7 +409,21 @@ private function handleFatalError(array $error = null): void
...
+        if (empty($error)) {
+            return;
+        }
+
+        if (
+            self::$lastCapturedException
+            && self::$lastCapturedException->getFile() === $error['file']
+            && self::$lastCapturedException->getLine() === $error['line']
+            && false !== strpos($error['message'], 'Uncaught ')
+            && false !== strpos($error['message'], self::$lastCapturedException->getMessage())
```
Jean85: Those two string matches are a bit "lazy", but it's to avoid issues with different variants of the message, like:
 * `Uncaught Error: `
 * `Uncaught Exception: `
 * `Error: Uncaught Error: `
etc

ste93cry:
It looks like by trying to fix one specific bug, you had put in the PR a lot of stuffs that do not strictly relate to the issue. Please revert all the changes and leave only the ones that fix the bug this PR originated from and instead let's open a separate issue/PR to refactor the tests or fix the other bugs you found out 🙏

src/ErrorHandler.php
```diff
@@ -397,19 +410,43 @@ private function handleFatalError(array $error = null): void
...
             $this->exceptionReflection->setValue($errorAsException, []);
 
             $this->invokeListeners($this->errorListeners, $errorAsException);
             $this->invokeListeners($this->fatalErrorListeners, $errorAsException);
         }
+
+        if ($shouldExit) {
+            // Force the process to exit with code 255 to respect what would
+            // happen natively if no handler was registered
+            register_shutdown_function('register_shutdown_function', static function (): void {
```
Jean85: I don't like this, it's dangerous and it could kill the app before other shutdown functions are executed.

Why did you add this? The exit with failure was the only thing that worked even before.

src/ErrorHandler.php
```diff
@@ -426,14 +463,17 @@ private function handleException(\Throwable $exception): void
...
         $previousExceptionHandlerException = $exception;
 
-        // Unset the previous exception handler to prevent infinite loop in case
-        // we need to handle an exception thrown from it
+        // Sets the previous handler so that the exception is printed on the
+        // screen the next time this method runs. This also breaks the infinite
+        // loop that would happen if the previous handler thrown a new exception
         $previousExceptionHandler = $this->previousExceptionHandler;
-        $this->previousExceptionHandler = null;
+        $this->previousExceptionHandler = [$this, 'printException'];
```
Jean85: This doesn't seem transparent to me.

src/ErrorHandler.php
ste93cry: As the comment says, if we don't do this then in case of a fatal error the process will exit with code 0. Instead, natively, it would exit with code 255

src/ErrorHandler.php
ste93cry: This is going to do the same thing that the native handler would do which is printing the exception. We have to do this because in case if there is a `$previousExceptionHandler` set at the first run we call it, so nothing gets printed. At the second run instead we do want to print the error. Also, this is needed to break the infinite loop in case `$previousExceptionHandler` throws an exception

src/ErrorHandler.php
Jean85: No it doesn't. The process was already exiting with 255, but without a fatal.

src/ErrorHandler.php
Jean85: I agree that breaking the loop is needed, but "stealing" the job from the native handler seems asking for trouble to me. We need to give the flow back to whatever is happening, not mimic it.

ste93cry: It's a lot more complicated than what it seems, because there are a lot of different cases that should be handled and there doesn't seems to be a solution that works for all of them without some sort of hack/workaround. If we don't print the exception ourselves then there may be cases where it doesn't get printed at all. Given that this code is again based on the SF error handler, although with some changes, and that I'm pretty sure that the people that worked on it discovered the same things, I firmly believe that we either choose to take some compromises or we don't fix the fact that an exception thrown inside the next handler after our has been called is not catched

Jean85: I agree. I tend to think that we can leave behind the case of "another handler messed up" if it forces us to implement dirty hacks. It wouldn't be our fault, and we can't cover 100% of the cases.

tests/phpt/error_handler_captures_exception_thrown_from_next_handler.phpt
```diff
@@ -0,0 +1,50 @@
+--TEST--
+
```
Jean85: Description missing

tests/phpt/error_handler_captures_rethrown_exception_once.phpt
```diff
@@ -1,58 +0,0 @@
---TEST--
-Test that a thrown exception is captured only once if rethrown from a previous exception handler
```
Jean85: Can we restore this test so we capture the current (even if not perfect) behavior, so we stop having another debate like this in the future? :smile:

src/ErrorHandler.php
```diff
@@ -376,19 +382,17 @@ private function handleError(int $level, string $message, string $file, int $lin
...
     /**
-     * Handles fatal errors by capturing them through the Raven client. This
-     * method is used as callback of a shutdown function.
-     *
-     * @param array|null $error The error details as returned by error_get_last()
+     * Handles fatal errors by capturing them through the client. This method
+     * is used as callback of a shutdown function.
      */
-    private function handleFatalError(array $error = null): void
+    private function handleFatalError(): void
```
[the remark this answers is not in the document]
ste93cry: I think that this was a leftover of some old implementation of this method that was `public` to allow unit testing. In reality it's never called with the `$error` parameter filled, so it can be safely removed

src/ErrorHandler.php
```diff
@@ -422,36 +423,57 @@ private function handleFatalError(array $error = null): void
...
+        try {
+            $this->handleException($previousExceptionHandlerException);
+        } finally {
+            $this->shouldInvokeListeners = $shouldInvokeListeners;
+        }
+    }
+
+    private function printException(\Throwable $exception): void
+    {
+        echo 'Fatal error: Uncaught '. $exception . PHP_EOL;
```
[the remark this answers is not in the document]
ste93cry: Yes, this interface defines the `__toString` method :smiley: However, this method and a lot of other changes are gonna be reverted given that we decided to not pursue the handling of an exception thrown from the next handler in the chain after our has been executed

tests/phpt/error_handler_captures_rethrown_exception_once.phpt
ste93cry: This test has been moved to `error_handler_captures_exception_rethrown_from_previous_handler.phpt`, not sure why it got marked as deleted instead of renamed

src/ErrorHandler.php
```diff
@@ -422,36 +423,57 @@ private function handleFatalError(array $error = null): void
      */
     private function handleException(\Throwable $exception): void
     {
-        $this->invokeListeners($this->exceptionListeners, $exception);
+        if ($this->shouldInvokeListeners) {
```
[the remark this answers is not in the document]
ste93cry: No, this flag was part of a bigger series of changes needed to handle an exception thrown from the next handler in the chain after our has been executed

tests/phpt/error_handler_captures_exception_thrown_from_next_handler.phpt
ste93cry: Gonna delete this test as part of the cleanup of this PR from the changes we decided to drop

src/ErrorHandler.php
```diff
@@ -9,11 +9,9 @@
...
 /**
  * This class implements a simple error handler that catches all configured
- * error types and logs them using a certain Raven client. Registering more
- * than once this error handler is not supported and will lead to nasty problems.
- * The code is based on the Symfony Debug component.
- *
- * @author Stefano Arlandini <sarlandini@alice.it>
+ * error types and logs them using a certain client. Registering more than
+ * once this error handler is not supported and will lead to nasty problems.
+ * The code is based on the Symfony ErrorHandler component.
```
stayallive:
```suggestion
 * error types relays them to all configured listeners. Registering this 
 * error handler more than once is not supported and will lead to nasty 
 * problems. The code is based on the Symfony ErrorHandler component.
```

src/ErrorHandler.php
```diff
@@ -376,19 +376,17 @@ private function handleError(int $level, string $message, string $file, int $lin
...
         return false;
     }
 
     /**
-     * Handles fatal errors by capturing them through the Raven client. This
-     * method is used as callback of a shutdown function.
-     *
-     * @param array|null $error The error details as returned by error_get_last()
+     * Handles fatal errors by capturing them through the client. This method
+     * is used as callback of a shutdown function.
```
stayallive:
```suggestion
     * Tries to handle a fatal error if any and relay them to the listeners.
     * It only tries to do this if we still have our reserved memory on hand.
     * This method is used as callback of a shutdown function.
```

src/ErrorHandler.php
```diff
@@ -434,16 +429,18 @@ private function handleException(\Throwable $exception): void
...
+
+                return;
             }
         } catch (\Throwable $previousExceptionHandlerException) {
-            // Do nothing, we just need to set the $previousExceptionHandlerException
-            // variable to the exception we just catched to compare it later
-            // with the original object instance
+            // This `catch` statement seems to be useless, but in reality its purpose
+            // is to set the $previousExceptionHandlerException variable to the exception
+            // we just catched
```
stayallive:
```suggestion
            // This `catch` statement is here to set the $previousExceptionHandlerException
            // variable to the exception we just catched
```

src/ErrorHandler.php
```diff
@@ -434,16 +429,18 @@ private function handleException(\Throwable $exception): void
...
         }
 
-        // If the exception object instance is the same as the one catched from
-        // the previous exception handler, if any, give it back to the native
-        // PHP handler to prevent infinite circular loop
+        // If the instance of the exception we're handling is the same as the one
+        // catched from the previous exception handler then we give it back to the
+        // native PHP handler to prevent infinite circular loop
```
stayallive:
```suggestion
        // If the instance of the exception we're handling is the same as the one
        // catched from the previous exception handler then we give it back to 
        // the native PHP handler to prevent an infinite loop
```

ste93cry:
>but feel free to ignore

Nah, they are valuable suggestions and I will take the chance to also add some tests that during the refactoring made the code coverage drop a little bit due to some now untested pieces of code

Jean85:
Ping @thomasvargiu, maybe you want to test this now that it's on master.

PR #975, opened 2020-02-11
Array and string offset access syntax with curly braces is deprecated

DeyV:
I know that version 1.0 is not designated for php 7.4 but this deprecation is older .

### DIFF
--- a/lib/Raven/Client.php
+++ b/lib/Raven/Client.php
@@ -328,7 +328,7 @@ class Raven_Client
         // we need app_path to have a trailing slash otherwise
         // base path detection becomes complex if the same
         // prefix is matched
-        if ($path{0} === DIRECTORY_SEPARATOR && substr($path, -1) !== DIRECTORY_SEPARATOR) {
+        if ($path[0] === DIRECTORY_SEPARATOR && substr($path, -1) !== DIRECTORY_SEPARATOR) {
             $path .= DIRECTORY_SEPARATOR;
         }
         return $path;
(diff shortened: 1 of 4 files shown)


ste93cry:
Build is broken due to the fact that PHP < 5.6 is only available in Trusty dist. Can you please add `dist: trusty` to `.travis.yml` and see if it works?

lib/Raven/Client.php
```diff
@@ -328,7 +328,7 @@ private static function _convertPath($value)
         // we need app_path to have a trailing slash otherwise
         // base path detection becomes complex if the same
         // prefix is matched
-        if ($path{0} === DIRECTORY_SEPARATOR && substr($path, -1) !== DIRECTORY_SEPARATOR) {
+        if ($path[0] === DIRECTORY_SEPARATOR && substr($path, -1) !== DIRECTORY_SEPARATOR) {
```
Jean85: This may generate a notice with an empty string: https://3v4l.org/GVBHT

Maybe we should use `substr` here too?

DeyV: https://3v4l.org/u8HWDh - the same problem with braces

stayallive:
@DeyV sorry about having to push a commit to your branch but to me this was the easiest way to fix a bug that occurs when running with xdebug >=2.8 which your PR uncovered.

Thanks for fixing the deprecations!

stayallive:
`1.11.0` was just release which includes this PR. Thanks again!

PR #979, opened 2020-02-13
Enforce a timeout for connecting to the server and for the requests instead of waiting indefinitely

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.4
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

In #967 we decided to enforce a default value for both the timeouts of the connection to the server and for the requests to avoid the application to wait indefinitely while sending an event in the case something goes wrong. This PR is fixing this for both the HTTP clients supported out-of-the-box. To customize the values of these options it's still recommended to instantiate your own HTTP client and use it through the transport factory

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Enforce a timeout for connecting to the server and for the requests instead of waiting indefinitely (#979)
+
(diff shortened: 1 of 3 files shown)


stayallive:
Need some phpstan exclusions but other than that 🚢

Issue #910, opened 2019-10-26
 Make sending behavior of the HttpTransport "delaySendingUntilShutdown" configurable (RFC) 


bzannah:
Clearly writing custom transport is not something as easy as it seems right now since you have to configure yourself the HTTP client with all the plugins needed for authentication. 

Also since the behavior of the HttpTransport by sending events immediately instead of waiting the shutdown of the application is deprecated and has a significant impact on performance. 

For applications with use cases where there is a need to send events immediately regardless of performance, this is a big dilemma. Although it is possible using a compiler pass or an extension that accesses the transport properties using reflection, the solution may seem hacky and perhaps unnecessary. 

Therefore the best outcome could be to give make this delay configurable so anyone can use it the way they want.

mfb:
As a Sentry user, this seems like a reasonable config - although every config does add overhead to the library in terms of maintainability, testing etc.

Sending events in parallel at shutdown or when flush() is called would be a reasonable default for me, since some of my apps send multiple events to Sentry fairly often and I need to minimize execution time. But other users might prefer to send events immediately, depending on their use case.

Jean85:
FTR, this was already explained by @ste93cry in https://github.com/getsentry/sentry-php/issues/811#issuecomment-490066559 (and probably elsewhere too).

ste93cry:
>Clearly writing custom transport is not something as easy as it seems right now since you have to configure yourself the HTTP client with all the plugins needed for authentication.

This is going to change with version `2.3`. You will be able to customize the HTTP client without worring about which plugins are required to make it working with Sentry

>Also since the behavior of the HttpTransport by sending events immediately instead of waiting the shutdown of the application is deprecated and has a significant impact on performance.

The next patch release is going to rollback the change of the default behavior of sending, however as explained in the comment @Jean85 linked above the deprecation will remain. This however does not means that we cannot think about making a new transport that sends events in parallel, asyncronously or whatever we think is appropriate

nenadalm:
So as I understand it, logs were not meant to be sent when request ends and there will be some work on it. Good. I am just checking  how to enable sentry in long running process (probably overwriting the `delaySendingUntilShutdown` via reflection).

Writing just to add the reason why one might want to send logs earlier (ideally that process would run without crash until new version of the app is available).

ste93cry:
>So as I understand it, logs were not meant to be sent when request ends and there will be some work on it.

[...]

PR #981, opened 2020-02-18
Changed HttpTransport to use sync requests instead of promises

not-implemented:
The problem of the current implementation of HttpTransport is:

- In the `send()` function, the request is immediately initiated using `sendAsyncRequest()`
- This already opens file-descriptors and uses resources for the connection
- As we do not have an event-loop, the initiated connection "hangs" until the shutdown-function is reached and `cleanupPendingRequests()` finalizes the requests

For "normal" PHP scripts, this may not be a big problem, but:

- For long-running scripts the HTTP connection may **timeout in the meantime**
- When generating more events than allowed **maximum filedescriptors** (normally 1024) in one script run, then everything will crash/hang/whatever with really strange errors
- A lot of events will lead to **_much_ more memory usage** while script run (e.g. ~250MB for 1024 PHP-Notices)

This patch just puts the requests into pendingRequests directly instead of promises, and uses `sendRequest()` later instead of `sendAsyncRequest()`.

I also noticed the deprecation of "delaySendingUntilShutdown" and i saw the RFC for making it configurable in #910 - I don't know what the final goal would be - personally I prefer a configuration option as both scenarios have pro's and con's. And with this patch, the "delaySendingUntilShutdown" should lead to less problems.

### DIFF
--- a/src/Transport/HttpTransport.php
+++ b/src/Transport/HttpTransport.php
@@ -4,11 +4,12 @@ declare(strict_types=1);
 
 namespace Sentry\Transport;
 
+use GuzzleHttp\Promise\EachPromise;
 use GuzzleHttp\Promise\FulfilledPromise;
 use GuzzleHttp\Promise\PromiseInterface;
 use Http\Client\HttpAsyncClient as HttpAsyncClientInterface;
 use Http\Message\RequestFactory as RequestFactoryInterface;
-use Http\Promise\Promise as HttpPromiseInterface;
+use Psr\Http\Message\RequestInterface;
 use Sentry\Event;
 use Sentry\Exception\MissingProjectIdCredentialException;
 use Sentry\Options;
@@ -38,7 +39,7 @@ final class HttpTransport implements TransportInterface, ClosableTransportInterf
     private $requestFactory;
 
     /**
-     * @var HttpPromiseInterface[] The list of pending promises
+     * @var RequestInterface[] The list of pending requests
      */
     private $pendingRequests = [];
 
@@ -112,13 +113,11 @@ final class HttpTransport implements TransportInterface, ClosableTransportInterf
             JSON::encode($event->toArray())
         );
 
-        $promise = $this->httpClient->sendAsyncRequest($request);
-
         if ($this->delaySendingUntilShutdown) {
-            $this->pendingRequests[] = $promise;
+            $this->pendingRequests[] = $request;
         } else {
             try {
-                $promise->wait();
+                $this->httpClient->sendAsyncRequest($request)->wait();
             } catch (\Throwable $exception) {
                 return null;
             }
@@ -138,8 +137,7 @@ final class HttpTransport implements TransportInterface, ClosableTransportInterf
     }
 
     /**
-     * Cleanups the pending promises by awaiting for them. Any error that occurs
-     * will be ignored.
+     * Sends the pending requests. Any error that occurs will be ignored.
      *
      * @deprecated since version 2.2.3, to be removed in 3.0. Even though this
      *             method is `private` we cannot delete it because it's used
@@ -148,13 +146,20 @@ final class HttpTransport implements TransportInterface, ClosableTransportInterf
      */
     private function cleanupPendingRequests(): void
     {
-        while ($promise = array_pop($this->pendingRequests)) {
-            try {
-                $promise->wait();
-            } catch (\Throwable $exception) {
-                // Do nothing because we don't want to break applications while
-                // trying to send events
-            }
+        try {
+            $requestGenerator = function (): \Generator {
+                foreach ($this->pendingRequests as $key => $request) {
+                    yield $key => $this->httpClient->sendAsyncRequest($request);
+                }
+            };
+
+            $eachPromise = new EachPromise($requestGenerator(), ['concurrency' => 30]);
+            $eachPromise->promise()->wait();
+        } catch (\Throwable $exception) {
+            // Do nothing because we don't want to break applications while
+            // trying to send events
         }
+
+        $this->pendingRequests = [];
     }
 }
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -4,8 +4,6 @@ declare(strict_types=1);
 
 namespace Sentry\Tests;
 
-use Http\Discovery\MessageFactoryDiscovery;
-use Http\Mock\Client as MockClient;
 use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
 use Sentry\ClientBuilder;
@@ -15,7 +13,6 @@ use Sentry\SentrySdk;
 use Sentry\Serializer\Serializer;
 use Sentry\Severity;
 use Sentry\Stacktrace;
-use Sentry\Transport\HttpTransport;
 use Sentry\Transport\TransportFactoryInterface;
 use Sentry\Transport\TransportInterface;
 
@@ -316,26 +313,18 @@ class ClientTest extends TestCase
      */
     public function testSampleRateAbsolute(float $sampleRate): void
     {
-        $httpClient = new MockClient();
-        $options = new Options(['dsn' => 'http://public:secret@example.com/1']);
-        $options->setSampleRate($sampleRate);
-        $transportFactory = $this->createTransportFactory(new HttpTransport($options, $httpClient, MessageFactoryDiscovery::find(), true, false));
+        $transport = $this->createMock(TransportInterface::class);
+        $transport->expects(0 == $sampleRate ? $this->never() : $this->exactly(10))
+            ->method('send');
 
-        $client = (new ClientBuilder($options))
+        $transportFactory = $this->createTransportFactory($transport);
+
+        $client = (new ClientBuilder(new Options(['sample_rate' => $sampleRate])))
             ->setTransportFactory($transportFactory)
             ->getClient();
 
         for ($i = 0; $i < 10; ++$i) {
             $client->captureMessage('foobar');
-        }
-
-        switch ($sampleRate) {
-            case 0:
-                $this->assertEmpty($httpClient->getRequests());
-                break;
-            case 1:
-                $this->assertNotEmpty($httpClient->getRequests());
-                break;
         }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Hard-limit concurrent requests in `HttpTransport` and removed pre-init of promises (fixes "too many open files" errors) (#981)
 - Fix `http_proxy` option not being applied (#978)
 
 ## 2.3.1 (2020-01-23)


not-implemented:
> The promise is not resolved until the call to wait, thus the request should not even start and consequently I would expect no file descriptor to be opened until that moment.

@ste93cry: That's not correct. You are right, that the promise is not **resolved** until the `wait()`-call. But the "synchronous part" of the promise is executed immediately - that means "initiating the connection" in this case. You can easily reproduce this by generating 10 Notices:

```php
\Sentry\init(['dsn' => '...']);

$x = [];
for ($i = 0; $i < 10; $i++) {
    $x[$i]++; // Notice
}

// analyze open PHP resource handle counts by type:
foreach (get_resources() as $resource) {
    $type = get_resource_type($resource);
    if (!isset($resourceTypes[$type])) $resourceTypes[$type] = 0;
    $resourceTypes[$type]++;
}

print_r($resourceTypes);
```

Before this patch I have 10 open curl handles here:
```
Array
(
    [stream] => 63
    [stream-context] => 1
    [curl] => 10
    [curl_multi] => 1
)
```

After this patch it is much cleaner:
```
Array
(
    [stream] => 23
    [stream-context] => 1
)
```

> Also the proposed changes are breaking because you are changing all typehints from HttpAsyncClientInterface to HttpClientInterface.

Ok I see, I have rewritten the patch now to use `$this->httpClient->sendAsyncRequest($request)->wait();` instead of `sendRequest()`, so we do not need an interface change.

There is just one failing test in `Sentry\Tests\ClientTest::testSampleRateAbsolute`: It checks against `$httpClient->getRequests()` it must be rewritten to check against `HttpTransport::$pendingRequests` array, but it is a private property. Maybe you have a good idea, how this test can be rewritten in an easy way :-)

ste93cry:
>But the "synchronous part" of the promise is executed immediately

Right, I checked the code of Guzzle and it's like you says. I honestly didn't expect that though, I don't know if it may be worth to report this upstream and see if there is room for changing this behavior as it may affect users in general and not only this library

>Maybe you have a good idea, how this test can be rewritten in an easy way

What about mocking the transport instead of the HTTP client? We're only interested in how many times the transport is called, so it should be enough to write the assertion against the `TransportInterface::send` method

not-implemented:
> I don't know if it may be worth to report this upstream

I don't think so. It is the way, promises work - the Guzzle implementation is not wrong here :-)

> What about mocking the transport instead of the HTTP client?

Ah, sure ... good idea, I'll try.

mfb:
A further improvement in cleanupPendingRequests() would be to send the requests with some concurrency - you wouldn't want to try to send 1400 requests in parallel, but a max concurrency of 5 or 8 could make sense.

not-implemented:
@mfb: Maybe, yes (but keep in mind, that it is easier to "overload" the server instances then). Maybe the protocol could also be improved, in the way that multiple events can be sent in one HTTP request.

But this is out of scope of this bug-report ;-)

not-implemented:
@ste93cry: Ok I changed `testSampleRateAbsolute` now - the build succeeds now :-)

But I'm stuck with the "codecov/project" check. It does not show me, in which line coverage is missing exactly. Can you help me here?

mfb:
Well, it seems like the server overload issue was already present, since without this PR there is no limit on the number of concurrent connections that will be opened.. :)

This PR reduces the concurrency to 1 - each request will have to wait for the previous one to complete before a new connection is opened - such that multiple events will take measurably longer to send, so I think it would be good to restore some concurrency either in this PR or as a followup.

I guess the Sentry team could recommend a reasonable max concurrent requests or it could be configurable.

mfb:
In my testing, a test script that sends 20 events to sentry.io is 0.3 seconds without this PR, 1.4 seconds with this PR.

not-implemented:
@mfb: Hmm okay. I didn't keep in mind, that `wait()` on one promise also executes the others in parallel - but yes, it makes sense.

@ste93cry: It would be easy with `\GuzzleHttp\Pool`. Would it be okay, to add this dependency here?

An alternative "poor man" implementation would be, to generate "chunks" of e.g. 5 promises, which are executed in parallel with `Promise\settle($promises)->wait()` - but it would not reach a real "pool" implementation.

ste93cry:
>But I'm stuck with the "codecov/project" check. It does not show me, in which line coverage is missing exactly. Can you help me here?

I don't really have any clue about what changed that is causing the status check to fail, anyway it's not required for merge so if we don't find the root cause it's not really an issue

>Well, it seems like the server overload issue was already present, since without this PR there is no limit on the number of concurrent connections that will be opened

True story, not sure if the HTTP clients or cURL itself (most of them use it) have some hard limit, we didn't set any anyway so this is a good point to fix

>I guess the Sentry team could recommend a reasonable max concurrent requests or it could be configurable.

I would like to avoid exposing any other option related specifically to the HTTP client from the options of the SDK because that's not the place where they belongs to

>It would be easy with `\GuzzleHttp\Pool`. Would it be okay, to add this dependency here?

I'm not a real expert about this class, but as far as I see `Promise\settle($promises)` uses `EachPromise` under the hood which in turn is also used by the `GuzzleHttp\Pool` class, so they are basically the same except for the configurable concurrency limit. If we want to support batching let's use the pool at this point. As an hard limit we may use 30, which is the same default concurrency limit used in the [Javascript SDK](https://github.com/getsentry/sentry-javascript/blob/a46f7eebb22d59f47ed955f5447fe2a2ef534c4d/packages/node/src/transports/base.ts#L39)

mfb:
> True story, not sure if the HTTP clients or cURL itself (most of them use it) have some hard limit, we didn't set any anyway so this is a good point to fix

I believe the limit would be openfiles limit that @not-implemented described above - which is often 1024 by default.  In my testing, the the requests seem to hang (due to Sentry server rate limiting?) and the server could have an out-of-memory situation, with the script being killed by the OOM reaper. In addition, when encountering this openfiles limit, the script is unable to load various exception class files such as guzzle ConnectException.php

A limit of 30 concurrent connections seems fine if other sentry SDKs are using it.

not-implemented:
@ste93cry @mfb Okay, I implemented concurrency now in a new commit. I used `EachPromise` directly because `GuzzleHttp\Pool` needs a `GuzzleHttp\ClientInterface` instead of a `HttpAsyncClientInterface`. But the implementation is still simple and short and should still be fully backward-compatible.

I hardcoded the concurrency limit to 30 now ... it is much better than before in any case ... (I still recommend to reduce it to e.g. 6, because the browser limits of concurrent HTTP connections to one host are similar ... but this is up to you ;-))

src/Transport/HttpTransport.php
```diff
@@ -148,13 +146,19 @@ public function close(?int $timeout = null): PromiseInterface
...
-                $promise->wait();
-            } catch (\Throwable $exception) {
-                // Do nothing because we don't want to break applications while
-                // trying to send events
-            }
+        try {
+            $requestGenerator = function () {
+                foreach ($this->pendingRequests as $key => $request) {
+                    yield $key => $this->httpClient->sendAsyncRequest($request);
+                }
```
Jean85: Can we simplify this with a `yield from $this->pendingRequests`?

src/Transport/HttpTransport.php
```diff
@@ -148,13 +146,19 @@ public function close(?int $timeout = null): PromiseInterface
...
     {
-        while ($promise = array_pop($this->pendingRequests)) {
-            try {
-                $promise->wait();
-            } catch (\Throwable $exception) {
-                // Do nothing because we don't want to break applications while
-                // trying to send events
-            }
+        try {
+            $requestGenerator = function () {
```
ste93cry: Would you mind typehinting the return type please?

src/Transport/HttpTransport.php
```diff
@@ -148,13 +146,19 @@ public function close(?int $timeout = null): PromiseInterface
...
+                foreach ($this->pendingRequests as $key => $request) {
+                    yield $key => $this->httpClient->sendAsyncRequest($request);
+                }
+            };
+            $eachPromise = new EachPromise($requestGenerator(), ['concurrency' => 30]);
```
ste93cry: Please add a blank line above this one to improve readability

src/Transport/HttpTransport.php
ste93cry: You can't, we are using the values stored in `$pendingRequests` to pass them to the `sendAsyncRequest` method and not just yielding them

tests/ClientTest.php
```diff
@@ -316,10 +313,12 @@ public function testSendChecksBeforeSendOption(): void
      */
     public function testSampleRateAbsolute(float $sampleRate): void
     {
-        $httpClient = new MockClient();
+        $transport = $this->createMock(TransportInterface::class);
+        $transport->expects(0 == $sampleRate ? $this->never() : $this->atLeastOnce())
```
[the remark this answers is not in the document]
not-implemented: Ok fine - but as we capture 10 events, I changed it to `$this->exactly(10)` :-)

tests/ClientTest.php
```diff
@@ -316,10 +313,12 @@ public function testSendChecksBeforeSendOption(): void
...
+            ->method('send');
         $options = new Options(['dsn' => 'http://public:secret@example.com/1']);
```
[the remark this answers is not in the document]
not-implemented: Done

src/Transport/HttpTransport.php
not-implemented: Done

src/Transport/HttpTransport.php
not-implemented: Done

src/Transport/HttpTransport.php
not-implemented: Oh cool, I didn't know about `yield from` ;-) ... but yes, in this case it does not help :-)

not-implemented:
Okay, I fixed the small issues and force pushed the 3 commits (also optimized the commit messages). Should I also add a Changelog-entry like:

- Changed `HttpTransport` to not pre-initialize promises to avoid resource-/file-descriptor-overhead (#981)

?

ste93cry:
>also optimized the commit messages

Thank you, although we squash and merge so the commit messages will not be retained anyway

>Should I also add a Changelog-entry like

What you proposed is ok, another variant that came up to my mind could be ``Hard-limit the number of concurrent requests sent by the `HttpTransport` transport``, but feel free to choose whichever you think is more appropriate

not-implemented:
Ok done (+ already squashed commits together) :-) ... so from my point of view, the PR is ready :-)

ste93cry:
Woops, build failed due to an error that sometimes happen. Would you mind force-push to retrigger the CI?

not-implemented:
Haha ... done ;-)

mfb:
Looks good in my testing - performance when sending 20 events is the same as master branch.  When sending 2000 events with concurrency of 30, sentry.io pretty quickly responds with a 429 error (presumably due to hitting my project's spike protection limit), HttpTransport catches the exception and stops sending. I guess that's fine - no need to keep attempting to send? It is quite different from the current behavior, where the 429 exception is caught and then the next pending request is sent.

ste93cry:
@not-implemented, sorry to ask again to force push but you did it while GitHub was having some issues with the webhooks and so it didn’t get notified that the AppVeyor build passed

mfb:
In current code, we wait, hit exception, then wait again. With this PR, the exception bubbles up, is caught and script ends, right? To corroborate this, the script ends in ~0.5 seconds if you try to send 2000 events but already hit your quota.

mfb:
Well I might be wrong about this, because it does take longer if I send more events. Maybe it is sending but the Sentry dashboard doesn't accurately report the rate limiting, it just ignores the events.

mfb:
Confirmed that requests are in fact sent - I jumped to conclusions based on sentry dashboard. Sorry for the noise...

ste93cry:
>In current code, we wait, hit exception, then wait again. With this PR, the exception bubbles up, is caught and script ends, right?

It theory yes, but the `EachPromise` class catches any rejected promise and ensures that the processing of the pending ones continues. I just did a few tests on my own and it seems to work as described, so we are good to go. As soon as the status checks are green I will merge

mfb:
For the record, after sending thousands of events to sentry.io, it reports 845 accepted, 257 rate limited, 1.1k total. So there is some non-transparent rate limiting in place, which makes sense, as accepting all those requests even just to count them would be resource intensive.

not-implemented:
Yes, the errors should be silently catched by `EachPromise`, but it does not stop sending. Maybe the try/catch around the `EachPromise` can even be completely removed as it is really misleading.

mfb:
> Maybe the try/catch around the `EachPromise` can even be completely removed as it is really misleading.

That try/catch is necessary because most users don't want Sentry to throw an additional exception while logging.  For example, it might be logging notices, hit a 4xx error, and throw the ClientErrorException exception, which is good for knowing that your logging isn't working, but generally bad for practical purposes...

not-implemented:
@mfb: That's what I mean with "misleading" :-) ... Any errors while processing the promises (like connection errors etc.) are catched by `EachPromise` itself. Normally you define a 'fulfilled' and 'rejected' callback via the constructor-config of `EachPromise` to get the (success or error) result of the promises (I just skipped that, because the result is not interesting here). So the only errors that the try/catch in `cleanupPendingRequests()` would catch, are e.g. interface errors while calling the `EachPromise` constructor (or on any bugs in `EachPromise` itself).

(It would be very bad architecture anyway, to "hard cancel" any other half-processed requests while just one is failing :-))

Maybe I remove the try/catch here and leave a short comment line there?

not-implemented:
I had another look into `EachPromise` ... there is indeed one case when the try/catch is needed: When `sendAsyncRequest()` inside the generator throws an error because `$request` is invalid for any case, then the `EachPromise` itself is rejected and throws an error because of `wait()`.

I force-pushed again now to trigger the CI ;-)

ste93cry:
We are using the `wait()` method on the promise which has an argument with the default value of `true`, which means that the result will be unwrapped and thus if the result is an exception it will be thrown. When I tried to test that all requests were sent regardless of one failing I saw sometimes the code inside the `catch` block to be executed, and even if I didn’t investigate why it proves that it may happen. In the past we also had problems with exceptions still being thrown while calling `wait(false)`, so I would not beat that it’s really safe

Jean85:
Thanks for this PR! For the `try-catch` scenario, I would also add that we could fill the `catch` in the future when relanding #907

PR #983, opened 2020-02-24
Gzip encoding should not unnecessarily read entire stream

willemstuursma:
Reading the entire stream will cause the stream to be loaded into memory, which may take up a lot of memory.

Instead, gzip the stream on demand by wrapping it into a `\Http\Message\Encoding\GzipEncodeStream`. This removes the necessity to read the entire stream at once.


ste93cry:
Unfortunately loading the entire stream in memory was necessary [to fix](https://github.com/getsentry/sentry-php/pull/877) a PHP bug, so we cannot change this

willemstuursma:
Hm that's really too bad.

willemstuursma:
Well thanks for the feedback.

PR #984, opened 2020-02-25
Allow passing request fetcher to RequestIntegration - closes #982

### LINKED ISSUE #982: RoadRunner integration
We're using [RoadRunner](https://roadrunner.dev/) to run our website. It bootstraps multiple instances of your application (called workers) that wait for incoming requests. RoadRunner will pass a request object to the worker and the worker will have to pass back a response object. Anyway, you can find out more about how it works on the website. The implication is that you cannot use the globals `$_GET`, `$_SERVER`, etc. as the workers are just CLI processes.

I saw that you can pass a `$request` object to [RequestIntegration::applyToEvent](https://github.com/getsentry/sentry-php/blob/9ea19135d1943593a1b9903dee069f15fc503301/src/Integration/RequestIntegration.php#L87) but it isn't clear to me who calls this function and how, if at all, I can pass my own request object.

I would appreciate some help, maybe this could also be added to the documentation.

Thank you!

iluuu1994:


iluuu1994:
Build failure seems to be unrelated.

ste93cry:
What about creating an interface for the request resolver instead of using a `callable`? The pros about this is that we would have a strictly-typed contract instead of relying on static-analysis tools to enforce the correct signature of a closure

iluuu1994:
We can do that. I'm guessing in 90% of cases I'd guess the code would look something like this though:

```php
$integration = new RequestIntegration(..., function () {
    return $this->getRequest();
});
```

With an interface this would have to be something like:

```php
$integration = new RequestIntegration(..., new class($this) implements RequestFactory {
    private $outer;

    public function __construct($outer)
    {
        $this->outer = $outer;
    }

    public function createRequest(): ServerRequestInterface
    {
        return $this->outer->getRequest();
    }
});
```

Which looks significantly more complex. I don't think it's worth the extra type hinting but you call the shots 🙂

ste93cry:
>Which looks significantly more complex

It is, but relying on typehints we can avoid checking if the value returned by the `callable` is of the correct type for example, and the contract is clear instead of just being written in the PHP DocBlock. Also I imagine that if you use a framework (e.g. Symfony or Laravel) then you will likely need to retrieve the request from the DI container (or some service that must be retrieved from it), so anyway a `callable` seems a bit too much simplified for these cases. I do think that from either POVs one solution looks goor or bad, but I strongly prefer the interface for the reason mentioned above

>Build failure seems to be unrelated.

It is, we deactivated Scrutinizer for this repository but the CI build for the `develop` branch has not been updated yet to take this into account. I will do it as soon as we release the next patch version and I merge the `master` branch which works fine into `develop`

iluuu1994:
Fair enough, there you go 😉

src/Integration/RequestFactory.php
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Sentry\Integration;
+
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ * Allows customizing the request information that is attached to the logged event.
+ * An implementation of this interface can be passed to RequestIntegration.
+ */
+interface RequestFactory
```
ste93cry: What about renaming this interface to `RequestFetcherInterface`? A factory in this case seems to imply that we are really creating a request while probably we just want to fetch it from some other place

src/Integration/RequestIntegration.php
```diff
@@ -98,7 +107,11 @@ public static function applyToEvent(self $self, Event $event, ?ServerRequestInte
     private function processEvent(Event $event, Options $options, ?ServerRequestInterface $request = null): void
     {
         if (null === $request) {
-            $request = isset($_SERVER['REQUEST_METHOD']) && \PHP_SAPI !== 'cli' ? ServerRequest::fromGlobals() : null;
+            if (null !== $this->requestFactory) {
+                $request = $this->requestFactory->createRequest();
+            } elseif (isset($_SERVER['REQUEST_METHOD']) && \PHP_SAPI !== 'cli') {
+                $request = ServerRequest::fromGlobals();
```
ste93cry: This code could be moved to a default implementation of the `RequestFactory` interface so that it can be unit tested more easily. One more reason I would like to extract this little piece of code is that it's not responsibility anymore of the integration to create the server request as now there is the injected factory

tests/Integration/RequestIntegrationTest.php
```diff
@@ -442,4 +443,22 @@ private function getStreamMock(int $size, string $content = ''): StreamInterface
 
         return $stream;
     }
+
+    private function createRequestFactory(ServerRequestInterface $request): RequestFactory
+    {
+        return new class($request) implements RequestFactory {
+            /** @var ServerRequestInterface */
```
ste93cry: :pray: format the DocBlock so that it's multiline to keep consistency with the rest of the codebase

tests/Integration/RequestIntegrationTest.php
```diff
@@ -74,9 +75,9 @@ public function applyToEventWithRequestHavingIpAddressDataProvider(): array
     public function testApplyToEvent(array $options, ServerRequestInterface $request, array $expectedResult): void
     {
         $event = new Event();
-        $integration = new RequestIntegration(new Options($options));
+        $integration = new RequestIntegration(new Options($options), $this->createRequestFactory($request));
 
-        RequestIntegration::applyToEvent($integration, $event, $request);
+        RequestIntegration::applyToEvent($integration, $event);
```
ste93cry: This change is causing an unwanted code coverage drop as we're not testing anymore that the `applyToEvent` method works when a custom request is passed to it

tests/Integration/RequestIntegrationTest.php
```diff
@@ -40,9 +41,9 @@ public function testInvokeWithRequestHavingIpAddress(bool $shouldSendPii, array
         $event->getUserContext()->setData(['foo' => 'bar']);
 
         $request = new ServerRequest('GET', new Uri('http://www.example.com/fo'), [], null, '1.1', ['REMOTE_ADDR' => '127.0.0.1']);
-        $integration = new RequestIntegration(new Options(['send_default_pii' => $shouldSendPii]));
+        $integration = new RequestIntegration(new Options(['send_default_pii' => $shouldSendPii]), $this->createRequestFactory($request));
 
-        RequestIntegration::applyToEvent($integration, $event, $request);
+        RequestIntegration::applyToEvent($integration, $event);
```
ste93cry: Can you please refactor this test method so that we really test the full flow of how the integration would be executed (take a look at the `IgnoreErrorsIntegrationTest` class to see what I mean)?

tests/Integration/RequestIntegrationTest.php
iluuu1994: There's a [fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.16/src/Fixer/Phpdoc/PhpdocLineSpanFixer.php) for that.

ste93cry: Nice to know, would you be so kind to add it to the PHPCS config to avoid further issues in the future?

tests/Integration/RequestIntegrationTest.php
iluuu1994: So what's your suggestion here? Copy the test and refactor that one to match `IgnoreErrorsIntegrationTest`?

ste93cry: I don't really expect this method to change in any way, it should just pass as it did before. The method that should be refactored is `testInvokeWithRequestHavingIpAddress`, and to clarify what I meant is that it should be changed to match what have been done in the `IgnoreErrorsIntegrationTest` class

src/Integration/RequestFetcher.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\Integration;
+
+use GuzzleHttp\Psr7\ServerRequest;
+use Psr\Http\Message\ServerRequestInterface;
+
+class RequestFetcher implements RequestFetcherInterface
```
ste93cry: Please mark this class as `final` and add a small DocBlock to explain what it does

src/Integration/RequestFetcher.php
```diff
@@ -0,0 +1,18 @@
...
+{
+    public function fetchRequest(): ?ServerRequestInterface
```
ste93cry: Please add the `{@inheritdoc}` tag

src/Integration/RequestFetcher.php
```diff
@@ -0,0 +1,18 @@
...
+    {
+        return isset($_SERVER['REQUEST_METHOD']) && \PHP_SAPI !== 'cli'
```
ste93cry: Can you please use the `if` statement instead of the ternary operator?

tests/Integration/RequestIntegrationTest.php
```diff
@@ -31,20 +36,29 @@ public function testConstructorThrowsDeprecationIfPassingOptions(): void
      * @group legacy
```
ste93cry: This method should not use anything deprecated, so this line should be unnecessary

tests/Integration/RequestIntegrationTest.php
```diff
@@ -31,20 +36,29 @@ public function testConstructorThrowsDeprecationIfPassingOptions(): void
...
-     * @expectedDeprecation The "Sentry\Integration\RequestIntegration::applyToEvent" method is deprecated since version 2.1 and will be removed in 3.0.
      */
     public function testInvokeWithRequestHavingIpAddress(bool $shouldSendPii, array $expectedValue): void
     {
         $event = new Event();
         $event->getUserContext()->setData(['foo' => 'bar']);
 
         $request = new ServerRequest('GET', new Uri('http://www.example.com/fo'), [], null, '1.1', ['REMOTE_ADDR' => '127.0.0.1']);
-        $integration = new RequestIntegration(new Options(['send_default_pii' => $shouldSendPii]));
+        $integration = new RequestIntegration(new Options(['send_default_pii' => $shouldSendPii]), $this->createRequestFetcher($request));
```
ste93cry: Instead of passing the options to the constructor which is deprecated, please set them on the client

tests/Integration/RequestIntegrationTest.php
```diff
@@ -31,20 +36,29 @@ public function testConstructorThrowsDeprecationIfPassingOptions(): void
...
+        $client->expects($this->once())
+            ->method('getIntegration')
+            ->willReturn($integration);
 
-        $this->assertEquals($expectedValue, $event->getUserContext()->toArray());
+        SentrySdk::getCurrentHub()->bindClient($client);
+
+        withScope(function (Scope $scope) use ($event, $expectedValue): void {
+            $event = $scope->applyToEvent($event, []);
+            $this->assertNotNull($event);
```
ste93cry: Since you're there, can you please add a blank line above this one to improve readability?

src/Integration/RequestFetcherInterface.php
```diff
@@ -0,0 +1,16 @@
...
+
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ * Allows customizing the request information that is attached to the logged event.
+ * An implementation of this interface can be passed to RequestIntegration.
+ */
+interface RequestFetcherInterface
+{
+    public function fetchRequest(): ?ServerRequestInterface;
```
ste93cry: Please add a small docblock to explain what this method does. It's kinda repetitive and I'm going to evaluate stopping asking this, but to keep consistency until we decide to change this everything should be documented

iluuu1994:
It'd also worth considering removing `{@inheritdoc}`. I has absolutely no effect nowadays, not in PHPStorm, PHPStan or Psalm.

src/Integration/RequestIntegration.php
```diff
@@ -41,18 +41,27 @@ final class RequestIntegration implements IntegrationInterface
...
+     *
+     * @var RequestFetcherInterface
+     */
+    private $requestFetcher;
+
     /**
      * Constructor.
      *
-     * @param Options $options The client options
+     * @param Options                      $options        The client options
```
Jean85: Isn't this PHPDoc wrong?

iluuu1994: Wrong because of the missing `null`?

iluuu1994: I added the missing null type.

iluuu1994:
Also added the missing CHANGELOG.md entry.

src/Integration/RequestFetcher.php
```diff
@@ -0,0 +1,26 @@
...
+
+declare(strict_types=1);
+
+namespace Sentry\Integration;
+
+use GuzzleHttp\Psr7\ServerRequest;
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ * Default implementation for RequestFetcherInterface. Creates a request object from the PHP superglobals.
```
ste93cry: Would you be so kind to split this description so that it desn't last longer than ~80 chars per line?

src/Integration/RequestIntegration.php
```diff
@@ -41,18 +41,27 @@ final class RequestIntegration implements IntegrationInterface
      */
     private $options;
 
+    /**
+     * Fetches a PSR-7 request object for the current request.
```
ste93cry: Please move this description on the same line as the `@var` tag (if you know another PHPCS rule to enforce it feel free to add it to the config). Also, this description is describing what the instance of the property is doing rather than what this property is, so I suggest to change it to something like `PSR-7 request fetcher`

src/Integration/RequestIntegration.php
```diff
@@ -41,18 +41,27 @@ final class RequestIntegration implements IntegrationInterface
...
+     * @var RequestFetcherInterface
+     */
+    private $requestFetcher;
+
     /**
      * Constructor.
      *
-     * @param Options $options The client options
+     * @param Options|null                 $options        The client options
+     * @param RequestFetcherInterface|null $requestFetcher Fetches a PSR-7 request object for the current request
```
ste93cry: You can reuse the same description of the property, see above for the reason of this change

iluuu1994:
Sorry for the delay, I'll do this tonight.

iluuu1994:
The Psalm failure is unrelated, probably caused by a new Psalm version.

ste93cry:
Yea, you can suppress the error with `@phpcs-suppress InvalidReturnType` in the docblock of the method. Once the build is green I will merge the PR

PR #989, opened 2020-03-14
Log internal debug and error messages to a PSR-3 compatible logger

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.4
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Starting from the work done in #907, I've added support for logging some debug and error messages to a PSR-3 compatible logger. In the future we may want to log additional info, for example when we will implement rate limit handling for the `HttpTransport` transport

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -7,6 +7,7 @@ namespace Sentry;
 use GuzzleHttp\Promise\FulfilledPromise;
 use GuzzleHttp\Promise\PromiseInterface;
 use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 use Sentry\Integration\Handler;
 use Sentry\Integration\IgnoreErrorsIntegration;
 use Sentry\Integration\IntegrationInterface;
@@ -47,7 +48,7 @@ final class Client implements FlushableClientInterface
     private $eventFactory;
 
     /**
-     * @var LoggerInterface|null The PSR-3 logger
+     * @var LoggerInterface The PSR-3 logger
      */
     private $logger;
 
@@ -72,7 +73,7 @@ final class Client implements FlushableClientInterface
         $this->transport = $transport;
         $this->eventFactory = $eventFactory;
         $this->integrations = Handler::setupIntegrations($options->getIntegrations());
-        $this->logger = $logger;
+        $this->logger = $logger ?? new NullLogger();
     }
 
     /**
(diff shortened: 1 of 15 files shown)


src/Client.php
```diff
@@ -58,13 +64,15 @@ final class Client implements FlushableClientInterface
...
+     * @param LoggerInterface|null  $logger       The PSR-3 logger
      */
-    public function __construct(Options $options, TransportInterface $transport, EventFactoryInterface $eventFactory)
+    public function __construct(Options $options, TransportInterface $transport, EventFactoryInterface $eventFactory, ?LoggerInterface $logger = null)
     {
         $this->options = $options;
         $this->transport = $transport;
         $this->eventFactory = $eventFactory;
         $this->integrations = Handler::setupIntegrations($options->getIntegrations());
+        $this->logger = $logger;
```
Jean85: Using the `NullLogger` as a `null` fallback would reduce null checks a lot in our code.

```suggestion
        $this->logger = $logger ?? new NullLogger();
```

src/Client.php
```diff
@@ -156,20 +164,24 @@ public function flush(?int $timeout = null): PromiseInterface
...
      * @return Event|null returns ready to send Event, however depending on options it can be discarded
      */
     private function prepareEvent(array $payload, ?Scope $scope = null): ?Event
     {
-        $sampleRate = $this->getOptions()->getSampleRate();
+        $sampleRate = $this->options->getSampleRate();
 
         if ($sampleRate < 1 && mt_rand(1, 100) / 100.0 > $sampleRate) {
+            if (null !== $this->logger) {
+                $this->logger->info('The event will be discarded because it has been sampled.');
```
Jean85: Maybe we should add the event itself (or at least some data about it) in the logger's context?

src/Integration/Handler.php
```diff
@@ -7,6 +7,8 @@
 /**
  * This class handles the state of already installed integrations.
  * It makes sure to call {@link IntegrationInterface::setupOnce} only once per integration.
+ *
+ * @internal since version 2.4
```
Jean85: Why? Isn't `final` enough?

src/Transport/DefaultTransportFactory.php
```diff
@@ -24,16 +25,23 @@ final class DefaultTransportFactory implements TransportFactoryInterface
...
      * @param MessageFactoryInterface    $messageFactory    The PSR-7 message factory
      * @param HttpClientFactoryInterface $httpClientFactory The HTTP client factory
+     * @param LoggerInterface|null       $logger            A PSR-3 logger
      */
-    public function __construct(MessageFactoryInterface $messageFactory, HttpClientFactoryInterface $httpClientFactory)
+    public function __construct(MessageFactoryInterface $messageFactory, HttpClientFactoryInterface $httpClientFactory, ?LoggerInterface $logger = null)
     {
         $this->messageFactory = $messageFactory;
         $this->httpClientFactory = $httpClientFactory;
+        $this->logger = $logger;
```
Jean85: `NullLogger` here too.

src/Transport/HttpTransport.php
```diff
@@ -79,6 +87,7 @@ public function __construct(
         $this->httpClient = $httpClient;
         $this->requestFactory = $requestFactory;
         $this->delaySendingUntilShutdown = $delaySendingUntilShutdown;
+        $this->logger = $logger;
```
Jean85: `NullLogger` here too.

src/Transport/HttpTransport.php
```diff
@@ -146,20 +157,40 @@ public function close(?int $timeout = null): PromiseInterface
      */
     private function cleanupPendingRequests(): void
     {
+        $requestGenerator = function (): \Generator {
+            foreach ($this->pendingRequests as $key => $request) {
+                yield $key => $this->httpClient->sendAsyncRequest($request);
+            }
+        };
+
```
Jean85: Why did you move this outside? Wouldn't this actively interrupt the cleanup at the first error?

src/Integration/Handler.php
ste93cry: It isn't, marking a class as `@internal` allow us to change it however we like, included deleting it or changing the public API however we like. We cannot do it as of now anyway as the class wasn't internal before, but letting people know that it's not mean to be part of the public API is a first step towards that

src/Transport/HttpTransport.php
ste93cry: No, this is just a closure declaration so nothing gets executed. I've moved it outside just to reduce the visual clutter of the `try` block

src/Client.php
ste93cry: Indeed it makes sense. I've added the event object to the context data

src/Client.php
ste93cry: Done. A downside that I see however is that we're instantiating the logger in multiple places, which I don't really like, and also it's not really explicit of course that the logger is doing nothing

tests/ClientTest.php
```diff
@@ -336,6 +338,68 @@ public function sampleRateAbsoluteDataProvider(): array
         ];
     }
 
+    public function testProcessDiscardsEventWhenItIsSampledDueToSampleRateOption(): void
```
Jean85: Test name not clear, it seems to me nearly the same meaning of line 316

tests/ClientTest.php
```diff
@@ -336,6 +338,68 @@ public function sampleRateAbsoluteDataProvider(): array
...
+    {
+        /** @var LoggerInterface&MockObject $logger */
```
Jean85: Do we really need those PHPDoc annotations?

ste93cry: Yes, as said multiple times in the past they help the IDE in autocompleting things and stopping warnings about types mismatch. Sure, there is a plugin for PHPStorm that does it without annotations, but it has not been updated since ages so it may cease to work at some point. Also this is not a PHPStorm-specific annotation, so it may help people using other IDEs too

tests/ClientTest.php
ste93cry: I've merged together the two tests and improved their code in the process 👍

PR #991, opened 2020-03-20
Use is_iterable instead of is_array

alexgt9:
What about using `is_iterable` to allow Traversable objects to be processed automatically?

If the check is only to be able to do a foreach, I think it's safe.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Enforce a timeout for connecting to the server and for the requests instead of waiting indefinitely (#979)
 - Add `RequestFetcherInterface` to allow customizing the request data attached to the logged event (#984)
 - Log internal debug and error messages to a PSR-3 compatible logger (#989)
+- Make `AbstractSerializer` to accept `Traversable` values using `is_iterable` instead of `is_array` (#991)
 
 ## 2.3.2 (2020-03-06)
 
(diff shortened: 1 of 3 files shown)


alexgt9:
If you think that this change deserve it's existence I can add tests

ste93cry:
Yes please, once the tests are in place I will gladly merge the PR

alexgt9:
@ste93cry I just committed the test 👍

ste93cry:
Instead of merging, can you please rebase on top of `develop` and consequently change the target branch to it?

alexgt9:
Sure, I just did it, I didn't know you were using develop

tests/Util/Fixtures/TraversableSerializableClass.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\Tests\Util\Fixtures;
+
+class TraversableSerializableClass implements \IteratorAggregate
```
ste93cry: Do we really need this custom class? Can't we use directly `ArrayIterator`?

alexgt9: Actually yes, you are right

alexgt9:
> We're missing a changelog entry! Apart from that, LGTM! 👍

Ouch!, added.

guilliamxavier:
I'm sorry but this seems to break Behat, please see #1028

PR #995, opened 2020-03-31
Extract all the logic related to the DSN to its own class

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.4
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| License       | MIT

I realized that the `Options` class was getting bigger and bigger and that some logic related to how the DSN is parsed was leaking in it, breaking the single-responsibility principle. Plus, there are currently some other issues:

- There is no proper way to pass around the DSN as it's stored as a simple `string`
- There is no way to retrieve some specific information like the port or the server in an independent manner. We are currently exposing only the public key, the secret key and the host where the events should be sent to

This led to me to think about refactoring the parsing logic to extract it into a value object. A few workarounds had to be taken to maintain BC that can be safely removed in `3.0`, but the code should look clean and readable anyway

### DIFF
--- a/src/Dsn.php
+++ b/src/Dsn.php
@@ -183,7 +183,7 @@ final class Dsn
     /**
      * Gets the URL of the API endpoint to use to post an event to Sentry.
      */
-    public function getApiEndpointUrl(): string
+    public function getStoreApiEndpointUrl(): string
     {
         $url = $this->scheme . '://' . $this->host;
 
--- a/src/Dsn.php
+++ b/src/Dsn.php
@@ -1,0 +1,228 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry;
+
+/**
+ * This class represents a Sentry DSN that can be obtained from the Settings
+ * page of a project.
+ *
+ * @author Stefano Arlandini <sarlandini@alice.it>
+ */
+final class Dsn
+{
+    /**
+     * @var string The protocol to be used to access the resource
+     */
+    private $scheme;
+
+    /**
+     * @var string The host that holds the resource
+     */
+    private $host;
+
+    /**
+     * @var int The port on which the resource is exposed
+     */
+    private $port;
+
+    /**
+     * @var string The public key to authenticate the SDK
+     */
+    private $publicKey;
+
+    /**
+     * @var string|null The secret key to authenticate the SDK
+     */
+    private $secretKey;
+
+    /**
+     * @var int The ID of the resource to access
+     */
+    private $projectId;
+
+    /**
+     * @var string The specific resource that the web client wants to access
+     */
+    private $path;
+
+    /**
+     * Class constructor.
+     *
+     * @param string      $scheme    The protocol to be used to access the resource
+     * @param string      $host      The host that holds the resource
+     * @param int         $port      The port on which the resource is exposed
+     * @param int         $projectId The ID of the resource to access
+     * @param string      $path      The specific resource that the web client wants to access
+     * @param string      $publicKey The public key to authenticate the SDK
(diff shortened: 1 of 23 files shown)


src/Exception/ExceptionInterface.php
```diff
@@ -8,6 +8,8 @@
  * This interface must be implemented by all exception classes of this library.
  *
  * @author Stefano Arlandini <sarlandini@alice.it>
+ *
+ * @deprecated since version 2.4, to be removed in 3.0
```
Jean85, line 12: Why?

src/Dsn.php
```diff
@@ -0,0 +1,228 @@
...
+     */
+    public function getSecretKey(): ?string
+    {
+        return $this->secretKey;
+    }
+
+    /**
+     * Gets the URL of the API endpoint to use to post an event to Sentry.
+     */
+    public function getApiEndpointUrl(): string
```
stayallive, line 186: Maybe we want to call this something like `getEventApiEndpointUrl` or even `getEventStoreApiEndpointUrl` since this method name seems a bit too generic for the specific endpoint it returns.

You might want to add a `$path` paramater to this method and add the `getEventStoreApiEndpointUrl` method too which call this one with a path (or leave that method out and call this one with a path from the http transport. This so we can use it to "generate" other API endpoints too. 

Little bit of future proofing for upcoming features we need to call API endpoints for.

src/Exception/ExceptionInterface.php
ste93cry: Because not all our exception classes implement it and I'm not sure it has some value in keeping it. But I'm open to reverting the change if you think it's worth

src/Dsn.php
ste93cry: I decided to call the method `getStoreApiEndpointUrl`, based on some similarity with what has been done in the Go SDK. 

>You might want to add a $path paramater to this method

I don't think that it's worth adding it now that we don't even have a case to justify it, plus if we will ever have the need of generating more endpoint URLs then it will probably be better to rely on separate methods.

src/Exception/ExceptionInterface.php
Jean85: If not all exceptions are implementing it, I think that that's the error that we should fix.

src/Dsn.php
stayallive: Fair enough 👌 Good change!

src/Exception/ExceptionInterface.php
ste93cry: The point is, does it makes sense to keep this interface? What are the pros of having it and what are the cons of dropping it?

Jean85: The advantage is giving the possibility to us and to the end user to catch all exceptions that we throw. I don't see downsides in keeping it.

ste93cry: Well then, I reverted the change 👍

Kocal:
Hi!

Currently there is no _easy_ way to get the Sentry option `dsn` without building the DSN manually: 
Sélection_590

I can't access `Options#options['dsn']` since it's private (in fact I can with some reflections or a Closure, but it's a no! :D), and building the DSN myself seems a bit risky. I can lose some data.

The best option is `$options->getDsn(false)->__toString()` from this PR to get the full DSN.

Do you know when a new release is planned? 
Thanks you!

ste93cry:
>Do you know when a new release is planned?

I'm sorry but there is no clear roadmap on when new versions are going to be released. You can either lock your `composer.json` to the branch/specific commit or wait for the release of version `2.4`

PR #1003, opened 2020-04-26
Fix the context_lines option not skipping reading the source code if its value is set to null

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.3
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

I noticed that setting the `context_lines` option to ~~`0`~~ `null` does not prevent the source code from being read and attached to each frame of a stacktrace, but the documentation says it should work this way. In addition to this, the validation was missing leading to bad things when such option was set to a value lower than zero, so I took the time to fix this too.

### DIFF
--- a/tests/Integration/ModulesIntegrationTest.php
+++ b/tests/Integration/ModulesIntegrationTest.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Sentry\Tests\Integration;
 
-use Jean85\PrettyVersions;
 use PHPUnit\Framework\TestCase;
 use Sentry\Event;
 use Sentry\Integration\ModulesIntegration;
@@ -20,8 +19,6 @@ final class ModulesIntegrationTest extends TestCase
 
         $modules = $event->getModules();
 
-        $this->assertArrayHasKey('sentry/sentry', $modules, 'Root project missing');
-        $this->assertArrayHasKey('ocramius/package-versions', $modules, 'Indirect dependency missing');
-        $this->assertEquals(PrettyVersions::getVersion('sentry/sentry'), $modules['sentry/sentry']);
+        $this->assertNotEmpty($modules);
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix the reading of the source code when the `context_lines` option is set to 0 and add missing validation (#1003)
+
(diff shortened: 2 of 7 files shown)


ste93cry:
>so, `null` is "Do not upload my source code" where instead `0` is "upload only the offending line, no context lines".

While you are right that the documentation says a different thing, and from what I see the Go SDK follows the same reasoning, I don't think that it really makes sense that setting the option to `0` still sends the offending line. As there is there is no distinction for the user between pre/post context lines and the context line itself, as a user I would not expect that `0` means `1` and `null` means `0`, wouldn't you?

Jean85:
It seems we're not agreeing on the meaning of "context". Context [is the surrounding of the subject](https://www.merriam-webster.com/dictionary/context), so in this case the lines above and below the one in the stacktrace. That's why I consider that the correct approach.

@HazAT can you weight in here? Is this interpretation the right and meant one?

ste93cry:
>Context [is the surrounding of the subject](https://www.merriam-webster.com/dictionary/context), so in this case the lines above and below the one in the stacktrace.

After reading the linked article I now think that you are proven right, but to be on the safe side let's wait anyway for a final word by @HazAT

HazAT:
So the `context_lines` are really focused around the `pre_context` and `post_context`. So setting this to `0` on makes `pre_context` & `post_context` empty.
`null` should not even send the offending line.

tests/Integration/ModulesIntegrationTest.php
```diff
@@ -20,8 +19,6 @@ public function testInvoke(): void
 
         $modules = $event->getModules();
 
-        $this->assertArrayHasKey('sentry/sentry', $modules, 'Root project missing');
-        $this->assertArrayHasKey('ocramius/package-versions', $modules, 'Indirect dependency missing');
-        $this->assertEquals(PrettyVersions::getVersion('sentry/sentry'), $modules['sentry/sentry']);
+        $this->assertNotEmpty($modules);
```
Jean85, line 22: Why did you suppress everything here? Last line could survive.

ste93cry: Yes, it could. However I don't think that it's worth keeping it because it doesn't add any value to the test and also because I don't want to the test to depend anymore on the names of the dependencies or of the root package because if those changes then our build shoud not break like it did in the past

PR #1008, opened 2020-05-01
jean85/pretty-package-versions <1.4

hebbet:
jean85/pretty-package-versions version 1.4 dropped support for composer 1.x

currently running "composer outdated --strict -m" returns that version is available, but it is not installed as `jean85/pretty-package-versions 1.4.0 requires composer-runtime-api ^2.0 -> no matching package found.`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "ext-mbstring": "*",
         "guzzlehttp/promises": "^1.3",
         "guzzlehttp/psr7": "^1.6",
-        "jean85/pretty-package-versions": "^1.2",
+        "jean85/pretty-package-versions": ">=1.2 <1.4",
         "php-http/async-client-implementation": "^1.0",
         "php-http/client-common": "^1.5|^2.0",
         "php-http/discovery": "^1.6.1",


hebbet:
>  If you're using Composer 1, it will install up to 1.3; 

yes, but `composer outdated --strict -m` fails and suggest to update to 1.4

we are using that check in many scripts for deploying stuff.

ste93cry:
Imho Composer should not show such update if it's not compatible with the requirements. @Seldaek sorry to bring you in, but I would like your opinion on whether this is a bug or something that we should really fix here

Seldaek:
I have to investigate if this can be fixed.. will get back to you in the coming week.

Jean85:
IMHO that could be fixed with `--ignore` as a workaround: https://github.com/composer/composer/issues/8317#issuecomment-542028840

Seldaek:
Yes --ignore is definitely a way to shove this under the rug for now :)

https://github.com/composer/composer/pull/8871 fixes it though, will be included in the 1.10.6 release which I hopefully can wrap up tomorrow.

ste93cry:
Wow, thank you for fixing this so fast! I will leave this PR open until @hebbet confirms it works

hebbet:
thanks for the bugfix in composer. closing

Seldaek:
1.10.6 is out now

PR #1009, opened 2020-05-03
Update PHPStan to version 0.12

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.4
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| License       | MIT

As the title says I updated PHPStan to version `0.12`. I was able to also deprecate the usage of the serializer in the `Stacktrace` class, which is a step forward making it a value object rather than the current mess that I'm not even sure can be called a service. I also stopped executing PHPStan on the tests folder: there are ~186 errors to solve, the most of which are about missing typehints, but I'm not sure it's worth solving them as it would imply adding a lot of docblocks which are rather useless


tests/ClientTest.php
```diff
@@ -462,50 +461,6 @@ public function convertExceptionDataProvider(): array
...
-        $serializer->setMbDetectOrder('ISO-8859-1, ASCII, UTF-8');
-
-        $client = ClientBuilder::create()
-            ->setTransportFactory($this->createTransportFactory($transport))
-            ->setSerializer($serializer)
-            ->getClient();
-
-        $client->captureException(require_once __DIR__ . '/Fixtures/code/Latin1File.php');
-    }
-
```
Jean85, line 508: Why is this removed?

ste93cry: It should be useless because of the `JSON` class now forcing the encoding of all strings to UTF-8 regardless of the PHP version

PR #1011, opened 2020-05-05
Add the FrameContextifierIntegration integration

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.4
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| License       | MIT

Porting of the [contextifyFramesIntegration integration](https://github.com/getsentry/sentry-go/blob/550a179aaa1655092edfba5d2ea4bfc1690736e5/integrations.go#L162) of the Go SDK. The main benefit of this change is that we're moving away from the `Stacktrace` class a lot of code that doesn't belongs there, advancing one more step towards making it a value object and DTO rather than a service that depends on the serializer, the client options, etc.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -8,43 +8,22 @@ parameters:
         - '/Argument of an invalid type object supplied for foreach, only iterables are supported/'
         - "/^Parameter #1 \\$function of function register_shutdown_function expects callable\\(\\): void, 'register_shutdown…' given\\.$/"
         -
-            message: /^Cannot assign offset 'os' to array\|string\.$/
-            path: src/Event.php
-        -
-            message: '/^array\|string does not accept array<string, string>\.$/'
-            path: src/Event.php
-        -
             message: '/^Argument of an invalid type array\|object supplied for foreach, only iterables are supported\.$/'
             path: src/Util/JSON.php
         -
             message: '/^Access to constant (?:PROXY|TIMEOUT|CONNECT_TIMEOUT) on an unknown class GuzzleHttp\\RequestOptions\.$/'
             path: src/HttpClient/HttpClientFactory.php
         -
-            message: '/^Parameter #2 \$str of function explode expects string, int\|string given\.$/'
-            path: src/Dsn.php
-        -
-            message: '/^Parameter #1 \$haystack of function strrpos expects string, int\|string given\.$/'
-            path: src/Dsn.php
-        -
-            message: '/^Parameter #1 \$str of function substr expects string, int\|string given\.$/'
-            path: src/Dsn.php
-        -
-            message: '/^Parameter #2 \$host of class Sentry\\Dsn constructor expects string, int\|string given\.$/'
-            path: src/Dsn.php
-        -
-            message: '/^Parameter #3 \$port of class Sentry\\Dsn constructor expects int, int\|string given\.$/'
-            path: src/Dsn.php
-        -
-            message: '/^Parameter #5 \$path of class Sentry\\Dsn constructor expects string, int\|string given\.$/'
-            path: src/Dsn.php
-        -
-            message: '/^Parameter #6 \$publicKey of class Sentry\\Dsn constructor expects string, int\|string given\.$/'
-            path: src/Dsn.php
-        -
-            message: '/^Parameter #7 \$secretKey of class Sentry\\Dsn constructor expects string\|null, int\|string\|null given\.$/'
-            path: src/Dsn.php
-        -
             message: '/^Parameter #1 \$c of function ctype_digit expects int\|string, string\|null given\.$/'
+            path: src/Dsn.php
+        -
+            message: "/^Offset 'scheme' does not exist on array\\(\\?'scheme' => string, \\?'host' => string, \\?'port' => int, \\?'user' => string, \\?'pass' => string, \\?'path' => string, \\?'query' => string, \\?'fragment' => string\\)\\.$/"
+            path: src/Dsn.php
+        -
+            message: "/^Offset 'path' does not exist on array\\(\\?'host' => string, \\?'port' => int, \\?'user' => string, \\?'pass' => string, \\?'path' => string, \\?'query' => string, \\?'fragment' => string, 'scheme' => 'http'|'https'\\)\\.$/"
+            path: src/Dsn.php
+        -
+            message: "/^Offset 'user' does not exist on array\\('scheme' => 'http'\\|'https', \\?'host' => string, \\?'port' => int, \\?'user' => string, \\?'pass' => string, \\?'path' => string, \\?'query' => string, \\?'fragment' => string\\)\\.$/"
             path: src/Dsn.php
         -
             message: '/^Method Sentry\\Monolog\\Handler::write\(\) has parameter \$record with no value type specified in iterable type array\.$/'
@@ -54,6 +33,12 @@ parameters:
             path: src/Serializer/RepresentationSerializer.php
         -
             message: '/^Method Sentry\\Client::getIntegration\(\) should return T of Sentry\\Integration\\IntegrationInterface\|null but returns Sentry\\Integration\\IntegrationInterface\|null\.$/'
+            path: src/Client.php
+        -
+            message: '/^Method Sentry\\EventFactoryInterface::createWithStacktrace\(\) invoked with 2 parameters, 1 required\.$/'
+            path: src/Client.php
+        -
+            message: '/^Method Sentry\\EventFactoryInterface::create\(\) invoked with 2 parameters, 1 required\.$/'
             path: src/Client.php
     excludes_analyse:
         - tests/resources
--- a/src/Client.php
+++ b/src/Client.php
@@ -8,6 +8,7 @@ use GuzzleHttp\Promise\FulfilledPromise;
 use GuzzleHttp\Promise\PromiseInterface;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
+use Sentry\Integration\FrameContextifierIntegration;
 use Sentry\Integration\Handler;
 use Sentry\Integration\IgnoreErrorsIntegration;
 use Sentry\Integration\IntegrationInterface;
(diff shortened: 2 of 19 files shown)


src/Client.php
```diff
@@ -167,8 +167,6 @@ public function flush(?int $timeout = null): PromiseInterface
      *
      * @param array<string, mixed> $payload The payload that will be converted to an Event
      * @param Scope|null           $scope   Optional scope which enriches the Event
-     *
-     * @return Event|null returns ready to send Event, however depending on options it can be discarded
```
Jean85, line 171: Why is this removed and not simply fixed?

src/EventFactory.php
```diff
@@ -61,15 +61,18 @@ public function __construct(SerializerInterface $serializer, RepresentationSeria
...
-
-            $event->setStacktrace($stacktrace);
+        if (!isset($payload['stacktrace']) || !$payload['stacktrace'] instanceof Stacktrace) {
+            $payload['stacktrace'] = Stacktrace::createFromBacktrace(
+                $this->options,
+                $this->serializer,
+                $this->representationSerializer,
+                debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS),
+                __FILE__,
+                __LINE__ - 6
```
Jean85, line 71: I would create this value as `__LINE__ + 1` one line before this call, to make it more robust and less change-prone while refactoring.

src/Integration/FrameContextifierIntegration.php
```diff
@@ -0,0 +1,163 @@
...
+     * @param LoggerInterface $logger       A PSR-3 logger
+     */
+    public function __construct(int $contextLines = 5, ?LoggerInterface $logger = null)
+    {
+        if ($contextLines < 0) {
+            throw new \InvalidArgumentException(sprintf('The value of the $maxLinesToFetch argument must be greater than or equal to 0. Got: "%d".', $contextLines));
+        }
+
+        $this->contextLines = $contextLines;
+        $this->logger = $logger;
```
Jean85, line 44: Please use the `NullLogger` here to simplify usage below.

src/Integration/FrameContextifierIntegration.php
```diff
@@ -0,0 +1,163 @@
...
+            if (null !== $event->getStacktrace()) {
+                $integration->addContextToStacktraceFrames($event->getStacktrace());
+            }
+
+            foreach ($event->getExceptions() as $exception) {
+                if (!isset($exception['stacktrace'])) {
+                    continue;
+                }
+
+                $integration->addContextToStacktraceFrames($exception['stacktrace']);
```
Jean85, line 78: Should you not check the type before passing it in?

src/Options.php
```diff
@@ -937,6 +951,24 @@ private function normalizeDsnOption(SymfonyOptions $options, $value): ?Dsn
...
+     * @deprecated since version 2.4, to be removed in 3.0
+     */
+    private function normalizeContextLinesOption(SymfonyOptions $options, ?int $value): ?int
+    {
+        if (null !== $value) {
+            @trigger_error('The "context_lines" option is deprecated since version 2.4 and will be removed in 3.0. Use the "Sentry\Integration\FrameContextifierIntegration" integration instead.', E_USER_DEPRECATED);
+        }
+
+        return $value;
+    }
```
Jean85, line 970: Dropping the options doesn't seem sensible to me. It would be a lot more difficult to change this default value.

src/Client.php
ste93cry: Nothing to fix, I just though it was kinda useless to explain why this method may return either an object or `null` as the official documentation is pretty clear itself and we don't usually document the return values. Do you think it adds value and therefore I should revert this?

src/EventFactory.php
ste93cry: Actually it doesn't really matter because:

- If we need to add one more argument to the method we still have to update this line taking into account the number of new lines
- It's safer to rely on the `-` because we cannot remove any argument to avoid BC, the only thing that can change is that we revert the formatting of this code to one-line again which is unlikely to happen at this point
- It's more appropriate to point to the line where the function name actually is rather than on its closing parenthesis

src/Integration/FrameContextifierIntegration.php
ste93cry: I will, but it's becoming more and more common to instantiate a `NullLogger` in every place where we now accept the logger and, even though I'm pretty sure it doesn't hurts performances, I don't think we should continue doing this rather than checking for `null` to avoid one more object

src/Integration/FrameContextifierIntegration.php
ste93cry: It's not worth it because:

- I plan to phase out all these arrays with value objects instead to solve the issue upstream
- If the type is wrong PHP will complain due to the typehinting of the parameter of the function, and imho it's enough. We have to remember that this information is coming from the internals of the event factory, so it's unlikely that is wrong

src/Options.php
ste93cry: I knew we would have discussed about this :grimacing: There are quite a few reasons I want to deprecate this option:

- It's one less option to hande around the codebase, to validate, normalize, etc
- Since the feature is now controlled with an integration, it doesn't make sense at all to have a SDK option that affects a specific integration
- By adding or removing the integration you control whether you want to send the context lines or not, therefore making the special handling of a `null` value useless and this in turn makes the feature clearer about how it works (if something isn't enabled then obviously it does nothing)
- The Go SDK does not have such option as well, and this implementation is just a porting of the way it's handled there here

src/Client.php
Jean85: It's just for consistency, since all other params are documented.

src/EventFactory.php
Jean85: I was not suggesting to change it inline, rather to create a temp variable before the call, and reference the call itself with the `+1`. This way we no longer have to worry about any of the concerns that you pointed out.

src/Options.php
Jean85: I would agree on this, but we have one drawback: this should be a default enabled integration. So you would have to re-add it to make it work? How do we handle adding an integration twice?

This does not seem smooth to me..

src/EventFactory.php
ste93cry: >This way we no longer have to worry about any of the concerns that you pointed out

It could very well be that I'm not understanding your concerns, but I cannot see how your proposal would solve the issue: even creating a temporary variable (which I would like to avoid as it's pretty useless) right before the function call, if someone adds a new line between them the sum will not be valid regardless. In the end, it will always matter how many lines there are between __LINE__ and the line where you call the function

src/Options.php
ste93cry: >I would agree on this, but we have one drawback: this should be a default enabled integration

It is at the time of writing, in the future we may play safer and not send by default the context lines as there may be some organizations not willing to share their source code with external applications, but this is out of topic here of course

>How do we handle adding an integration twice?

We shouldn't anyway handle it probably as there may be some legitimate case for having the same integration added more than once per client (I'm not talking specifically about default integrations)

src/EventFactory.php
Jean85: Sure. But putting a line in between those two calls is calling for trouble and you would have to do it intentionally. Instead, with this situation, this could break up with automatic CS fixes too.

BTW, I would be fine with either if we have a unit test that checks this, it's an important information to get right.

src/Options.php
Jean85: Still, it would become really complicated just to switch this off. I'm still :-1: on removing the option.

src/EventFactory.php
ste93cry: >But putting a line in between those two calls is calling for trouble and you would have to do it intentionally

Also adding a new argument or line before this or removing one argument must be done intentionally, so either way we always go back to the same conclusion

>I would be fine with either if we have a unit test that checks this

I don't think it's possible to test this as it would mean hardcode the line number into the test itself, which is something I'm pretty sure neither of us wants. We never bothered before about this kind of things, so I strongly think that it's worthless continuing the discussion

src/Options.php
ste93cry: As @HazAT said that for the moment I should refrain from deprecating the option, I reverted the changes :confused:

phpstan.neon
```diff
@@ -7,44 +7,23 @@ parameters:
...
-            message: '/^Parameter #6 \$publicKey of class Sentry\\Dsn constructor expects string, int\|string given\.$/'
+            message: "/^Offset 'scheme' does not exist on array\\(\\?'scheme' => string, \\?'host' => string, \\?'port' => int, \\?'user' => string, \\?'pass' => string, \\?'path' => string, \\?'query' => string, \\?'fragment' => string\\)\\.$/"
             path: src/Dsn.php
         -
-            message: '/^Parameter #7 \$secretKey of class Sentry\\Dsn constructor expects string\|null, int\|string\|null given\.$/'
+            message: "/^Offset 'path' does not exist on array\\(\\?'host' => string, \\?'port' => int, \\?'user' => string, \\?'pass' => string, \\?'path' => string, \\?'query' => string, \\?'fragment' => string, 'scheme' => 'http'|'https'\\)\\.$/"
             path: src/Dsn.php
         -
-            message: '/^Parameter #1 \$c of function ctype_digit expects int\|string, string\|null given\.$/'
+            message: "/^Offset 'user' does not exist on array\\('scheme' => 'http'\\|'https', \\?'host' => string, \\?'port' => int, \\?'user' => string, \\?'pass' => string, \\?'path' => string, \\?'query' => string, \\?'fragment' => string\\)\\.$/"
```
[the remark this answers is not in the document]
ste93cry: I'm ignoring these errors because PHPStan is not able to detect a `isset` statement inside a `loop`, so they are all false positives

src/Stacktrace.php
```diff
@@ -18,6 +18,16 @@ class Stacktrace implements \JsonSerializable
 {
     private const INTERNAL_FRAME_FILENAME = '[internal]';
 
+    /**
+     * @var bool Flag indicating whether it's responsibility of this class to
+     *           read the source code excerpts for each frame
+     *
+     * @internal
+     *
+     * @deprecated since version 2.4, to be removed in 3.0
```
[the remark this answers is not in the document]
ste93cry: >Why deprecated?

Because its only purpose is to trigger the deprecation error when users try to fill the context lines of a frame using this class instead of the newer integration.

>why `protected`?

Because the rest of the class methods are `protected` and if someone overridden any of them he may want to access the property, even if it's marked as `internal`. Do you think it's not worth keeping it as is?

phpstan.neon
```diff
@@ -7,44 +7,23 @@ parameters:
...
-            path: src/Dsn.php
-        -
-            message: '/^Parameter #2 \$host of class Sentry\\Dsn constructor expects string, int\|string given\.$/'
-            path: src/Dsn.php
-        -
-            message: '/^Parameter #3 \$port of class Sentry\\Dsn constructor expects int, int\|string given\.$/'
-            path: src/Dsn.php
-        -
-            message: '/^Parameter #5 \$path of class Sentry\\Dsn constructor expects string, int\|string given\.$/'
+            message: '/^Parameter #1 \$c of function ctype_digit expects int\|string, string\|null given\.$/'
```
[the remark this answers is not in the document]
ste93cry: I agree, but it's a false-positive :disappointed:

src/Stacktrace.php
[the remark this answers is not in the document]
Jean85: I'm undecided. Feel free to do as you wish here.

PR #1012, opened 2020-05-05
Allow to pass either a PSR-17 or a Httplug stream factory to the constructor of the GzipEncoderPlugin class

### LINKED ISSUE #1006: GzipEncoderPlugin incorrectly claims to use PSR-17 stream factory
https://github.com/getsentry/sentry-php/blob/f7645df23d00a1102db3f8ca5de997763e599a21/src/HttpClient/Plugin/GzipEncoderPlugin.php#L21

The doc comment says that the plugin is using a PSR-17 stream factory. However it is actually using [`Http\Message\StreamFactory`](https://github.com/getsentry/sentry-php/blob/f7645df23d00a1102db3f8ca5de997763e599a21/src/HttpClient/Plugin/GzipEncoderPlugin.php#L9) which is something else. PSR-17 factory is `\Psr\Http\Message\StreamFactoryInterface`.

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.4
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes (kind of)
| License       | MIT

Fixes #1006 by allowing to pass to the constructor of the `GzipEncoderPlugin` class either a PSR-17 stream factory or Httplug's stream factory. In `3.0` we hopefully can drop support for Httplug `1.x` and rely exclusively on the PSR standards

### DIFF
--- a/src/HttpClient/Plugin/GzipEncoderPlugin.php
+++ b/src/HttpClient/Plugin/GzipEncoderPlugin.php
@@ -6,9 +6,10 @@ namespace Sentry\HttpClient\Plugin;
 
 use Http\Client\Common\Plugin as PluginInterface;
 use Http\Discovery\StreamFactoryDiscovery;
-use Http\Message\StreamFactory as StreamFactoryInterface;
+use Http\Message\StreamFactory as HttplugStreamFactoryInterface;
 use Http\Promise\Promise as PromiseInterface;
 use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\StreamFactoryInterface;
 
 /**
  * This plugin encodes the request body by compressing it with Gzip.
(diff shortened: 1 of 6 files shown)


src/HttpClient/Plugin/GzipEncoderPlugin.php
```diff
@@ -18,25 +19,27 @@
...
+    public function __construct($streamFactory = null)
     {
         if (!\extension_loaded('zlib')) {
             throw new \RuntimeException('The "zlib" extension must be enabled to use this plugin.');
         }
 
         if (null === $streamFactory) {
             @trigger_error(sprintf('A PSR-17 stream factory is needed as argument of the constructor of the "%s" class since version 2.1.3 and will be required in 3.0.', self::class), E_USER_DEPRECATED);
+        } elseif (!$streamFactory instanceof HttplugStreamFactoryInterface && !$streamFactory instanceof StreamFactoryInterface) {
+            throw new \InvalidArgumentException(sprintf('The $streamFactory argument must be an instance of either the "%s" or the "%s" interface.', HttplugStreamFactoryInterface::class, StreamFactoryInterface::class));
```
Jean85, line 42: I would like to throw a different message when HTTPlug is used here, to deprecate its direct usage. Is it possible?

ste93cry: Yes it is, I didn't because of a few reasons:

- We still didn't decide if we want to stop supporting Httplug `1.x` in `3.0` regardless of the switch to PSR classes, although I'm strongly in favor of doing it
- Deprecating Httplug here means that it will be the only place in the entire codebase where it's discouraged to rely on the legacy Httplug factories, which sounds strange and incoherent. We can treat is as the starting point for deprecating Httplug `1.x`, but we must all agree on this
- Deprecating Httplug here means that we have to handle skipping the throwing of the error if the plugin is instantiated from Sentry itself, and to do it we have to rely on `boolean` flags like we do in other places of the codebase. It's fine because we will have to do it at some point if we decide to deprecate the support of Httplug `1.x`, but again we have to decide this before imho

Jean85: Ok I'm fine on postponing the deprecation discussion and merge this as is.

PR #1013, opened 2020-05-05
Encapsulate the event ID into its own value object

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.4
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes (kind of)
| License       | MIT

Small refactor to start encapsulating the ID of an event inside its own value object. The idea behind this change is to typehint this information in `3.0` in all methods that accepts or returns it instead of accepting a generic `string` value that noone can guarantee to be valid without validating it every time

### DIFF
--- a/src/Event.php
+++ b/src/Event.php
@@ -172,12 +172,10 @@ final class Event implements \JsonSerializable
      *
      * @return string|EventId
      */
-    public function getId(bool $returnAsString = true, bool $throwDeprecation = true)
+    public function getId(bool $returnAsString = true)
     {
         if ($returnAsString) {
-            if ($throwDeprecation) {
-                @trigger_error(sprintf('Calling the method %s() and expecting it to return a string is deprecated since version 2.4 and will stop working in 3.0.', __METHOD__), E_USER_DEPRECATED);
-            }
+            @trigger_error(sprintf('Calling the method %s() and expecting it to return a string is deprecated since version 2.4 and will stop working in 3.0.', __METHOD__), E_USER_DEPRECATED);
 
             return (string) $this->id;
         }
--- a/src/Event.php
+++ b/src/Event.php
@@ -19,7 +19,7 @@ use Sentry\Context\UserContext;
 final class Event implements \JsonSerializable
 {
     /**
-     * @var string The UUID
+     * @var EventId The ID
      */
     private $id;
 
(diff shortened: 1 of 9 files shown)


src/EventId.php
```diff
@@ -0,0 +1,48 @@
...
+     * Generates a new event ID.
+     */
+    public static function generate(): self
+    {
+        return new self(str_replace('-', '', uuid_create(UUID_TYPE_RANDOM)));
+    }
+
+    /**
+     * @see https://www.php.net/manual/en/language.oop5.magic.php#object.tostring
+     */
```
Jean85, line 43: This PHPDoc seems superfluous to me
```suggestion
```

src/Event.php
```diff
@@ -173,7 +172,7 @@ public function __construct()
      */
     public function getId(): string
     {
-        return $this->id;
+        return (string) $this->id;
     }
```
Jean85, line 176: Maybe we could deprecate this method and add a `getEventId`?

src/EventId.php
```diff
@@ -0,0 +1,48 @@
...
+declare(strict_types=1);
+
+namespace Sentry;
+
+/**
+ * This class represents an event ID.
+ *
+ * @author Stefano Arlandini <sarlandini@alice.it>
+ */
+final class EventId
```
stayallive, line 12: Would possibly be nice to also add a `getId` getter to this class so you don't have to use a cast or call `__toString()`.

ste93cry: It's possible of course, but I don't see a real value in having it because of a few reasons:

- what are the advantages of calling `getId` which returns a `string` vs casting to `string` the object?
- the value object represents the ID, so it doesn't really makes sense imho to have a method that basically returns itself

src/Event.php
ste93cry: I though about this, but I didn't like it because what would happen is `$event->getEventId()` which is redundant. Instead, I propos to deprecate the returning of a `string` from this method like we did for `Options::getDsn()` e.g.:

```php
if ($returnAsString) {
    return (string) $this->id;
} else {
    return $this->id;
}
```

Jean85: That's doable, but only if we suppress all deprecations around in our code too.

ste93cry: Obviously 👍 Done in the latest commit

src/Event.php
```diff
@@ -170,9 +169,19 @@ public function __construct()
 
     /**
      * Gets the UUID of this event.
+     *
+     * @return string|EventId
      */
-    public function getId(): string
+    public function getId(bool $returnAsString = true, bool $throwDeprecation = true)
```
Jean85, line 175: Why adding that second parameter? Calling `getId(true, false)` would NOT throw a deprecation but break in the next major. IMHO calling  `getId(false)->__toString()` obtains the same but avoiding the risk of silent future breakage.

ste93cry: >Why adding that second parameter?

I wrongly assumed that it was needed to make some tests work, but you're right in saying that `getId(false)->__toString()` is more than enough to not trigger any deprecation in the unit tests and thus it's useless

PR #1018, opened 2020-05-15
Fix BC break in Stacktrace::getSourceCodeExcerpt() method caused by #1003

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 2.3
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

After merging I realized that #1003 caused a BC in the `Stacktrace::getSourceCodeExcerpt()` method which I didn't notice at the time that was `protected`. Technically speaking there is also another BC in the `Options::getContextLines` method which can now return `null`, but I think that since it's necessary to fix the bug it's unavoidable and for this reason should not be treated as a real BC

### DIFF
--- a/src/Stacktrace.php
+++ b/src/Stacktrace.php
@@ -141,18 +141,21 @@ class Stacktrace implements \JsonSerializable
         }
 
         $frame = new Frame($functionName, $this->stripPrefixFromFilePath($file), $line);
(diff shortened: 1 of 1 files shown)


ste93cry:
There is no reason to mention that I fixed a BC break because the code that caused it is not released yet. For the BC break of the `Options` class, being it part of the bug fix of #1003 I don’t think it’s worth mentioning it

PR #1020, opened 2020-06-03
feat: Span and SpanContext Class

HazAT:
Adds Span and SpanContext class

### DIFF
--- a/src/EventId.php
+++ b/src/EventId.php
@@ -9,12 +9,12 @@ namespace Sentry;
  *
  * @author Stefano Arlandini <sarlandini@alice.it>
  */
-final class EventId
+class EventId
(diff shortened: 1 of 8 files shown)


src/EventId.php
```diff
@@ -9,12 +9,12 @@
  *
  * @author Stefano Arlandini <sarlandini@alice.it>
  */
-final class EventId
+class EventId
```
Jean85, line 12: I would prefer to not do this. This would mean being able to use a `SpanId` in place of an `EventId`, which should not be possible.

HazAT:
The test failures have nothing to do with this PR -> this will be merged into the collection PR/branch.

PR #1022, opened 2020-06-04
feat: Add Transaction and TransactionContext

HazAT:


src/Event.php
```diff
@@ -24,10 +25,17 @@ final class Event implements \JsonSerializable
...
+     * @var string|float The date and time of when this event was generated
      */
     private $timestamp;
 
+    /**
+     * This property is used if it's a Transaction event together with $timestamp it's the duration of the transaction.
+     *
+     * @var string|float|null The date and time of when this event was generated
+     */
+    private $startTimestamp;
```
Jean85, line 37: Instead of having such a polymorphic behavior, maybe we can split it in two class, with a possible common ancestor? WDYT?

src/Event.php
```diff
@@ -674,6 +709,12 @@ public function toArray(): array
             $data['breadcrumbs']['values'] = $this->breadcrumbs;
         }
 
+        if (!empty($this->spans)) {
+            $data['spans'] = array_map(function ($span) {
```
ste93cry, line 713: The `Closure` can be `static`, and it's missing both the typehint of the parameter and of the return value

src/Tracing/Transaction.php
```diff
@@ -0,0 +1,108 @@
...
+     * @param HubInterface|null       $hub     Instance of a hub to flush the transaction
+     *
+     * @internal
+     */
+    public function __construct(?TransactionContext $context = null, ?HubInterface $hub = null)
+    {
+        parent::__construct($context);
+        $this->hub = $hub;
+
+        $this->name = $context->name ?? '<unlabled transaction>';
```
ste93cry, line 38: Typo: `unlabled` instead of `unlabeled`

src/Tracing/Transaction.php
```diff
@@ -0,0 +1,108 @@
...
+     * @param TransactionContext|null $context The context to create the transaction with
+     * @param HubInterface|null       $hub     Instance of a hub to flush the transaction
+     *
+     * @internal
+     */
+    public function __construct(?TransactionContext $context = null, ?HubInterface $hub = null)
+    {
+        parent::__construct($context);
+        $this->hub = $hub;
+
```
ste93cry, line 37: Please remove this blank line and put it before the first variable assignation instead

src/Tracing/Transaction.php
```diff
@@ -0,0 +1,108 @@
...
+        $this->name = $name;
+    }
+
+    /**
+     * @return array<string, mixed> Returns the trace context
+     */
+    public function getTraceContext(): array
+    {
+        $data = $this->toArray();
+        unset($data['start_timestamp']);
```
ste93cry, line 55: Please add  blank line above this one to improve readability and context-switching

src/Tracing/Transaction.php
```diff
@@ -0,0 +1,108 @@
...
+
+        parent::finish($endTimestamp);
+
+        if (true !== $this->sampled) {
+            // At this point if `sampled !== true` we want to discard the transaction.
+            return null;
+        }
+
+        $hub = $this->hub;
+        if (null !== $hub) {
```
ste93cry, line 79: No need to reassign the variable, just always use `$this->`

src/Tracing/TransactionContext.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\Tracing;
+
+class TransactionContext extends SpanContext
```
ste93cry, line 7: Please make this class `final`

src/Tracing/TransactionContext.php
```diff
@@ -0,0 +1,13 @@
...
+{
+    /**
+     * @var string|null Name of the transaction
+     */
+    public $name;
```
ste93cry, line 12: Please make this property `private` and add the appropriate `getter` and `setter`

src/Tracing/SpanRecorder.php
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Adds a span to the array.
+     */
+    public function add(Span $span): void
+    {
+        if (\count($this->spans) > $this->maxSpans) {
+            $span->spanRecorder = null;
+        } else {
+            array_push($this->spans, $span);
```
ste93cry, line 38: Please use the `[]` operator to push to the array instead of `array_push`

src/Tracing/Span.php
```diff
@@ -15,57 +15,64 @@ class Span implements \JsonSerializable
...
      */
-    private $endTimestamp;
+    protected $endTimestamp;
+
+    /**
+     * @var SpanRecorder|null Reference instance to the SpanRecorder
+     *
+     * @internal
+     */
+    public $spanRecorder;
```
ste93cry, line 75: What's the reason of storing the recorder here? We're creating a circular reference between it and the span itself, plus I cannot find any usage of this variable in this class

src/Event.php
```diff
@@ -596,6 +611,22 @@ public function setType(?string $type): void
         $this->type = $type;
     }
 
+    /**
+     * @param string|float|null $startTimestamp The start time of the event
+     */
+    public function setStartTimestamp($startTimestamp): void
+    {
+        $this->startTimestamp = $startTimestamp;
```
ste93cry, line 619: If we really want to support the union as typehint, we should throw an exception if the variable is not of the expected type

src/Tracing/Transaction.php
```diff
@@ -0,0 +1,108 @@
...
+    public function jsonSerialize(): array
+    {
+        $event = new Event();
+        $event->setType('transaction');
+        $event->getTagsContext()->merge($this->tags->toArray());
+        $event->setTransaction($this->name);
+        $event->setStartTimestamp($this->startTimestamp);
+        $event->setContext('trace', $this->getTraceContext());
+        $spanRecorder = $this->spanRecorder;
+        if (null != $spanRecorder) {
```
ste93cry, line 100: Please add a blank line both before this line and after the closing brace

src/Tracing/SpanRecorder.php
HazAT: Rusty PHP :) 
Fixed

src/Tracing/Span.php
HazAT: I added implementation now, at least part of it.

src/Tracing/TransactionContext.php
HazAT: For now we want to have simple objects carrying the properties.
Having getter/setters produces a lot more code, let's see we can still add this later if we really want to.

ste93cry: This is a bad practice, plus in PHP we don't have typed properties before PHP 7.4 so there is nothing enforcing the value to be of the expected type. Also, if in the future we want to change the implementation of the class we cannot do it because everything is `public`

PR #1025, opened 2020-06-09
feat: startTransaction

HazAT:


src/State/HubInterface.php
```diff
@@ -135,4 +137,23 @@ public static function setCurrent(self $hub): self;
...
+     * Every child span must be finished before the transaction is finished,
+     * otherwise the unfinished spans are discarded.
+     *
+     * The transaction must be finished with a call to its `finish()` method, at
+     * which point the transaction with all its finished child spans will be sent to
+     * Sentry.
+     *
+     * @param TransactionContext $context properties of the new `Transaction`
+     */
+    public function startTransaction(TransactionContext $context): Transaction;
```
Jean85, line 158: This is a breaking change.

ste93cry: The method declaration should be moved to the annotation (see the `ClientBuilderInterface` for an example), and we should also check whether the method really exists with `method_exists` before calling it

Jean85: Or we could extend this interface with a new one that would be merged in the next major.

HazAT: Adding something shouldn't be a breaking change not for users using the SDK.
It might be for someone extending the SDK but they should be aware of when we are changing.

ste93cry: Adding anything to an interface is a breaking change, regardless of how this interface is used or from whom. That's the rule 😉

HazAT: That's just not true. How is this breaking for someone that is using the SDK?
No code change needed for any user.

This is breaking for anyone that implements the interface, yes but this is us internally and anyone who extends the SDK.

ste93cry: >This is breaking for anyone that implements the interface, yes but this is us internally and anyone who extends the SDK

The definition of BC change cannot be used on the base that the code may or may not be used by the user. Unless the class, the function or whatever is marked as `@internal`, then it is part of the public API and we must respect this, regardless of whether or not someone may be extending it

Jean85: I agree with @ste93cry. As a bare minimum, as a user I'm using the interface for test, and this will break them for me.

It will probably break the integrations (Symfony, Laravel) tests too.

PR #1030, opened 2020-06-23
Revert "Add support for iterables in the serializer (#991)"

### LINKED ISSUE #1028: After 2.3.2->2.4.0 update, Behat stops on the first feature involving an exception
Refs https://github.com/Behat/Behat/issues/1302

After updating `sentry/sentry` to 2.4.0 (from 2.3.2), when Behat encounters a feature that [contains a scenario that] involves an exception (for example
```gherkin
    When I send a "GET" request to "/dummy"
    Then the response status code should be 404
```
with Symfony), then it stops after this file, not running the remaining features in the test suite.

I've tracked the regression down to commit 498a41cc1dbf3c403c10a098d8994af6848f3685 (PR #991).

After some debugging, it appears that the problem occurs when `$value` is an instance of
`\Behat\Behat\Gherkin\Specification\LazyFeatureIterator` and/or
`\Behat\Testwork\Specification\GroupedSpecificationIterator`
(I guess the serialization exhausts the current suite iterator).

guilliamxavier:
Fixes #1028 "After 2.3.2->2.4.0 update, Behat stops on the first feature involving an exception"

Reverts #991 "Use is_iterable instead of is_array" and adds a non-regression test

### DIFF
--- a/tests/Serializer/AbstractSerializerTest.php
+++ b/tests/Serializer/AbstractSerializerTest.php
@@ -65,6 +65,41 @@ abstract class AbstractSerializerTest extends TestCase
         $this->assertSame(['key' => 'value'], $result);
     }
 
+    public function traversableDataProvider(): \Generator
+    {
+        $getArray = static function (): array {
+            return [
+                'a' => 'foo',
+                'b' => 'bar',
+            ];
+        };
+
+        yield [$getArray(), new \ArrayIterator($getArray())];
+
+        // Also test with a non-rewindable non-cloneable iterator:
+        $newGenerator = static function (array $array): \Generator {
+            yield from $array;
+        };
+
+        yield [$getArray(), $newGenerator($getArray())];
+    }
+
+    /**
+     * @dataProvider traversableDataProvider
+     */
+    public function testTraversablesAreNotConsumed(array $array, \Traversable $traversable): void
+    {
+        $serializer = $this->createSerializer();
+
+        $traversed = [];
+        foreach ($traversable as $k => $v) {
+            $traversed[$k] = $v;
+            $this->invokeSerialization($serializer, $traversable);
+        }
+
+        $this->assertSame($array, $traversed);
+    }
+
     /**
      * @dataProvider serializeAllObjectsDataProvider
      */
--- a/tests/Serializer/AbstractSerializerTest.php
+++ b/tests/Serializer/AbstractSerializerTest.php
@@ -65,32 +65,46 @@ abstract class AbstractSerializerTest extends TestCase
         $this->assertSame(['key' => 'value'], $result);
     }
 
-    public function traversableDataProvider(): \Generator
-    {
-        yield [new \ArrayIterator(['foo', 'bar']), ['foo', 'bar']];
-
-        // Also test with a non-rewindable non-cloneable iterator:
-        $newGenerator = static function (array $array): \Generator {
-            yield from $array;
-        };
-
-        yield [$newGenerator(['foo', 'bar']), ['foo', 'bar']];
-    }
-
     /**
-     * @dataProvider traversableDataProvider
+     * @dataProvider iterableDataProvider
      */
-    public function testTraversablesAreNotConsumed(\Traversable $traversable, array $array): void
+    public function testIterablesAreNotConsumed(iterable $iterable, array $input): void
     {
         $serializer = $this->createSerializer();
+        $output = [];
 
-        $traversed = [];
-        foreach ($traversable as $k => $v) {
-            $traversed[$k] = $v;
-            $this->invokeSerialization($serializer, $traversable);
+        foreach ($iterable as $k => $v) {
+            $output[$k] = $v;
+
+            $this->invokeSerialization($serializer, $iterable);
         }
 
-        $this->assertSame($array, $traversed);
+        $this->assertSame($input, $output);
+    }
+
+    public function iterableDataProvider(): \Generator
+    {
+        yield [
+            'iterable' => ['value1', 'value2'],
+            'input' => ['value1', 'value2'],
+        ];
+
+        yield [
+            'iterable' => new \ArrayIterator(['value1', 'value2']),
+            'input' => ['value1', 'value2'],
+        ];
+
+        // Also test with a non-rewindable non-cloneable iterator:
+        yield [
+            'iterable' => (static function (): \Generator {
+                yield 'value1';
+                yield 'value2';
+            })(),
+            'input' => [
+                'value1',
+                'value2',
+            ],
+        ];
     }
 
     /**
--- a/tests/Serializer/AbstractSerializerTest.php
+++ b/tests/Serializer/AbstractSerializerTest.php
@@ -67,27 +67,20 @@ abstract class AbstractSerializerTest extends TestCase
 
     public function traversableDataProvider(): \Generator
     {
-        $getArray = static function (): array {
-            return [
-                'a' => 'foo',
-                'b' => 'bar',
-            ];
-        };
-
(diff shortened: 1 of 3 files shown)


guilliamxavier:
About broken CI, why is b586900f6912a69c7743dbed268981b10500e586 only on `develop` and cd80b9b681b92716e807a08a19a84d0c3355a79a only on `hazat/am` (not `master`)?

tests/Serializer/AbstractSerializerTest.php
```diff
@@ -65,6 +65,41 @@ public function testObjectsAreNotStrings(): void
         $this->assertSame(['key' => 'value'], $result);
     }
 
+    public function traversableDataProvider(): \Generator
+    {
+        $getArray = static function (): array {
```
ste93cry, line 70: What's the purpose of this anonymous function? I suggest to either inline the array in every place where it's used (easiest solution and also probably the cleanest given how little it is) or to declare it outside the function which to me looks as a pretty useless wrapper

tests/Serializer/AbstractSerializerTest.php
```diff
@@ -65,6 +65,41 @@ public function testObjectsAreNotStrings(): void
...
+            return [
+                'a' => 'foo',
```
ste93cry, line 72: I think that for this test it's enough to use a simple list instead of an hashmap

tests/Serializer/AbstractSerializerTest.php
```diff
@@ -65,6 +65,41 @@ public function testObjectsAreNotStrings(): void
...
+
+    /**
+     * @dataProvider traversableDataProvider
+     */
+    public function testTraversablesAreNotConsumed(array $array, \Traversable $traversable): void
+    {
+        $serializer = $this->createSerializer();
+
+        $traversed = [];
+        foreach ($traversable as $k => $v) {
```
ste93cry, line 95: Not sure about what you are trying to achieve here, but it seems wrong to iterate the traversable and call for each iteration the serialization of the traversable itself: just compare the expected result with the call to `$this->invokeSerialization($serializer, iterator_to_array($traversable))`

tests/Serializer/AbstractSerializerTest.php
guilliamxavier: I had started with `$array = [ /* ... */ ];` (and use `$array` in place of `$getArray()` below) but then wanted to make sure to get a new copy each time (to avoid any risk of messing the internal pointer). I also wanted to avoid code duplication, but I will follow your suggestion and inline it :+1:

tests/Serializer/AbstractSerializerTest.php
guilliamxavier: You're right (and that'll make it even easier to inline) :+1:

tests/Serializer/AbstractSerializerTest.php
guilliamxavier: I'm reproducing #1028: Behat traverses the features through an iterator (taken as a function argument), and for each feature, if a scenario triggers a catchable exception (like the expected `NotFoundHttpException` in my case of testing an HTTP 404), it is captured by Sentry, which ends up serializing the stacktrace (that includes the iterator).
So here I don't care whether the serialization result is `[1, 2, 3]` or `'Object ArrayIterator'`, I just want to test that the serialization of the iterator doesn't cause the iteration to end prematurely (by default, i.e. not enabling `serializeAllObjects`). Do you see a better way to achieve that?

guilliamxavier:
(rebased on top of `master` updated with #1032, CI green)

tests/Serializer/AbstractSerializerTest.php
ste93cry: >Do you see a better way to achieve that?

Nothing worth changing the test logic itself. I just asked for a few CS fixes, and then LGTM

tests/Serializer/AbstractSerializerTest.php
```diff
@@ -65,6 +65,34 @@ public function testObjectsAreNotStrings(): void
...
+    }
+
+    /**
+     * @dataProvider traversableDataProvider
+     */
+    public function testTraversablesAreNotConsumed(\Traversable $traversable, array $array): void
+    {
+        $serializer = $this->createSerializer();
+
+        $traversed = [];
```
ste93cry, line 87: Please move the blank line above this one below this one

tests/Serializer/AbstractSerializerTest.php
```diff
@@ -65,6 +65,34 @@ public function testObjectsAreNotStrings(): void
         $this->assertSame(['key' => 'value'], $result);
     }
 
+    public function traversableDataProvider(): \Generator
+    {
+        yield [new \ArrayIterator(['foo', 'bar']), ['foo', 'bar']];
```
ste93cry, line 70: Can you please split each yielded value on its own line to improve readability?

tests/Serializer/AbstractSerializerTest.php
```diff
@@ -65,6 +65,34 @@ public function testObjectsAreNotStrings(): void
...
+
+        // Also test with a non-rewindable non-cloneable iterator:
+        $newGenerator = static function (array $array): \Generator {
+            yield from $array;
+        };
+
+        yield [$newGenerator(['foo', 'bar']), ['foo', 'bar']];
```
ste93cry, line 77: Can you please split each yielded value on its own line to improve readability? Also, you can inline the generator by wrapping the lambda with round brackets and yielding the values directly from inside it

tests/Serializer/AbstractSerializerTest.php
```diff
@@ -65,6 +65,34 @@ public function testObjectsAreNotStrings(): void
...
+     * @dataProvider traversableDataProvider
+     */
+    public function testTraversablesAreNotConsumed(\Traversable $traversable, array $array): void
+    {
+        $serializer = $this->createSerializer();
+
+        $traversed = [];
+        foreach ($traversable as $k => $v) {
+            $traversed[$k] = $v;
+            $this->invokeSerialization($serializer, $traversable);
```
ste93cry, line 90: Please add a blank line above this one to improve readability

tests/Serializer/AbstractSerializerTest.php
```diff
@@ -65,6 +65,34 @@ public function testObjectsAreNotStrings(): void
...
+            yield from $array;
+        };
+
+        yield [$newGenerator(['foo', 'bar']), ['foo', 'bar']];
+    }
+
+    /**
+     * @dataProvider traversableDataProvider
+     */
+    public function testTraversablesAreNotConsumed(\Traversable $traversable, array $array): void
```
ste93cry, line 83: As previously we were serializing any `iterable` types I think that it would be more approprite to change the typehint accordingly. Also :pray: change `$array` to `$input` to keep consistency with the rest of the test methods of this class

tests/Serializer/AbstractSerializerTest.php
```diff
@@ -65,6 +65,34 @@ public function testObjectsAreNotStrings(): void
         $this->assertSame(['key' => 'value'], $result);
     }
 
+    public function traversableDataProvider(): \Generator
```
ste93cry, line 68: Please move this method below the test method, and replace the `traversable` word with `iterable`

guilliamxavier:
@ste93cry: Thank you for the careful review. Does it look good now?

ste93cry:
It does! Thank you for your effort, LGTM

Jean85:
We forgot the changelog entry :(

ste93cry:
Ops, I will never stop forgetting to update the CHANGELOG with each PR :disappointed: Anyway, I will do it once we prepare the release

PR #1034, opened 2020-06-29
[FIX] Add http timeouts even if no http_proxy is used

### LINKED ISSUE #1033: http_timeout and connection_timeout only working with a http proxy
This pull request proposed to add timeouts so that long running requests (e.g. your sentry is offline) do not block the whole app.
https://github.com/getsentry/sentry-php/pull/967

It got changed and merged:
https://github.com/getsentry/sentry-php/pull/979

But the current version does not work as intended:
If you do not have a http_proxy the timeouts do not get set, because `getHttpProxy()` is `null`.
https://github.com/getsentry/sentry-php/blob/b2cfb8b9e524727db63592fd5183e7a48058bb73/src/HttpClient/HttpClientFactory.php#L113

etlam:
Fixes #1033

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix timeouts for connecting to the server and for the requests only being applied if a http proxy was specified (#1033)
+
 ### 2.4.0 (2020-05-21)
 
 - Enforce a timeout for connecting to the server and for the requests instead of waiting indefinitely (#979)
--- a/src/HttpClient/HttpClientFactory.php
+++ b/src/HttpClient/HttpClientFactory.php
@@ -112,11 +112,9 @@ final class HttpClientFactory implements HttpClientFactoryInterface
 
         if (null === $httpClient) {
             if (class_exists(GuzzleHttpClient::class)) {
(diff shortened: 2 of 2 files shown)


etlam:
I don*t know how to UndefinedClass-error, maybe someone can help

ste93cry:
>I don't know how to suppress the UndefinedClass error in psalm

Just move the annotation before the declaration of the `$guzzleConfig` variable and drop all the remaining unnecessary annotations

>I would prefer the second, since it would always work, in local too

We could install those dependencies as part of `require-dev`, but then it would be harder to test the SDK with another HTTP client as the order in which the clients are discovered is not something we can decide. Even moving the installation of suchs deps to the CI means that we still have to suppress the errors locally

etlam:
> Just move the annotation before the declaration of the `$guzzleConfig` variable and drop all the remaining unnecessary annotations

Thanks!

How can I help to get the patch into the next release?

ste93cry:
Would you mind adding also an entry to the CHANGELOG? I always forget about asking this 🤦 

>How can I help to get the patch into the next release?

As soon as we merge it it will be already part of the next patch release which I would like to release shortly after

PR #1037, opened 2020-07-04
Cleanup the deprecations related to the error handler

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| License       | MIT

As the title suggests, I'm just cleaning up the codebase of `3.x` from the deprecations introduced in #788 related to the error handler


src/Options.php
```diff
@@ -815,7 +815,7 @@ private function configureOptions(OptionsResolver $resolver): void
                 return $event;
             },
             'tags' => [],
-            'error_types' => E_ALL,
+            'error_types' => error_reporting(),
```
stayallive, line 818: We should also document this change in the upgrade docs since this might be unexpected.

ste93cry: Nice catch. I would like to open a discussion about retaining this option, but for the moment I agree that this change should be documented

PR #1044, opened 2020-07-06
Remove deprecated methods from the Options class related to getting the DSN

ste93cry:
| Q             | A
| ------------- | ---
| Branch?       | 3.0
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| License       | MIT

Cleanup of deprecated methods of the `Options` class introduced in #995


src/Dsn.php
```diff
@@ -211,23 +217,17 @@ protected function getBaseEndpointUrl(): string
...
 
         return $url;
     }
 
     /**
-     * @see https://www.php.net/manual/en/language.oop5.magic.php#object.tostring
+     * Returns the base url to Sentry from the DSN.
      */
-    public function __toString(): string
+    private function getBaseEndpointUrl(): string
```
Jean85, line 228: This change is not documented

ste93cry: I just moved the functions around, there is no change to document

Jean85: You made it private, but it's final (I just noticed), so it doesn't count.

PR #1048, opened 2020-07-14
Ability to pass fingerprint via Monolog context

voodoo-dn:
Hi.

Will be great if developer can pass fingerprint for Sentry event via Monolog context.

`$this->logger->error('some error', ['fingerprint' => ['group','by','that']]);`

Thanks for yours time.

### DIFF
--- a/src/Monolog/Handler.php
+++ b/src/Monolog/Handler.php
@@ -70,6 +70,10 @@ final class Handler extends AbstractProcessingHandler
                 }
             }
 
+            if (isset($record['context']['fingerprint']) && \is_array($record['context']['fingerprint'])) {
+                $scope->setFingerprint($record['context']['fingerprint']);
+            }
+
             $this->hub->captureEvent($payload);
         });
     }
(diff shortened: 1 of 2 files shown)


stayallive:
Oh, we would like to request one more change, can you add an entry to the `CHANGELOG.md`?

ste93cry:
A long time ago we decided that we don't want to pull out data from unstructured context and I'm not sure for this reason that this change, although legit and looking good, should be merged

voodoo-dn:
@Jean85 I can say the same about **extra**, **tags** fields in context. Maybe we can choose another name of fingerprint key?

ste93cry:
In fact, somewhere in the past we discussed about removing support for those keys in the next major as we cannot do any BC break before that and it's highly likely we will do it. The solution to your problem is to create a decorator for the handler that uses `withScope` to set the fingerprint on the scope

ste93cry:
Closing as solution has been provided and in `3.0` we removed the remaining bits that extracted non-obvious data from the Monolog context

PR #1051, opened 2020-07-16
Add regression test

Jean85:
This is just a regression test. I may not have the time to fix it, feel free to push it forward.

### DIFF
--- a/tests/phpt/1050_silenced_error_regression.phpt
+++ b/tests/phpt/1050_silenced_error_regression.phpt
@@ -7,6 +7,8 @@ declare(strict_types=1);
 
 namespace Sentry\Tests;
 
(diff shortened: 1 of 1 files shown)


mfb:
I think Sentry is creating a FatalErrorException for E_COMPILE_WARNING (and E_CORE_WARNING) because an error handler function cannot handle these error types.  I don't think it should though, as these are by definition not fatal errors, so it just creates confusion for the user.

mfb:
By the way, this error is also captured by Sentry 1.x, but as an ErrorException with level "warning", rather than a FatalErrorException with level "fatal".

Jean85:
Closing as already fixed in #1183

Issue #1041, opened 2020-07-06
Sentry seems to be breaking with php-http/curl-client


wagnerjsilva:
This issue has now started to happen in one of our system, and the stack trace points to sentry.

Any ideas? 

```
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Http\Client\Curl\Client::__construct() must be an instance of Psr\Http\Message\ResponseFactoryInterface or null, instance of Nyholm\Psr7\Factory\HttplugFactory given, called in /home/wagner_silva/testing-engine/vendor/sentry/sentry/src/HttpClient/HttpClientFactory.php on line 139 and defined in /home/wagner_silva/testing-engine/vendor/php-http/curl-client/src/Client.php:79
Stack trace:
#0 /home/wagner_silva/testing-engine/vendor/sentry/sentry/src/HttpClient/HttpClientFactory.php(139): Http\Client\Curl\Client->__construct(Object(Nyholm\Psr7\Factory\HttplugFactory), Object(Nyholm\Psr7\Factory\HttplugFactory), Array)
getsentry/sentry#1 /home/wagner_silva/testing-engine/vendor/sentry/sentry/src/Transport/DefaultTransportFactory.php(58): Sentry\HttpClient\HttpClientFactory->create(Object(Sentry\Options))
getsentry/sentry#2 /home/wagner_silva/testing-engine/vendor/sentry/sentry/src/ClientBuilder.php(307): Sentry\Transport\DefaultTransportFactory->create(Object(Sentry\Options))
getsentry/sentry#3 /home/wagner_silva/testing in /home/wagner_silva/testing-engine/vendor/php-http/curl-client/src/Client.php on line 79

untitaker:
Please file SDK issues against the correct repo.

wagnerjsilva:
Hi Unitaker,

Not sure where else I'm supposed to file this issue.

I'm a sentry customer, and the UI pointed me to log this issue here.

I've also logged a support request, as Sentry is part of a reporting system, which is broken because of this problem.

I'm not entirely sure it's related to Sentry, but the stack trace leads me to believe so.

ste93cry:
This is the almost the same error being reported in #1039, thus I think that the solution is the same. Would you mind checking if you are using an old version of `sentry/sdk` and in case update it?

wagnerjsilva:
It's indeed very similar. 

I've tried removing sentry all together and putting it back following the documented instructions in GH.

[...]

Issue #1039, opened 2020-07-05
[v2.4.1] ResponseFactoryInterface argument error


MGatner:
Hi Sentry! I updated via Composer yesterday from `2.4.0` to `2.4.1` and my entire dev site went down. Logs are below, but it looks like an issue of argument type inheritance upstream from Guzzle? Regardless I'm surprised a minor point upgrade like this would cause a failure with no other implementation changes so I thought I would report it.

Workaround for now is to leave the SDK definition `"sentry/sdk": "2.0.3"` but specify the older framework: `"sentry/sentry": "2.4.0"`. At some point I would like to update though so a fix or explanation would be greeted warmly.

PHP: 7.2.31 (shared hosting 😭)
OS: Ubuntu 18.04
Server: Apache

```
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Http\Client\Curl\Client::__construct() must be an instance of Psr\Http\Message\ResponseFactoryInterface or null, instance of Http\Message\MessageFactory\GuzzleMessageFactory given, called in /var/www/project/vendor/sentry/sentry/src/HttpClient/HttpClientFactory.php on line 139 and defined in /var/www/project/vendor/php-http/curl-client/src/Client.php:79

Stack trace:
#0 /var/www/project/vendor/sentry/sentry/src/HttpClient/HttpClientFactory.php(139): Http\Client\Curl\Client->__construct(Object(Http\Message\MessageFactory\GuzzleMessageFactory), Object(Http\Message\StreamFactory\GuzzleStreamFactory), Array)

#1 /var/www/project/vendor/sentry/sentry/src/Transport/DefaultTransportFactory.php(58): Sentry\HttpClient\HttpClientFactory->create(Object(Sentry\Options))

#2 /var/www/project/vendor/sentry/sentry/src/ClientBuilder.php(307): Se in /var/www/project/vendor/php-http/curl-client/src/Client.php on line 79
```

ste93cry:
Would you mind sharing your `composer.json` and the code you use to initialize the SDK if it's different from relying on `Sentry\init()`?

MGatner:
Sure. I shared the relevant parts of **composer.json** above. The project is not open source but here's the redacted version in JSON format:
```
{
    "name": "project/project",
    "type": "project",
    "description": "Project",
    "minimum-stability": "dev",
    "prefer-stable": true,
    "require": {
        "php": ">=7.2",
        "twbs/bootstrap": "^4.3",
[...]

PR #1052, opened 2020-07-16
Fix passing incorrect PSR factories to curl client

stayallive:
We are passing our internal response and stream factories to the curl http client, which are deprecated, however it expects actual PSR namespaced classes we are not using yet internally (because BC).

Setting them to null causes the curl client to discover it's own fixing: #1039 & #1041.

I think this should be released as `1.4.2` as soon as we can 😄

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix typehint errors while instantiating the Httplug cURL client by forcing the usage of PSR-17 complaint factories (#1052)
+
 ### 2.4.1 (2020-07-03)
 
 - Fix HTTP client connection timeouts not being applied if an HTTP proxy is specified (#1033)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix incorrectly instantiating `php-http/curl-client` causing argument errors (#1052)
```
ste93cry, line 5: What about `Fix typehint errors while instantiating the Httplug cURL client by forcing the usage of PSR-17 complaint factories`?

stayallive: That actually sounds pretty good, adjusted!

richardhj:
Hi 👋
When do you release the next version? The current release is blocking my deployment atm 🙂

stayallive:
@richardhj `2.4.2` was released this morning so you should be good to go.

PR #1055, opened 2020-07-24
Fix unreliable test comparing timestamps

stayallive:
The amount of times this test fails annoys me and since there is no easy way for us to restart tests it's just better to fix the "problem".

Example (but there are many): https://ci.appveyor.com/project/sentry/sentry-php/builds/34276992/job/mi61ululink3kbfl

No changelog needed, just a QoL improvement.

### DIFF
--- a/tests/EventTest.php
+++ b/tests/EventTest.php
@@ -12,6 +12,7 @@ use Sentry\Event;
 use Sentry\Severity;
 use Sentry\Stacktrace;
 use Sentry\Util\PHPVersion;
+use Symfony\Bridge\PhpUnit\ClockMock;
 
 /**
  * @group time-sensitive
@@ -40,6 +41,8 @@ final class EventTest extends TestCase
 
     public function testToArray(): void
     {
+        ClockMock::register(Event::class);
+
         $event = new Event();
 
         $expected = [
@@ -70,6 +73,8 @@ final class EventTest extends TestCase
 
     public function testToArrayMergesCustomContextsWithDefaultContexts(): void
     {
+        ClockMock::register(Event::class);
+
         $event = new Event();
         $event->setContext('foo', ['foo' => 'bar']);
         $event->setContext('bar', ['bar' => 'foo']);


stayallive:
Feel free to make it more reliable some other way but it must be fixed one way or another.

Since we're using internal PHP functions there is no real use in testing that the value is the exact same, we are testing Sentry code not if `gmdate` is doing it's job. 

One other thing we could do is regex the value to make sure it adheres to the expected formatting but if that also wouldn't be acceptable please come up with a way to make this test reliable.

ste93cry:
>Since we're using internal PHP functions there is no real use in testing that the value is the exact same, we are testing Sentry code not if `gmdate` is doing it's job.

As for any other value being asserted, this one is no different and we should compare that the current time equals the value returned by the `getTimestamp()` method. We don't care about how such value is computed of course, as it's part of the internal implementation, but we care about the value being the one we expect

>Feel free to make it more reliable some other way but it must be fixed one way or another

What about adding an optional `$occurredAt` argument to the constructor so that we can generate an event with a fixed date?

O.T. I wonder if Symfony builds fail too so often as ours for this issue :thinking:

stayallive:
Also fine with that (good option to have overal), made that change instead.

And I have no clue if Symfony or others have similar issues and to be honest don't think it's important. Our test suite should work depending on the code and not the time it's running at, possibly others can live with this, I for one cannot 😄

ste93cry:
>And I have no clue if Symfony or others have similar issues and to be honest don't think it's important

>We could check the timestamp compared before and after the test call, or use Symfony's ClockMock as an alternative

We're using it, this is why I was wondering if the Symfony builds are broken as often as ours...

ste93cry:
@stayallive after looking at the documentation of `ClockMock`, I think that I've misused it and what is being mocked is the `Sentry\Tests` namespace while the function `gmdate` is called from the `Sentry` namespace. To solve the issue it should be enough to call `ClockMock::register(Event::class)` at the beginning of the test method.

https://symfony.com/doc/current/components/phpunit_bridge.html#time-sensitive-tests
>The `@group time-sensitive` annotation is equivalent to calling `ClockMock::register(MyTest::class)`. If you want to mock a function used in a different class, do it explicitly using `ClockMock::register(MyClass::class)`:

stayallive:
I've registered the `Event` class with `ClockMock` as suggested by @ste93cry, this looks correct to me too, but I guess we will have to see since there is no good way to trigger this bug AFAIK. But at least this change should be good to go.

PR #1057, opened 2020-07-28
Allow setting the environment to null

jvasseur:
Null is a valid value for the environment option (it means no environment is specified).

This allow setting the environment to null in the Option class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


jvasseur:
@stayallive I've added the line to the changelog.

ste93cry:
@jvasseur would you mind force-pushing to trigger again the CI? The issue with Psalm should be solved due to its new release, and once the build is green I will merge

stayallive:
No need @ste93cry, retriggered the build. 

Thanks again for this PR 👌

PR #1063, opened 2020-08-08
Refactor the Stacktrace and Frame classes

ste93cry:
One of the biggest mistakes I found out after the `2.0` release of the SDK was that the `Stacktrace` class was depending on a bunch of stuffs like the serializer and the SDK options. This is inherently wrong because of a few reasons:

- The class should be a DTO or a VO rather than a full blown class that also acts as a factory of itself and of the `Frame` objects
- It's really hard to test the class due to all the dependencies it has
- The code is a mess to maintain because it's still based on some legacy code coming from SDK `1.x`
- It's impossible to evolve further the class without breaking BC because it got marked as final only after version `2.3`

This PR is an attempt to fix all the issues reported above by creating builder classes for both the `Stacktrace` and the `Frame` class. Since it was a low-cost effort, I also implemented support for the exception mechanism and I enhanced how we store some data by switching from plain arrays to fully typehinted objects


src/EventFactory.php
```diff
@@ -50,30 +50,26 @@ final class EventFactory implements EventFactoryInterface
     public function __construct(SerializerInterface $serializer, RepresentationSerializerInterface $representationSerializer, Options $options, string $sdkIdentifier, string $sdkVersion)
     {
         $this->serializer = $serializer;
-        $this->representationSerializer = $representationSerializer;
         $this->options = $options;
         $this->sdkIdentifier = $sdkIdentifier;
         $this->sdkVersion = $sdkVersion;
+        $this->stacktraceBuilder = new StacktraceBuilder($options, $representationSerializer);
```
Jean85, line 56: This is still a hidden dependency like this... How can we improve this?

ste93cry: One of the things I learned in `2.x` is that the way an event data is collected or constructed is often something internal and the user shouldn't worry about, so I don't think we should expose such dependencies. Also, making the builders part of the public API will probably get us back to the point where it will be harder to improve the API or do bug fixes because we can't break BC

PR #1068, opened 2020-08-19
Cleanup the Monolog handler from deprecated features

ste93cry:
With this PR I'm going to revert the support of our Monolog handler for setting tags and extras on the events. The reason of this change is that, as discussed multiple times in the past, we don't want to rely on pulling more information than necessary from the unstructured data of the record's context because the same keys may be used by the user for things that have nothing to do with Sentry


Jean85 (changes requested):
Code LGTM :+1:  but I would add a section in the upgrade guide on how to decorate Monolog to achieve the same. This is because it's an highly requested feature, and having a piece of text to point to will be valuable in the future.

ste93cry:
>This is because it's an highly requested feature, and having a piece of text to point to will be valuable in the future

Absolutely agree, but is this really a thing written in the UPGRADE file or in the official docs?

Jean85 (approved):

Issue #985, opened 2020-02-26
php_uname is used without checking whether the function exists


joejansen:
We are migrating a Wordpress app that uses sentry-php to [Flywheel](https://getflywheel.com/). Flywheel [disables `php_uname` along with a number of other functions](https://getflywheel.com/wordpress-support/what-php-functions-are-disabled/).

In core Wordpress the use of `php_uname` is [wrapped in a check to see whether it exists](https://core.trac.wordpress.org/browser/tags/5.3/src/wp-admin/includes/class-wp-debug-data.php#L575).

Could we do the same thing for the [use in the `ServerOsContext` class](https://github.com/getsentry/sentry-php/blob/master/src/Context/ServerOsContext.php#L182-L187)?

ste93cry:
This really looks like a case of "security through obscurity", but I'm not a security expert so I won't judge. I'm open to change this, so if you want to submit a PR I will gladly review it. Note that the change is a bit more complicated than what it seems at a first glance as we have to ensure that an empty context is not sent at all and we cannot make the parameters and return types of the getters/setters of the `ServerOsContext` class nullable or we would break BC, so the only thing that I can think of is to set the data to an empty string and override the `isEmpty` method to work accordingly

Jean85:
If avoiding creating the context is too complex, we could fall back to a fixed string, or use the `PHP_OS` constant, which doesn't have security implications.

ste93cry:
>If avoiding creating the context is too complex

With the current architecture it's a bit complicated indeed, and for the future we may evaluate to (I was already thinking about this anyway) to stop creating all the contexts in the constructor of the `Event` class and instead populate them on-demand using integrations so that users can workaround these issues by disabling what they don't want/don't need. Anyway it should be feasible, we just have to accept the downside of having some messy code here and there until we find out the time to refactor this part of the codebase

>we could fall back to a fixed string

I would like to avoid this because useless and wrong data would then be reported to Sentry

>use the `PHP_OS` constant

I though about that too, but there are two issues: the first one is that we should parse ourselves the value to extract each single information, and the second one (and most important) is that this constant returns the details of the OS PHP was built on, which is different from `php_uname` that returns instead the details of the current OS

fuzzy76:
> In core Wordpress the use of `php_uname` is [wrapped in a check to see whether it exists](https://core.trac.wordpress.org/browser/tags/5.3/src/wp-admin/includes/class-wp-debug-data.php#L575).

[...]

PR #1072, opened 2020-08-23
Refactor the context classes

ste93cry:
In this PR I refactored the `*Context` classes by optimizing them to store data into typed properties rather than plain arrays: this allowed me to also remove their dependency on the [OptionsResolver](https://symfony.com/doc/current/components/options_resolver.html) component of Symfony and to drop the base `Context` class because of a few reasons:

- The `extra` field is not really a context as it's a top-level field of the event payload. Also, since the field can basically contain anything it didn't really make sense to store its data in anything else than a plain array in the `Event` class
- The `user` field is not really a context for the same reason as above, therefore having a `UserContext` class was semantically wrong

I also extracted the intialization of the default values of the contexts into an `EnvironmentIntegration` integration so that it's easier to solve issues like #985 by just disabling the integration.


UPGRADE-3.0.md
```diff
@@ -37,3 +37,59 @@
...
+- The `OsContext` and `RuntimeContext` classes do not implement anymore the `ArrayAccess`, `IteratorAggregate` and `JsonSerializable` interfaces
+- The following methods have been removed from the `OsContext` and `RuntimeContext` classes:
+  - `*Context::merge()`
+  - `*Context::setData()`
+  - `*Context::replaceData()`
+  - `*Context::clear()`
+  - `*Context::isEmpty()`
+  - `*Context::toArray()`
+- Removed the `UserContext` class, use `UserDataBag` instead
+- The signature of the `RuntimeContext` class changed to require a bunch of parameters in place of an array of initial data
```
Jean85, line 51: It also became `final`

UPGRADE-3.0.md
```diff
@@ -37,3 +37,59 @@
 - The signature of the `HttpClientFactory::__construct()` method changed to accept instances of the PSR-17 factories in place of Httplug's ones
 - The signature of the `DefaultTransportFactory::__construct()` method changed to accept instances of the PSR-17 factories in place of Httplug's ones
 - The signature of the `GzipEncoderPlugin::__construct()` method changed to accept an instance of the `Psr\Http\Message\StreamFactoryInterface` interface only
+- Removed the `TagsContext`, `ExtraContext` and `Context` classes, data is now stored in a plain array
```
Jean85, line 40: `TagContext` no longer does merge, always overrides right?

Jean85 (changes requested):

UPGRADE-3.0.md
ste93cry: `TagsContext` used to provide a `merge()` function, but now both the `Event` and `Scope` classes provide a setter that imho is quite self-explanatory. Of course when applying a scope on an event the tags still get merged

Jean85 (approved):

stayallive (approved):
🚢

PR #1074, opened 2020-08-26
feat: Tracing getter/setters

HazAT:
Move Span/Transaction to hub instead of scope

### DIFF
--- a/src/State/Hub.php
+++ b/src/State/Hub.php
@@ -264,6 +264,7 @@ final class Hub implements HubInterface
 
     /**
      * {@inheritdoc}
+     *
      * @psalm-suppress MoreSpecificReturnType
      * @psalm-suppress LessSpecificReturnStatement
      */
@@ -289,7 +290,7 @@ final class Hub implements HubInterface
     /**
      * {@inheritdoc}
      */
-    public function setSpan(?Span $span): self
+    public function setSpan(?Span $span): HubInterface
     {
         $this->span = $span;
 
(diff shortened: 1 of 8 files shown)


src/State/Hub.php
```diff
@@ -207,6 +213,22 @@ public function getIntegration(string $className): ?IntegrationInterface
...
+    {
+        return $this->getStackTop()->getScope();
+    }
+
+    /**
+     * Gets the topmost client/layer pair in the stack.
+     */
+    private function getStackTop(): Layer
+    {
+        return $this->stack[\count($this->stack) - 1];
```
Jean85, line 229: The stack can be empty, right?

src/State/Hub.php
```diff
@@ -241,18 +263,37 @@ public function startTransaction(TransactionContext $context): Transaction
...
+     * @psalm-suppress MoreSpecificReturnType
+     * @psalm-suppress LessSpecificReturnStatement
      */
-    private function getScope(): Scope
+    public function getTransaction(): ?Transaction
     {
-        return $this->getStackTop()->getScope();
+        $span = $this->span;
+        if (null !== $span && null !== $span->spanRecorder) {
+            /** @phpstan-ignore-next-line */
```
Jean85, line 275: All those ignores are hiding a real problem, it may be empty!

HazAT: Fixed in `3.x` thx @Jean85

PR #1077, opened 2020-08-29
Extract the logic to serialize the event into an event serializer

ste93cry:
One more piece of refactoring: I've extracted the logic to serialize the event into its own class, therefore creating an event serializer service. I have some doubts whether making it a dependency of the HTTP transport rather than just instantiating it in the constructor. The pros of having it passed as an argument is of course testability, the cons is that I don't think there will ever be another implementation and so it may be useless to be able to set the instance from outside the transport. Any opinion is more than welcome!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,7 @@
         "guzzlehttp/promises": "^1.3",
         "guzzlehttp/psr7": "^1.6",
         "jean85/pretty-package-versions": "^1.2",
+        "ocramius/package-versions": "^1.8",
         "php-http/async-client-implementation": "^1.0",
         "php-http/client-common": "^1.5|^2.0",
         "php-http/discovery": "^1.6.1",
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -188,7 +188,13 @@ final class ClientBuilder implements ClientBuilderInterface
         $this->serializer = $this->serializer ?? new Serializer($this->options);
         $this->representationSerializer = $this->representationSerializer ?? new RepresentationSerializer($this->options);
 
-        return new EventFactory($this->serializer, $this->representationSerializer, $this->options, $this->sdkIdentifier, $this->sdkVersion);
+        return new EventFactory(
+            $this->serializer,
+            $this->representationSerializer,
+            $this->options,
+            $this->sdkIdentifier,
+            $this->sdkVersion
+        );
     }
 
     /**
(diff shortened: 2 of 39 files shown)


src/EventType.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        return self::getInstance('transaction');
+    }
+
+    public function __toString(): string
+    {
+        return $this->value;
+    }
+
+    private static function getInstance(string $value): self
```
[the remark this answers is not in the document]
ste93cry: Because this enum (and I will refactor the old ones too in the near future) is now implemented using the flyweight pattern so that you always get the same instance for the same value, which means improved memory performances and allows comparing it using `===`

Jean85 (approved):

HazAT (changes requested):
Can we rename `EventSerializer` to `PayloadSerializer`?
Since we already send envelopes and maybe in the future sessions and not just Events. `Payload` is more generic.

composer.json
```diff
@@ -26,6 +26,7 @@
         "guzzlehttp/promises": "^1.3",
         "guzzlehttp/psr7": "^1.6",
         "jean85/pretty-package-versions": "^1.2",
+        "ocramius/package-versions": "^1.8",
```
stayallive, line 29: Why are we adding this explicitly? We only interface with `jean85/pretty-package-versions` right?

src/ClientBuilder.php
```diff
@@ -206,6 +212,10 @@ private function createDefaultTransportFactory(): DefaultTransportFactory
             $this->sdkVersion
         );
 
-        return new DefaultTransportFactory($streamFactory, Psr17FactoryDiscovery::findRequestFactory(), $httpClientFactory, $this->logger);
+        return new DefaultTransportFactory(
+            $streamFactory, Psr17FactoryDiscovery::findRequestFactory(),
+            $httpClientFactory,
+            $this->logger
+        );
```
stayallive, line 219: I think this should've been:

```suggestion
        return new DefaultTransportFactory(
            $streamFactory,
            Psr17FactoryDiscovery::findRequestFactory(),
            $httpClientFactory,
            $this->logger
        );
```

composer.json
ste93cry: This change isn't really related to this PR, anyway its reason is that we use the `PackageVersions\Versions::VERSION` constant and thus we should reference its package in our `composer.json` to avoid breaking of our code if all the other dependencies stops requiring it

HazAT (approved):

stayallive (approved):

PR #1080, opened 2020-09-02
Remove the SpoolTransport transport

ste93cry:
The `SpoolTransport` was introduced as an attempt to have a transport that was able to persist events to a storage rather than sending them straight away. I now think that the implementation is not really useful, so I decided to cleanup all the codebase and stop offering it out of the box. If someone wants to delay sending of the events for some reason it's better to implement the feature directly in the project


stayallive (approved):
👍 good plan!

Jean85 (changes requested):
I'm not sure about this... Wouldn't the `MemorySpool` still be useful for sending events at the end of the request?

ste93cry:
> Wouldn't the `MemorySpool` still be useful for sending events at the end of the request

Yes it could, but we don't have anymore a transport (at least implemented in the `3.x` branch) that is able to send events at the end of the request so keeping this code does not really makes sense. Also, we can always store the delayed events into the transport itself rather than relying on external storage

Jean85 (approved):

PR #1083, opened 2020-09-07
feat: Add TracesSampler

HazAT:
Adds `TracesSampler` aka `traces_sampler` option for people to define a callable to filter out Transactions.

You can define it like: 

```php
Sentry\init([
  'traces_sampler' => static function (SamplingContext $samplingContext): float {
    return 1;
  }
]);
```


src/Tracing/SamplingContext.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+namespace Sentry\Tracing;
+
+class SamplingContext
```
ste93cry, line 5: Please make the class `final`

src/State/Hub.php
```diff
@@ -231,21 +234,35 @@ public function startTransaction(TransactionContext $context): Transaction
...
-                    // if false = we don't want to have it
-                    $context->sampled = false;
-                } else {
-                    $context->sampled = true;
-                }
             }
         }
 
+        if ($sampleRate < 1 && mt_rand(1, 100) / 100.0 > $sampleRate) {
+            // if true = we want to have the transaction
```
ste93cry, line 259: I think that these comments are unnecessary, to me the code is pretty self-explanatory, isn't it?

src/Options.php
```diff
@@ -658,6 +659,35 @@ public function setClassSerializers(array $serializers): void
...
+
+    /**
+     * Sets a callback that will be invoked when we take the sampling decision for Transactions.
+     * Return a number between 0 and 1 to define the sample rate for the provided SamplingContext.
+     *
+     * @param callable $sampler The sampler
+     *
+     * @psalm-param \Sentry\Tracing\TracesSamplerInterface|callable(\Sentry\Tracing\SamplingContext): float $sampler
+     */
+    public function setTracesSampler($sampler): void
```
ste93cry, line 684: What's the reason this option can be all those things? Wouldn't a simple `callable` be enough? If someone wants to use a class he can either code it to make it self-invokable or he can wrap the call to the class into the `callable` itself

src/Options.php
```diff
@@ -675,6 +705,7 @@ private function configureOptions(OptionsResolver $resolver): void
             'prefixes' => explode(PATH_SEPARATOR, get_include_path()),
             'sample_rate' => 1,
             'traces_sample_rate' => 0,
+            'traces_sampler' => null,
```
ste93cry, line 708: Missing the validation of the option, `string|callable|object` should work

src/State/Hub.php
```diff
@@ -231,21 +234,35 @@ public function startTransaction(TransactionContext $context): Transaction
     {
         $client = $this->getClient();
```
ste93cry, line 236: Please remove this blank line

src/State/Hub.php
```diff
@@ -231,21 +234,35 @@ public function startTransaction(TransactionContext $context): Transaction
...
 
+        $sampleRate = null;
+
+        if (null !== $client) {
+            $sampler = $client->getOptions()->getTracesSampler();
+            if (null !== $sampler) {
```
ste93cry, line 241: Please add a blank line above this one to improve readability

src/State/Hub.php
```diff
@@ -231,21 +234,35 @@ public function startTransaction(TransactionContext $context): Transaction
...
+                    $sampleRate = $sampler(SamplingContext::getDefault($context));
+                } elseif ($sampler instanceof TracesSamplerInterface) {
+                    $sampleRate = $sampler->sample(SamplingContext::getDefault($context));
+                }
+            }
+        }
+
         // Roll the dice for sampling transaction, all child spans inherit the sampling decision.
-        if (null === $context->sampled) {
+        // And only if $$sampleRate == null because then the traces_sampler wasn't defined and we need to roll the dice
```
ste93cry, line 251: Can you please rephrase better this comment? I'm having an hard time understanding it

src/Tracing/SamplingContext.php
```diff
@@ -0,0 +1,42 @@
...
+{
+    /**
+     * @var TransactionContext|null TransactionContext
```
ste93cry, line 8: Please change the description of the variable to `The context of the transaction`

src/Tracing/SamplingContext.php
```diff
@@ -0,0 +1,42 @@
...
+     */
+    private $transactionContext;
+
+    /**
+     * Returns the default instance of for the SamplingContext
+     *
+     * @param TransactionContext $transactionContext
```
ste93cry, line 15: There is no much value in having this annotation without a description of the parameter. Please either add it or drop the line altogether as the parameter is self-explanatory

src/Tracing/SamplingContext.php
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public function getTransactionContext(): ?TransactionContext
+    {
+        return $this->transactionContext;
+    }
+
+    /**
+     * @param TransactionContext|null $transactionContext
+     */
+    public function setTransactionContext(?TransactionContext $transactionContext): void
```
ste93cry, line 37: There is no much value in having this annotation without a description of the parameter and of the function. Please either add them or drop the DocBlock altogether

ste93cry (changes requested):

src/Options.php
HazAT: The reason was mostly how people like to configure these things in laravel like:
`'traces_sampler' => App\Helpers\General\TracesSampler::class`

While you are right, they can just create a callable class which might even be better since it's less code.

ste93cry: Please typehint the parameter as `?callable`

src/Options.php
```diff
@@ -658,6 +658,31 @@ public function setClassSerializers(array $serializers): void
         $this->options = $this->resolver->resolve($options);
     }
 
+    /**
+     * Gets a callback that will be invoked when we sample a Transaction.
+     *
+     * @psalm-return null|callable(\Sentry\Tracing\SamplingContext): float
+     */
+    public function getTracesSampler()
```
ste93cry, line 666: Please typehint the return tpe as `?callable`

src/State/Hub.php
```diff
@@ -230,22 +231,35 @@ private function getStackTop(): Layer
...
     {
         $client = $this->getClient();
+        $sampleRate = null;
+
+        if (null !== $client) {
+            $sampler = $client->getOptions()->getTracesSampler();
+
+            if (null !== $sampler) {
+                if (\is_callable($sampler)) {
+                    $sampleRate = \call_user_func($sampler, SamplingContext::getDefault($context));
```
ste93cry, line 241: Please use the `$sampler(...)` syntax to call the user function

src/State/Hub.php
```diff
@@ -230,22 +231,35 @@ private function getStackTop(): Layer
     public function startTransaction(TransactionContext $context): Transaction
     {
         $client = $this->getClient();
+        $sampleRate = null;
+
+        if (null !== $client) {
+            $sampler = $client->getOptions()->getTracesSampler();
+
+            if (null !== $sampler) {
+                if (\is_callable($sampler)) {
```
ste93cry, line 240: Once we typehint the getter and the setter this check can be replaced with a simpler `null !== $sampler`

ste93cry (changes requested):
There are three things missing yet: the CHANGELOG, a test for the new option and a test that assert that `startTransaction` works correctly with all cases of the `traces_sampler` option

HazAT:
@ste93cry Added tests and fixed your comments, thanks

ste93cry (approved):

src/Options.php
Jean85: Symfony too is not so keen on using callables since it has a lot of YAML config by default. I could write a callable inside the bundle to handle FQCN string though.

PR #1086, opened 2020-09-13
Refactor the setup process of the integrations

ste93cry:
I took the occasion to refactorize the code that handles the setup of the integrations. Since this is an internal part of the API it should not really affect the users, the only visible thing that changed is that the integrations are run in a different order than before due to some changes in how we prepare the list of those to setup. I don't know if it's worth documenting it because it should not really matter, but if we think we should I have nothing cons


src/Integration/IntegrationRegistry.php
```diff
@@ -0,0 +1,141 @@
...
+
+        return $integrations;
+    }
+
+    private function setupIntegration(IntegrationInterface $integration, LoggerInterface $logger): void
+    {
+        $integrationName = \get_class($integration);
+
+        if (isset($this->integrations[$integrationName])) {
+            return;
```
Jean85, line 69: Missing unit test.

src/Integration/IntegrationRegistry.php
```diff
@@ -0,0 +1,141 @@
...
+                if (!isset($pickedIntegrationsClasses[$integrationClassName])) {
+                    $integrations[] = $userIntegration;
+                    $pickedIntegrationsClasses[$integrationClassName] = true;
+                }
+            }
+        } else {
+            $integrations = $userIntegrations($defaultIntegrations);
+
+            if (!\is_array($integrations)) {
+                throw new \UnexpectedValueException(sprintf('Expected the callback set for the "integrations" option to return a list of integrations. Got: "%s".', get_debug_type($integrations)));
```
Jean85, line 115: Missing unit test.

Jean85 (approved):
Apart from test coverage, LGTM :+1:

stayallive (approved):
Looks great 👍 

I don't think documenting the possible order change is really needed, don't think it's relevant either.

PR #1088, opened 2020-09-13
Update some deps and require at least PHP 7.2

ste93cry:
Not much to say, minimum supported PHP version for `3.x` will be `7.2`. I also decided to update the required version of the `symfony/options-resolver` paclkage so that we are able to use the standard `value[]` syntax to validate that an option of type `array` contains only values of that type rather than writing our own validator function

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -7,12 +7,6 @@ image: Visual Studio 2019
 
 environment:
     matrix:
-        - PHP_VERSION: 7.1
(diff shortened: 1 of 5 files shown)


stayallive (approved):
Seems reasonable, good 👍

Jean85 (approved):
Acceptable :+1: It should be noted that we dropped support only for very old Symfony versions.

Jean85 (changes requested):
Wops! Changelog entry missing!

Jean85 (approved):

PR #1091, opened 2020-09-21
ref: Remove EventFactory

HazAT:


src/Client.php
```diff
@@ -56,21 +56,53 @@ final class Client implements ClientInterface
...
         $this->options = $options;
         $this->transport = $transport;
-        $this->eventFactory = $eventFactory;
         $this->logger = $logger ?? new NullLogger();
         $this->integrations = IntegrationRegistry::getInstance()->setupIntegrations($options, $this->logger);
+        $this->serializer = $serializer ?? new Serializer($this->options);
+        $this->representationSerializer = $representationSerializer ?? new RepresentationSerializer($this->options);
+        $this->stacktraceBuilder = new StacktraceBuilder($options, $this->representationSerializer);
+        $this->sdkIdentifier = $sdkIdentifier ?? self::SDK_IDENTIFIER;
+        $this->sdkVersion = $sdkVersion ?? '0.0.0';
```
Jean85, line 105: Can we use `PrettyVersions` here?

src/Client.php
```diff
@@ -206,4 +238,104 @@ private function prepareEvent($payload, ?Scope $scope = null): ?Event
...
+
+    /**
+     * Create an {@see Event} with a stacktrace attached to it.
+     *
+     * @param array<string, mixed>|Event $payload The data to be attached to the Event
+     */
+    private function buildEventWithStacktrace($payload): Event
+    {
+        if ($payload instanceof Event) {
+            return $this->buildEvent($payload);
```
Jean85, line 250: What about the stacktrace here?

src/Client.php
```diff
@@ -56,21 +56,53 @@ final class Client implements ClientInterface
...
+     * @param Options                                $options                  The client configuration
+     * @param TransportInterface                     $transport                The transport
+     * @param string|null                            $sdkIdentifier            The Sentry SDK identifier
+     * @param string|null                            $sdkVersion               The Sentry SDK version
+     * @param SerializerInterface|null               $serializer               The serializer
+     * @param RepresentationSerializerInterface|null $representationSerializer The serializer for function arguments
+     * @param LoggerInterface|null                   $logger                   The PSR-3 logger
      */
-    public function __construct(Options $options, TransportInterface $transport, EventFactoryInterface $eventFactory, ?LoggerInterface $logger = null)
+    public function __construct(Options $options, TransportInterface $transport, ?string $sdkIdentifier, ?string $sdkVersion, ?SerializerInterface $serializer, ?RepresentationSerializerInterface $representationSerializer, ?LoggerInterface $logger = null)
```
Jean85, line 95: I would allow a `null` default for those additional dependencies
```suggestion
    public function __construct(
        Options $options,
        TransportInterface $transport,
        ?string $sdkIdentifier = null,
        ?string $sdkVersion = null,
        ?SerializerInterface $serializer = null,
        ?RepresentationSerializerInterface $representationSerializer = null,
        ?LoggerInterface $logger = null
    )
```

tests/ClientTest.php
```diff
@@ -68,7 +66,10 @@ public function setupOnce(): void
                 'integrations' => [$integration],
             ]),
             $this->createMock(TransportInterface::class),
-            $eventFactory,
+            null,
+            null,
+            null,
+            null,
```
Jean85, line 72: Having put the `null`s default up, there would be no longer necessary

Jean85 (changes requested):

tests/ClientTest.php
HazAT: Can't because we test something in the logger here 😅

src/Client.php
HazAT: I will add a test case for this, this is not obvious how it works.

tests/ClientTest.php
Jean85: I don't understand... I'm just talking about omitting the `null`s here.

ste93cry: You missed that there is one last argument after the line you commented on, which is `$logger` and whose value comes from the data provider of the test

src/Client.php
HazAT: Was dead code after all

src/Client.php
```diff
@@ -56,21 +57,60 @@ final class Client implements ClientInterface
...
         $this->options = $options;
         $this->transport = $transport;
-        $this->eventFactory = $eventFactory;
         $this->logger = $logger ?? new NullLogger();
         $this->integrations = IntegrationRegistry::getInstance()->setupIntegrations($options, $this->logger);
+        $this->serializer = $serializer ?? new Serializer($this->options);
+        $this->representationSerializer = $representationSerializer ?? new RepresentationSerializer($this->options);
+        $this->stacktraceBuilder = new StacktraceBuilder($options, $this->representationSerializer);
+        $this->sdkIdentifier = $sdkIdentifier ?? self::SDK_IDENTIFIER;
+        $this->sdkVersion = $sdkVersion ?? PrettyVersions::getVersion('sentry/sentry')->getPrettyVersion();
```
ste93cry, line 113: We should be able to use `Versions::rootPackageName()` in place of hardcoding the package name

ste93cry (changes requested):

ste93cry (approved):

PR #1092, opened 2020-09-21
Polish the API related to APM

ste93cry:
As the title says, I had a look at the JS SDK and noticed some differences in key locations of the Tracing API, so I tried to align our codebase to that one. I refactored and improved some tests and I tried to ensure the state of the classes is exposed properly using getters/setters and that it doesn't leak from them in an uncontrolled way as it was before.

### DIFF
--- a/src/Tracing/Span.php
+++ b/src/Tracing/Span.php
@@ -425,4 +425,18 @@ class Span
     {
         $this->spanRecorder = null;
     }
+
+    /**
+     * Returns a string that can be used for the `sentry-trace` header.
+     */
+    public function toTraceparent(): string
+    {
+        $sampled = '';
+
+        if (null !== $this->sampled) {
+            $sampled = $this->sampled ? '-1' : '-0';
+        }
+
+        return sprintf('%s-%s%s', (string) $this->traceId, (string) $this->spanId, $sampled);
+    }
 }
--- a/src/Tracing/Span.php
+++ b/src/Tracing/Span.php
@@ -32,7 +32,7 @@ class Span
     protected $op;
 
     /**
-     * @var string|null Completion status of the Span
+     * @var SpanStatus|null Completion status of the Span
      */
     protected $status;
 
@@ -68,13 +68,11 @@ class Span
 
     /**
      * @var SpanRecorder|null Reference instance to the SpanRecorder
-     *
-     * @internal
      */
-    public $spanRecorder;
+    protected $spanRecorder;
 
     /**
-     * Span constructor.
+     * Constructor.
      *
      * @param SpanContext|null $context The context to create the span with
      *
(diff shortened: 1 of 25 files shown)


src/Tracing/Span.php
```diff
@@ -174,19 +182,6 @@ public function getEndTimestamp(): ?float
...
-    public function toTraceparent(): string
-    {
-        $sampled = '';
-        if (null !== $this->sampled) {
-            $sampled = $this->sampled ? '-1' : '-0';
-        }
-
-        return $this->traceId . '-' . $this->spanId . $sampled;
-    }
-
```
HazAT, line 189: We cannot remove this function, we need it in Laravel.

src/Tracing/SpanContext.php
```diff
@@ -184,35 +182,4 @@ public function setEndTimestamp(?float $endTimestamp): void
...
-        if (mb_strlen($matches['span_id']) > 0) {
-            $context->parentSpanId = new SpanId($matches['span_id']);
-        }
-
-        if (\array_key_exists('sampled', $matches)) {
-            $context->sampled = '1' === $matches['sampled'];
-        }
-
-        return $context;
-    }
```
HazAT, line 217: We also need this function.

HazAT (changes requested):

src/Tracing/SpanContext.php
HazAT: + Restore tests

src/Tracing/Span.php
```diff
@@ -375,26 +386,57 @@ public function finish(?float $endTimestamp = null): ?EventId
...
+    public function toTraceparent(): string
+    {
+        $sampled = '';
+
+        if (null !== $this->sampled) {
+            $sampled = $this->sampled ? '-1' : '-0';
+        }
+
+        return sprintf('%s-%s%s', (string) $this->traceId, (string) $this->spanId, $sampled);
+    }
```
rhcarvalho, line 441: This produces a different result from [the Python implementation](https://github.com/getsentry/sentry-python/blob/f6f3525f8812f60911573a8b7f71807ecf2e1052/sentry_sdk/tracing.py#L298-L305) when the sampling decision has not been made (trailing `-`):

```py
    def to_traceparent(self):
        # type: () -> str
        sampled = ""
        if self.sampled is True:
            sampled = "1"
        if self.sampled is False:
            sampled = "0"
        return "%s-%s-%s" % (self.trace_id, self.span_id, sampled)
```

@HazAT we need to define what is the correct behavior and document in https://github.com/getsentry/develop/pull/147

HazAT: I think we need to fix this in Python and document it properly.

HazAT (approved):
👍

stayallive (approved):
Looks good 👍

PR #1093, opened 2020-09-23
Update PHPUnit and remove unnecessary code left behind after updating PHP minimum required version

ste93cry:
As per title, I've updated PHPUnit to the greatest version still supported on PHP 7.2 and I cleaned up some leftovers of the update of the minimum required version of PHP


tests/Util/JSONTest.php
```diff
@@ -100,49 +99,6 @@ public function encodeSubstitutesInvalidUtf8CharactersOnPhp72OrGreaterDataProvid
...
-     * @requires PHP < 7.2
-     *
-     * @dataProvider encodeSubstitutesInvalidUtf8CharactersOnPhp71OrLowerDataProvider
-     */
-    public function testEncodeSubstitutesInvalidUtf8CharactersOnPhp71OrLower($value, string $expectedResult): void
-    {
-        $this->assertSame($expectedResult, JSON::encode($value));
-    }
-
-    public function encodeSubstitutesInvalidUtf8CharactersOnPhp71OrLowerDataProvider(): \Generator
```
Jean85, line 113: Dropping this should mean that we could improve the `Json` class by using native escaping, right?

composer.json
```diff
@@ -47,8 +47,8 @@
         "phpstan/extension-installer": "^1.0",
         "phpstan/phpstan": "^0.12",
         "phpstan/phpstan-phpunit": "^0.12",
-        "phpunit/phpunit": "^7.5.18",
-        "symfony/phpunit-bridge": "^4.3|^5.0",
+        "phpunit/phpunit": "^8.5",
```
Jean85, line 50: Should we allow PHPUnit 9? So we could test under PHP 8 later

Jean85 (changes requested):

tests/Util/JSONTest.php
ste93cry: You're right, I cleaned up the tests but not the code :pensive:

Jean85 (approved):

stayallive (approved):
Looks good 👍

PR #1094, opened 2020-09-23
Only accept Event on captureEvent methods

stayallive:
The EventHint class was introduced to pass around extra data needed to correctly hydrate an event, this method of "hinting" was copied from the JavaScript SDK.

Chosen for a dedicated class to have type safety built-in instead of having an array and adding type checks and isset everywhere, this looks cleaner internally. To make it easier to use I've added a simple `fromArray` static method to have "the best of both worlds" I think, if users need it they can use the `extra` property on the hint to add arbitrary data.

### DIFF
--- a/UPGRADE-3.0.md
+++ b/UPGRADE-3.0.md
@@ -113,3 +113,5 @@
 - The signature of `ClientInterface::captureEvent` changed to `ClientInterface::captureEvent(Event $event, ?EventHint $hint = null, ?Scope $scope = null)`
 - The signature of `HubInterface::captureEvent` changed to `HubInterface::captureEvent(Event $event, ?EventHint $hint = null)`
 - The signature of `captureEvent` changed to `captureEvent(Event $event, ?EventHint $hint = null)`
+- The signature of `Scope::applyToEvent` changed to `Scope::applyToEvent(Event $event)`
+- Global and scope event processors will no longer receive a second argument, callable should now have the signature `callable(Event $event)`
--- a/UPGRADE-3.0.md
+++ b/UPGRADE-3.0.md
@@ -110,8 +110,8 @@
 - Removed the `Stacktrace::toArray()` and `Stacktrace::jsonSerialize()` methods
 - Removed the `SpoolTransport` class and the `SpoolInterface` interface with related implementation
 - Made the `Event::__construct()` method `private`, use the named constructors instead
-- The signature of `ClientInterface::captureEvent` changed to `ClientInterface::captureEvent(Event $event, ?EventHint $hint = null, ?Scope $scope = null)`
-- The signature of `HubInterface::captureEvent` changed to `HubInterface::captureEvent(Event $event, ?EventHint $hint = null)`
-- The signature of `captureEvent` changed to `captureEvent(Event $event, ?EventHint $hint = null)`
-- The signature of `Scope::applyToEvent` changed to `Scope::applyToEvent(Event $event)`
-- Global and scope event processors will no longer receive a second argument, callable should now have the signature `callable(Event $event)`
+- The signature of `ClientInterface::captureEvent()` changed to `ClientInterface::captureEvent(Event $event, ?EventHint $hint = null, ?Scope $scope = null)`
+- The signature of `HubInterface::captureEvent()` changed to `HubInterface::captureEvent(Event $event, ?EventHint $hint = null)`
+- The signature of `captureEvent()` changed to `captureEvent(Event $event, ?EventHint $hint = null)`
+- The signature of `Scope::applyToEvent()` changed to `Scope::applyToEvent(Event $event, ?EventHint $hint = null)`
+- Global and scope event processors will now receive a `EventHint` as the second parameter, callable should now have the signature `callable(Event $event, EventHint $hint)`
--- a/UPGRADE-3.0.md
+++ b/UPGRADE-3.0.md
@@ -102,7 +102,7 @@
 - Removed the `Event::getTagsContext()` method, use `Event::getTags()` instead
 - Removed the `Event::getUserContext()` method, use `Event::getUser()` instead
 - Renamed the `Event::getServerOsContext()` method to `Event::getOsContext()`
-- The signature of the `Scope::setUser()` method changed ot accept a plain array
+- The signature of the `Scope::setUser()` method changed to accept a plain array
 - Removed the `FlushableClientInterface` and `ClosableTransportInterface` interfaces. Their methods have been moved to the corresponding `ClientInterface` and `TransportInterface` interfaces
 - Removed the `Event::toArray()` and `Event::jsonSerialize()` methods
 - Removed the `Breadcrumb::toArray()` and `Breadcrumb::jsonSerialize()` methods
@@ -110,3 +110,6 @@
 - Removed the `Stacktrace::toArray()` and `Stacktrace::jsonSerialize()` methods
 - Removed the `SpoolTransport` class and the `SpoolInterface` interface with related implementation
 - Made the `Event::__construct()` method `private`, use the named constructors instead
+- The signature of `ClientInterface::captureEvent` changed to `ClientInterface::captureEvent(Event $event, ?EventHint $hint = null, ?Scope $scope = null)`
+- The signature of `HubInterface::captureEvent` changed to `HubInterface::captureEvent(Event $event, ?EventHint $hint = null)`
+- The signature of `captureEvent` changed to `captureEvent(Event $event, ?EventHint $hint = null)`
--- a/src/Client.php
+++ b/src/Client.php
@@ -215,7 +215,7 @@ final class Client implements ClientInterface
     {
         $event = $this->buildEvent($payload, $hint);
 
-        $this->addMissingStacktraceToEvent($payload, $hint);
+        $this->addMissingStacktraceToEvent($event, $hint);
 
         $sampleRate = $this->options->getSampleRate();
 
--- a/src/Client.php
+++ b/src/Client.php
@@ -8,7 +8,6 @@ use GuzzleHttp\Promise\PromiseInterface;
 use Jean85\PrettyVersions;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
-use Sentry\Exception\EventCreationException;
 use Sentry\Integration\IntegrationInterface;
 use Sentry\Integration\IntegrationRegistry;
 use Sentry\Serializer\RepresentationSerializer;
@@ -126,10 +125,12 @@ final class Client implements ClientInterface
      */
     public function captureMessage(string $message, ?Severity $level = null, ?Scope $scope = null): ?EventId
     {
-        return $this->captureEvent($this->buildEvent([
-            'message' => $message,
-            'level' => $level,
-        ]), null, $scope);
+        $event = Event::createEvent();
+
+        $event->setMessage($message);
+        $event->setLevel($level);
+
+        return $this->captureEvent($event, null, $scope);
     }
 
     /**
@@ -205,15 +206,30 @@ final class Client implements ClientInterface
     /**
      * Assembles an event and prepares it to be sent of to Sentry.
      *
-     * @param array<string, mixed>|Event $payload The payload that will be converted to an Event
-     * @param EventHint|null             $hint    May contain additional information about the original exception
-     * @param Scope|null                 $scope   Optional scope which enriches the Event
+     * @param Event          $event The payload that will be converted to an Event
+     * @param EventHint|null $hint  May contain additional information about the original exception
+     * @param Scope|null     $scope Optional scope which enriches the Event
      *
      * @return Event|null The prepared event object or null if it must be discarded
      */
-    private function prepareEvent($payload, ?EventHint $hint = null, ?Scope $scope = null): ?Event
+    private function prepareEvent(Event $event, ?EventHint $hint = null, ?Scope $scope = null): ?Event
     {
-        $event = $this->buildEvent($payload, $hint);
+        if (null !== $hint) {
+            if (null !== $hint->exception && empty($event->getExceptions())) {
+                $this->addThrowableToEvent($event, $hint->exception);
+            }
+
+            if (null !== $hint->stacktrace && null === $event->getStacktrace()) {
+                $event->setStacktrace($hint->stacktrace);
+            }
+        }
+
+        $event->setSdkIdentifier($this->sdkIdentifier);
+        $event->setSdkVersion($this->sdkVersion);
+        $event->setServerName($this->options->getServerName());
+        $event->setRelease($this->options->getRelease());
+        $event->setTags($this->options->getTags());
+        $event->setEnvironment($this->options->getEnvironment());
 
         $this->addMissingStacktraceToEvent($event, $hint);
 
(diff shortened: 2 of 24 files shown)


UPGRADE-3.0.md
```diff
@@ -102,11 +102,16 @@
...
-- The signature of the `Scope::setUser()` method changed ot accept a plain array
+- The signature of the `Scope::setUser()` method changed to accept a plain array
 - Removed the `FlushableClientInterface` and `ClosableTransportInterface` interfaces. Their methods have been moved to the corresponding `ClientInterface` and `TransportInterface` interfaces
 - Removed the `Event::toArray()` and `Event::jsonSerialize()` methods
 - Removed the `Breadcrumb::toArray()` and `Breadcrumb::jsonSerialize()` methods
 - Removed the `Frame::toArray()` and `Frame::jsonSerialize()` methods
 - Removed the `Stacktrace::toArray()` and `Stacktrace::jsonSerialize()` methods
 - Removed the `SpoolTransport` class and the `SpoolInterface` interface with related implementation
 - Made the `Event::__construct()` method `private`, use the named constructors instead
+- The signature of `ClientInterface::captureEvent` changed to `ClientInterface::captureEvent(Event $event, ?EventHint $hint = null, ?Scope $scope = null)`
```
ste93cry, line 113: Please add the missing round brackets to the method name

UPGRADE-3.0.md
```diff
@@ -102,11 +102,16 @@
...
+- The signature of `HubInterface::captureEvent` changed to `HubInterface::captureEvent(Event $event, ?EventHint $hint = null)`
```
ste93cry, line 114: Please add the missing round brackets to the method name

UPGRADE-3.0.md
```diff
@@ -102,11 +102,16 @@
...
+- The signature of `captureEvent` changed to `captureEvent(Event $event, ?EventHint $hint = null)`
```
ste93cry, line 115: Please add the missing round brackets to the method name

UPGRADE-3.0.md
```diff
@@ -102,11 +102,16 @@
...
+- The signature of `Scope::applyToEvent` changed to `Scope::applyToEvent(Event $event)`
```
ste93cry, line 116: Please add the missing round brackets to the method name

src/functions.php
```diff
@@ -43,11 +43,12 @@ function captureException(\Throwable $exception): ?EventId
 /**
  * Captures a new event using the provided data.
  *
- * @param array<string, mixed> $payload The data of the event being captured
+ * @param Event                  $event The event being captured
+ * @param \Sentry\EventHint|null $hint  May contain additional information about the original exception
  */
-function captureEvent(array $payload): ?EventId
+function captureEvent(Event $event, ?EventHint $hint = null): ?EventId
```
ste93cry, line 49: @HazAT It looks like some of the SDKs are accepting the `$hint` parameter (e.g. Python) and some aren't (e.g. JS). I have nothing cons this change, just want to be sure that we are doing well

src/functions.php
```diff
@@ -43,11 +43,12 @@ function captureException(\Throwable $exception): ?EventId
 /**
  * Captures a new event using the provided data.
  *
- * @param array<string, mixed> $payload The data of the event being captured
+ * @param Event                  $event The event being captured
+ * @param \Sentry\EventHint|null $hint  May contain additional information about the original exception
```
ste93cry, line 47: Since the class is already imported you can just reference it by its name rather than FQCN

UPGRADE-3.0.md
```diff
@@ -102,11 +102,16 @@
...
+- Global and scope event processors will no longer receive a second argument, callable should now have the signature `callable(Event $event)`
```
ste93cry, line 117: What's the reason of this change? The second argument is the only way to pass additional data from the location where an event has been captured down to the event processors and I see no reason to drop it. Also we had at least one issue where this saved us, so :-1: on removing it

src/ClientInterface.php
```diff
@@ -47,10 +47,11 @@ public function captureLastError(?Scope $scope = null): ?EventId;
     /**
      * Captures a new event using the provided data.
      *
-     * @param array<string, mixed>|Event $payload The data of the event being captured
-     * @param Scope|null                 $scope   An optional scope keeping the state
+     * @param Event          $event The event being captured
+     * @param EventHint|null $hint  May contain additional information about the original exception
```
ste93cry, line 51: The description is not accurate, the fact that this object may contain the original exception is just a detail, we should describe what the `$hint` is and not what it contains

src/State/Scope.php
```diff
@@ -303,10 +303,9 @@ public function clear(): self
      * Applies the current context and fingerprint to the event. If the event has
      * already some breadcrumbs on it, the ones from this scope won't get merged.
      *
-     * @param Event                      $event   The event object that will be enriched with scope data
-     * @param array<string, mixed>|Event $payload The raw payload of the event that will be propagated to the event processors
+     * @param Event $event The event object that will be enriched with scope data
      */
-    public function applyToEvent(Event $event, $payload): ?Event
```
ste93cry, line 309: See one of the comment above about why this argument should not be dropped, but it should became the `$hint` instead

tests/State/HubTest.php
```diff
@@ -257,22 +257,24 @@ public function testCaptureLastError(): void
...
         $client->expects($this->once())
             ->method('captureEvent')
-            ->with(['message' => 'test'])
+            ->with($event)
             ->willReturn($eventId);
 
         $hub = new Hub();
 
-        $this->assertNull($hub->captureEvent([]));
+        $this->assertNull($hub->captureEvent(Event::createEvent()));
```
ste93cry, line 273: You can use `$event` too here

tests/Monolog/HandlerTest.php
```diff
@@ -19,14 +20,24 @@ final class HandlerTest extends TestCase
...
             ->method('captureEvent')
-            ->with($expectedPayload, $this->callback(function (Scope $scopeArg) use ($expectedExtra): bool {
-                $event = $scopeArg->applyToEvent(Event::createEvent(), []);
+            ->with($this->callback(function (Event $event) use ($expectedEvent): bool {
+                $this->assertEquals($expectedEvent->getLevel(), $event->getLevel());
+                $this->assertEquals($expectedEvent->getMessage(), $event->getMessage());
+                $this->assertEquals($expectedEvent->getLogger(), $event->getLogger());
+
+                return true;
+            }), $this->callback(function (EventHint $hint) use ($expectedHint): bool {
```
ste93cry, line 35: You can just pass `$expectedHint` instead of `$this->callback(...)`

tests/Monolog/HandlerTest.php
```diff
@@ -49,11 +60,16 @@ public function handleDataProvider(): iterable
                 'context' => [],
                 'extra' => [],
             ],
-            [
-                'level' => Severity::debug(),
-                'message' => 'foo bar',
-                'logger' => 'monolog.channel.foo',
-            ],
+            (static function (): Event {
+                $event = Event::createEvent();
```
ste93cry, line 64: If you move the instantiation of the event before the `yield` statement we can avoid creating all these self-invokable lambdas which makes the code less readable

tests/Integration/TransactionIntegrationTest.php
```diff
@@ -49,22 +49,25 @@ public function setupOnceDataProvider(): \Generator
             Event::createEvent(),
             true,
             [],
-            [],
             null,
         ];
 
         yield [
-            Event::createEvent(),
+            (static function () {
```
ste93cry, line 56: If you move the instantiation of the event before the `yield` statement we can avoid creating all these self-invokable lambdas which makes the code less readable

tests/FunctionsTest.php
```diff
@@ -67,18 +67,20 @@ public function testCaptureException(): void
...
+
+        $event->setMessage('foo');
 
         /** @var ClientInterface|MockObject $client */
         $client = $this->createMock(ClientInterface::class);
         $client->expects($this->once())
-            ->method('captureEvent')
-            ->with(['message' => 'foo'])
-            ->willReturn($eventId);
+               ->method('captureEvent')
```
ste93cry, line 77: Please revert this change, methods were already correctly indented (4 spaces)

tests/FunctionsTest.php
```diff
@@ -67,18 +67,20 @@ public function testCaptureException(): void
 
     public function testCaptureEvent(): void
     {
-        $eventId = EventId::generate();
+        $event = Event::createEvent($eventId = EventId::generate());
+
+        $event->setMessage('foo');
```
ste93cry, line 72: I think you can remove this line, we were capturing the message because it was the only way to know if the event was being created correctly but we can now compare the instances

tests/ClientTest.php
```diff
@@ -144,22 +146,22 @@ public function testCaptureEvent(): void
...
-        $inputData = [
-            'transaction' => 'foo bar',
-            'level' => Severity::debug(),
-            'logger' => 'foo',
-            'tags_context' => ['foo', 'bar'],
-            'extra_context' => ['foo' => 'bar'],
-            'user_context' => ['bar' => 'foo'],
-        ];
+        $event = Event::createEvent();
+
```
ste93cry, line 150: You can remove this blank line as the statements below are related to the preceding line

src/Monolog/Handler.php
```diff
@@ -44,21 +46,23 @@ public function __construct(HubInterface $hub, $level = Logger::DEBUG, bool $bub
      */
     protected function write(array $record): void
     {
-        $payload = [
-            'level' => self::getSeverityFromLevel($record['level']),
-            'message' => $record['message'],
-            'logger' => 'monolog.' . $record['channel'],
-        ];
+        $event = Event::createEvent();
+
```
ste93cry, line 50: You can remove this blank line as the statements below are related to the one above

src/Client.php
```diff
@@ -205,19 +206,33 @@ public function flush(?int $timeout = null): PromiseInterface
...
-        if ($this->options->shouldAttachStacktrace() && !($payload instanceof Event) && !isset($payload['exception']) && !isset($payload['stacktrace'])) {
-            $event = $this->buildEventWithStacktrace($payload);
-        } else {
-            $event = $this->buildEvent($payload);
+        if (null !== $hint) {
+            if (null !== $hint->exception && empty($event->getExceptions())) {
+                $this->addThrowableToEvent($event, $hint->exception);
+            }
+
+            if (null !== $hint->stacktrace && null === $event->getStacktrace()) {
```
ste93cry, line 222: Shouldn't the stacktrace be set only when no exception is set? I remember I read something like this somewhere in the Unified API docs, not sure where thought

src/Client.php
```diff
@@ -126,28 +125,30 @@ public function getOptions(): Options
      */
     public function captureMessage(string $message, ?Severity $level = null, ?Scope $scope = null): ?EventId
     {
-        $payload = [
-            'message' => $message,
-            'level' => $level,
-        ];
+        $event = Event::createEvent();
```
ste93cry, line 129: You can remove this blank line as the statements below are related to the line above

src/EventHint.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry;
+
+/**
+ * This class represents hints on how to process an event.
+ */
+final class EventHint
```
ste93cry, line 10: While I absolutely love typed things, I don't think that in this case we should typehint the `$hint` argument, mainly because it should allow users to carry any extra data they want and in this case it's not possible without setting a property dinamically (class is `final` and the `fromArray` method does not accept unknown properties) that makes this object much more similar to a plain array

ste93cry (changes requested):

UPGRADE-3.0.md
stayallive: Because there is no longer that second argument since we only pass around Event objects now. So there is no "payload" anymore?

stayallive: We could pass the new $hint instead of the $payload I guess?

ste93cry: Yea, this is what it's done in the other SDKs

src/functions.php
HazAT: This is good!

src/EventHint.php
HazAT: Can we leave it typed but not throw an exception and just discard unknown args?
I think that might be a better solution.

ste93cry: >Can we leave it typed but not throw an exception and just discard unknown args?

Of course we can, but I really can't find any good reason to have an object and then treat it like a plain array like in this case. With [array shapes](https://phpstan.org/writing-php-code/phpdoc-types#array-shapes) we can document the known keys that should be in the array so that static analyzers understand the minimal payload that the argument can carry, and for the rest it's up to the user what to pass and we shouldn't put any limit on it

src/Client.php
stayallive: You are correct at least according to the old code.

tests/Integration/TransactionIntegrationTest.php
```diff
@@ -48,34 +49,49 @@ public function setupOnceDataProvider(): \Generator
...
+        ];
+
         $event = Event::createEvent();
         $event->setTransaction('/foo/bar');
 
         yield [
             $event,
             true,
-            [],
+            null,
```
ste93cry, line 94: Something is wrong here, this test case looks like it's repeating the one at [line 57](https://github.com/getsentry/sentry-php/pull/1094/files#diff-9198a26b9171256987d7a092025d8ba6R57) 🤔

tests/Integration/TransactionIntegrationTest.php
```diff
@@ -86,15 +102,15 @@ public function setupOnceDataProvider(): \Generator
...
-            ['/foo/bar/baz'],
+            null,
             [],
             '/foo/bar',
         ];
 
         yield [
             Event::createEvent(),
             false,
-            ['transaction' => '/foo/bar'],
```
ste93cry, line 97: Same as above, this test case is now repeating one already written above

src/Client.php
ste93cry: This doesn't looks resolved to me. You changed the code in the `addMissingStacktraceToEvent` method to handle this, but here you are still setting the stacktrace regardless

src/State/Scope.php
```diff
@@ -347,8 +347,13 @@ public function applyToEvent(Event $event, $payload): ?Event
             $event->setContext($name, $data);
         }
 
+        // We create a empty `EventHint` instance to allow processors to always recieve a `EventHint` instance even if there wasn't one
```
ste93cry, line 350: Typo: `recieve` instead of `receive`

tests/State/ScopeTest.php
```diff
@@ -274,11 +275,32 @@ public function testAddEventProcessor(): void
             return null;
         });
 
-        $this->assertNull($scope->applyToEvent($event, []));
+        $this->assertNull($scope->applyToEvent($event));
         $this->assertTrue($callback2Called);
         $this->assertFalse($callback3Called);
     }
 
+    public function testEventProcessorRecievesTheEventAndEventHint(): void
```
ste93cry, line 283: Typo: `recieve` instead of `receive`

ste93cry (changes requested):

ste93cry (approved):

PR #1104, opened 2020-09-30
Fix FrameContextifier

### LINKED ISSUE #1103: [3.0] Frame context is not populated
This is a regression introduced in v3.0.0 during Frame refactoring.
When stacktrace file name and stripped file name is same, the context is not populated by FrameContextifierIntegration.

See:
https://github.com/getsentry/sentry-php/blob/65030a1376b900aee06a4f477bbf320bc8dc9c52/src/FrameBuilder.php#L87
https://github.com/getsentry/sentry-php/blob/65030a1376b900aee06a4f477bbf320bc8dc9c52/src/Integration/FrameContextifierIntegration.php#L81-L83

I would suggest to fallback to `$frame->getFile()` when `$frame->getAbsoluteFilePath()` is `null` or removing the conditions while building an instance of `Frame`.

IonBazan:
This PR fixes #1103 when `$file` is same as absolute path.

### DIFF
--- a/src/FrameBuilder.php
+++ b/src/FrameBuilder.php
@@ -84,7 +84,7 @@ final class FrameBuilder
             $strippedFilePath,
             $line,
             $rawFunctionName,
-            Frame::INTERNAL_FRAME_FILENAME !== $file && $strippedFilePath !== $file ? $file : null,
+            Frame::INTERNAL_FRAME_FILENAME !== $file ? $file : null,
             $this->getFunctionArguments($backtraceFrame),
             $this->isFrameInApp($file, $functionName)
         );
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix stacktrace frames were not contextified (#1104)
+
 ## 3.0.2 (2020-10-02)
 
 - fix: Use the traces sample rate for traces instead of the event sample rate (#1106)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Fix stacktrace frames were not contextified (#1104)
+- Fix missing source code excerpts for stacktrace frames whose absolute file path is equal to the file path (#1104)
 
 ## 3.0.2 (2020-10-02)
 
--- a/src/StacktraceBuilder.php
+++ b/src/StacktraceBuilder.php
@@ -70,7 +70,7 @@ final class StacktraceBuilder
         }
 
         // Add a final stackframe for the first method ever of this stacktrace
-        array_unshift($frames, new Frame(null, $file, $line));
+        array_unshift($frames, $this->frameBuilder->buildFromBacktraceFrame($file, $line, []));
 
         return new Stacktrace($frames);
     }
--- a/tests/FrameBuilderTest.php
+++ b/tests/FrameBuilderTest.php
@@ -36,7 +36,7 @@ final class FrameBuilderTest extends TestCase
                 'line' => 20,
                 'function' => 'test_function',
             ],
-            new Frame('test_function', '/path/to/file', 10),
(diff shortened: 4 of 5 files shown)


src/Frame.php
```diff
@@ -92,7 +92,7 @@ public function __construct(?string $functionName, string $file, int $line, ?str
         $this->file = $file;
         $this->line = $line;
         $this->rawFunctionName = $rawFunctionName;
-        $this->absoluteFilePath = $absoluteFilePath;
+        $this->absoluteFilePath = $absoluteFilePath ?? $file;
```
ste93cry, line 95: In case the frame is internal then `$file` would be set to `[internal]` because we need a placeholder somehow but reporting that value as absolute path of the file would not be correct

ste93cry (changes requested):
Rather than making the `abs_path` attribute required, which in the case of internal frames would report an incorrect value, I would rather change the frame builder to set its value regardless of whether the `$file` is equal or not to its stripped value. Technically speaking if `$file` is not an absolute path we would still report an incorrect value and the integration would try to read a non-existent file, but it should not happen if the frame is built from a real backtrace

IonBazan:
@ste93cry Thanks for the review. I've applied the changes. Not sure if I should pass `$file` as `$absoluteFilePath` here too:
https://github.com/getsentry/sentry-php/blob/65030a1376b900aee06a4f477bbf320bc8dc9c52/src/StacktraceBuilder.php#L73

src/FrameBuilder.php
```diff
@@ -84,7 +84,7 @@ public function buildFromBacktraceFrame(string $file, int $line, array $backtrac
             $strippedFilePath,
             $line,
             $rawFunctionName,
-            Frame::INTERNAL_FRAME_FILENAME !== $file && $strippedFilePath !== $file ? $file : null,
+            Frame::INTERNAL_FRAME_FILENAME !== $file ? $file : null,
```
IonBazan, line 87: I was also thinking about moving this check to `Frame` constructor.

ste93cry: I prefer to leave the responsibility of deciding whether the `$absoluteFilePath` argument should be filled or not to the builder

ste93cry:
>Not sure if I should pass `$file` as `$absoluteFilePath` here too

Strictly speaking yes, but we should also strip the prefix from the path like we do for the rest of the frames. We should probably call the frame builder also for this "fake" frame :thinking:

src/Integration/FrameContextifierIntegration.php
```diff
@@ -78,11 +78,12 @@ public function setupOnce(): void
     private function addContextToStacktraceFrames(int $maxContextLines, Stacktrace $stacktrace): void
     {
         foreach ($stacktrace->getFrames() as $frame) {
-            if ($frame->isInternal() || null === $frame->getAbsoluteFilePath()) {
+            if ($frame->isInternal()) {
                 continue;
             }
 
-            $sourceCodeExcerpt = $this->getSourceCodeExcerpt($maxContextLines, $frame->getAbsoluteFilePath(), $frame->getLine());
```
ste93cry, line 85: Shouldn't we revert every change in this file now that we always set the `$absoluteFilePath` field except when the frame is internal?

IonBazan:
@ste93cry You're right. I've squashed my commits so only the required changes are there. Added necessary assertions to `StacktraceBuilderTest` to make sure it won't happen again.

ste93cry (approved):
Changes look great, thank you for both spotting and solving this bug :sunglasses: Can you please add a CHANGELOG entry before I merge?

ste93cry:
@IonBazan maybe it passed under the radar because I edited the comment after posting it, but would you be so kind to add an entry to the CHANGELOG?

IonBazan:
@ste93cry Added the changelog entry - feel free to modify it before release if it does not match the standards.

PR #1105, opened 2020-10-01
feat: Expose setDsn in Options

HazAT:
We need this in Laravel for the test command

https://github.com/getsentry/sentry-laravel/pull/392


HazAT:
Found another solution

PR #1108, opened 2020-10-02
Fix criteria for valid user data

### LINKED ISSUE #1107: Identifying User by username produces an Error
According to the docs, one of `id`, `username`, `email`, or `ip_address` are required:
https://docs.sentry.io/platforms/javascript/guides/react/enriching-events/identify-user/

However, including only a `username` produces an error due to this line:
https://github.com/getsentry/sentry-php/blob/17588a2a00d2c4f3f55d847cdb5684ee676c4e1e/src/UserDataBag.php#L74

ahaurat:
Fixes #1107 to allow `username` or `email` to identify a user in addition to `id` or `ip_address`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 - Fix missing source code excerpts for stacktrace frames whose absolute file path is equal to the file path (#1104)
-- Fix criteria for valid user data (#1108)
+- Fix requirements to construct a valid object instance of the `UserDataBag` class (#1108)
 
 ## 3.0.2 (2020-10-02)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix criteria for valid user data (#1108)
+
 ## 3.0.2 (2020-10-02)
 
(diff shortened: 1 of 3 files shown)


Jean85 (changes requested):
Thank you for your PR; please add at least a test and a changelog entry.

ste93cry:
@ahaurat after the answer of @kamilogorek in the referenced issue, there is some more work to do, if you are willing to contribute it. We should:

- deprecate the static named constructors which only purpose was to ensure that the object was created in a consistent state with the minimum amount of required information
- remove all validation logic that assumes that either the ID or the IP address fields must be set
- make the constructor `public` and add all the arguments to it

In the next major version considering that none of the attributes of the `UserDataBag` class is required we can probably replace its usages with a simple plain array

stayallive:
@ahaurat I hope you don't min me jumping in making changes so we can push this over the finish line 👍 

After a discussion with @ste93cry we've decided to not deprecate the static named constructors yet, we might do that later and just keep them as nice ways to quickly construct a user data bag.

ahaurat:
@stayallive no problem at all; thanks for tackling it. Apologies I didn't have time to get to it this week.

ste93cry (approved):

Jean85 (approved):

PR #1116, opened 2020-10-14
Set the environment of the event to production if it has not been overridden explicitly

ste93cry:
According to https://github.com/getsentry/develop/pull/169 the `environment` field is now always required. Since we have to maintain the compatibility we cannot just make the property `non-nullable`, so I added our very first deprecation of the `3.x` series that will suggest users to always pass a `string`. I also changed the default of the option as per documentation

### DIFF
--- a/src/Serializer/PayloadSerializer.php
+++ b/src/Serializer/PayloadSerializer.php
@@ -44,10 +44,6 @@ final class PayloadSerializer implements PayloadSerializerInterface
             ],
         ];
 
-        if (null !== $event->getEnvironment()) {
-            $result['environment'] = $event->getEnvironment();
-        }
-
         if (null !== $event->getStartTimestamp()) {
             $result['start_timestamp'] = $event->getStartTimestamp();
         }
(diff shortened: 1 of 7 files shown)


stayallive (approved):
👍

stayallive (changes requested):
Soooo actually there is some discussion about the actual value maybe needing to be `production` instead of `default`, see discussion in [#php on discord](https://discord.gg/TjCmyGC). Let's hold of a little bit before merging.

stayallive (changes requested):
So that way quick: https://github.com/getsentry/develop/commit/676ab2bc2818d91ba8aa78dbf45768121f835420

We should change it to use `production` not `default` by default and allow the user to pass `null` if they want to (https://discordapp.com/channels/621778831602221064/621964286952210432/766052439660560384).

I guess that postpones our first deprecation for a bit 😉

src/Serializer/PayloadSerializer.php
```diff
@@ -68,10 +72,6 @@ private function serializeAsEvent(Event $event): string
             $result['release'] = $event->getRelease();
         }
 
-        if (null !== $event->getEnvironment()) {
```
Jean85, line 71: Can we revert this so we reduce the diff?

Jean85 (changes requested):

stayallive (approved):

Issue #1118, opened 2020-10-23
Sentry can't detect errors if a request body size is zero


Bziks:
Sentry can't detect errors if a request body size is zero.
Example:
```php
do {
    try {
        abs();
    } catch (Throwable $e) {
        $s .= str_repeat('*', 1e3);
    }
} while (++$i < 1e6);
```
I get an error "Allowed memory size..."

I found place, where can I fix it, class **RequestIntegration**:
```php
if (
    null === $requestBodySize ||
>>>    0 === $requestBodySize ||   <<<
    'none' === $maxRequestBodySize ||
    ('small' === $maxRequestBodySize && $requestBodySize > self::REQUEST_BODY_SMALL_MAX_CONTENT_LENGTH) ||
    ('medium' === $maxRequestBodySize && $requestBodySize > self::REQUEST_BODY_MEDIUM_MAX_CONTENT_LENGTH)
) {
    return null;
}
```

Probably you add this small fix in the next release?

ste93cry:
Thank you for reporting the bug, a contribution would be happily accepted. Would you be the hero that fixes the bug by opening two PRs, one targeting the `master` branch and one the `2.x` branch? Please also remember to add a test case to the `RequestIntegrationTest` class

Bziks:
Sure! How can I get access?

ste93cry:
You have to fork this repository, make a branch with your changes and then open a PR. If it's your first time contributing you can find a ton of articles on the web that explains all steps: [here](https://www.dataschool.io/how-to-contribute-on-github/) is one, picked totally randomly

ste93cry:
This has been fixed in both `2.x` and `3.x`, thank you for the contribution

PR #1119, opened 2020-10-25
Detect errors if a request body size is zero

Bziks:
Issue - https://github.com/getsentry/sentry-php/issues/1118

### DIFF
--- a/src/Integration/RequestIntegration.php
+++ b/src/Integration/RequestIntegration.php
@@ -160,8 +160,7 @@ final class RequestIntegration implements IntegrationInterface
         $requestBodySize = $request->getBody()->getSize();
 
         if (
-            null === $requestBodySize ||
-            0 === $requestBodySize ||
+            !$requestBodySize ||
             'none' === $maxRequestBodySize ||
             ('small' === $maxRequestBodySize && $requestBodySize > self::REQUEST_BODY_SMALL_MAX_CONTENT_LENGTH) ||
             ('medium' === $maxRequestBodySize && $requestBodySize > self::REQUEST_BODY_MEDIUM_MAX_CONTENT_LENGTH)
--- a/src/Integration/RequestIntegration.php
+++ b/src/Integration/RequestIntegration.php
@@ -161,6 +161,7 @@ final class RequestIntegration implements IntegrationInterface
 
         if (
             null === $requestBodySize ||
+            0 === $requestBodySize ||
             'none' === $maxRequestBodySize ||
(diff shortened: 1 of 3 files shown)


src/Integration/RequestIntegration.php
```diff
@@ -161,6 +161,7 @@ private function captureRequestBody(Options $options, ServerRequestInterface $re
 
         if (
             null === $requestBodySize ||
```
ste93cry, line 163: What about merging this and the next condition to `!$requestBodySize`?

tests/Integration/RequestIntegrationTest.php
```diff
@@ -459,6 +462,25 @@ public function invokeDataProvider(): iterable
...
+                    'Host' => ['www.example.com'],
+                    'Content-Type' => ['application/json'],
+                ],
+            ],
+            null,
+            null,
+        ];
     }
 
     private function getStreamMock(?int $size, string $content = ''): StreamInterface
```
ste93cry, line 486: Since the `$size` depends in the majority of the cases from the `$content` argument, WDYT of removing it, make `$content` nullable and compute `$size` accordingly?

tests/Integration/RequestIntegrationTest.php
```diff
@@ -304,7 +304,8 @@ public function invokeDataProvider(): iterable
             (new ServerRequest('POST', new Uri('http://www.example.com/foo')))
                 ->withUploadedFiles([
                     'foo' => new UploadedFile('foo content', 123, UPLOAD_ERR_OK, 'foo.ext', 'application/text'),
-                ]),
+                ])
+                ->withBody($this->getStreamMock(123 + 321)),
```
ste93cry, line 308: What's the reason of adding `withBody()` to all test cases? The only change I would expect is the new test case that asserts that when the `$requestBodySize` is 0 then everything works fine

ste93cry (changes requested):
Build is broken, probably due to an update of PHPStan. Can you please remove the ignored error from the `phpstan.neon` file and see if it works?

ste93cry:
@Bziks are you willing to make the changes to make this PR mergeable? Otherwise I will take care of them for you

tests/Integration/RequestIntegrationTest.php
ste93cry: I realized only now that it's needed because the handling of the uploaded files happen only if the stream size is greater than 0, so this is fine

ste93cry (approved):
Thank you for your initial work, since I want to release the next patch version asap I took the time to fix the remaining issues. I will also take care of backporting this fix to the 2.x series

PR #1123, opened 2020-11-01
Send stacktrace with events, not just exceptions

mfb:
I noticed that if I set a stacktrace on an event, it is not sent to Sentry.  This patch should resolve the issue.

### DIFF
--- a/src/Serializer/PayloadSerializer.php
+++ b/src/Serializer/PayloadSerializer.php
@@ -152,6 +152,12 @@ final class PayloadSerializer implements PayloadSerializerInterface
             $result['spans'] = array_values(array_map([$this, 'serializeSpan'], $event->getSpans()));
         }
 
+        if ($stacktrace = $event->getStacktrace()) {
+            $result['stacktrace'] = [
+                'frames' => array_map([$this, 'serializeStacktraceFrame'], $stacktrace->getFrames()),
+            ];
+        }
+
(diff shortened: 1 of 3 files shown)


src/Serializer/PayloadSerializer.php
```diff
@@ -152,6 +152,12 @@ private function serializeAsEvent(Event $event): string
             $result['spans'] = array_values(array_map([$this, 'serializeSpan'], $event->getSpans()));
         }
 
+        if ($stacktrace = $event->getStacktrace()) {
```
ste93cry, line 155: Can you please avoid assigning the variable in the `if` statement and check for `null` explicitly?

ste93cry (changes requested):
Good catch! Can you please add a CHANGELOG entry and also a test case for this?

Jean85 (approved):

ste93cry:
I pushed a commit to the `master` branch to solve the broken CI, can you please rebase and try again if it works?

ste93cry (approved):

mkopinsky:
We've found (on an old version of the SDK - ^1.10) that for fatal errors/OOMs, sentry only gets the error but not the stack trace. Does this PR fix that? (Has it already been fixed since ^1.10?)

Jean85:
This is on a completely different version (3.0), the issue is completely unrelated.

PR #1124, opened 2020-11-01
Add getter for $client->stacktraceBuilder

mfb:
Is there any interest in making the StacktraceBuilder available from the client, so that we don't have to create our own?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Add `toArray` public method in `PayloadSerializer` to be able to re-use Event serialization
 - The `withScope` methods now return the callback's return value (#1263)
 - Set the event extras by taking the data from the Monolog record's context (#1244)
-- StacktraceBuilder is no longer marked @internal, and a new public method Client::getStacktraceBuilder() allows custom stacktraces to be built and added to events (#1124)
+- Make the `StacktraceBuilder` class part of the public API and add the `Client::getStacktraceBuilder()` method to build custom stacktraces (#1124)
 
 ## 3.3.7 (2022-01-19)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Add `toArray` public method in `PayloadSerializer` to be able to re-use Event serialization
 - The `withScope` methods now return the callback's return value (#1263)
 - Set the event extras by taking the data from the Monolog record's context (#1244)
-- New public method Client::getStacktraceBuilder() allows custom stacktraces to be built and added to events.
+- StacktraceBuilder is no longer marked @internal, and a new public method Client::getStacktraceBuilder() allows custom stacktraces to be built and added to events.
 
 ## 3.3.7 (2022-01-19)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Add `toArray` public method in `PayloadSerializer` to be able to re-use Event serialization
 - The `withScope` methods now return the callback's return value (#1263)
 - Set the event extras by taking the data from the Monolog record's context (#1244)
-- StacktraceBuilder is no longer marked @internal, and a new public method Client::getStacktraceBuilder() allows custom stacktraces to be built and added to events.
+- StacktraceBuilder is no longer marked @internal, and a new public method Client::getStacktraceBuilder() allows custom stacktraces to be built and added to events (#1124)
 
 ## 3.3.7 (2022-01-19)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Add `toArray` public method in `PayloadSerializer` to be able to re-use Event serialization
 - The `withScope` methods now return the callback's return value (#1263)
 - Set the event extras by taking the data from the Monolog record's context (#1244)
+- New public method Client::getStacktraceBuilder() allows custom stacktraces to be built and added to events.
 
 ## 3.3.7 (2022-01-19)
 
--- a/src/Client.php
+++ b/src/Client.php
@@ -200,6 +200,14 @@ final class Client implements ClientInterface
     }
 
     /**
+     * {@inheritdoc}
+     */
+    public function getStacktraceBuilder(): StacktraceBuilder
+    {
+        return $this->stacktraceBuilder;
+    }
+
+    /**
      * Assembles an event and prepares it to be sent of to Sentry.
      *
      * @param Event          $event The payload that will be converted to an Event
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -803,9 +803,11 @@ final class ClientTest extends TestCase
             $this->createMock(RepresentationSerializerInterface::class)
         );
 
-        $event = Event::createEvent();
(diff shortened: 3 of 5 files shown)


ste93cry:
I don't think that this kind of things should be exposed from the client, they are dependencies that are required to make it work and are marked as `@internal` because we want to have the freedom of evolving them without worring too much about BC

mfb:
I see now that StacktraceBuilder and FrameBuilder are marked `@internal` - that's a little unfortunate for integration developers as they would need to duplicate the code and build stacktraces manually.

(It looks like RepresentationSerializer is not `@internal` so we could perhaps add a getter for that, for what it's worth.)

ste93cry:
>they would need to duplicate the code and build stacktraces manually

Do you have a real use-case in mind about why you need to build the `Stacktrace` object manually? The only reason I see is when building the event from the grounds up, but I don't think it's a common thing as most people just rely on the `captureXXX()` methods

>It looks like RepresentationSerializer is not `@internal`

Because the serializer is injected through the constructor the default implementation cannot be marked as such

mfb:
Here's my real-world use case: https://git.drupalcode.org/project/raven/-/blob/3.x/src/Logger/Raven.php#L236

In my case, either the backtrace already exists in the context of the log message, or I need to create it (removing the unnecessary stack frames related to the logging itself), depending on the type of log message. Then I attach the stacktrace to the event and call `\Sentry\captureEvent()`.

ste93cry:
I definitely see the use-case and it was the one I had in mind, however I still strongly believe that it would not be appropriate to expose such builder from the client object, at least using a getter, because it would be a leak of an internal dependency of the class. What about passing the stacktrace as property of the event hint rather than setting it directly on the event object and then use a custom event processor or `before_send` callback to remove the unwanted frames?

mfb:
> What about passing the stacktrace as property of the event hint rather than setting it directly on the event object and then use a custom event processor or `before_send` callback to remove the unwanted frames?

Well I think either way I need to instantiate a StacktraceBuilder and call buildFromBacktrace()?

ste93cry:
Mmm yes, I though the hint supported the backtrace but I was wrong and it only accepts an already built stacktrace object. However, I think there would be nothing cons if we add such property that will then be translated into a stacktrace using the builder. Do you think it could work?

mfb:
Well, it doesn't really make a difference to me whether I can call a method or bundle data directly into an object/array.

In my case I am passing in dummy file name and line number, since I don't necessarily know or care where the backtrace is coming, and then removing the last frame - perhaps it's a good idea to build in this functionality as well..?

```
$stacktrace = $stacktraceBuilder->buildFromBacktrace($backtrace, '', 0);
$stacktrace->removeFrame(count($stacktrace->getFrames()) - 1);
$event->setStacktrace($stacktrace);
```

Jean85:
It seems that what you're interested in could be summarized into a "named costructor", a static public method on the Stacktrace to create it directly. Do you agree? Do you want to change this PR in that way?

mfb:
> It seems that what you're interested in could be summarized into a "named costructor", a static public method on the Stacktrace to create it directly. Do you agree? Do you want to change this PR in that way?

A static public method on the Stacktrace to create the StacktraceBuilder? It's been a while since I looked at this so probably totally misinterpreting :)

Jean85:
No I meant to create a "dummy" `Stacktrace` as you need.

ste93cry:
After re-reading again this issue and considering the real use-case example that @mfb linked above, maybe after all it makes sense to expose the stacktrace builder (removing the `@internal` annotation). The reason is that we have the `captureEvent()` method, but if you want to pass an event that has a custom stacktrace you cannot easily because you don't have an easy way to build it. @Jean85 WDYT?

mfb:
To create a Stacktrace from a backtrace, I need the Client options, FrameBuilder, RepresentationSerializer and the StacktraceBuilder instance, on which I call the buildFromBacktrace() method.  I'm not sure how a Stacktrace static method would give me what I need, but maybe some pseudocode would help me grok :)

My initial proposal - a getter for `$client->stacktraceBuilder` seemed like the easiest and most performant solution, as these things are already instantiated with the proper configuration inside the client.

src/Client.php
```diff
@@ -202,6 +202,14 @@ public function flush(?int $timeout = null): PromiseInterface
         return $this->transport->close($timeout);
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function getStacktraceBuilder(): StacktraceBuilder
```
ste93cry, line 208: We should make this method part of the interface. To avoid a BC, it should be done with the `@method` annotation

ste93cry (changes requested):
As commented above, we should remove the `@internal` annotation from the `StacktraceBuilder` class to officialize it as part of the public API

src/Client.php
mfb: Does this do it? I don't have a PhpStorm etc. on this laptop to test

tests/ClientTest.php
```diff
@@ -770,6 +770,45 @@ public function testBuildWithStacktrace(): void
...
+                $stacktrace = $event->getStacktrace();
+
+                $this->assertInstanceOf(Stacktrace::class, $stacktrace);
+
+                /** @var Frame $lastFrame */
+                $lastFrame = array_reverse($stacktrace->getFrames())[0];
+
+                $this->assertSame(
+                    'MyApp.php',
+                    basename($lastFrame->getFile())
```
ste93cry, line 791: Do we really need a call to `basename()` here? Isn't this value exactly the one you pass to the `buildFromBacktrace()` method?

src/Client.php
ste93cry: Yes it does 👍

tests/ClientTest.php
mfb: probably not, I copy pasted this code

mfb:
> Please also add a CHANGELOG entry

OK added

mfb:
> > Please also add a CHANGELOG entry
> 
> OK added

Added a mention that StacktraceBuilder is no longer internal

Jean85 (changes requested):
Code looks fine to me :+1: you should fix the PHPStan error, you just need to remove 2 lines from the baseline (or regenerate it to have the same effect)

mfb:
> Code looks fine to me +1 you should fix the PHPStan error, you just need to remove 2 lines from the baseline (or regenerate it to have the same effect)

hmm as far as I can tell, the PHPStan errors are unrelated to this PR, those lines should also be removed on develop and master branches

ste93cry:
Would you mind rebasing the PR? I've updated the baseline, so build should be green now

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Add `toArray` public method in `PayloadSerializer` to be able to re-use Event serialization
 - The `withScope` methods now return the callback's return value (#1263)
 - Set the event extras by taking the data from the Monolog record's context (#1244)
+- StacktraceBuilder is no longer marked @internal, and a new public method Client::getStacktraceBuilder() allows custom stacktraces to be built and added to events (#1124)
```
ste93cry, line 9:
```suggestion
- Make the `StacktraceBuilder` class part of the public API and add the `Client::getStacktraceBuilder()` method to build custom stacktraces (#1124)
```

ste93cry (changes requested):
A really small change, then LGTM for me

ste93cry (approved):

Jean85 (approved):

PR #1126, opened 2020-11-04
Add Scope::removeTag() method

vojtech-dobes:
This was cheap to put together, so feel free to cheaply kill this. But I can't find other way how to remove tag other than clear the whole `Scope`, which is something I don't believe is desired (I don't want to lose default tags + I don't want to lose tags set by myself in higher level of the code).

### DIFF
--- a/tests/State/ScopeTest.php
+++ b/tests/State/ScopeTest.php
@@ -49,6 +49,27 @@ final class ScopeTest extends TestCase
         $this->assertSame(['foo' => 'bar', 'bar' => 'baz'], $event->getTags());
     }
 
+    public function testRemoveTag(): void
+    {
+        $scope = new Scope();
+        $event = $scope->applyToEvent(Event::createEvent());
+
+        $scope->setTag('foo', 'bar');
+        $scope->setTag('bar', 'baz');
+
+        $event = $scope->applyToEvent(Event::createEvent());
+
+        $this->assertNotNull($event);
+        $this->assertSame(['foo' => 'bar', 'bar' => 'baz'], $event->getTags());
+
+        $scope->removeTag('foo');
+
+        $event = $scope->applyToEvent(Event::createEvent());
+
+        $this->assertNotNull($event);
+        $this->assertSame(['bar' => 'baz'], $event->getTags());
+    }
+
     public function testSetAndRemoveContext(): void
     {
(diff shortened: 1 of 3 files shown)


tests/State/ScopeTest.php
```diff
@@ -49,6 +49,29 @@ public function testSetTags(): void
         $this->assertSame(['foo' => 'bar', 'bar' => 'baz'], $event->getTags());
     }
 
+    public function testRemoveTag(): void
+    {
+        $scope = new Scope();
+        $event = $scope->applyToEvent(Event::createEvent());
+
+        $this->assertNotNull($event);
+        $this->assertEmpty($event->getTags());
```
ste93cry, line 58: There is no need for these lines as the "base" case is that a scope has some tag and they get removed, we don't care to assert that the scope is initially empty

ste93cry (changes requested):
Apart from fixing the CS and adding an entry in the CHANGELOG, changes are fine for me, however if I don't recall wrong there has been some discussion in the past about adding such removal methods. I'm sure @HazAT can give us some advice on whether we want or not such feature, but I am personally not contrary though

tests/State/ScopeTest.php
vojtech-dobes: Thanks, I've removed these 2 lines. Please let me know if you would further cleanup the case from unnecessary stuff.

ste93cry: I think that it's fine, the only thing I forgot to write is that since the newly added method is removing a specific tag it would be better to add to the scope two tags so that we can assert that only the right one is removed

ste93cry:
Can you please target this PR against the `develop` branch and can you please reference the PR in the CHANGELOG?

vojtech-dobes:
@ste93cry Feedback incorporated (changelog, removing one of two, base on develop).

HazAT (approved):
Thx :)

ste93cry:
@vojtech-dobes can you 🙏 rebase the PR once more? Build should now be green

vojtech-dobes:
@ste93cry done :)

ste93cry (approved):
Thank you for your hard work!

vojtech-dobes:
You're welcome!

PR #1132, opened 2020-11-13
build(ci): Use GitHub Actions

billyvg:
`travis-ci.org` is shutting down at the end of the year. Here is a draft to use GitHub Actions instead. Another option is to migrate to `travis-ci.com`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: test
+name: ci
 
 on:
   push:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,71 @@
+name: test
+
+on:
+  push:
+    branches:
+      - master
+      - develop
+  pull_request:
+
+jobs:
+  static-analysis:
+    name: Code style and static analysis
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        php: [7.4]
+        script: [phpcs, phpstan, psalm]
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup PHP ${{ matrix.php }}
+        uses: shivammathur/setup-php@c596331f84de248c01c26db5a184880781e96e36
+        with:
+          php-version: ${{ matrix.php }}
+
+      - name: Install dependencies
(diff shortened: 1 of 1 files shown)


billyvg:
@HazAT who can I tag to take a look at this o.o

ste93cry:
Thank you for the heads up on this, honestly I knew that `travis-ci.org` was shutting down but not when. While I don't have anything in contrary in the switch, I'm not sure that doing it is worth the time as we already have a fully working config for two CIs and I cannot think of any advantage of GitHub actions over Travis CI for our use-case. However, since I believe that you may want to standardize the services used across all SDK projects, I will gladly accept any outcome of this discussion 👍

Jean85:
IIRC Travis is also limiting the hours of execution for CI projects...
Also, Actions, due to the native integration, are pretty responsive compared to any other solution.

ste93cry:
>IIRC Travis is also limiting the hours of execution for CI projects

I don't know, but we never had a problem with this

>are pretty responsive compared to any other solution

I have to admit that I don't see a difference between Travis and GitHub, and I feel that the UI of the GitHub Actions is more cucumbersome and slow because each step is collapsed and opening them take time to load the data. Anyway, if we decide to switch we should also switch the Windows builds over here as I don't see a benefit of keeping AppVeyor at that point

Jean85:
Having a single CI with both Linux & Windows builds would be a nice perk, indeed!

ste93cry:
I just saw in a recent PR that the project already got migrated to `travis-ci.com` by the way 😉

billyvg:
@ste93cry yeah we mass migrated all of the repos over to `.com`, it doesn't look like there's too much interest in moving though, so we can close this. Feel free to ping me if this isn't the case and I can help finish this up.

billyvg:
@ste93cry I'm re-opening this because of billing issues with Travis, we now definitely need to move off of Travis. They pulled some pretty shady stuff with billing and we're now being charged for all of our open source repos (with no warning afaik).

ste93cry:
We switched to GitHub Actions already some time ago. Closing this again 😃

billyvg:
🙈 Thanks!

PR #1133, opened 2020-11-16
Refactor HttpClientFactory

IonBazan:
This PR moves HTTP client resolution to a protected method so to avoid `elseif` statements and simplify the code.


src/HttpClient/HttpClientFactory.php
```diff
@@ -175,4 +128,61 @@ public function create(Options $options): HttpAsyncClientInterface
...
+            if (null !== $options->getHttpProxy()) {
+                $curlConfig[CURLOPT_PROXY] = $options->getHttpProxy();
+            }
+
+            /** @psalm-suppress InvalidPropertyAssignmentValue */
+            return new CurlHttpClient(null, null, $curlConfig);
+        }
+
+        if (null !== $options->getHttpProxy()) {
+            throw new \RuntimeException('The "http_proxy" option requires either the "php-http/curl-client" or the "php-http/guzzle6-adapter" package to be installed.');
```
Jean85, line 183: We should update this message, we now support the Symfony client too.

Jean85 (changes requested):

src/HttpClient/HttpClientFactory.php
```diff
@@ -175,4 +128,61 @@ public function create(Options $options): HttpAsyncClientInterface
 
         return new PluginClient($httpClient, $httpClientPlugins);
     }
+
+    /**
+     * @param Options $options
+     *
+     * @return ClientInterface|HttpAsyncClient
```
ste93cry, line 135: The method return type should be `HttpAsyncClientInterface` as the whole SDK still works using such interface rather than the PSR-18 non-async interface

ste93cry (changes requested):
Since this part is not tested, we have to be really careful changing it as we risk to break people's apps like it happened in the past ⚠️ Said this, apart from a few changes it looks fine to me

IonBazan:
@ste93cry Thank you for review - I've applied the changes requested. I was thinking about writing a proper test for this method as it is very important but not covered. That would require to add `php-http/curl-client`, `php-http/guzzle6-adapter` and `symfony/http-client` to `require-dev` and some autoloader magic together with running the tests in separate processes to handle this situations properly so I'm not sure if it's worth it.

ste93cry:
>That would require to add `php-http/curl-client`, `php-http/guzzle6-adapter` and `symfony/http-client` to `require-dev`

I think we could do this while running the CI and then making the test suite skip each test if the package is not installed, similar to the `@requires` annotation of PHPUnit. However, the main issue I never got around is how to really test that each HTTP client is configured correctly, because it's simple to assert that the instance of the returned value is of the expected type but it's much more difficult to assert that the config is the expected one

tests/HttpClient/HttpClientFactoryTest.php
```diff
@@ -101,4 +116,79 @@ public function testCreateThrowsIfHttpProxyOptionIsUsedWithCustomHttpClient(): v
...
+    {
+        $httpClientFactory = new HttpClientFactory(
+            Psr17FactoryDiscovery::findUrlFactory(),
+            Psr17FactoryDiscovery::findResponseFactory(),
+            Psr17FactoryDiscovery::findStreamFactory(),
+            null,
+            'sentry.php.test',
+            '1.2.3'
+        );
+        class_exists(HeaderSetPlugin::class);
```
ste93cry, line 170: Correct me if I'm wrong, but I think that with all these `class_exists` function calls you are trying to assert that those classes got loaded by the HTTP client factory. However, I see at least two problems here: the first is that the first time such classes are loaded is when calling the `HttpClientFactory::create()` method and not its constructor, and the second is that the default value of the `$autoload` parameter of `class_exists` is `true`, so there is no assurance that those classes have been loaded before. As a last thing, testing the internal state of a class is bad for a lot of reasons, we did it in the past and I would like to avoid it since than. I wonder if this class has too many responsibilities and if a different approach would benefit both the maintenance and the tests (e.g. having a factory for each type of HTTP client we support)

ste93cry (changes requested):
Thank you for the effort you put into this PR, however many things are hard to understand (e.g. the tricks with the autoloader) or we would lock us into a specific version of an HTTP client while we cannot test another one. Since like in the past it I don't think there is an easy solution to test this part, either I prefer you to rollback changes so that  we merge just the refactoring you did of the factory or leave this PR pending until we came up with a better architecture design that will prevent the issues reported in the code review

ste93cry:
@IonBazan as I don't feel comfortable to merge the PR as-is with all that reflection and autoload related code, would you mind reverting the changes to stick to the original purpose of this PR which was only the refactoring of the `HttpClientFactory` class?

IonBazan:
@ste93cry Sure, I will replace this PR with a new one this week.

ste93cry:
No need to open a new PR, just push over here the changes

IonBazan:
@ste93cry I guess that week lasted 5 months 🙈  Anyway, removed these fragile tests and resolved all conflicts. PTAL

ste93cry (approved):
Looks fine to me 👍

Jean85 (approved):
LGTM too! :+1: 
Thank you!

PR #1134, opened 2020-11-16
Deprecate SpanContext::fromTraceparent() and allow passing custom sampling context data to startTransaction()

HazAT:
This is a PR to align the PHP SDKs to be equal on the implementation level as other SDKs.

The change to `fromTraceParent` is strictly speaking breaking and we also need to update Laravel SDK.
https://github.com/getsentry/sentry-laravel/pull/420


ste93cry:
With the premise that I haven't looked at the code completely yet, why is it breaking? It looks to me that you just moved `SpanContext::fromTraceparent()` to `TransactionContext::fromSentryTrace()`, so we could handle this in a BC-compatible way by reporting a deprecation error in the parent method and overriding it in the subclass :thinking:

src/State/HubInterface.php
```diff
@@ -139,8 +140,9 @@ public function getIntegration(string $className): ?IntegrationInterface;
      * Sentry.
      *
      * @param TransactionContext $context Properties of the new transaction
+     * @param array<string, mixed> $customSamplingContext Additional context that will be passed to the SamplingContext
      */
-    public function startTransaction(TransactionContext $context): Transaction;
+    public function startTransaction(TransactionContext $context, ?array $customSamplingContext): Transaction;
```
ste93cry, line 145: This is a huge breaking change, which is not acceptable without releasing a new major version. Rather than adding this argument for real, you can comment it and handle it with `func_get_arg` in the implementation class to avoid the break: we already did it in the past and it worked even though it was a bit messy

src/Tracing/SamplingContext.php
```diff
@@ -16,6 +16,12 @@ final class SamplingContext
      */
     private $parentSampled;
 
+    /**
+     * @var array<string, mixed>|null Additional context, depending on where the SDK runs
+     */
+    private $additionalContext;
+
```
ste93cry, line 23: One blank line is more than sufficient 😄

src/Tracing/SamplingContext.php
```diff
@@ -48,4 +54,20 @@ public function setParentSampled(?bool $parentSampled): void
     {
         $this->parentSampled = $parentSampled;
     }
+
+    /**
+     * @param array<string, mixed>|null $additionalContext
```
ste93cry, line 59: Can you please add a little description of what this function does, e.g. `Sets additional optional context data.`?

src/Tracing/SamplingContext.php
```diff
@@ -48,4 +54,20 @@ public function setParentSampled(?bool $parentSampled): void
...
+     */
+    public function setAdditionalContext(?array $additionalContext): void
+    {
+        $this->additionalContext = $additionalContext;
+    }
+
+    /**
+     * @return array<string, mixed>|null
```
ste93cry, line 67: Can you please add a little description of what this function does, e.g. `Gets additional optional context data.`?

src/Tracing/SpanContext.php
```diff
@@ -6,8 +6,6 @@
 
 class SpanContext
 {
-    private const TRACEPARENT_HEADER_REGEX = '/^[ \\t]*(?<trace_id>[0-9a-f]{32})?-?(?<span_id>[0-9a-f]{16})?-?(?<sampled>[01])?[ \\t]*$/i';
```
ste93cry, line 9: This is a breaking change ⚠️ Instead of removing it, please add the following DocBlock to it

```php
/**
 * @deprecated since version 3.1, to be removed in 4.0
 */
```

src/Tracing/SpanContext.php
```diff
@@ -184,35 +182,4 @@ public function setEndTimestamp(?float $endTimestamp): void
...
     }
-
-    /**
-     * Returns a context populated with the data of the given header.
-     *
-     * @param string $header The sentry-trace header from the request
-     *
-     * @return static
-     */
-    public static function fromTraceparent(string $header)
```
ste93cry, line 195: This is a breaking change and cannot be done. Please leave the method as-is, mark is as deprecated and throw a `E_DEPRECATED` error following the standard of what has been done in `2.x`

src/State/Hub.php
```diff
@@ -227,24 +227,28 @@ public function startTransaction(TransactionContext $context): Transaction
         }
 
         $samplingContext = SamplingContext::getDefault($context);
+        $samplingContext->setAdditionalContext($customSamplingContext);
         $tracesSampler = $options->getTracesSampler();
```
ste93cry, line 231: Please add a blank line above this one to improve readability

ste93cry (changes requested):

src/State/Hub.php
```diff
@@ -313,8 +317,16 @@ private function getSampleRate(?bool $hasParentBeenSampled, float $fallbackSampl
         return $fallbackSampleRate;
     }
 
-    private function isValidSampleRate(float $sampleRate): bool
+    /**
+     * @param mixed $sampleRate
```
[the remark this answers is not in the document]
HazAT: We decided for `mixed` case in theory since it's a user defined function the user could return anything.

HazAT:
Appveyor is failing because of rate limiting

src/Tracing/SpanContext.php
```diff
@@ -191,9 +194,12 @@ public function setEndTimestamp(?float $endTimestamp): void
      * @param string $header The sentry-trace header from the request
      *
      * @return static
+     *
+     * @deprecated since version 3.1, to be removed in 4.0 use {@link TransactionContext::fromSentryTrace()}
      */
     public static function fromTraceparent(string $header)
     {
+        @trigger_error(sprintf('The %s() method is deprecated since version 3.1 and will be removed in 4.0. Use TransactionContext::fromSentryTrace() instead.', __METHOD__), E_USER_DEPRECATED);
```
ste93cry, line 202: Please add a blank line below to improve readability. Some day you will get this right 😄

src/Tracing/TransactionContext.php
```diff
@@ -65,4 +66,34 @@ public function setParentSampled(?bool $parentSampled): void
...
+    /**
+     * Returns a context populated with the data of the given header.
+     *
+     * @param string $header The sentry-trace header from the request
+     *
+     * @return static
+     */
+    public static function fromSentryTrace(string $header)
+    {
+        $context = new static();
```
ste93cry, line 79: Since the class is `final` we can now transform this to `new self()` and also add the typehint to the method rather than using the annotation

src/Tracing/TransactionContext.php
```diff
@@ -6,6 +6,7 @@
 
 final class TransactionContext extends SpanContext
 {
+    private const TRACEPARENT_HEADER_REGEX = '/^[ \\t]*(?<trace_id>[0-9a-f]{32})?-?(?<span_id>[0-9a-f]{16})?-?(?<sampled>[01])?[ \\t]*$/i';
```
ste93cry, line 9: Please add a blank line below to improve readability.

src/State/Hub.php
```diff
@@ -216,6 +216,12 @@ public function getIntegration(string $className): ?IntegrationInterface
      */
     public function startTransaction(TransactionContext $context): Transaction
     {
+        $numArgs = \func_num_args();
+        $customSamplingContext = null;
+        if ($numArgs > 1) {
```
ste93cry, line 221: Both add a new blank line above this one to improve readability and inline the `func_num_args()` function call

src/Tracing/SamplingContext.php
```diff
@@ -48,4 +55,24 @@ public function setParentSampled(?bool $parentSampled): void
     {
         $this->parentSampled = $parentSampled;
     }
+
+    /**
+     * Anything that will be provided as an second argument in {@link startTransaction()} will be set here.
```
ste93cry, line 60: I think a better description would be `Sets additional data that will be provided as a second argument to {@link \Sentry\startTransaction()}.`

src/Tracing/SamplingContext.php
```diff
@@ -48,4 +55,24 @@ public function setParentSampled(?bool $parentSampled): void
...
+     *
+     * @param array<string, mixed>|null $additionalContext
+     */
+    public function setAdditionalContext(?array $additionalContext): void
+    {
+        $this->additionalContext = $additionalContext;
+    }
+
+    /**
+     * Anything that was provided as a second argument by {@link startTransaction()} will be returned here.
```
ste93cry, line 70: I think a better description would be `Gets the additional data that will be provided as a second argument to {@link \Sentry\startTransaction()}.`

tests/DsnTest.php
```diff
@@ -171,6 +171,16 @@ public function createFromStringThrowsExceptionIfValueIsInvalidDataProvider(): \
             'tcp://public:secret@example.com/1',
             'The scheme of the "tcp://public:secret@example.com/1" DSN must be either "http" or "https".',
         ];
+
+        yield 'valid project id' => [
```
ste93cry, line 175: As far as I know you cannot have two test cases with the same name/description. Please change this description to `invalid project ID (char instead of number)`

tests/DsnTest.php
```diff
@@ -171,6 +171,16 @@ public function createFromStringThrowsExceptionIfValueIsInvalidDataProvider(): \
...
+            'http://public:secret@example.com/j',
+            'DSN must contain a valid project ID.',
+        ];
+
+        yield 'valid project id' => [
```
ste93cry, line 180: Please change this description to `invalid project ID (negative number)` to make it different from the one before

ste93cry (changes requested):
Two things are still missing: the CHANGELOG entry and the `@expectedDeprecation` annotation on the failing test method (you can check the `2.x` branch to see how it works)

src/Tracing/SamplingContext.php
```diff
@@ -48,4 +53,24 @@ public function setParentSampled(?bool $parentSampled): void
...
+
+    /**
+     * Sets additional data that will be provided as a second argument to {@link \Sentry\startTransaction()}.
+     *
+     * @param array<string, mixed>|null $additionalContext
+     */
+    public function setAdditionalContext(?array $additionalContext): void
+    {
+        $this->additionalContext = $additionalContext;
+    }
```
lobsterkatie, line 65: IDK how much this difference matters, but in the other SDKs, the context data provided in `customSamplingContext` is spread into `samplingContext`, whereas here it seems like it's being added as a single separate entry called `additionalContext`.

tests/Tracing/TransactionContextTest.php
```diff
@@ -22,4 +24,60 @@ public function testGettersAndSetters(): void
...
+        if (null !== $expectedSpanId) {
+            $this->assertEquals($expectedSpanId, $spanContext->getParentSpanId());
+        }
+
+        if (null !== $expectedTraceId) {
+            $this->assertEquals($expectedTraceId, $spanContext->getTraceId());
+        }
+
+        $this->assertSame($expectedSampled, $spanContext->getParentSampled());
+    }
```
lobsterkatie, line 44:
```suggestion
    public function testFromTraceparent(string $header, ?SpanId $expectedSpanId, ?TraceId $expectedTraceId, ?bool $expectedParentSampled): void
    {
        $spanContext = TransactionContext::fromSentryTrace($header);

        if (null !== $expectedSpanId) {
            $this->assertEquals($expectedSpanId, $spanContext->getParentSpanId());
        }

        if (null !== $expectedTraceId) {
            $this->assertEquals($expectedTraceId, $spanContext->getTraceId());
        }

        $this->assertSame($expectedParentSampled, $spanContext->getParentSampled());
    }
```

src/Tracing/SamplingContext.php
HazAT: I would say this is ok to be not 100% the same, spreading this in PHP isn't something that feels natural in PHP.

ste93cry: My two cents: it's not really important whether we want to use the spreading operator or not, the key question is if this method should merge the new data with the existing one or not

lobsterkatie: That (merging) is how it works elsewhere. I feel like it's just a smidge nicer not to have to go digging any deeper than you need to for stuff, so I'd be in favor of the spread, but I don't feel strongly.

ste93cry: >spreading this in PHP isn't something that feels natural in PHP

Reading again @HazAT comment I think that what he meant is that in PHP a setter sets data, does not merge it, at least usually. I agree with this, but I don't have a strong opinion so I will let you two choose what to do. As far as I'm concerned, PR looks good and I'm waiting this discussion to reach an agreement before giving the final approval

lobsterkatie: >  in PHP a setter sets data, does not merge it

Ah, I see. Thanks for the explanation. Any reason we couldn't do away with this method entirely and just do

```PHP
$samplingContext = array_merge(SamplingContext::getDefault($context), $customSamplingContext));
```

in `startTransaction()`?

ste93cry: We cannot because `SamplingContext::getDefault()` return an object instance, so `array_merge` cannot work because you are trying to merge it with an array. Either we don't bother about the setter merging data (it's not a rule written in the stone that a setter must overwrite data rather than merging it) or we could rename the method to something else, e.g. `mergeAdditionalContext`

lobsterkatie: Okay. I poked at it a bit and I guess the real answer here would be to have `SamplingContext` just be a hashmap rather than a full-blown object instance, thereby allowing for arbitrary keys, but I assume that would be a totally breaking change, so... I think it's fine. Let's just leave it.

Thanks for the explanations - it's helpful to know a tiny bit about how PHP works.

lobsterkatie (approved):

ste93cry (approved):

PR #1138, opened 2020-11-19
Allow passing the event hint to the captureException method and pass it down to the before_send callback

jarstelfox:
I have a use case where we have extended Exception to have some extra data. Ex:

```php
class ExtraDataException {
   private $extraData = [];
   public function addExtraData($data) {
      $extraData[] = $data;
   }
   public function getExtraData(): array {
      return $this->extraData;
   }
}
```

This is ideologically different from how sentry works. Namely, the behaviour is to wrap with a scope to be able to `$scope->setExtra('data', $data)`.

I am not opposed to this at all. In fact, I like it better than our current approach.

However, I have issues being able to shim in the functionality I need at a global level, as I lose context of the exception right here.

With this change I can do something like:

```php
'before_send' => function (Event $event): Event {
   $data = array_map(function(ExceptionDataBag $e) {
      return $e->getThrowable()->getExtraData();
   }, $event->getExceptions());
   $event->setExtra($data);
   return $event;
}
```

The main reason for this pull is to allow us to shim our existing code.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -123,23 +123,39 @@ final class Client implements ClientInterface
     /**
      * {@inheritdoc}
      */
-    public function captureMessage(string $message, ?Severity $level = null, ?Scope $scope = null): ?EventId
+    public function captureMessage(string $message, ?Severity $level = null, ?Scope $scope = null/*, ?EventHint $hint = null*/): ?EventId
     {
+        $hint = \func_num_args() > 3 ? func_get_arg(3) : null;
+
+        if (null !== $hint && !$hint instanceof EventHint) {
+            throw new \InvalidArgumentException(sprintf('The $hint argument must be an instance of the "%s" class. Got: "%s".', EventHint::class, get_debug_type($hint)));
+        }
+
         $event = Event::createEvent();
         $event->setMessage($message);
         $event->setLevel($level);
 
-        return $this->captureEvent($event, null, $scope);
+        return $this->captureEvent($event, $hint, $scope);
     }
 
     /**
      * {@inheritdoc}
      */
-    public function captureException(\Throwable $exception, ?Scope $scope = null): ?EventId
+    public function captureException(\Throwable $exception, ?Scope $scope = null/*, ?EventHint $hint = null*/): ?EventId
     {
-        return $this->captureEvent(Event::createEvent(), EventHint::fromArray([
-            'exception' => $exception,
-        ]), $scope);
+        $hint = \func_num_args() > 2 ? func_get_arg(2) : null;
+
+        if (null !== $hint && !$hint instanceof EventHint) {
+            throw new \InvalidArgumentException(sprintf('The $hint argument must be an instance of the "%s" class. Got: "%s".', EventHint::class, get_debug_type($hint)));
+        }
+
+        $hint = $hint ?? new EventHint();
+
+        if (null === $hint->exception) {
+            $hint->exception = $exception;
+        }
+
+        return $this->captureEvent(Event::createEvent(), $hint, $scope);
     }
 
     /**
@@ -170,8 +186,9 @@ final class Client implements ClientInterface
     /**
      * {@inheritdoc}
      */
-    public function captureLastError(?Scope $scope = null): ?EventId
+    public function captureLastError(?Scope $scope = null/*, ?EventHint $hint = null*/): ?EventId
     {
+        $hint = \func_num_args() > 1 ? func_get_arg(1) : null;
         $error = error_get_last();
 
         if (null === $error || !isset($error['message'][0])) {
@@ -180,7 +197,7 @@ final class Client implements ClientInterface
 
         $exception = new \ErrorException(@$error['message'], 0, @$error['type'], @$error['file'], @$error['line']);
 
-        return $this->captureException($exception, $scope);
+        return $this->captureException($exception, $scope, $hint);
     }
 
     /**
@@ -267,7 +284,7 @@ final class Client implements ClientInterface
 
         if (!$isTransaction) {
             $previousEvent = $event;
-            $event = ($this->options->getBeforeSendCallback())($event);
+            $event = ($this->options->getBeforeSendCallback())($event, $hint);
 
             if (null === $event) {
                 $this->logger->info('The event will be discarded because the "before_send" callback returned "null".', ['event' => $previousEvent]);
--- a/src/Client.php
+++ b/src/Client.php
@@ -123,14 +123,8 @@ final class Client implements ClientInterface
     /**
      * {@inheritdoc}
      */
-    public function captureMessage(string $message, ?Severity $level = null, ?Scope $scope = null/*, ?EventHint $hint = null*/): ?EventId
+    public function captureMessage(string $message, ?Severity $level = null, ?Scope $scope = null, ?EventHint $hint = null): ?EventId
     {
-        $hint = \func_num_args() > 3 ? func_get_arg(3) : null;
-
-        if (null !== $hint && !$hint instanceof EventHint) {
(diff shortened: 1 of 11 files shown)


ste93cry:
Sorry for not answering your question, I forgot it (shame on me). I'm not sure I understand why you cannot set the data on the scope but you can on the event hint. As a side note, it looks like in the JS SDK you can set the [`captureContext`](https://github.com/getsentry/sentry-javascript/blob/d8243fde0845901c5a78dcb18e46c3b00d7302ae/packages/types/src/event.ts#L52) key which is defined as [`Scope | Partial<ScopeContext> | ((scope: Scope) => Scope)`](https://github.com/getsentry/sentry-javascript/blob/d8243fde0845901c5a78dcb18e46c3b00d7302ae/packages/types/src/scope.ts#L12). Would it be enough for you to solve your issue in a more general way without having a specific field for the exception?

jarstelfox:
> As a side note, it looks like in the JS SDK you can set the [`captureContext`](https://github.com/getsentry/sentry-javascript/blob/d8243fde0845901c5a78dcb18e46c3b00d7302ae/packages/types/src/event.ts#L52) key which is defined as [`Scope | Partial<ScopeContext> | ((scope: Scope) => Scope)`](https://github.com/getsentry/sentry-javascript/blob/d8243fde0845901c5a78dcb18e46c3b00d7302ae/packages/types/src/scope.ts#L12). Would it be enough for you to solve your issue in a more general way without having a specific field for the exception?

That would work! For reference, how I solved this for us was a custom integration.

```php
class DebugExceptionIntegration extends AbstractErrorListenerIntegration {
   /**
    * {@inheritdoc}
    */
   protected function captureException(HubInterface $hub, Throwable $exception): void {
      $hub->withScope(function (Scope $scope) use ($hub, $exception): void {
         $scope->addEventProcessor(\Closure::fromCallable([$this, 'addExceptionMechanismToEvent']));

         if ($exception instanceof DebugException && $exception->hasDebugInfo()) {
            $scope->setExtra('DebugInfo', $exception->getDebugInfoStr());
         }

         $hub->captureException($exception);
      });
   }
```

Some background of *why* I opened this PR
---

Our previous error handler expected to get the exception at a similar time as `before_send`. When working with sentry I looked as the callback but [noticed it is not passed the hint (which has the exception)](https://github.com/jarstelfox/sentry-php/blob/master/src/Client.php#L255). I also noticed that the ExcpetionBag does not save the full exception and thought I may be missing something. After some digging, I was fairly certain that I did not have access to the exception during the on `before_send` function. Hence the pull.

It's possible I am still missing something and the $event does have access to the hint. If I missed It, I suspect other people would as well.

tl;dr
---

Any way to access the actual exception object just before reporting would be great!

ste93cry:
>That would work! For reference, how I solved this for us was a custom integration.

That was the solution I was going to suggest you as soon as I realized that what I wrote before did not make sense because I confused the `ExceptionDataBag` object with the `EventHint` object. They are indeed two distinct objects with two distinct purposes: while the first is the data object model of the API payload for the [Exception interface](https://develop.sentry.dev/sdk/event-payloads/exception/), the latter is an object that should be passed along to carry information to the event processors and to the `before_send` callback

>I was fairly certain that I did not have access to the exception during the on `before_send` function

And you were right. The key point is that we forgot to pass the hint to the callback like [it's done](https://github.com/getsentry/sentry-javascript/blob/d8243fde0845901c5a78dcb18e46c3b00d7302ae/packages/core/src/baseclient.ts#L507) in the JS SDK. If we did this, we would then be able fill the `EventHint::$exception` property and pass the object to the `captureException` method which in turn would carry that information over up to the `before_send` callback. If you are willing to reopen this PR and change it to make it work as expected I would gladly accept it.

jarstelfox:
> If you are willing to reopen this PR and change it to make it work as expected I would gladly accept it.

Will do!

ste93cry:
There will be a few things to take into account, more specifically the fact that the captureException method of both the `HubInterface` and `ClientInterface` interfaces cannot be changed without breaking BC, so you will have to do some magic with `func_get_args()` to handle the new argument. Also, this PR should target the `develop` branch because it will be released in the next minor version. Thank you very much in advance for the patience and the willingness of still helping out even if you didn't get an answer at first

jarstelfox:
@ste93cry I seem to be hitting a weird CI issue with xdebug + code coverage. As you can see, the code changes would have no effect on these. However, I am failing CI on:
```
Use of undefined constant XDEBUG_CC_UNUSED - assumed 'XDEBUG_CC_UNUSED' (this will throw an Error in a future version of PHP)
```

WOuld you mind helping me out here?

ste93cry:
I believe it's because Xdebug 3 is not compatible with PHPUnit 7.5, but PHP 7.3.25 on Travis uses it. The issue was introduced with php-build/php-build@ac92f4e0e8a5860f663c2ece71123500a0db73d2 as a consequence of php-build/php-build#650

jarstelfox:
@ste93cry Thanks for the links. I was looking around for some breadcrumbs as to what happened. I saw your fix was merged. Would you mind re triggering a build on this pull? I don't see the option in Travis, I assume I don't have enough permission.

ste93cry:
I restarted the build but it still doesn't work because of Xdebug 3 that is not supported by PHPUnit 7.5. At this point, since the update of that package is already done in the PR to support PHP8, my suggestion is to wait until it gets merged. Please also remember that these changes are not sufficient yet as you should add the hint to the `HubInterface::captureException()` method too

ste93cry:
`develop` branch should again be green, if you don't mind rebasing and making the last changes required to get this PR working I will be happy to review it once more

ste93cry:
@jarstelfox are you still interested in completing the work? Otherwise, would you mind if I do it for you?

jarstelfox:
@ste93cry feel free to take it up. Sorry between work and the holidays, I forgot about this

ste93cry:
Even if the CI is broken (until I merge `master` into `develop` to port the switch from Travis to GA), I expect everything to work, so if you want to give it a shot to see if it helps you with your original issue then it would be 🆒 

Side note: I rebased on the wrong branch, in the tentative of fixing my mistake I lost your initial commit. I will anyway ensure to give you the proper credits for your contribution in the commit

ste93cry (approved):

src/Client.php
```diff
@@ -123,23 +123,39 @@ public function getOptions(): Options
     /**
      * {@inheritdoc}
      */
-    public function captureMessage(string $message, ?Severity $level = null, ?Scope $scope = null): ?EventId
+    public function captureMessage(string $message, ?Severity $level = null, ?Scope $scope = null/*, ?EventHint $hint = null*/): ?EventId
```
Jean85, line 126: The client is `final`, why are we doing this dance here? Isn't having a default value enough for BC?

Jean85 (changes requested):
Sorry for making you backtrack again, but I fear that @ste93cry's suggestion was wrong: adding an argument to the interfaces is a breaking change and you correctly, but nearly (or all?) the implementing classes are final, so adding the argument there without `func_get_arg()` should be fine.

Do you agree @ste93cry?

src/Client.php
ste93cry: No, because the methods must have the same signature of the ones in the interface

ste93cry:
>Do you agree @ste93cry?

Yes, I wasn't aware that adding an argument with a default value even though the signature is different from the interface was fine. I now changed the implementation to get rid of all the calls to `func_get_arg` 👍

Jean85 (approved):

stayallive (approved):
Looks great, thanks all, good addition 👍

PR #1139, opened 2020-11-20
Fix regression introduced in #1119 that prevents capturing of the request body

### LINKED ISSUE #1135: Sentry no longer capturing request body
It appears that after #1119 the RequestIntegration is no longer capturing the request body.

As far as I can tell, there **is** a request body, i.e. getBody() returns e.g. a URL-encoded POST request, but the getSize() method returns zero.

ste93cry:
This PR fixes a regression introduced in #1119 and closes #1135: in particular, the `php://input` stream is not a file and `fstat` does not work on it and `StreamInterface::getSize()` always return `0`. In reality, using the PSR-7 implementation from Guzzle that method does work as you read from the stream, but it does just because it remembers how many bytes it read until that point. Other SDKs implementations use the `Content-Length` header, I decided to do the same even though it may report incorrect data (which is unlikely anyway)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## Unreleased
 
-## 3.0.4 (2020-11-6)
+- Fix capturing of the request body in the `RequestIntegration` integration (#1139)
+
+## 3.0.4 (2020-11-06)
 
 - Fix stacktrace missing from payload for non-exception events (#1123)
 - Fix capturing of the request body in the `RequestIntegration` integration when the stream is empty (#1119)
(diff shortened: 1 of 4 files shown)


mfb:
it's possible for sentry to be installed w/ older versions of guzzlehttp/psr7 which have the now-deprecated copy_to_string() instead of Utils::copyToString() https://github.com/guzzle/psr7/blob/1.x/src/functions.php#L194

ste93cry:
Thanks for noticing it, I didn't see it because PHPUnit emitted a warning and the build was green!

Jean85 (approved):

stayallive (approved):
Looks good! 👍

PR #1087, opened 2020-09-13
PHP 8 Support

Jean85:
This is temporarily blocked by https://github.com/Jean85/pretty-package-versions/pull/25 (which in turn is blocked by https://github.com/phpspec/prophecy/issues/491)

PR #1140, opened 2020-11-26
PHP8 Support

KartaviK:


ste93cry:
Duplicate of #1087. TLDR, we are working since a while to support PHP 8 but there are still a few issues with dependencies

KartaviK:
@ste93cry Can i help with something?

ste93cry:
Last time I had a look at the referenced PR we were waiting for PHPCS to be compatible with PHP8 and there were some failing tests due to some change in PHPUnit and XDebug. The latter issue should be solved with XDebug 3 but Travis CI does not install it yet, so we just have to wait. As soon as all those issues will be solved we will release the new version 😃

PR #1143, opened 2020-11-27
Fix regression introduced in #1119 that prevents capturing of the request body

ste93cry:
Backporting of #1139 from `3.x` to the target branch


stayallive (approved):
👌

PR #1150, opened 2020-12-04
Switch from Travis CI and AppVeyor to GitHub Actions

Jean85:
Includes #1087.


Jean85:
...and that's why I prefer moving to Github Actions. I got a green build in less than 10 minutes of work, build time included.

We could even migrate the Windows build there too, but I don't want to pollute this PR further.

ste93cry:
I have nothing against GitHub Actions, but at the same time I would like to remind that nothing cames for free and at some point in the future it's probable that limitations will be put in place, simply because otherwise it would not be sustainable. I said in the past that I don't agree with moving off from Travis not because I'm tied with it, but mainly because it works well for our use-case like AppVeyor does and to be honest I don't see any thing that we could benefit from after the change. Said this, it's undoubtful that Travis is having some issues with getting PHP8 work, although they are working on it. I also see that the build is passing, but just because you changed from Xdebug to PCOV which is less accurate and is less active in the development, so we can say that you worked around the issue rather than solving it. Anyway, subjective opinions aside, I suggest to wait until the next week to see if situation improves on Travis, otherwise let's do the switch to GA and release this long awaited update to support PHP 8. I don't think that neither the weekend nor a few more days of wait will change the life of people

>We could even migrate the Windows build there too, but I don't want to pollute this PR further

I would prefer to make the changes in the original PR. Also, please, let's avoid in the future the opening of more PR with the same changes as it makes more difficult for people to follow the development, and code reviews may get splitted among different places

ste93cry:
Situation on Travis improved as expected and PHP 8 builds are fine, however we hit the OSS credits limit and I honestly didn't expected it at all since we are not a project with a lot of builds. Anyway, in addition to this, since a while AppVeyor builds fail because of rate limiting on Chocolatey (it's so damn low!), so let's make this switch to GitHub Actions to be happy

@HazAT can you please disable both Travis and AppVeyor? For the latter the checks still run despite me deleting the config file

Jean85:
I had permissions on Appveyor, I probably fixed it (but it will stop working on other branches :( )

PHPStan is probably failing due to different function signatures under PHP 8, where they come from core and no longer from the polyfill. Should we switch back to 7.4 for the time being?

ste93cry:
>but it will stop working on other branches :(

I'm not worried about this, we are going to switch to GitHub actions for all branches of all versions aren't we?

>PHPStan is probably failing due to different function signatures under PHP 8, where they come from core and no longer from the polyfill

They look more like regressions or bugs. I opened phpstan/phpstan#4203 to track the issue with the `get_include_path` function for now

Jean85:
I've also opened https://github.com/phpstan/phpstan/issues/4205 for the same reason.

ste93cry (approved):

PR #1155, opened 2020-12-10
Fix wrong deprecation message for the MissingPublicKeyCredentialException class

ste93cry:
While making some changes to the tests in another PR I found out that the `@expectedDeprecationMessage` annotation apparently was not working at all. Once solved the issue, I discovered a bug in a deprecation error that refers to the wrong class name


stayallive (approved):
👍

Jean85 (approved):

PR #1158, opened 2020-12-27
Avoid calling the before_send callback with transaction events

ste93cry:
Reading the [Tracing documentation](https://develop.sentry.dev/sdk/unified-api/tracing/#interaction-with-beforesend-and-event-processors) I found out that the `before_send` callback shouldn't be called for the transactions. This PR fixes the behaviour, but it kinda of breaks the BC because users that were relying on this behaviour will not be able to do it anymore. Since this is a bug I consider the BC justified and I don't see any other way to avoid it anyway

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Unreleased
+
+- Fix unwanted call to the `before_send` callback with transaction events (#1158)
+
 ## 3.1.1 (2020-12-07)
 
 - Add support for PHP 8.0 (#1087)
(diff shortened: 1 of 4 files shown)


stayallive (approved):
Maybe you could point people to `traces_sampler` in case they were using `before_send` before?

But otherwise nice catch!

ste93cry:
>Maybe you could point people to traces_sampler in case they were using `before_send` before?

Do you mean in the CHANGELOG?

stayallive:
Yeah, no sense putting it in the UPGRADE, maybe something like:

```
- Fix unwanted call to the `before_send` callback with transaction events, use `traces_sampler` instead to filter transactions (#1158)
```

ste93cry:
Makes sense, done!

PR #1163, opened 2021-01-06
Failing test for extraneous path normalization

AndreKR:
This adds a (failing) test that demonstrates how normalization of the
prefixes makes it impossible to remove leading slashes from reported
stack frame paths.


Jean85:
This seems a good reproduction to me, but I'm a bit unsure about how this should work in relation to the server side of things.

@HazAT does this make sense to you? Are we misusing the option? I can't find it in the docs.

AndreKR:
Not sure what your question is, but FYI on the server the difference is visible in whether the file names start with a slash:

image

Jean85:
Sorry for taking so long, I've stumbled onto getsentry/sentry-symfony#431 that made me realize the real impact of this. I've tried to fix it but stacktraces paths are impacted too, and I'm not sure what happens on the UI with those `\0` chars...

Jean85:
Closing this as stale, and I'm still not sure this is the right way to fix this.
Please open an issue if this is still a problem.

PR #1164, opened 2021-01-06
Add Yii2 integration

rustamwin:


ste93cry (approved):
Thank you for the contribution and the effort of supporting Sentry on one more platform 😃

PR #1167, opened 2021-01-09
Deprecate the logger option

ste93cry:
As mentioned in #1165, while fixing a bug involving the `logger` option I found out that at some point in time both the JS and Python SDK stopped supporting it (or they never did, to be honest I don't have the full historic context to say it). Anyway, since the code of those libraries is mainly our reference and I don't think that having this option adds much value as there is usually one logger per application, I decided to deprecate the option and remove it in the next major version. The field with the same name in the event object is not affected.

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -270,9 +270,15 @@ final class Options
 
     /**
      * Gets the logger used by Sentry.
+     *
+     * @deprecated since version 3.2, to be removed in 4.0
      */
-    public function getLogger(): string
+    public function getLogger(/*bool $triggerDeprecation = true*/): string
     {
+        if (0 === \func_num_args() || (\func_num_args() > 0 && false !== func_get_arg(0))) {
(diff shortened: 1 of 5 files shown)


src/Options.php
```diff
@@ -270,19 +270,29 @@ public function setInAppIncludedPaths(array $paths): void
 
     /**
      * Gets the logger used by Sentry.
+     *
+     * @deprecated since version 3.2, to be removed in 4.0
      */
-    public function getLogger(): string
+    public function getLogger(/*bool $triggerDeprecation = true*/): string
     {
+        if (0 === \func_num_args() || (\func_num_args() > 0 && false !== func_get_arg(0))) {
```
Jean85, line 278: Double-checking the number of arguments is superflouous
```suggestion
        if (0 === \func_num_args() || false !== func_get_arg(0))) {
```

Jean85 (approved):
LGTM :+1: I would require a :+1: from someone like @HazAT too to be sure that removing this option doesn't have some strange side-effect or limitation server-side.

HazAT (approved):
~Kill~ Deprecate it with 🔥

stayallive (approved):
🔥 🚢

PR #1169, opened 2021-01-14
Fix default for client version

Jean85:
The usage of the `PrettyVersion` class there is wrong: the root package is the project using Sentry, not our lib.


ste93cry (approved):
Believe me, I noticed the same thing yesterday and I was wondering if it was wrong 😃

stayallive (approved):
Great to have the author jump in and correct that for us 😬😘😊

Jean85:
CI seems broken due to PHPStan update, unrelated :(

PR #1170, opened 2021-01-15
Allow PrettyVersion 2.0

Jean85:
This allows https://github.com/Jean85/pretty-package-versions/releases/tag/2.0.0, which I released with very few BC breaks, so that so we can use both versions without issues. In this PR I'm also able to drop the indirect dependency, so we can move forward toward Composer 2.

I'll probably backport this to the 2.x branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Allow jean85/pretty-package-versions 2.0 (#1170)
+
 ## 3.1.3 (2021-01-25)
 
 - Fix the fetching of the version of the SDK (#1169)
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,6 @@
         "guzzlehttp/promises": "^1.4",
         "guzzlehttp/psr7": "^1.7",
         "jean85/pretty-package-versions": "^1.5|^2.0",
-        "ocramius/package-versions": "^1.8",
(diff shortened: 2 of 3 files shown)


Jean85:
This breaks everything as discussed in https://github.com/getsentry/sentry-symfony/issues/383#issuecomment-760898738 :cry:

Jean85:
This should work with Composer 2.0.9 and PHPStan 0.12.70 as reported in https://github.com/getsentry/sentry-symfony/issues/383#issuecomment-768394725

Jean85:
Ok found the issue: the `ModulesIntegration` is requesting the version of a replaced package (`ocramius/package-version` in this case). I need to understand how to avoid doing that.

Jean85:
@ste93cry @stayallive this is now ready for review! I had to [release a patch](https://github.com/Jean85/pretty-package-versions/releases/tag/2.0.1) to solve this :smile:

stayallive (approved):
Not knowing enough to know if this is 100% correct and all but I assume you know best in this case 👍

ste93cry (changes requested):
Do we also need a CHANGELOG entry? Not sure about this, it looks like a change of internals that should not impact users :thinking:

Jean85:
I agree that a changelog entry is not needed, this is an internal thing only; the only effect on the outside is that it allows newer version of the dep, and removes an other one.

ste93cry (approved):

PR #1171, opened 2021-01-15
Allow jean85/pretty-package-versions 2.0 (backport of #1170)

Jean85:


Jean85:
@ste93cry @stayallive this is ready for review too!

stayallive (approved):
If you filter out all the `\` changes this isn't so big as I thought 😉

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Allow jean85/pretty-package-versions 2.0 (#?, backport of #1170 from 3.x)
```
ste93cry, line 5: Missing the reference to the PR, there is just an :hash:

src/Integration/ModulesIntegration.php
```diff
@@ -63,11 +65,33 @@ public static function applyToEvent(self $self, Event $event): void
     private function processEvent(Event $event): void
     {
         if (empty(self::$loadedModules)) {
-            foreach (Versions::VERSIONS as $package => $rawVersion) {
-                self::$loadedModules[$package] = PrettyVersions::getVersion($package)->getPrettyVersion();
+            foreach (self::getInstalledPackages() as $package) {
+                try {
+                    self::$loadedModules[$package] = PrettyVersions::getVersion($package)->getPrettyVersion();
+                } catch (\OutOfBoundsException | VersionMissingExceptionInterface $exception) {
```
ste93cry, line 71: Would it makes sense to catch any `Throwable`? Since we're inside an integration, if something fails it's really difficult to know what went wrong. In addition to that, regardless of what happens, if we don't report the version of a certain package it's not the end of the world

ste93cry (changes requested):

src/Integration/ModulesIntegration.php
Jean85: I agree. I also ported the same change into the original PR #1170 too with 43bee01.

ste93cry (approved):

PR #1172, opened 2021-01-15
Update Handler.php

wirwolf:
Remove final class. Can not extend Handler for overwriting some logic. For example recombine $record in write method


ste93cry:
What you cannot do by decorating the handler rather than extending it?

ste93cry:
I'm closing this issue because almost a month passed without answer, and probably the decorator pattern is a safer and quicker way for most of the customizations of the handler.

PR #1174, opened 2021-01-16
Deprecate the tags option

ste93cry:
As it was done for the `logger` option, also this one is going to be deprecated. Instead, the tags should be set either on the event object or on the scope. Both the Python and JS SDK already do not support this option anymore, although I'm not sure if they ever did it or they stopped doing it recently

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Deprecate the `logger` option (#1167)
 - Pass the event hint from the `capture*()` methods down to the `before_send` callback (#1138)
+- Deprecate the `tags` option (#1174)
 
 ## 3.1.2 (2021-01-08)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Deprecate the `logger` option (#1167)
 - Pass the event hint from the `capture*()` methods down to the `before_send` callback (#1138)
-- Deprecate the `tags` option (#1174)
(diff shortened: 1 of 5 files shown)


Jean85 (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Deprecate the `logger` option (#1167)
 - Pass the event hint from the `capture*()` methods down to the `before_send` callback (#1138)
+- Deprecate the `tags` option (#1174)
```
stayallive, line 7:
```suggestion
- Deprecate the `tags` option, see the [docs](https://docs.sentry.io/platforms/php/guides/laravel/enriching-events/tags/) for other ways to set tags (#1174)
```

Not sure this is the correct way but I like adding a little bit of “upgrade docs” to the changelog. Could also be mentioned in the 4.0 upgrade docs instead of here so people can prepare and use the new methods already.

stayallive (approved):
Changelog change would be nice but other than that 👍

CHANGELOG.md
ste93cry: >Could also be mentioned in the 4.0 upgrade docs

I don't know if it's would be correct for a few reaons:

- People that already upgraded will likely missing the new changes to that document
- The old way of setting the tags using the options works and it's fine until `5.0`, so at best this should be reported in a `UPGRADE-5.0.md` document

stayallive: 4 is the next release right? Not 5? Otherwise your deprecation exceptions are also wrong 😅

It should be mentioned in the upgrade docs for whatever version the tags option is actually removed, let's say it like that 😬

ste93cry: >4 is the next release right? Not 5?

Yeah sorry, my mistake typing the message

PR #1181, opened 2021-02-02
Add a GitHub Action workflow to release a new version

ste93cry:
As per title, this workflow comes straight from the repo of the Javascript SDK and should enable releasing a new version without depending on @HazAT

### DIFF
--- a/.github/workflows/publish-release.yaml
+++ b/.github/workflows/publish-release.yaml
@@ -1,0 +1,29 @@
+name: Prepare Release
+
+on:
+    workflow_dispatch:
+        inputs:
+            version:
+                description: Version to release
+                required: true
+            force:
+                description: Force a release even when there are release-blockers (optional)
+                required: false
+
+jobs:
+    release:
+        runs-on: ubuntu-latest
+        name: Release version
+        steps:
+            - uses: actions/checkout@v2
+              with:
+                  token: ${{ secrets.GH_RELEASE_PAT }}
+                  fetch-depth: 0
+
+            - name: Prepare release
+              uses: getsentry/action-prepare-release@v1
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GH_RELEASE_PAT }}
+              with:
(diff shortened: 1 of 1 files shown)


Jean85:
Woha, this seems great! How much stuff is done by that action? And how would it be triggered?

ste93cry:
>How much stuff is done by that action?

[This](https://github.com/getsentry/action-prepare-release) is the action, I had a quick look yesterday at the code and it seemed to me that it does everything using the Craft tool under the hood, so it should be able to automatically tag the branch and publish the release

>And how would it be triggered?

This is the only screenshot that @HazAT posted on Discord to show how it works. Basically, it should be sufficient to merge this PR and then when we want to release a new version go to the Actions tab and trigger the workflow manually by specifying the version to release and from which branch to start it

immagine

Jean85 (approved):
That's great! Can't wait to port it to `sentry-symfony` too!

ste93cry:
Mmm, I noticed in the action that it looks like it doesn't publish the version, but it rather open an issue that once merged unlock the release. @HazAT am I wrong?

https://github.com/getsentry/action-prepare-release/blob/d4e63a91f26f0ad81ab25f38ebd8f3f172128870/action.yml#L86-L88

HazAT:
@ste93cry That is correct, someone of the release managers at Sentry need to approve. But a simple ping is enough and this process is quicker and we have a track record.

stayallive (approved):
Looks like a great addition 👍

ste93cry:
>someone of the release managers at Sentry need to approve

Understood, I was hoping we could became independent in releasing but this is anyway an improvement over now. Do we have to keep pinging you on Discord or is there anyone else to bother?

HazAT:
In theory, there are notification but I can't keep up ^^
A ping in discord is simple, it's really not more than a click for me then (way better than before)

Jean85:
Oook so I'll port this to `sentry/sentry-symfony` too! :tada:

PR #1183, opened 2021-02-07
Fix incorrect detection of silenced errors (by the `@` operator)

stayallive:
As can be found here the `@` operator no longer silences _all_ error types starting with PHP 8: https://php.watch/versions/8.0/fatal-error-suppression.

Initially I thought the following was happening but that is not true: ~~So when capturing a silenced error the `error_reporting()` will not report `0` on PHP 8 but `4437` (which is `E_ERROR | E_PARSE | E_CORE_ERROR | E_COMPILE_ERROR | E_USER_ERROR | E_RECOVERABLE_ERROR`, check for yourself: https://maximivanov.github.io/php-error-reporting-calculator/).~~

When capturing a silenced error the `error_reporting()` will not report `0` on PHP 8 but `E_ERROR | E_PARSE | E_CORE_ERROR | E_COMPILE_ERROR | E_USER_ERROR | E_RECOVERABLE_ERROR` without the ones you excluded in an earlier `error_reporting()` call yourself (for example `error_reporting(E_ALL & ~E_USER_ERROR)` would result in `error_reporting()` being: `E_ERROR | E_PARSE | E_CORE_ERROR | E_COMPILE_ERROR | E_RECOVERABLE_ERROR` when handling a silenced error on PHP 8 (missing the `E_USER_ERROR`)).

The current behaviour makes it so that `error_reporting()` is leading in determining what is a "silenced error" which is simply not true under PHP 8 (or 7 for that matter).

Currently when you set `error_reporting(E_ALL & ~E_NOTICE)` and `'error_types' => E_ALL` it will currently report all `E_NOTICE` as silenced errors (thus not report them at all) which is incorrect and something I consider to be a bug and the incorrect behaviour. 

This PR aims to resolve the above.

There is 1 edge case in this "detection" of silenced errors. If the user set's their `error_reporting()` to 0 on PHP 7 and below or anything only containing the error types that cannot be silenced on PHP 8 and up we detect all errors as silenced, but I consider this acceptable since this was always the way it was before the broken code [was introduced](https://github.com/getsentry/sentry-php/pull/1087/files#diff-1b2bb2c6377ff609b2d252005cc6151091a6a904b51993bb607120b3e5dd62c2L268-R268). If we find this unacceptable we need to remove the silenced errors detection all together.


tests/phpt/error_handler_respects_error_types_option_regardless_of_error_reporting.phpt
```diff
@@ -0,0 +1,67 @@
...
+
+SentrySdk::getCurrentHub()->bindClient($client);
+
+echo 'Triggering E_USER_NOTICE error' . PHP_EOL;
+
+trigger_error('A E_USER_NOTICE which will be reported by Sentry!', E_USER_NOTICE);
+
+echo 'Triggering E_USER_WARNING error' . PHP_EOL;
+
+trigger_error('A E_USER_WARNING which won\'t be reported by Sentry!', E_USER_WARNING);
```
Jean85, line 62: Shouldn't we add more cases here with the `@` operator? Or do we have those cases in other tests already?

Jean85 (commented):
Thank you for picking this really nasty issue!

tests/phpt/error_handler_respects_error_types_option_regardless_of_error_reporting.phpt
stayallive: We have this one already that I did not break and found a good test, I added this to test the specific "bug" introduced and thought the other test case would be enough for the `@` alone.

https://github.com/getsentry/sentry-php/blob/master/tests/phpt/error_handler_respects_error_reporting.phpt

src/ErrorHandler.php
```diff
@@ -265,7 +265,13 @@ public function addExceptionHandlerListener(callable $listener): void
      */
     private function handleError(int $level, string $message, string $file, int $line, ?array $errcontext = []): bool
     {
-        if (0 === (error_reporting() & $level)) {
+        // The error reporting `error_reporting()` will have when an error is captured that was silenced with the `@` operator
+        // is different on PHP 8 versus earlier versions where it's 0. On PHP 8 it defaults to errors that are not silencable.
+        $expectedSilencedErrorReportingLevel = \PHP_MAJOR_VERSION >= 8
```
ste93cry, line 270: This value is not fixed, according to what I'm observing [here](https://3v4l.org/QGU0X), `error_reporting()` is returning the bitmask of all fatal error types  minus the ones that are excluded (`E_USER_ERROR` in that case)

ste93cry (changes requested):
I have to admit I'm having a really hard time understanding the issue without a real code example at disposal, but (correct me if I'm wrong) what you are saying is that we are reporting some errors as silenced when they are not, right?

src/ErrorHandler.php
stayallive: Jep I am seeing that too now, I'm trying to figure out why that didn't come up in testing.

But to drive the point home, this is still very much incorrect way to check if an error is silenced: https://3v4l.org/eQj7W.

I'm seeing if I can add a test for this and if I can resolve this point.

stayallive:
It looks like as @ste93cry mentioned I overlooked that the `error_reporting()` set when reporting on silenced errors is not fixed as I thought, so I created a workaround for that too (the joys of error handling code).

Also to show that the difference between what PHP 7 and 8 reports but with out current code they are both wrong I have this little proof of concept (adapted from @ste93cry): https://3v4l.org/nct1E (I would expect it to report the silent status correctly which is does not).

I have also tried to add more tests and better word what we I am doing and why I am dong it.

src/ErrorHandler.php
```diff
@@ -265,7 +270,24 @@ public function addExceptionHandlerListener(callable $listener): void
...
+            $errorWasSilencedByUser = 0 === (error_reporting() & ~self::PHP8_ERROR_TYPES_THAT_CANNOT_BE_SILENCED);
+
+            // However this will mean we will erroneously mark errors that cannot be silenced as silenced so if the error
+            // level is on the list of error types that cannot be silenced we make sure the error is not marked as silent
+            if ($level === (self::PHP8_ERROR_TYPES_THAT_CANNOT_BE_SILENCED & $level)) {
+                $errorWasSilencedByUser = false;
+            }
+        } else {
+            // On PHP 7 and below when an error is silenced with the `@` operator `error_reporting()` will always return `0`
+            $errorWasSilencedByUser = 0 === error_reporting();
```
Jean85, line 287: Why this is different from before? It was:
```suggestion
            $errorWasSilencedByUser = 0 === error_reporting() & $level;
```

tests/phpt/error_handler_captures_errors_not_silencable_on_php_8_and_up.phpt
```diff
@@ -0,0 +1,71 @@
...
+            {
+                return new FulfilledPromise(true);
+            }
+        };
+    }
+};
+
+error_reporting(E_ALL & ~E_USER_ERROR);
+
+$client = ClientBuilder::create(['error_types' => E_ALL])
```
Jean85, line 58: Shouldn't we ensure that silenced errors should not be reported in Sentry's options here?

Jean85 (changes requested):

src/ErrorHandler.php
stayallive: No. That is exactly were the bug came from :) 

You can see that that does not work: https://3v4l.org/nct1E

Neither on PHP 7 or 8, but on PHP 7 and below we can safely assume it's always a literal 0.

tests/phpt/error_handler_captures_errors_not_silencable_on_php_8_and_up.phpt
stayallive: You are probably right just in case the default changes in the future, I've explicitly added the option 👍

src/ErrorHandler.php
Jean85: Holy crap this is really messy! Thanks for solving it.

Jean85 (approved):

Jean85:
We should backport those tests in the 2.x branch and apply fixes if needed.

stayallive:
There is no need to backport because this was introduced in #1087 and that change only went in 3.x so we're good!

ste93cry (approved):
After talking on Discord with @stayallive, I'm gonna approve this PR to not block it further, however I totally disagree with what has been done. Taking as example the following code, what would happen is that since the `error_reporting()` is configured to not report the `E_USER_NOTICE` errors, the error generated by the undefined `$helloworld` variable will not be shown to the user, therefore acting as if it was silenced using the `@` operator

```php
error_reporting(E_ALL & ~E_NOTICE);

echo $helloworld;
```

What's happening right now with Sentry is that the error will be reported as a `SilencedErrorException`, which I agree it may be misleading in the strict terms as it hasn't been silenced manually but rather using `error_reporting()`. Obviously, apart from an incorrect labeling of the error type, the real issue is that users may lose the error if the don't have `capture_silenced_errors = true`. However, with this patch, what will happen is that the error will be reported as a `ErrorException`, leading users into thinking that this error **is** not suppressed at all and consequently has been displayed to their customers. According to the documentation and also to the behaviour that SDKs have had until now, this is expected. This is where I disagree: for me, the behaviour has always been wrong, we changed it starting from version `3.0` (as a consequence of another change, but it is so) and now we want to go back to that old behaviour.

What I would do instead is fixing the documentation, and in case investigate if it's possible to distinguish between errors silenced using the `@` operator and errors suppressed by the `error_reporting` setting in order to make the `capture_silenced_errors` option not affecting this last case. But even if that's not possible, I cannot think of an error that has not been shown as an error that is not suppressed :-1:

ste93cry:
FYI: this is a bugfix and as such should have been merged into the `master` branch, not the `develop` branch. I cherry-picked the patch into the `master` branch with 2c59c6c

PR #1191, opened 2021-03-03
Update list of 3rd party integrations

robertlemke:
Added a new section for integrations using the 2.x SDK and included an alternative package for Neos Flow (previously "Flow Framework") which supports Flow 7.x and Sentry SDK 3.x.


ste93cry (approved):
Thank you!

stayallive (approved):
Thanks 👍

PR #1196, opened 2021-03-11
Fix `error_reporting` value not evaluated on the moment of the error being reported by default

### LINKED ISSUE #1195: Suppressed PHP warning is send to Sentry
### Environment

**How do you use Sentry?**
Self hosted 21.2.0

**Which SDK and version?**
```
bobvandevijver@webdev:/var/www/dev$ composer show | grep sentry
sentry/sdk                              3.1.0     This is a metapackage shipping sentry/sentry with a recommended HTTP client.
sentry/sentry                           3.2.0     A PHP SDK for Sentry (http://sentry.io)
sentry/sentry-symfony                   4.0.3     Symfony integration for Sentry (http://getsentry.com)
```

We're using PHP 7.3.27 on Debian 10. The issue started after upgrading to v4 of the `sentry/sentry-symfony`. This upgrade included:

| Package | Old version | New version |
| -- | -- | -- |
| sentry/sdk | 2.2.0 | 3.1.0
| sentry/sentry | 2.5.2 | 3.2.0
| sentry/sentry-symfony | 3.5.3 | 4.0.3 |

### Steps to Reproduce

Symfony tries to silence an include warning during the HTTP Kernel initialization, as the container cache is allowed to not exist. If that is the case, the container variables is reset and the application should continue without any notification whatsoever. The error reporting is restored directly after the "critical" part in the finally handler.

See the [Kernel](https://github.com/symfony/symfony/blob/fc016ddd922c955ccf4ea342110c32a4da5c9df4/src/Symfony/Component/HttpKernel/Kernel.php#L443) class implementation of Symfony, line 443 for the error level adjustment, line 469 for the line that generates the warning and the finally block on 480-482 that restores the error reporting.

This means that #1183 and #1087 are probably related to this particular issue.

For this particular case the line marked as [broken code](https://github.com/getsentry/sentry-php/pull/1087/files#diff-1b2bb2c6377ff609b2d252005cc6151091a6a904b51993bb607120b3e5dd62c2L268-R268) in #1183 was actually doing its job:

image

So, a small reproduction would be:

```php
$errorLevel = error_reporting(\E_ALL ^ \E_WARNING);
try {
  include 'non-existing-file.php';
} finally {
  error_reporting($errorLevel);
}
```

### Expected Result

The error should not be logged into Sentry, as it is suppressed using an explicit `error_reporting` call.

### Actual Result

But is isn't...

image

stayallive:
This PR aims to fix a bug where `error_reporting()` is copied on initialisation of the SDK and not evaluated on the moment of the error capture.

There are 2 changed to the behaviour of the SDK after this bug fix:

- When not explicitly setting the `error_types` option on the SDK the `error_reporting()` value is used to determine if an error should be reported to Sentry
- Errors silenced by the `@` operator are reported regardless of their level, only taking `capture_silenced_errors` (defaults to `false`) into consideration

The last point is a side effect (IMO perfectly acceptable) because we cannot know the actual error reporting level while handling silenced (by the `@` operator) errors because PHP temporarily modifies the `error_reporting()` level while handling such errors.

Fixes #1195.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Fix `error_reporting()` not being evaluated when handling errors when the `error_types` option is not set (#1196)
+- Changes behaviour of `error_types` option when not set: before it defaulted to `error_reporting()` statically at SDK initialization; now it will be evaluated each time during error handling to allow silencing errors temporarily (#1196)
 
 ## 3.2.0 (2021-03-03)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix `error_reporting()` not being evaluated when handling errors when the `error_types` option is not set (#1196)
+
 ## 3.2.0 (2021-03-03)
 
 - Make the HTTP headers sanitizable in the `RequestIntegration` integration instead of removing them entirely (#1161)
--- a/tests/OptionsTest.php
+++ b/tests/OptionsTest.php
@@ -539,23 +539,6 @@ final class OptionsTest extends TestCase
         $this->assertSame('0.0.1', (new Options())->getRelease());
     }
 
-    public function testErrorTypesIsDynamiclyReadFromErrorReportingLevelWhenUnset(): void
-    {
-        $options = new Options(['error_types' => null]);
-
-        $errorReportingBeforeTest = error_reporting($currentErrorReporting = \E_NOTICE & \E_USER_NOTICE);
-
-        $this->assertEquals($currentErrorReporting, $options->getErrorTypes());
-
-        error_reporting($currentErrorReporting = \E_WARNING & \E_USER_WARNING);
-
-        $this->assertEquals($currentErrorReporting, $options->getErrorTypes());
-
-        error_reporting($errorReportingBeforeTest);
-
-        $this->assertEquals($errorReportingBeforeTest, $options->getErrorTypes());
-    }
-
     public function testErrorTypesIsNotDynamiclyReadFromErrorReportingLevelWhenSet(): void
     {
         $errorReportingBeforeTest = error_reporting(\E_ERROR);
--- a/tests/OptionsTest.php
+++ b/tests/OptionsTest.php
@@ -539,16 +539,18 @@ final class OptionsTest extends TestCase
         $this->assertSame('0.0.1', (new Options())->getRelease());
     }
 
-    public function testErrorTypesIsNotDynamiclyReadFromErrorReportingLevelWhenSet(): void
+    public function testErrorTypesOptionIsNotDynamiclyReadFromErrorReportingLevelWhenSet(): void
     {
         $errorReportingBeforeTest = error_reporting(\E_ERROR);
 
-        $options = new Options(['error_types' => $currentErrorTypes = \E_NOTICE & \E_USER_NOTICE]);
+        $errorTypesOptionValue = \E_NOTICE;
 
-        $this->assertEquals($currentErrorTypes, $options->getErrorTypes());
+        $options = new Options(['error_types' => $errorTypesOptionValue]);
+
+        $this->assertEquals($errorTypesOptionValue, $options->getErrorTypes());
 
         error_reporting($errorReportingBeforeTest);
 
-        $this->assertEquals($currentErrorTypes, $options->getErrorTypes());
+        $this->assertEquals($errorTypesOptionValue, $options->getErrorTypes());
     }
 }
--- a/tests/OptionsTest.php
+++ b/tests/OptionsTest.php
@@ -519,4 +519,34 @@ final class OptionsTest extends TestCase
 
         $this->assertSame('0.0.1', (new Options())->getRelease());
     }
+
+    public function testErrorTypesIsDynamiclyReadingErrorReportingLevelWhenUnset(): void
+    {
+        $options = new Options(['error_types' => null]);
+
+        $errorReportingBeforeTest = error_reporting($currentErrorReporting = E_NOTICE & E_USER_NOTICE);
+
+        $this->assertEquals($currentErrorReporting, $options->getErrorTypes());
+
+        error_reporting($currentErrorReporting = E_WARNING & E_USER_WARNING);
+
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix `error_reporting()` not being evaluated when handling errors when the `error_types` option is not set (#1196)
```
Jean85, line 5: I think we should clarify this better, with something like this:
```suggestion
- Changes behaviour of `error_types` option when not set: before it defaulted to `error_reporting()` statically at SDK initialization; now it will be evaluated each time during error handling to allow silencing error temporary dynamically (#1196)
```

tests/OptionsTest.php
```diff
@@ -519,4 +519,34 @@ public function testReleaseOptionDefaultValueIsGotFromEnvironmentVariable(): voi
 
         $this->assertSame('0.0.1', (new Options())->getRelease());
     }
+
+    public function testErrorTypesIsDynamiclyReadingErrorReportingLevelWhenUnset(): void
```
Jean85, line 523:
```suggestion
    public function testErrorTypesIsDynamiclyReadFromErrorReportingLevelWhenUnset(): void
```

tests/OptionsTest.php
```diff
@@ -519,4 +519,34 @@ public function testReleaseOptionDefaultValueIsGotFromEnvironmentVariable(): voi
...
+    {
+        $options = new Options(['error_types' => null]);
+
+        $errorReportingBeforeTest = error_reporting($currentErrorReporting = \E_NOTICE & \E_USER_NOTICE);
```
Jean85, line 527: Since we're messing with `error_reporting`, we should probably add a restore mechanism in `setup` and `teardown` of this test class.

tests/phpt/error_handler_respects_current_error_reporting_level.phpt
```diff
@@ -0,0 +1,79 @@
...
+
+$vendor = __DIR__;
+
+while (!file_exists($vendor . '/vendor')) {
+    $vendor = \dirname($vendor);
+}
+
+require $vendor . '/vendor/autoload.php';
+
+$transportFactory = new class implements TransportFactoryInterface {
```
Jean85, line 32: Wouldn't a `before_send` callback obtain the same objective and be less verbose?

CHANGELOG.md
stayallive: Sounds pretty good 👍

tests/phpt/error_handler_respects_current_error_reporting_level.phpt
stayallive: Yeah sure. I just copied this from a few other tests, but I'll change it 👍

tests/OptionsTest.php
```diff
@@ -14,6 +14,27 @@ final class OptionsTest extends TestCase
...
+        // There are tests changing the error reporting level in this class so we store the current value before running the tests
+        $this->errorReportingOnSetup = error_reporting();
+    }
+
+    protected function tearDown(): void
+    {
+        parent::tearDown();
+
+        // There are tests changing the error reporting level in this class so we restore it to the value before running the tests
+        error_reporting($this->errorReportingOnSetup);
```
Jean85, line 35: Order of calls should be reversed, we need to intervene before other teardown operations are triggered.

Also, comments are probably superfluous here, it's pretty clear what we're doing.

Jean85 (changes requested):

tests/OptionsTest.php
stayallive: Ah yeah makes sense, didn't think about that, switched it around!

Jean85 (approved):

tests/OptionsTest.php
```diff
@@ -519,4 +538,34 @@ public function testReleaseOptionDefaultValueIsGotFromEnvironmentVariable(): voi
 
         $this->assertSame('0.0.1', (new Options())->getRelease());
     }
+
+    public function testErrorTypesIsDynamiclyReadFromErrorReportingLevelWhenUnset(): void
```
ste93cry, line 542: I don't understand the usefulness of neither this nor the next test. As far as I understand from the changes, it should be enough to test just these cases (using a data provider):

- the getter returns `E_*` when `error_types = E_*`
- the getter returns the current value of `error_reporting()` when `error_types = null`

All the code to test that when changing `error_reporting()` the value obtained from the getter is the right one is made superfluous by the first test case above, isn't it?

ste93cry (changes requested):

tests/OptionsTest.php
Jean85: The difference is that the value is not statically saved anywhere, but it's fetched from `error_reporting()` each time.

stayallive: Yeah that was the intention of the 2 tests to validate that in the first when `error_types` is `null` it changes depending on the current value of `error_reporting` and in the second it doesn't matter what `error_reporting()` says it's always the same.

I thought it be easier to read like this too without a data provider since we need to modify the `error_reporting()` to make sure we see the "live" value in the first case.

But feel free to refactor it to something else if that makes more sense.

ste93cry: >The difference is that the value is not statically saved anywhere, but it's fetched from error_reporting() each time.

But this is asserted in the second case I mentioned. Let's try to rewrite the following code to analyze the possible outputs:

https://github.com/getsentry/sentry-php/blob/41bf1efb217f1fa98658d59104505b28ea63e314/src/Options.php#L414

It becomes:

```php
if ($this->options['error_types'] !== null) {
    return $this->options['error_types'];
}

return error_reporting();
```

There are just two exit points from the function, and they are both covered with the test cases mentioned above. There is no point in checking here that if you set a different value for the `error_reporting` then the new value is reported because, well, there is no other way it could not 😃

ste93cry: >It does if you consider the previous behavior right?

Yes, I think that what you say makes sense. But in turn this is tested in the PHPT which is a real integration test, isn't it?

ste93cry (approved):

bobvandevijver:
@ste93cry Do you have any prediction on when this will be tagged and released?

ste93cry:
Released right now as `3.2.1`, sorry for the delay

PR #1198, opened 2021-03-12
Remove final for State/Scope

4n70w4:
Hi! I make error handler for different logic for different environments and other conditions.
While writing unit tests, I ran into the problem that I cannot make a mock for the State/Scope because it is final.
I propose to remove the final.

If you are against, then please add an interface that can be mocked.

### DIFF
--- a/src/State/Scope.php
+++ b/src/State/Scope.php
@@ -16,7 +16,7 @@ use Sentry\UserDataBag;
  * The scope holds data that should implicitly be sent with Sentry events. It
  * can hold context data, extra parameters, level overrides, fingerprints etc.
  */
-final class Scope
+class Scope
 {
     /**
      * @var Breadcrumb[] The list of breadcrumbs recorded in this scope


Jean85:
I'm :-1: on this; `Scope` is basically a data object with no behaviour, you shouldn't mock it.

ste93cry:
I tend to agree, we are usually a lot conservative on removing `final` from classes because there is usually a better extension point and/or a better way to solve the issue (e.g. decorator pattern). Now, in the case of the `Scope` class, the missing interface is indeed a blocker to either mock, replace or decorate it. There is [this](https://ocramius.github.io/blog/when-to-declare-classes-final/) great article explaining in details the pros/cons of `final` and when it should be used. At the same time, the rules declared in that article to me are more of a guideline that something written in the stone, so I think that there are exceptions to consider like in everything. In this particular case, I would reason on two things:

- Would it make sense to add a `ScopeInterface`? If the answer is yes, why and what are the benefits of having it?
- Why you need to mock the `Scope` class? We don't have any test that does it, and they still works well. I also don't think that the business logic of this class is something that should be changed, there are already some extensions points (e.g. the `before_send`) callback and the event processors and this until now has always worked well

4n70w4:
Example. I have legacy code. I don’t know what it is and why, but there is no task of changing or deleting it.
I want to cover it with tests to avoid regression in the future.

```php
        $sentry->withScope(static function(Scope $scope) use ($error, &$eventId, $sentry) {
            $scope->setExtra('details', FormattedError::createFromException($error, Debug::INCLUDE_DEBUG_MESSAGE));
            $scope->setExtra('clientSafe', $error->isClientSafe());

            try {
                $eventId = $sentry->captureException($error);
                Log::error("Sentry Id: {$eventId}. {$error->getMessage()}");
            } catch(Exception $e) {
                Log::critical($error->getMessage() );
            }

        });
```

I make test:

```php
...
        $scope = Mockery::mock(Scope::class);
        $sentry->shouldReceive('setExtra')->with('details', 'test')->once()->andReturn();
...
```

I have no other idea how to check this behavior in tests.

The absence of final or the presence of an interface would easily have solved the situation. Or the absence of this class.

ste93cry:
While I don't know `Mockery`, assuming that `$sentry` is an instance of `HubInterface`, here's an example (not meant to be fully working) of how you can do the same test without mocking the `Scope` class:

```php
$expectedScope = new Scope();
$expectedScope->setExtra('details', ...);

$sentry = Mockery::mock(HubInterface::class);
$sentry->shouldReceive('withScope')->once()->andReturnUsing(function (Scope $scope) use ($expectedScope): void {
    $this->assertEquals($expectedScopem $scope);
});
```

Another alternative, similar to the one above, is to call `$scope->applyToEvent()` from inside the closure and then assert that the data of the scope has been set on the event object. We use both ways in our own tests 😉

4n70w4:
@ste93cry thanks for the examples, I will try them soon.

4n70w4:
None of the options came up. The ability to mock `Scope` is simple and obvious.

ste93cry:
I'm sorry, I forgot the most important part of the example, you should of course call your own `withScope` callback 😅 I updated the example, please give it another try

```php
class Foo
{
    public function testMethod(): void
    {
        $sentry = \Sentry\SentrySdk::getCurrentHub();
        $sentry->withScope(static function(\Sentry\State\Scope $scope): void {
            $scope->setExtra('details', 'foo');
        });
    }
}

class FooTest extends \PHPUnit\Framework\TestCase
{
    use \Mockery\Adapter\Phpunit\MockeryPHPUnitIntegration;

    public function testTestMethodWithMockedWithScopeFunctionAndAssretingScopeEquals(): void
    {
        $expectedScope = new \Sentry\State\Scope();
        $expectedScope->setExtra('details', 'foo');

        $foo = new Foo();
        $hub = \Mockery::mock(\Sentry\State\HubInterface::class);
        $hub->shouldReceive('withScope')->once()->andReturnUsing(function (callable $callback) use ($expectedScope): void {
            $scope = new Scope();

            $callback($scope);

            $this->assertEquals($expectedScope, $scope);
        });

        SentrySdk::setCurrentHub($hub);

        $foo->testMethod();
    }

    public function testTestMethodWithMockedWithScopeFunctionAndAssertingEventData(): void
    {
        $foo = new Foo();
        $hub = \Mockery::mock(\Sentry\State\HubInterface::class);
        $hub->shouldReceive('withScope')->once()->andReturnUsing(function (callable $callback): void {
            $scope = new Scope();

            $callback($scope);

            $event = $scope->applyToEvent(Event::createEvent());

            $this->assertSame(['details' => 'foo'], $event->getExtra());
        });

        SentrySdk::setCurrentHub($hub);

        $foo->testMethod();
    }
}
```

4n70w4:
Another case. I want to make a `partial mock` sentry (`Hub::class`), in this case the interface will not help and I need a class without a final.

ste93cry:
All these classes are not meant to be extended, you should not mock them but rather use them, and there is usually little to no logic in them. I don't know why you cannot mock the `HubInterface` interface and you want instead to mock the implementation, but since in that case an interface is available there is no reason that I can think of to do it except if you typehinted against the class, which is wrong

Jean85:
I would also advise against partial mocks: they break the encapsulation of objects, hence risking unpredictable behaviour.

ste93cry:
Almost a month has passed without further interaction, so I believe that you solved the issue one way or another. I'm gonna close this issue, but feel free to reopen should you have more questions

PR #1199, opened 2021-03-17
Add breadcrumb monolog handler

Jean85:
This is an interesting additional feature, that would be a required stepping stone for me to proceed with the strategy outlined in https://github.com/getsentry/sentry-symfony/issues/337#issuecomment-800920045.

PR #1200, opened 2021-03-17
Try to fix the CI build for PHP 8 + Windows

Jean85:
We have a PHPT test that is randomly failing under PHP 8 + Windows due to memory constraints:
 * https://github.com/getsentry/sentry-php/pull/1199/checks?check_run_id=2129595114#step:9:43
 * https://github.com/getsentry/sentry-php/runs/2095230888?check_suite_focus=true#step:9:42
 * https://github.com/getsentry/sentry-php/runs/2085509080?check_suite_focus=true#step:9:43
 * https://github.com/getsentry/sentry-php/runs/2084796288?check_suite_focus=true#step:9:43

Since we can alter that memory constraint once the error is triggered, this should fix the test for good.

### DIFF
--- a/tests/phpt/error_handler_captures_out_of_memory_fatal_error.phpt
+++ b/tests/phpt/error_handler_captures_out_of_memory_fatal_error.phpt
@@ -1,7 +1,5 @@
 --TEST--
 Test catching out of memory fatal error
---INI--
-memory_limit=64M
 --FILE--
 <?php
 
@@ -34,6 +32,7 @@ $errorHandler->addExceptionHandlerListener(static function (): void {
     echo 'Exception listener called (it should not have been)' . PHP_EOL;
 });
 
+ini_set('memory_limit', '64M');
 $foo = str_repeat('x', 1024 * 1024 * 70);
 ?>
 --EXPECTF--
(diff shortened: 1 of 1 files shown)


stayallive:
If only it worked 😅 

Another idea could be to increase the `$reservedMemorySize` for the fatal error handler?

Jean85:
Nope, not even reserving memory works :cry:

ste93cry:
May it be because of the `memory_limit` cap in the `--INI--` setting of the PHPT?

Jean85:
Nope, it doesn't change anything... I'll try to dig the installed dependencies and see which got updated and could be the probable culprit.

ste93cry:
>Nope, it doesn't change anything...

I didn't mean to set the limit dynamically, but rather I was suggesting if the issue could be that it's too low after some dependency update. The trial I would have done would have been to increase the limit of `memory_limit`

Jean85:
It worked! :tada:

stayallive (approved):
Re-triggered the test suite twice and it still works without fail 👌

ste93cry (approved):

PR #1214, opened 2021-04-20
Fatal Error Listener: Allow configuring of error limit

jarstelfox:
From the looks of the code, this was meant to be configurable, but
was never lifted up into a user option. We ran into a case where we needed
to bump the reserved memory (currently doing so with a custom integration).

This change does not affect the interface of registerOnceFatalErrorHandler
thus is backwards compatible.

I believe the code can be a bit cleaner if we can change the param to be ?int

### DIFF
--- a/src/Integration/FatalErrorListenerIntegration.php
+++ b/src/Integration/FatalErrorListenerIntegration.php
@@ -22,7 +22,7 @@ final class FatalErrorListenerIntegration extends AbstractErrorListenerIntegrati
     {
         $client = SentrySdk::getCurrentHub()->getClient();
         $reservedMemory = $client
-         ? $client->getOptions()->getFatalReservedMemory()
+         ? $client->getOptions()->getFatalErrorHandlerReservedMemoryAmount()
          : ErrorHandler::DEFAULT_RESERVED_MEMORY_SIZE;
         $errorHandler = ErrorHandler::registerOnceFatalErrorHandler($reservedMemory);
         $errorHandler->addFatalErrorHandlerListener(static function (FatalErrorException $exception): void {
--- a/src/Integration/FatalErrorListenerIntegration.php
+++ b/src/Integration/FatalErrorListenerIntegration.php
@@ -20,7 +20,11 @@ final class FatalErrorListenerIntegration extends AbstractErrorListenerIntegrati
      */
     public function setupOnce(): void
     {
-        $errorHandler = ErrorHandler::registerOnceFatalErrorHandler();
+        $client = SentrySdk::getCurrentHub()->getClient();
+        $reservedMemory = $client
+         ? $client->getOptions()->getFatalReservedMemory() ?? ErrorHandler::DEFAULT_RESERVED_MEMORY_SIZE
+         : ErrorHandler::DEFAULT_RESERVED_MEMORY_SIZE;
+        $errorHandler = ErrorHandler::registerOnceFatalErrorHandler($reservedMemory);
         $errorHandler->addFatalErrorHandlerListener(static function (FatalErrorException $exception): void {
             $currentHub = SentrySdk::getCurrentHub();
             $integration = $currentHub->getIntegration(self::class);
--- a/src/Integration/FatalErrorListenerIntegration.php
+++ b/src/Integration/FatalErrorListenerIntegration.php
@@ -22,7 +22,7 @@ final class FatalErrorListenerIntegration extends AbstractErrorListenerIntegrati
     {
         $client = SentrySdk::getCurrentHub()->getClient();
         $reservedMemory = $client
-         ? $client->getOptions()->getFatalReservedMemory() ?? ErrorHandler::DEFAULT_RESERVED_MEMORY_SIZE
+         ? $client->getOptions()->getFatalReservedMemory()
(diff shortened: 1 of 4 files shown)


src/Options.php
```diff
@@ -686,6 +686,24 @@ public function setTracesSampler(?callable $sampler): void
         $this->options = $this->resolver->resolve($options);
     }
 
+    /**
+     * Gets the amount of memory to reserve in case of a fatal error.
+     * The number is in bytes. Null indicated the default should be used.
+     */
+    public function getFatalReservedMemory(): ?int
```
ste93cry, line 693: I suggest to rename this method to `getFatalErrorHandlerReservedMemoryAmount()`. Also, why not returning `ErrorHandler::DEFAULT_RESERVED_MEMORY_SIZE` instead of null to indicate the default value?

src/Options.php
```diff
@@ -686,6 +686,24 @@ public function setTracesSampler(?callable $sampler): void
...
+    {
+        return $this->options['fatal_error_reserved_memory'];
```
ste93cry, line 695: Doesn't makes more sense to call this option `fatal_error_handler_reserved_memory_amount`? I don't like it being so long, but the memory is clearly affecting the error handler

ste93cry (changes requested):
Can you please rebase on top of the `develop` branch and add a changelog entry with the reference to this PR?

src/Options.php
jarstelfox: Absolutely. I wasn't sure about the long name as well. I don't have a better name. Change incoming.

src/Options.php
jarstelfox: Fair point. Change incoming.

src/Options.php
```diff
@@ -729,6 +747,7 @@ private function configureOptions(OptionsResolver $resolver): void
             'capture_silenced_errors' => false,
             'max_request_body_size' => 'medium',
             'class_serializers' => [],
+            'fatal_error_handler_reserved_memory_amount' => null,
```
ste93cry, line 750: We can set this value to `ErrorHandler::DEFAULT_RESERVED_MEMORY_SIZE`. This way, we can revert the change to the allowed types and in the end we can remove the fallback to the constant's value in the getter

src/Integration/FatalErrorListenerIntegration.php
```diff
@@ -20,7 +20,11 @@ final class FatalErrorListenerIntegration extends AbstractErrorListenerIntegrati
      */
     public function setupOnce(): void
     {
-        $errorHandler = ErrorHandler::registerOnceFatalErrorHandler();
+        $client = SentrySdk::getCurrentHub()->getClient();
+        $reservedMemory = $client
+         ? $client->getOptions()->getFatalErrorHandlerReservedMemoryAmount()
+         : ErrorHandler::DEFAULT_RESERVED_MEMORY_SIZE;
+        $errorHandler = ErrorHandler::registerOnceFatalErrorHandler($reservedMemory);
```
ste93cry, line 27:
```suggestion

        $errorHandler = ErrorHandler::registerOnceFatalErrorHandler($reservedMemory);
```

ste93cry (changes requested):
The update of the CHANGELOG is still missing. In the meanwhile, I just realized why this option was not exposed as a user option: it's basically unmodifiable after it has been set: changing it will have no effect on the error handler, so probably it doesn't make sense to have a setter for it. Also, since the handler is shared among all SDK clients, the first client that gets created will actually define the amount of memory reserved for the handler. For these reasons, I had left out the option and instead opted to make users register the listener by themselves like you do

jarstelfox:
>For these reasons, I had left out the option and instead opted to make users register the listener by themselves like you do

Ah, that's totally reasonable. Do you think it is still a good idea to merge this pull? I could go either way. From our perspective, we would rather not have any custom listeners to reduce coupling with possible future changes.

ste93cry:
>Do you think it is still a good idea to merge this pull?

Yes, I can see the usefulness of what you are asking to support, even though wanting to change this option is not so common I believe. Unfortunately, while reviewing the code I missed the fact that the `setupOnce()` method of the integrations is called from the constructor of the client, but before the client is bound to the hub: in fact, the real code of the integration is itself called only when an event is being captured and its execution is delayed until the very last moment thanks to the use of the lambda. What this means is that when you call `SentrySdk::getCurrentHub()->getClient()` from inside the `setupOnce()` method but outside the event processor, the client is not bound yet and the method will always return `null`. The only workaround I can think of is to not have this option as a configurable option of the client but rather as a configurable option of the integration: this will still require a bunch of custom code to override the instance of the integration that is normally set by the client builder, but it should work

ste93cry:
May I ask if you plan to finish the work or if this PR is abandoned?

jarstelfox:
I have abandoned this PR for 3 reasons:

* I may have figured out the bloat in our app, thus not requiring the change in defaults
* I've been busy (so say we all)
* The change seems invasive considering that I may not even need it for our own use case

ste93cry:
Ok. Although I really think the change is genuinely something that I would find useful, I will close this PR. If in the future more users will ask for it we can always put it back on the table

PR #1221, opened 2021-06-05
Add user capturing for Monolog handler

andriichuk:
Hi! It would be great to have user identification feature for Monolog handler.

https://docs.sentry.io/platforms/php/enriching-events/identify-user/


ste93cry:
Thank you for the contribution. Unfortunately, we declined in the past other PR similar to this one because we don't want to support anymore extracting unstructured data from Monolog. You can find a lenghty discussion in getsentry/sentry-php#848 about the reasons of this choice. I suggest you to either implement your own handler or decorate our and call `withScope()` yourself

PR #1223, opened 2021-06-17
Fix missing frames's arguments when calling `captureEvent()` without explicit stacktrace or exception

### LINKED ISSUE #1220: Local variables are not showing on stack trace
Which SDK and version?
PHP SDK v3.2.2
PHP v7.2.34

A user is reporting an issue with the local variables passed as function arguments are not present in their stack traces in Sentry.

I was unable to reproduce the issue. When testing the feature, the function arguments are present in my stack trace, as you can see in the screenshot. The difference in my case is that I'm using PHP v8.0.3.

Screenshot 2021-05-27 at 14 46 07

Digging the code, I found this function that captures the stack trace and add it to the event that uses the constant `DEBUG_BACKTRACE_IGNORE_ARGS`, which according to PHP docs control if the function arguments should are omitted or not

https://github.com/getsentry/sentry-php/blob/4f6f8fa701e5db53c04f471b139e7d4f85831f17/src/Client.php#L289-L305

https://www.php.net/manual/en/function.debug-backtrace.php

Is this issue indeed related to the constant and it is configured locally by the user? Also, does the PHP version have an impact on this behaviour?

ste93cry:
Fixes #1220 by removing the `DEBUG_BACKTRACE_IGNORE_ARGS` constant from the call to the `debug_backtrace()` function. I think I've used it to lower the memory usage, however it makes sense to collect these information.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix missing collecting of frames's arguments when using `captureEvent()` without expliciting a stacktrace or an exception (#1223)
+
 ## 3.3.0 (2021-05-26)
 
 - Allow setting a custom timestamp on the breadcrumbs (#1193)
--- a/src/Client.php
+++ b/src/Client.php
@@ -298,7 +298,7 @@ final class Client implements ClientInterface
         }
 
         $event->setStacktrace($this->stacktraceBuilder->buildFromBacktrace(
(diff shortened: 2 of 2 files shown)


Jean85 (approved):
LGTM :+1: so this would impact only manual capturing or `captureMessage` with `attach_stacktrace === true`?

ste93cry:
Exactly 😉

Issue #1179, opened 2021-01-27
Monolog handler does not report $record['context'] other than $record['context']['exception']


bits4breakfast:
👋  - I've just (finally) upgraded the SDK to the latest version straight from the legacy one (1.8) and to my surprise the Monolog handler does not copy anything from the `$record` context other than the exception - everything else is discarded. I don't know if this is a known issue/conscious choice but I couldn't find any context (no pun intended) around that. I can eventually make changes myself however I wanted to make sure this was not something removed in the past.

Cheers

Jean85:
Yes this was intentional. See discussions in #848 for explanations.

Issue #1145, opened 2020-12-01
Missing PSR context properties


mrtus:
In older versions the PSR context properties were always available on events in the "extra" bag.
In the latest versions this is no longer the case.

As per [PSR-3 1.2](https://www.php-fig.org/psr/psr-3/#12-message): `Placeholder names MUST correspond to keys in the context array.`

I would make a PR to add this ability back to the `Sentry\Monolog\Handler` by using `$record['context']` and put those values in the "extra" bag.

```php
$this->hub->withScope(function (Scope $scope) use ($record, $event, $hint): void {
  $scope->setExtra('monolog.channel', $record['channel']);
  $scope->setExtra('monolog.level', $record['level_name']);

  foreach($record['context'] as $contextKey => $contextValue) {
    // TODO what about the exception itself? Log the class or completely ignore?
    $scope->setExtra($contextKey, $contextValue);
  }

  $this->hub->captureEvent($event, $hint);
});
```

What about the exception itself? Log the class or completely ignore as it is already processed through
```php
if (isset($record['context']['exception']) && $record['context']['exception'] instanceof \Throwable) {
  $hint->exception = $record['context']['exception'];
}
```

Would this contribution be accepted or is this ability available and am I simply missing something?

I'm looking forward to your response.

ste93cry:
It looks like we aren't able to avoid this issue popping up from time to time :disappointed: Please take a look at #848 (one of the many similar issues) to understand the reasoning about why we won't revert the decision of not blindly adding the context data of Monolog to the Sentry extras

mrtus:
> It looks like we aren't able to avoid this issue popping up from time to time 😞 Please take a look at #848 (one of the many similar issues) to understand the reasoning about why we won't revert the decision of not blindly adding the context data of Monolog to the Sentry extras

I'll have to respectfully disagree and while I do understand the reasoning behind this decision, this issue, I reckon, will keep popping up, it will probably for the reason as mentioned in the description.

> As per PSR-3 1.2: Placeholder names MUST correspond to keys in the context array.

While these are indeed recommendations they are there to allow a better design of frameworks, allow plug-and-play between different implementations, easy adaptability by other developers and frameworks, and many more. These PSR were thoughtfully compiled into a set of rules.

The [PSR-3 1.3](https://www.php-fig.org/psr/psr-3/#13-context) also states that:
[...]

Issue #1014, opened 2020-05-07
Additional data with Monolog


enumag:
Some of my calls to Monolog contain additional data in the `$context` array. I need to get some of these data into the `\Sentry\Event` object. I want to use an event processor for that since it gets both the Event instance and the original `$payload` array (see #1010).

The issue is that the data I need don't make it into the `$payload` array at the moment. The Monolog Handler only copies the exception but the rest of `$record['context']` is thrown away:

https://github.com/getsentry/sentry-php/blob/97bbca2c828651dfbd04769c94c1cc8b37e096a3/src/Monolog/Handler.php#L53-L55

Do I have to write my own Monolog Handler for this or can we improve the existing one in this library to pass the data into the payload?

enumag:
Related to https://github.com/amphp/http-server/pull/310.

Jean85:
Changing the handler was previously shot down due (and explained) in #844 & #848; TL;DR, we cannot assume what's inside Monolog context, so extracting info bu default could lead to unwanted stuff.

There's also https://github.com/B-Galati/monolog-sentry-handler that could help you in getting more information from Monolog.

Our handler is final, but you can surely use nested handler to avoid rewriting it.

enumag:
> Our handler is final, but you can surely use nested handler to avoid rewriting it.

I don't see how in this case.

Jean85:
In the same way that special handler do, decorating the original handler: https://seldaek.github.io/monolog/doc/02-handlers-formatters-processors.html#wrappers--special-handlers

enumag:
I managed to get the data through the existing Handler by implementing `Monolog\Processor\ProcessorInterface`:

```php
namespace App\Library\Monolog;

use Monolog\Processor\ProcessorInterface;

final class RequestExtraProcessor implements ProcessorInterface
{
	public function __invoke(array $record): array
	{
		if (array_key_exists('request', $record['context'])) {
			$record['context']['extra']['request'] = $record['context']['request'];
		}

		return $record;
	}
}

nnatter:
For completion; As far as I understand, the solution of registering a monolog processor described in https://github.com/getsentry/sentry-php/issues/1014#issuecomment-626673372 **does not work with version 4 of this bundle**. 

If you want to do this with version 4 of this bundle, you need to decorate the monolog handler (or register your own handler) like described in https://github.com/getsentry/sentry-php/pull/848#issuecomment-605322356.

PR #1226, opened 2021-07-09
Set scope context based on record

gabrielkoerich:
This PR sets the context on the log handler, getting it from the current record.

We have some warning that uses context and it's better to understand what was going on.

Here's how it looks when passing big contexts:
Screen Shot 2021-07-09 at 16 22 05

### DIFF
--- a/src/Monolog/Handler.php
+++ b/src/Monolog/Handler.php
@@ -60,6 +60,7 @@ final class Handler extends AbstractProcessingHandler
         $this->hub->withScope(function (Scope $scope) use ($record, $event, $hint): void {
             $scope->setExtra('monolog.channel', $record['channel']);
             $scope->setExtra('monolog.level', $record['level_name']);
+            $scope->setContext('context', $record['context']);
 
             $this->hub->captureEvent($event, $hint);
         });


iansltx:
This lets Sentry pick up data passed in the context param on PSR-3 log calls (third param for `log()`, second for `info()` etc.); if this needs to be put behind a config flag, reply back, since this _does_ transmit more info than before (though since folks are passing context in I think they'd expect the info to be passed upstream).

ste93cry:
We declined this feature in the past a lot of times, a simple search would have returned #1179, #1145 and #1014 as results with details on why we won't accept anything like this. If you want to send this data, either implement your own Monolog handler or decorate ours to add the data to the event using `withScope()`

iansltx:
Yep, sorry. Wasn't in the docs. Would y'all be interested in a PR that adds an explanation in the docs, along with a link to a package with a tweaked handler etc., to avoid someone opening the same issue in another six months?

iansltx:
Mentioning #848 here too as that has a much more detailed discussion on why this isn't implemented in the base package. Question still stands on whether y'all would accept a PR to the readme explaining why context is skipped, and how to opt into a workaround if someone so desires.

ste93cry:
A PR is more than welcome to amend the docs of course, but I fear this is not the right place. The official documentation is hosted on [`getsentry/sentry-docs`](https://github.com/getsentry/sentry-docs), so that's the best place where to address this issue

iansltx:
Got it. Will see if I can contribut there. Figured putting the info in the readme here would meet people where they're most likely to be looking when troubleshooting, but adding to the explicit docs repo works too.

ste93cry:
>Figured putting the info in the readme here would meet people where they're most likely to be looking when troubleshooting

I agree with you, but the Sentry guys as far as I know prefer to have the documentation in one place only. For sure it makes sense to put the info in that repo, but if they agree on duplicating it here I'm fine with it

PR #1228, opened 2021-07-18
Moved Silverstripe adaptor into supported v3 README section

phptek:


ste93cry (approved):
Thanks for the contribution :muscle:

PR #1229, opened 2021-07-18
Allow `psr/log` 2 and 3

derrabus:
👋🏻 

Versions 2 and 3 of the `psr/log` interfaces have been released with updated type declarations. This library does not provide any own implementations of those interfaces and thus should be compatible right away.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "php-http/message": "^1.5",
         "psr/http-factory": "^1.0",
         "psr/http-message-implementation": "^1.0",
-        "psr/log": "^1.0",
+        "psr/log": "^1.0|^2.0|^3.0",
         "symfony/options-resolver": "^3.4.43|^4.4.11|^5.0.11",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -33,7 +33,7 @@
         "php-http/message": "^1.5",
         "psr/http-factory": "^1.0",
         "psr/http-message-implementation": "^1.0",
-        "psr/log": "^1.0",
+        "psr/log": "^1|^2|^3",
```
ste93cry, line 36: I know it does not matter in this case, but can we stick with the convention of requiring at least the minor version as well?

ste93cry (changes requested):

composer.json
derrabus: Sure.

ste93cry (approved):

ste93cry:
Thanks, LGTM 🚢

PR #1235, opened 2021-08-16
nikic/php-parser dependency resolution to major version

jroszkiewicz:
There is no reason to lock this dependency on patch version, because minor changes don't have breaking changes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
         "friendsofphp/php-cs-fixer": "^2.17",
         "http-interop/http-factory-guzzle": "^1.0",
         "monolog/monolog": "^1.3|^2.0",
-        "nikic/php-parser": "^4.10.3",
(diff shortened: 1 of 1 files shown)


stayallive:
That is not what it means, it means that we require at least `4.10.3` or higher not only `4.10.3`.

Your change only allows `4.10.0`, `4.10.1`, `4.10.2` in addition to `4.10.3` and higher so I see no reason for us to make this change since this was either done explicitly because of a bug in earlier releases or because it was installed as that version.

See [this](https://semver.mwl.be/?package=nikic%2Fphp-parser&constraint=%5E4.10.3&stability=stable) vs. [this](https://semver.mwl.be/?package=nikic%2Fphp-parser&constraint=%5E4.10&stability=stable).

Unless there is something I'm missing I see no reason to make this change 👍

Issue #870, opened 2019-08-22
[Monolog] Handler not including context


apapsch:
We have many log calls that are structured in the [suggested way](https://github.com/Seldaek/monolog/blob/master/doc/01-usage.md#using-the-logging-context), i.e. context is an array with arbitrary fields:

```
                    $this->logger->error('Could not find channel of order', [
                        'order_id' => $orderId,
                        'channel_id' => $channelId,
                    ]);
```

Up until this week, we were still using Sentry 1.10 and the deprecated handler from Monolog upstream. The handler was configured with a LineFormatter and the line format `%message% %context% %extra%`. Sentry received messages like:

```
Could not find channel of order {"order_id": "FOO-1", "channel_id": "42"} []
```

Now we upgraded to Sentry 2.x and switched to the `Sentry\Monolog\Handler`. The message is now:

```
Could not find channel of order
```

The context is missing and also cannot be found under "Additional Data". Looking in the source of the integrated handler, the reasons become clear:

* `l. 49`: The Sentry field `message` ist set using the Monolog field `message`, which ignores the `formatted` field set by the LineFormatter.
* The arbitary context fields are not set in the Sentry scope.

The old behavior with the context added to the message as JSON object was not ideal, because Sentry would have had a hard time automatically merging the issues, resulting in many separate issues that needed manual merging. Therefore I would send a PR that adds the arbitrary context fields to the Sentry scope via `setExtra`.

ste93cry:
This has already been discussed in #848 and #844 and reasons have been explained about why we won't support anymore out-of-the-box blindly setting what is in the context and/or in the `extra` field in the event. There is a [package](https://github.com/B-Galati/monolog-sentry-handler) developed by @B-Galati that can help you by restoring pretty much what was done in `1.x`.

apapsch:
Good stuff, I will check it out. Closing then.

nuryagdym:
seems like this feature is supported now. 

in symfony sentry bundle with following config worked:
```
monolog:
    handlers:
        sentry:
            type: sentry
            dsn: "%env(SENTRY_DSN)%"
            level: !php/const Monolog\Logger::ERROR
            fill_extra_context: true #<- monolog data can be logged with this flag
```

for more info you can check `Sentry\Monolog\Handler`

PR #1238, opened 2021-09-03
Allow passed context data to be sent into sentry

noahduncan:
Currently if you pass additional context to monolog (through `$logger->error('My error', ['my_context' => 'data'])`) the current sentry handler ignores that information.  With this proposed change, we can maintain the functionality around context with exceptions, but if additional context is passed, we send that to sentry as well.

### DIFF
--- a/src/Monolog/Handler.php
+++ b/src/Monolog/Handler.php
@@ -51,6 +51,11 @@ final class Handler extends AbstractProcessingHandler
 
         if (isset($record['context']['exception']) && $record['context']['exception'] instanceof \Throwable) {
             $hint->exception = $record['context']['exception'];
+            unset($record['context']['exception']);
+        }
+
+        if (!empty($record['context'])) {
+            $event->setContext('context', $record['context']);
         }
 
         $this->hub->withScope(function (Scope $scope) use ($record, $event, $hint): void {


ste93cry:
This has been discussed and declined in the past for a series of reasons, please check #870 and referenced issues/PR to understand why. There are no plans to change mind on this

noahduncan:
Ah I see. And I see the recommended approach to do it myself. Thanks and
sorry to rehash and old decision!

devnix:
Hi @ste93cry! I'm looking the suggested solutions but I don't see a way to add Monolog's context on `^3.0` version of the SDK.

ste93cry:
You can take a look at https://github.com/getsentry/sentry-php/pull/848#issuecomment-605902783 to see an example of how decoration of the handler can be useful in this case

devnix:
I've been trying to work with that with no luck. Here's where I'm stuck with a more recent attemp:

image

```yaml
monolog:
    handlers:
        sentry:
            type: service
            level: !php/const Monolog\Logger::WARNING
            hub_id: Sentry\State\HubInterface
            id: App\Infrastructure\Sentry\MonologHandler

services:
    App\Infrastructure\Sentry\MonologHandler:
        decorates: Sentry\Monolog\Handler
        arguments: ['@App\Infrastructure\Sentry\MonologHandler.inner']
```

What I really feel is that we are missing updated official docs, or maybe this functionality built in: https://github.com/getsentry/sentry-docs/issues/4125

ste93cry:
The error is itself clear: you're decorating a service that doesn't exist. If you don't use the `sentry` Monolog handler type which instantiates the handler itself, you will have to manually set up things yourself. You are also mixing options like `hub_id` which is specific of the `sentry` handler with the options of the `service` handler... Here two different examples on how I would make this work, but I wrote them on the fly so they may need some tweaks to work:

```yaml
monolog:
    handlers:
        sentry:
            type: service
            id: App\Infrastructure\Sentry\MonologHandler

services:
    Sentry\Monolog\Handler:
        arguments:
            $level: !php/const Monolog\Logger::WARNING

    App\Infrastructure\Sentry\MonologHandler:
        decorates: Sentry\Monolog\Handler
```

```yaml
monolog:
    handlers:
        sentry:
            type: sentry
            level: !php/const Monolog\Logger::ERROR
            hub_id: Sentry\State\HubInterface

services:
    App\Infrastructure\Sentry\MonologHandler:
        decorates: monolog.handler.sentry # The name should come from https://github.com/symfony/monolog-bundle/blob/91a6bcca0e13d4182ebab28deabaef4eb03ca549/DependencyInjection/MonologExtension.php#L942
```

>What I really feel is that we are missing updated official docs

We will update the docs to mention how to make this work, but beware that this is just a basic Symfony service registration process with a decorator class, nothing special after all 😀

devnix:
Thank you @ste93cry, I've never worked with service decoration before.

It seems to be working now with both approaches (currently using the second one you just provided), but seems like now the stacktrace is not registered in Sentry. Do I have to send them through the decorator too?

ste93cry:
Glad to see you solved the issue 👍 Should you need more help though, please open a separate issue or use the [Discussions](https://github.com/getsentry/sentry-php/discussions) feature that got just enabled so that we avoid polluting the inbox of people that don't care about your specific problem

>This is the kind of documentation I would loved to have on the Symfony guide

The `in_app_exclude` option is documented as far as I know, but of course since it's not tied specifically to Monolog you won't find any mention of it in its documentation

borodulin:
My solution to this problem. Just added this class to the project:

```php
<?php

declare(strict_types=1);

namespace App\Infrastructure\Monolog;

use Monolog\Processor\ProcessorInterface;

class ExtraMessageProcessor implements ProcessorInterface
{
    public function __invoke(array $record): array
    {
        if (isset($record['context'])) {
            $record['context']['extra'] = $record['context'];
        }

        return $record;
    }
}

```

Symfony Monolog Bundle registers processors (ProcessorInterface) for autoconfiguration.

PR #1239, opened 2021-09-06
Replace findUrlFactory with findUriFactory for Psr17FactoryDiscovery

gclove:
https://github.com/php-http/discovery/blob/master/src/Psr17FactoryDiscovery.php#L125-L135

```php
    /**
     * @return UriFactoryInterface
     *
     * @throws Exception\NotFoundException
     *
     * @deprecated This will be removed in 2.0. Consider using the findUriFactory() method.
     */
    public static function findUrlFactory()
    {
        return static::findUriFactory();
    }
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "jean85/pretty-package-versions": "^1.5|^2.0.1",
         "php-http/async-client-implementation": "^1.0",
         "php-http/client-common": "^1.5|^2.0",
-        "php-http/discovery": "^1.6.1",
+        "php-http/discovery": "^1.11",
         "php-http/httplug": "^1.1|^2.0",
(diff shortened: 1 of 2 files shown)


gclove:
Sorry, this change is from php-http/discovery `1.11.0` .

https://github.com/php-http/discovery/pull/181

ste93cry:
I think there is no problem in bumping the minimum required version to `1.11`, if it makes it easier to then also support the next major :wink:

gclove:
Thank you, I actually think so too 😂 
May you help me re-run the CI workflows check?

ste93cry (approved):

ste93cry:
Thank you for the contribution!

PR #1244, opened 2021-09-24
Set the values specified in the Monolog's context in the extra context of the event

ste93cry:
Given the discussion we had in getsentry/sentry-docs#4125 and the fact that this feature is requested repeatedly, let's add it back to the SDK. As a side note, the documentation warns users that the extra context is deprecated, so this is one more reason I strongly believe we should not add this because we cannot know which structured context we would have to create on behalf of the users.

>In addition to structured contexts, Sentry supports adding unstructured "Additional Data" via set_extra. Additional Data is deprecated in favor of structured contexts and should be avoided when possible.

https://docs.sentry.io/platforms/php/enriching-events/context/#additional-data

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Update Guzzle tracing middleware to meet the [expected standard](https://develop.sentry.dev/sdk/features/#http-client-integrations) (#1234)
 - Add `toArray` public method in `PayloadSerializer` to be able to re-use Event serialization
 - The `withScope` methods now return the callback's return value (#1263)
+- Set the event extras by taking the data from the Monolog record's context (#1244)
 
 ## 3.3.6 (2022-01-14)
 
@@ -37,7 +38,7 @@
 ## 3.3.0 (2021-05-26)
 
 - Allow setting a custom timestamp on the breadcrumbs (#1193)
-- Add option `ignore_tags` to `IgnoreErrorsIntegration` in order to ignore exceptions by tags values. (#1201)
+- Add option `ignore_tags` to `IgnoreErrorsIntegration` in order to ignore exceptions by tags values (#1201)
 
 ## 3.2.2 (2021-05-06)
 
--- a/src/EventId.php
+++ b/src/EventId.php
@@ -35,7 +35,7 @@ final class EventId implements \Stringable
      */
     public static function generate(): self
     {
-        return new self(str_replace('-', '', uuid_create(UUID_TYPE_RANDOM)));
+        return new self(str_replace('-', '', uuid_create(\UUID_TYPE_RANDOM)));
     }
 
     public function __toString(): string
--- a/src/EventId.php
+++ b/src/EventId.php
@@ -35,7 +35,7 @@ final class EventId implements \Stringable
      */
     public static function generate(): self
     {
-        return new self(str_replace('-', '', uuid_create(\UUID_TYPE_RANDOM)));
+        return new self(str_replace('-', '', uuid_create(UUID_TYPE_RANDOM)));
     }
 
     public function __toString(): string
--- a/src/Monolog/Handler.php
+++ b/src/Monolog/Handler.php
@@ -57,6 +57,15 @@ final class Handler extends AbstractProcessingHandler
             $scope->setExtra('monolog.channel', $record['channel']);
             $scope->setExtra('monolog.level', $record['level_name']);
 
+            foreach ($record['context'] as $key => $value) {
+                // We skip the `exception` field because it's put in the event's hint
+                if ('exception' === $key) {
+                    continue;
+                }
+
+                $scope->setExtra((string) $key, $value);
+            }
+
             $this->hub->captureEvent($event, $hint);
         });
     }
--- a/src/Monolog/Handler.php
+++ b/src/Monolog/Handler.php
@@ -20,21 +20,29 @@ use Sentry\State\Scope;
  */
 final class Handler extends AbstractProcessingHandler
 {
+    private const CONTEXT_EXCEPTION_KEY = 'exception';
+
     /**
      * @var HubInterface
      */
     private $hub;
 
     /**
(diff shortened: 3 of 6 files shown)


Jean85 (changes requested):
Wait, this is not what I was advocating for in the discussion. Making this the default is still not a smart move: my suggestion was to port the handler from the `sentry-laravel` codebase, and leave it unhooked, or put it under an option disabled by default. In that way, we could solve the users' needs, without making it a default behavior of the SDK.

CHANGELOG.md
```diff
@@ -16,7 +17,7 @@
 ## 3.3.0 (2021-05-26)
 
 - Allow setting a custom timestamp on the breadcrumbs (#1193)
-- Add option `ignore_tags` to `IgnoreErrorsIntegration` in order to ignore exceptions by tags values. (#1201)
+- Add option `ignore_tags` to `IgnoreErrorsIntegration` in order to ignore exceptions by tags values (#1201)
```
glensc, line 20: outside scope of this merge request, move to a separate pull request

glensc:
As noted by @Jean85 and by [me](https://github.com/getsentry/sentry-php/pull/1271#issuecomment-1000749018) this needs to be opt-in, add constructor parameter to the Handler. Otherwise, it's a breaking change.

src/Monolog/Handler.php
```diff
@@ -57,6 +57,15 @@ protected function write(array $record): void
             $scope->setExtra('monolog.channel', $record['channel']);
             $scope->setExtra('monolog.level', $record['level_name']);
 
+            foreach ($record['context'] as $key => $value) {
```
Wick96, line 60: @glensc Hi i have two ideas :)

The first one is to maybe just add logger context as  `$scope->setExtra('monolog.context', $record['context']);` and then the user can freely filter or anonymize this array later in `before_send` hook if he needs it.

Sec idea... maybe for now just add context to hint `$hint->extra = ['monolog.context', $record['context']];` so the user can then manually filter & add data he needs to event in `before_send` hook.

Just "my two cents".

src/EventId.php
```diff
@@ -35,7 +35,7 @@ public function __construct(string $value)
      */
     public static function generate(): self
     {
-        return new self(str_replace('-', '', uuid_create(UUID_TYPE_RANDOM)));
+        return new self(str_replace('-', '', uuid_create(\UUID_TYPE_RANDOM)));
```
glensc, line 38: unrelated change to the pull request. should be submitted separately

CHANGELOG.md
ste93cry: I'm just removing the final dot to keep consistency with the rest of the points of the list, it's not worth opening a separate PR for this 😅

glensc: it's not about worthiness, it's about quality. you don't stuff unrelated changes together.

please read about atomic commits:
- https://www.freshconsulting.com/insights/blog/atomic-commits/

so, now consider your changeset needs to be reverted, but since it's all same commit, your unrelated change will be reverted too with `git revert`.

also, unrelated noise makes PR more readable and the style change PR gets merged very likely sooner as it's not affecting functionality.

src/EventId.php
glensc: and since already nit-picking about commits quality, having commit titled "CR fixes" does not give any value to commit history, since you already rebased and force pushed, squash it with original commit.

CHANGELOG.md
ste93cry: I understand your motivation and I agree with your reasoning when talking about big changes not related to the PR's purpose, but commenting that adding a dot to a phrase is something that should be done separately has no value for the code review in my opinion and it's just nitpicking

src/EventId.php
ste93cry: >should be submitted separately

This change actually is unwanted, it seems that when `ext-uuid` extension is installed (like on my environment) `PHP-CS-Fixer` consider the constant as available natively while when not present (but the polyfill is) it doesn't.

src/Monolog/Handler.php
ste93cry: @Jean85 @stayallive opinions about the first point? I'm kind of undecided if we should put everything under the extra context or if as, suggested, it would be wiser to group everything under monolog.context

Jean85: Maybe the second option, a dedicated context, would be better? That way we could avoid situation where we accidentally override user-submitted data.

ste93cry: Done!

glensc: @ste93cry if you inow how to rebase and force push, why don't you already cleanup the history? like why have commit stuff, revert stuff? just squash so that it's a commit that adds stuff with the reverted stuff in same commit?

ste93cry: Keeping the history of commits in the PR is sometimes useul to keep track of the process that led to the final version of the code to review, giving more context to the reviewer. Besides that, squashing is done when merging into the target branch so it doesn't really matter how many commits there are here, the history of the repository will be clean anyway :wink:

_Funny fact: you were the one advocating for commits to be as atomic as possible so that it's possible to revert them without losing other unrelated changes_

glensc: That completely changes the story if the pull request is squash merged. I wish github/gitlab present it to reviewers that squash merge is used, then could just omit any commit-related quality checks.

as for the atomic part, for the purposes of merging cleaner history, I would exclude (with rebase) changes from merge that are removed rather than merging revert commits. as everything before merge can be rewritten. if you want to preserve historic decisions, then maybe start with new pull request leave old one for posterity.

so far both github/gitlab still don't present well previous versions of pushes. gitlab at least has merge request versions, but in github you have to dig through timeline find the "has force pushed" and then can find previous versions.

Jean85 (approved):

stayallive (approved):
Looks good 👍

Issue #1243, opened 2021-09-24
Verify `php_uname` function availability before calling `EnvironmentIntegration.setOsContext()`


giilby:
Similarly to #985, I'm using sentry-php (via [wp-sentry](https://github.com/stayallive/wp-sentry)) with a WordPress site hosted on Flywheel. Unfortunately, wp-sentry doesn't currently provide a way to properly override the default integrations set up by sentry-php. Looking at the changes made in #1072, it seems like a straightforward way to remedy this issue would be to  wrap `EnvironmentIntegration:31` with a check to see if the `php_uname` function is available. Is there any reason not to do this? I'm happy to submit a PR with the change if it makes sense. Thanks!

ste93cry:
As I mentioned in the past, I think this is the classic case of "security through obscurity" and by wrapping the function calls with a `function_exist` check may open the door to end up like WordPress that checks that all functions exist. Anyway, I would happily accept a PR if you want to propose the change 😄

ste93cry:
The fix has been released in version `3.3.3`

PR #1246, opened 2021-09-28
Verify availability of php_uname in EnvironmentIntegration

giilby:
Per #1243, some hosting environments disable the use of `php_uname`. When using `EnvironmentIntegration` (enabled by default) in these environments, errors like this occur:

```
PHP Fatal error: Uncaught TypeError: Argument 1 passed to Sentry\Context\OsContext::__construct() must be of the type string, null given, called in [...]/src/Integration/EnvironmentIntegration.php on line 46 and defined in [...]/src/Context/OsContext.php:37
```

This small change updates `EnvironmentIntegration` to check for the existence of `php_uname` before calling it.

### DIFF
--- a/src/Integration/EnvironmentIntegration.php
+++ b/src/Integration/EnvironmentIntegration.php
@@ -50,8 +50,8 @@ final class EnvironmentIntegration implements IntegrationInterface
 
     private function updateServerOsContext(?OsContext $osContext): ?OsContext
     {
-        if (!function_exists('php_uname')) {
-            return null;
+        if (!\function_exists('php_uname')) {
+            return $osContext;
         }
 
         if (null === $osContext) {
--- a/src/Integration/EnvironmentIntegration.php
+++ b/src/Integration/EnvironmentIntegration.php
@@ -28,7 +28,10 @@ final class EnvironmentIntegration implements IntegrationInterface
 
             if (null !== $integration) {
                 $event->setRuntimeContext($integration->updateRuntimeContext($event->getRuntimeContext()));
-                $event->setOsContext($integration->updateServerOsContext($event->getOsContext()));
+
+                if (function_exists('php_uname')) {
+                    $event->setOsContext($integration->updateServerOsContext($event->getOsContext()));
+                }
             }
 
             return $event;
--- a/src/Integration/EnvironmentIntegration.php
+++ b/src/Integration/EnvironmentIntegration.php
@@ -28,10 +28,7 @@ final class EnvironmentIntegration implements IntegrationInterface
 
             if (null !== $integration) {
                 $event->setRuntimeContext($integration->updateRuntimeContext($event->getRuntimeContext()));
-
-                if (function_exists('php_uname')) {
-                    $event->setOsContext($integration->updateServerOsContext($event->getOsContext()));
-                }
+                $event->setOsContext($integration->updateServerOsContext($event->getOsContext()));
             }
 
             return $event;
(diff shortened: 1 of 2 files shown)


Jean85 (changes requested):
Thank you for you contribution! Could you please add a changelog entry?

Also, maybe we could fallback to the PHP constant if it doesn't work? @ste93cry WDYT?

giilby:
@Jean85 Done! RE: falling back to the PHP constant, are you referring to `PHP_OS`? My understanding is that it reports the OS that PHP was built on, which isn't as useful as what's reported by `php_uname`. In my situation, I don't mind lacking OS information in order to have error reporting work. I think we should keep this change simple as it currently is, but am happy to make more changes if you prefer. Thanks!

Jean85 (approved):
You're right! I though the constant was populated at start time, instead it's added at compile time: https://www.php.net/manual/en/reserved.constants.php#constant.php-os

LGTM :+1:

src/Integration/EnvironmentIntegration.php
```diff
@@ -28,7 +28,10 @@ public function setupOnce(): void
 
             if (null !== $integration) {
                 $event->setRuntimeContext($integration->updateRuntimeContext($event->getRuntimeContext()));
-                $event->setOsContext($integration->updateServerOsContext($event->getOsContext()));
+
+                if (function_exists('php_uname')) {
```
ste93cry, line 32: What about moving this condition inside the `updateServerOsContext` function? In fact, we could simply have an early return before [line 56](https://github.com/getsentry/sentry-php/blob/master/src/Integration/EnvironmentIntegration.php#L53)

ste93cry (changes requested):

stayallive (approved):
Thanks 👍

giilby:
I see that Codecov is complaining about line 54 not being covered by unit tests...I would've created a unit test for this case, but I don't know of a reasonable way to temporarily remove the `php_uname` function. Not sure how you want to resolve this...the way I had the code before would avoid complaints from Codecov...

src/Integration/EnvironmentIntegration.php
```diff
@@ -48,8 +48,12 @@ private function updateRuntimeContext(?RuntimeContext $runtimeContext): RuntimeC
         return $runtimeContext;
     }
 
-    private function updateServerOsContext(?OsContext $osContext): OsContext
+    private function updateServerOsContext(?OsContext $osContext): ?OsContext
     {
+        if (!function_exists('php_uname')) {
+            return null;
```
stayallive, line 54: This should probably just return `$osContext` since it could contain context from other sources.

And added a backslash to keep CI happy :)

```suggestion
        if (!\function_exists('php_uname')) {
            return $osContext;
```

stayallive (changes requested):

stayallive:
This can probably be tested using a phpt test, but I'm not sure it's worth a full test file for this.

ste93cry:
I agree, if we want 100% coverage the easiest way is to implement a PHPT test, however it's probably not worth it

src/Integration/EnvironmentIntegration.php
giilby: Good call on returning `$osContext` here...I should've done that to begin with. Just committed these changes...thanks!

stayallive (approved):

ste93cry (approved):
Thanks, LGTM

PR #1247, opened 2021-10-07
Support PHP 8.1 for sentry-php 2.x

annuh:
We can't update to sentry-php 3.x, because of this issue: https://github.com/getsentry/sentry-symfony/issues/421.

With this change we can still use Sentry and upgrade to the latest PHP 8.1. 🙂


ste93cry:
There is already open #1245 for PHP 8.1, I will keep working on it once I have some spare time

PR #1249, opened 2021-10-08
Add the ablitity to transform the event to array

BafS:
This PR doesn't have a BC but it adds a public function to be able to serliaize an event as an array.

**Context:**
We are using the backend as a proxy and we decorate a few part of the payload (re-set the project ID for example).
In a previous version it was possible to re-use more logic but with the current code the method `serialize` returns a string which makes it not possible to re-use (except if we decode it again...).
We are currently forced to implements `PayloadSerializerInterface` and to copy/paste everything from `PayloadSerializer` just to be able to get the payload as an array and reconstruct the `$record` to decorate it and save it as a file, having a public method would make our implementation less fragile.

### DIFF
--- a/src/Serializer/PayloadSerializer.php
+++ b/src/Serializer/PayloadSerializer.php
@@ -39,6 +39,9 @@ final class PayloadSerializer implements PayloadSerializerInterface
         return JSON::encode($result);
     }
 
+    /**
+     * @return array<string, mixed>
+     */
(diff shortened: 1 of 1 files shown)


ste93cry:
It looks like there is still something to rebase, as I see a commit that shouldn't be here 🤔

BafS:
Thanks, hopefully it should be green now

PR #1250, opened 2021-10-10
Avoid error when IP address is null for UserDataBag

alexander-schranz:
Think it should also be creatable with a null IP Address? 

Used in Symfony getClientIP() can return no IP if no is set: https://github.com/getsentry/sentry-symfony/blob/42787b4f8a8fbff03da56b7291d3de59fb669bfe/src/EventListener/RequestListener.php#L63

### DIFF
--- a/src/UserDataBag.php
+++ b/src/UserDataBag.php
@@ -62,7 +62,7 @@ final class UserDataBag
      *
      * @param string $ipAddress The IP address of the user
      */
-    public static function createFromUserIpAddress(string $ipAddress): self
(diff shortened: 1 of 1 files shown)


ste93cry:
I don't think that we should change a method intended to create a valid instance of the value object from an IP address by making it accept an invalid value for it. However, it would be better to avoid this situation in the Symony SDK instead

alexander-schranz:
Created a PR there: https://github.com/getsentry/sentry-symfony/pull/565

PR #1253, opened 2021-10-20
Return EventId only on successful response

tutNichts:
Related to this one: https://github.com/getsentry/sentry-symfony/issues/372

I wasn't able to test this properly yet (will do asap), but at least PHPUnit says "ok".

Feel free to reject it since it feels a bit too simple.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -162,7 +162,7 @@ final class Client implements ClientInterface
             $response = $this->transport->send($event)->wait();
             $event = $response->getEvent();
 
-            if (null !== $event) {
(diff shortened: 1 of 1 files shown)


ste93cry:
No. To be clear, I'm not against the proposed change, I just wanna be sure that the current behavior is not the expected one, then we can proceed with the work

tutNichts:
Anyone wants to add something here?

ste93cry:
From my side, as said before I agree on the usefulness of this change, but on the other side I don't want this SDK to be different than the other ones. @bruno-garcia I will leave the final word about this to you

bruno-garcia:
I'd defer this to @smeubank

PR #1263, opened 2021-11-15
Return original return value.

vaibhavpandeyvpz:
The `withScope` method is very useful to push local scope to error that happen during specific executions. But what if we rely on return values from the execution? It can be handled in a ugly way as below using current code.

```php
$success = false;

Sentry\withScope(function (Sentry\State\Scope $scope) use ($processor, $job, &$success) {
    $scope->setUser([
        'username' => $job->user->username,
    ]);
    $success = $processor->processJob($job);
});

if ($success) {
    // do whatever
}
```

The goal of this PR to simplify above as below:

```php
$success = Sentry\withScope(function (Sentry\State\Scope $scope) use ($processor, $job) {
    $scope->setUser([
        'username' => $job->user->username,
    ]);
    return $processor->processJob($job);
});
```

IMO this is cleaner and doesn't introduce any backwards incompatibility with existing code as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,12 @@
 
 ## 3.3.4 (2021-11-08)
 
-- Update the Guzzle tracing middleware to meet the [expected standard](https://develop.sentry.dev/sdk/features/#http-client-integrations) (#1234)
+- Update Guzzle tracing middleware to meet the [expected standard](https://develop.sentry.dev/sdk/features/#http-client-integrations) (#1234)
 - Run the test suite against PHP `8.1` (#1245)
 - Avoid overwriting the error level set by the user on the event when capturing an `ErrorException` exception (#1251)
 - Allow installing the project alongside Symfony `6.x` components (#1257)
 - Add `toArray` public method in `PayloadSerializer` to be able to re-use Event serialization
+- The `withScope` methods now return the callback's return value (#1263)
 
 ## 3.3.3 (2021-10-04)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,9 @@
 
 ## 3.3.4 (2021-11-08)
 
-- Run the test suite against PHP `8.1` (#1245)
 - Avoid overwriting the error level set by the user on the event when capturing an `ErrorException` exception (#1251)
 - Allow installing the project alongside Symfony `6.x` components (#1257)
+- Run the test suite against PHP `8.1` (#1245)
 
 ## 3.3.3 (2021-10-04)
 
--- a/src/State/HubInterface.php
+++ b/src/State/HubInterface.php
@@ -54,8 +54,9 @@ interface HubInterface
      * is automatically removed once the operation finishes or throws.
      *
      * @param callable $callback The callback to be executed
+     * @return mixed|void The callback's return value
      */
-    public function withScope(callable $callback): void;
+    public function withScope(callable $callback);
 
     /**
      * Calls the given callback passing to it the current scope so that any
--- a/src/State/HubInterface.php
+++ b/src/State/HubInterface.php
@@ -56,6 +56,10 @@ interface HubInterface
      * @param callable $callback The callback to be executed
      *
      * @return mixed|void The callback's return value, upon successful execution
+     *
+     * @phpstan-template T
+     * @phpstan-param callable(Scope): T $callback
+     * @phpstan-return T
      */
     public function withScope(callable $callback);
 
--- a/src/State/HubInterface.php
+++ b/src/State/HubInterface.php
@@ -54,7 +54,8 @@ interface HubInterface
      * is automatically removed once the operation finishes or throws.
      *
      * @param callable $callback The callback to be executed
-     * @return mixed|void The callback's return value
+     *
+     * @return mixed|void The callback's return value, upon successful execution
      */
     public function withScope(callable $callback);
 
--- a/src/functions.php
+++ b/src/functions.php
@@ -94,7 +94,8 @@ function configureScope(callable $callback): void
  * is automatically removed once the operation finishes or throws.
  *
  * @param callable $callback The callback to be executed
- * @return mixed|void The callback's return value
+ *
+ * @return mixed|void The callback's return value, upon successful execution
  */
 function withScope(callable $callback)
 {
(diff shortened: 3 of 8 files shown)


Jean85 (changes requested):
Using short closures would be a less-ugly way to do something like this, but IMO this feature still has its merits!

To go forward, could you please:
 * rebase on `develop` (@ste93cry already retargeted your MR)
 * fix CS
 * add a test
 * add a changelog entry

HazAT:
@ste93cry We had a quick internal discussion and it's fine ✔️ 
We can have small QoL changes like these if it makes usage of our SDK easier and more convenient

vaibhavpandeyvpz:
@Jean85 Ready for review again.

tests/FunctionsTest.php
```diff
@@ -206,6 +206,15 @@ public function testWithScope(): void
         $this->assertTrue($callbackInvoked);
     }
 
+    public function testWithScopeWithCallback(): void
```
ste93cry, line 209: It doesn't make sense to pass the variable as reference to the lambda function if you then assign the return value of the callback to it from the outer scope. This test can be merged with the previous one instead and we can simply check the return value

tests/State/HubAdapterTest.php
```diff
@@ -105,6 +105,23 @@ public function testWithScope(): void
         HubAdapter::getInstance()->withScope($callback);
     }
 
+    public function testWithScopeWithReturnValue(): void
```
ste93cry, line 108: As written above, this test can be squashed with the previous one and you can just check that the value returned by the callback is the one you're expecting

tests/State/HubAdapterTest.php
```diff
@@ -105,6 +105,23 @@ public function testWithScope(): void
...
+    {
+        $callback = static function () {
+            return true;
+        };
+
+        $hub = $this->createMock(HubInterface::class);
+        $hub->expects($this->once())
+            ->method('withScope')
+            ->with($callback)
+            ->willReturn(true);
```
ste93cry, line 118: It would be wiser to use `willReturnArgument()` so that we won't have to bother changing the argument in a future refactoring if we will decide to change the expected value

vaibhavpandeyvpz:
> There are still commits not belonging to this PR that need to be removed 😬

I don't understand how they came in though. Maybe because I did cut the branch off `master`.

CHANGELOG.md
```diff
@@ -2,8 +2,14 @@
...
 
-- Update the Guzzle tracing middleware to meet the [expected standard](https://develop.sentry.dev/sdk/features/#http-client-integrations) (#1234)
+## 3.3.4 (2021-11-08)
+
+- Update Guzzle tracing middleware to meet the [expected standard](https://develop.sentry.dev/sdk/features/#http-client-integrations) (#1234)
+- Run the test suite against PHP `8.1` (#1245)
+- Avoid overwriting the error level set by the user on the event when capturing an `ErrorException` exception (#1251)
+- Allow installing the project alongside Symfony `6.x` components (#1257)
 - Add `toArray` public method in `PayloadSerializer` to be able to re-use Event serialization
+- The `withScope` methods now return the callback's return value (#1263)
```
ste93cry: This change should go under the `Unreleased` section

ste93cry:
It happened because the branch was originally branched from `master`, which contained more commits than `develop`. When I changed the target branch, the commits remained. I've now merged `master` into `develop`, but a rebase is still needed

src/functions.php
```diff
@@ -94,10 +94,12 @@ function configureScope(callable $callback): void
  * is automatically removed once the operation finishes or throws.
  *
  * @param callable $callback The callback to be executed
+ *
+ * @return mixed|void The callback's return value, upon successful execution
```
ste93cry, line 98: Same as above

CHANGELOG.md
```diff
@@ -2,14 +2,15 @@
 
 ## Unreleased
 
-- Update the Guzzle tracing middleware to meet the [expected standard](https://develop.sentry.dev/sdk/features/#http-client-integrations) (#1234)
+- Update Guzzle tracing middleware to meet the [expected standard](https://develop.sentry.dev/sdk/features/#http-client-integrations) (#1234)
 - Add `toArray` public method in `PayloadSerializer` to be able to re-use Event serialization
+- The `withScope` methods now return the callback's return value (#1263)
```
ste93cry, line 7: This change should go under the `Unreleased` section

CHANGELOG.md
```diff
@@ -2,14 +2,15 @@
 
 ## Unreleased
 
-- Update the Guzzle tracing middleware to meet the [expected standard](https://develop.sentry.dev/sdk/features/#http-client-integrations) (#1234)
```
ste93cry, line 5: 🙏 revert all unnecessary changes to this file

src/State/HubInterface.php
```diff
@@ -54,8 +54,10 @@ public function popScope(): bool;
      * is automatically removed once the operation finishes or throws.
      *
      * @param callable $callback The callback to be executed
+     *
+     * @return mixed|void The callback's return value, upon successful execution
      */
-    public function withScope(callable $callback): void;
+    public function withScope(callable $callback);
```
ste93cry, line 60: Please add PHPStan generics to tie the signature of the callable to the return value of this function:

```
@phpstan-template T

@phpstan-param callable(Scope): T $callback

@phpstan-return T
```

tests/State/HubTest.php
```diff
@@ -167,6 +167,35 @@ public function testWithScope(): void
         $this->assertTrue($callbackInvoked);
     }
 
+    public function testWithScopeWithReturnValue(): void
```
ste93cry, line 170: This test can be merged with the previous one, as suggested for the other similar places

ste93cry (changes requested):

CHANGELOG.md
vaibhavpandeyvpz: I had moved this already.

ste93cry (changes requested):
Just some more quick changes and we should be good to go! Thanks for the patience

ste93cry:
@vaibhavpandeyvpz are you still interested in completing the work?

ste93cry:
Closing as superseded by @HypeMC's work

PR #1264, opened 2021-11-17
Support PHP 8.1 for sentry-php 2.x 

annuh:
We can't update to sentry-php 3.x, because of this issue: https://github.com/getsentry/sentry-symfony/issues/421.

With this change we can still use Sentry and upgrade to the latest PHP 8.1. 🙂 

Fixes:

```
Fatal error:  During inheritance of ArrayAccess: Uncaught ErrorException: Return type of Sentry\Context\Context::offsetGet($offset) should either be compatible with ArrayAccess::offsetGet(mixed $offset): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/html/vendor/sentry/sentry/src/Context/Context.php:152
```

(follow-up from https://github.com/getsentry/sentry-php/pull/1247)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         }
     ],
     "require": {
-        "php": "^7.1",
+        "php": "^7.1|^8.0",
         "ext-json": "*",
         "ext-mbstring": "*",
         "guzzlehttp/promises": "^1.3",
--- a/src/Context/Context.php
+++ b/src/Context/Context.php
@@ -149,7 +149,8 @@ class Context implements \ArrayAccess, \JsonSerializable, \IteratorAggregate
     /**
      * {@inheritdoc}
      */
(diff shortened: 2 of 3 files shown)


ste93cry:
As written in https://github.com/getsentry/sentry-php/pull/1247#issuecomment-937902900, not supporting PHP 8 on the `2.x` version was a deliberate choice because we don't have the resources to maintain that SDK if issues arise due to incompatibility or other things like that. Moreover, we did some changes in error handling to support PHP8 and we didn't backport them to older versions, so while it may seem that adding a few attributes here and there will make things work, there are other issues hidden elsewhere that will hit you in the long run. For this reason, I'm going to close the issue. I've left a comment on the referenced issue in the `getsentry/sentry-symfony` repo with a workaround that I think may work (not tested though)

PR #1267, opened 2021-11-30
Bump minor to include fix for composer #9816

99linesofcode:
Didn't run the tests yet. Hoping that CI will catch any boo boo's.

Composer https://github.com/composer/composer/releases/tag/2.0.14 introduced a breaking change to something that `pretty-package-versions` depends on. They fixed it in their `2.0.4` release. See:

https://github.com/Jean85/pretty-package-versions/pull/39

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "ext-mbstring": "*",
         "guzzlehttp/promises": "^1.4",
         "guzzlehttp/psr7": "^1.7|^2.0",
-        "jean85/pretty-package-versions": "^1.5|^2.0.1",
+        "jean85/pretty-package-versions": "^1.5|^2.0.4",
         "php-http/async-client-implementation": "^1.0",
         "php-http/client-common": "^1.5|^2.0",
         "php-http/discovery": "^1.6.1",


99linesofcode:
`sentry-laravel` will also need to be updated to incorporate these changes. Is that on your radar already?

Jean85 (approved):
Thanks!

ste93cry:
Thinking more about this, I have two questions:

- @Jean85, why your package doesn't require the minimum version of Composer that prevents the issue, forcing this package instead to require it?
- @99linesofcode, why should we update `sentry-larave`l? That package depends on this one, so once we release this change the Laravel users will benefit from it in cascade

ste93cry:
>@Jean85, why your package doesn't require the minimum version of Composer that prevents the issue, forcing this package instead to require it?

I will answer myself to this question: because the issue was not with Composer itself but with your package. I misread the package name in the `composer.json` and thought that `2.0.4` was the version of Composer. Sorry, long day today 🤯

ste93cry (approved):

Jean85:
> `sentry-laravel` will also need to be updated to incorporate these changes. Is that on your radar already?

Does it? I don't think it uses anything directly, it's this base SDK that does everything, so it should be fixed without further intervention.

99linesofcode:
> Does it? I don't think it uses anything directly, it's this base SDK that does everything, so it should be fixed without further intervention.

Yeah my bad. I thought maybe we had to bump the `sentry/sdk` version but of course its version range constrained 👍

Thanks for the quick 1-2 guys!

99linesofcode:
@ste93cry could it be that we still need to tag this in order for packagist to pick up on it properly?

Jean85:
This is not tagged yet, yes, but you can apply the same constraint to your projects as a workaround for now.

ste93cry:
Ops you're right, I completely forgot to release it 🤦‍♂️ Done now with version [`3.3.5`](https://github.com/getsentry/sentry-php/releases/tag/3.3.5)

PR #1272, opened 2022-01-06
Update PHPStan, Psalm and PHP-CS-Fixer to the latest versions

ste93cry:
As per title, I've updated the development tools to their latest version, taking the occasion to remove all Psalm error suppressions annotations and introducing the baseline instead. I've also updated the level of PHPStan to 9, which is the maximum level available at the moment of writing.

### DIFF
--- a/.github/workflows/static-analysis.yaml
+++ b/.github/workflows/static-analysis.yaml
@@ -17,8 +17,6 @@ jobs:
 
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
-        with:
-          php-version: '7.4'
 
       - name: Install dependencies
         run: composer update --no-progress --no-interaction --prefer-dist
--- a/.github/workflows/static-analysis.yaml
+++ b/.github/workflows/static-analysis.yaml
@@ -17,6 +17,8 @@ jobs:
 
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.1'
 
       - name: Install dependencies
         run: composer update --no-progress --no-interaction --prefer-dist
--- a/composer.json
+++ b/composer.json
@@ -45,8 +45,8 @@
         "nikic/php-parser": "^4.10.3",
         "php-http/mock-client": "^1.3",
         "phpstan/extension-installer": "^1.0",
-        "phpstan/phpstan": "^0.12",
-        "phpstan/phpstan-phpunit": "^0.12",
+        "phpstan/phpstan": "^1.3",
+        "phpstan/phpstan-phpunit": "^1.0",
         "phpunit/phpunit": "^8.5.14|^9.4",
         "symfony/phpunit-bridge": "^5.2|^6.0",
         "vimeo/psalm": "^4.2"
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         "symfony/polyfill-uuid": "^1.13.1"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "^3.4",
+        "friendsofphp/php-cs-fixer": "^2.19|^3.4",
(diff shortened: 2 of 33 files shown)


.github/workflows/static-analysis.yaml
```diff
@@ -17,8 +17,6 @@ jobs:
 
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
-        with:
-          php-version: '7.4'
```
Jean85, line 21: I would not remove this, I would just bump the version, or we will get another problem here next november...

composer.json
```diff
@@ -39,17 +39,17 @@
         "symfony/polyfill-uuid": "^1.13.1"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "^2.17",
+        "friendsofphp/php-cs-fixer": "^3.4",
```
Jean85, line 42: Adding this will disallow any Symfony < 4.4, did we check for this in the CI prefer-lowest jobs?

src/Util/JSON.php
```diff
@@ -22,12 +22,16 @@ final class JSON
      * @param int   $options  Bitmask consisting of JSON_* constants
      * @param int   $maxDepth The maximum depth allowed for serializing $data
      *
-     * @return mixed
+     * @return false|string
```
Jean85, line 25: `false` should not be possible here...

Jean85 (changes requested):

src/Util/JSON.php
ste93cry: Why not? `"false"` is a valid JSON string and so also `boolean(false)` is a valid return value

Jean85: Then any other `scalar` is fair game...
Since we bumped PHP requirement, couldn't we drop this in favor of  `JSON_THROW_ON_ERROR` and call it a day?

ste93cry: ` JSON_THROW_ON_ERROR` has nothing to do with this. As I said, the word `false` is deserialized to a `boolean(false)` because it's valid, not because an error happened. Anyway, there is one thing that is actually wrong: this method should return indeed `mixed` as before, so I'm not sure why I changed this besides thinking that PHPStan was complaining or that I misread its output. I will check it again and revert the change if this turns out to be true

ste93cry: The [PHP documentation](https://www.php.net/manual/en/function.json-encode.php) says that `false|string` is the return type for `json_encode()`, and I think that I did this change because PHPStan was complaining due to it not understanding the conditional check on `json_last_error()`

>Since we bumped PHP requirement, couldn't we drop this in favor of `JSON_THROW_ON_ERROR` and call it a day?

Indeed this or suppressing the PHPStan error are the proper ways to fix the issue. The problem with your suggestion is that that constant has been introduced in PHP `7.3`, while we still support PHP `7.2`, so the only thing left is suppressing the error for now

composer.json
ste93cry: Right, I didn't think about this. I relaxed the constraint to allow both version `2.x` and `3.x` to be installed

composer.json
```diff
@@ -39,17 +39,17 @@
         "symfony/polyfill-uuid": "^1.13.1"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "^2.17",
+        "friendsofphp/php-cs-fixer": "^2.19|^3.4",
```
Jean85, line 42: This is not ideal, since the config will break under 2.x, but that's not a big problem..

Jean85 (approved):

composer.json
ste93cry: Yes I know, but `3.x` does not support Symfony `3.4` anymore and it didn't make sense to do more work to have two distinct config files as the older version is unmaintained and we always run the tool with the newest dependencies

Issue #1273, opened 2022-01-08
`Sentry\Tracing\Span::__construct()` and `Sentry\Tracing\Span#startChild()` taking around 3% of execution time


Ocramius:
### Environment

 * relatively boring `symfony/*:^6` application with `doctrine/dbal` and `doctrine/orm`
 * `sentry/sentry-symfony:4.2.5`
 * `sentry/sdk:3.1.1` (yes, I know newer versions are out - checked diffs, didn't find anything related to this issue)

### Steps to Reproduce

Ran a full integration test suite with `XDEBUG_MODE=profile vendor/bin/phpunit`, produced a huge `cachegrind.out` file that I analyzed ( https://twitter.com/Ocramius/status/1479844147427172360 _

### Expected Result

Sentry should pretty much be non-existing in such a profile :D 

### Actual Result

 * `Sentry\Tracing\Span::__construct()` is one of the most time-consuming operations
 * `Sentry\Tracing\Span#startChild()` is one of the most time-consuming operations
 
 Here's a screenshot for clarity:

![Selection_877](https://user-images.githubusercontent.com/154256/148654480-9e1221c7-be2b-4581-afa3-6db49e5a8e73.png)

In practice, this means that `Sentry\Tracing\Span` is potentially heavier than most ORM operations, which I find quite weird.
Obviously, it is called 850000 times, in the context of the test suite, which matches query start/end, plus HTTP request start/end (where I'd expect the spans to be placed, kinda).

Looking at the implementation itself, I don't see anything evidently "high consumption", other than perhaps querying the system for `microtime(true)`:

https://github.com/getsentry/sentry-php/blob/c186c44c32899ad0cf5b4e942d71035f01b87b64/src/Tracing/Span.php#L81-L111
https://github.com/getsentry/sentry-php/blob/c186c44c32899ad0cf5b4e942d71035f01b87b64/src/Tracing/Span.php#L388-L409

I wonder if:

 * accessing `$context` could be slow?
 * cloning `$context` could be slow?
 * `microtime()` calls are not cached (because of missing FQN reference)?

Obviously, the resolution of the profile I have is low, because it covers a test suite run with thousands of tests, but I figured it would be a good idea to report this here, since I really wasn't expecting the `Span` API to be part of the higher-up time consumers.

ste93cry:
[...]

PR #1274, opened 2022-01-11
Optimize span constructor

Jean85:
This started as an investigation over #1273. I tried to run PHPBench over the `Span` constructor, and this is the result applying 5f94026:
immagine

Results seems promising, but the operation seems very small to obtain a consistent bench; @ocramius any suggestions? You can also try this branch if you want to check if there's an improvement on your profiled test suite.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -86,7 +86,11 @@
         ]
     },
     "config": {
-        "sort-packages": true
+        "sort-packages": true,
+        "allow-plugins": {
+            "composer/package-versions-deprecated": true,
+            "phpstan/extension-installer": true
+        }
     },
     "prefer-stable": true,
     "extra": {
--- a/composer.json
+++ b/composer.json
@@ -44,6 +44,7 @@
         "monolog/monolog": "^1.3|^2.0",
         "nikic/php-parser": "^4.10.3",
         "php-http/mock-client": "^1.3",
+        "phpbench/phpbench": "^1.2",
         "phpstan/extension-installer": "^1.0",
         "phpstan/phpstan": "^0.12",
         "phpstan/phpstan-phpunit": "^0.12",
--- a/src/Tracing/Span.php
+++ b/src/Tracing/Span.php
@@ -80,25 +80,17 @@ class Span
      */
     public function __construct(?SpanContext $context = null)
     {
-        $this->traceId = TraceId::generate();
-        $this->spanId = SpanId::generate();
-        $this->startTimestamp = microtime(true);
-
         if (null === $context) {
+            $this->traceId = TraceId::generate();
+            $this->spanId = SpanId::generate();
+            $this->startTimestamp = microtime(true);
+
             return;
         }
 
-        if (null !== $context->getTraceId()) {
-            $this->traceId = $context->getTraceId();
-        }
-
-        if (null !== $context->getSpanId()) {
-            $this->spanId = $context->getSpanId();
-        }
-
-        if (null !== $context->getStartTimestamp()) {
-            $this->startTimestamp = $context->getStartTimestamp();
-        }
+        $this->traceId = $context->getTraceId() ?? TraceId::generate();
+        $this->spanId = $context->getSpanId() ?? SpanId::generate();
+        $this->startTimestamp = $context->getStartTimestamp() ?? microtime(true);
 
         $this->parentSpanId = $context->getParentSpanId();
         $this->description = $context->getDescription();
--- a/tests/Benchmark/SpanBench.php
+++ b/tests/Benchmark/SpanBench.php
@@ -10,7 +10,7 @@ class SpanBench
 {
     /**
      * @Revs(10000)
-     * @Iterations(5)
+     * @Iterations(10)
      */
     public function benchConstructor(): void
     {
--- a/tests/Benchmark/SpanBench.php
+++ b/tests/Benchmark/SpanBench.php
@@ -11,10 +11,14 @@ class SpanBench
 {
     /** @var TransactionContext */
     private $context;
+    /** @var TransactionContext */
+    private $contextWithTimestamp;
 
     public function __construct()
     {
         $this->context = TransactionContext::fromSentryTrace('566e3688a61d4bc888951642d6f14a19-566e3688a61d4bc8-0');
+        $this->contextWithTimestamp = TransactionContext::fromSentryTrace('566e3688a61d4bc888951642d6f14a19-566e3688a61d4bc8-0');
+        $this->contextWithTimestamp->setStartTimestamp(microtime(true));
     }
 
     /**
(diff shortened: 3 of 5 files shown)


Ocramius:
@Jean85 this is a good start, and we can iterate on it after the benchmarks are committed to mainline

Jean85:
@Ocramius do you agree that we shouldn't put this in CI? In that case, how would you generate a sensible baseline? We would have to re-run them every time manually?

Ocramius:
> @Ocramius do you agree that we shouldn't put this in CI?

TBH, having it in CI is generally good value: obviously trouble for older PHP versions, but it doesn't need to run on `7.2`.

The reason for having it in CI is not to verify performance, but just to make sure that the benchmarks can still be run, even after years: otherwise, they quickly decay, and become unusable.

Ocramius:
What you can do in CI is running with `--revs=1 --iterations=1` - that makes sure that the bench is still usable.

stayallive (approved):
This seems like a good idea 👍 happy to see you guys diving so deep in the performance 💪

composer.json
```diff
@@ -86,7 +87,11 @@
         ]
     },
     "config": {
-        "sort-packages": true
+        "sort-packages": true,
+        "allow-plugins": {
+            "composer/package-versions-deprecated": true,
+            "phpstan/extension-installer": true
+        }
```
ste93cry, line 94: This is an important change and should be done in a separate PR to keep track of it in the history of Git

tests/Benchmark/SpanBench.php
```diff
@@ -0,0 +1,56 @@
...
+        $this->contextWithTimestamp->setStartTimestamp(microtime(true));
+    }
+
+    /**
+     * @Revs(100000)
+     * @Iterations(10)
+     */
+    public function benchConstructor(): void
+    {
+        $span = new Span();
```
ste93cry, line 36: We can remove the assignation, as it's useless. Same thing for the other methods 😄

tests/Benchmark/SpanBench.php
```diff
@@ -0,0 +1,56 @@
...
+use Sentry\Tracing\Span;
+use Sentry\Tracing\TransactionContext;
+
+class SpanBench
+{
+    /**
+     * @var TransactionContext
+     */
+    private $context;
+    /**
```
ste93cry, line 18:
```suggestion

    /**
```

tests/Benchmark/SpanBench.php
```diff
@@ -0,0 +1,56 @@
...
+declare(strict_types=1);
+
+namespace Sentry\Tests\Benchmark;
+
+use PhpBench\Benchmark\Metadata\Annotations\Iterations;
+use PhpBench\Benchmark\Metadata\Annotations\Revs;
+use Sentry\Tracing\Span;
+use Sentry\Tracing\TransactionContext;
+
+class SpanBench
```
ste93cry, line 12: Can this class be made `final` or is there some reason for not doing it (I never used `PhpBench`)?

.github/workflows/ci.yaml
```diff
@@ -75,3 +72,7 @@ jobs:
         uses: codecov/codecov-action@v1
         with:
           file: build/coverage-report.xml
+
+      - name: Check benchmarks
+        run: vendor/bin/phpbench run --revs=1 --iterations=1
```
ste93cry, line 77: If we don't want to run the bench except for ensuring that it still works (not sure why it should not though, even in the future), wouldn't make more sense to have a separate workflow that can run every now and then?

src/Tracing/Span.php
```diff
@@ -80,25 +80,17 @@ class Span
...
-            $this->spanId = $context->getSpanId();
+            return;
         }
 
-        if (null !== $context->getStartTimestamp()) {
-            $this->startTimestamp = $context->getStartTimestamp();
-        }
+        $this->traceId = $context->getTraceId() ?? TraceId::generate();
+        $this->spanId = $context->getSpanId() ?? SpanId::generate();
+        $this->startTimestamp = $context->getStartTimestamp() ?? microtime(true);
```
ste93cry, line 94: Please remove this blank line, it doesn't make sense

```suggestion

```

ste93cry (changes requested):

composer.json
Ocramius: This has no effect on consumers

.github/workflows/ci.yaml
Ocramius: > not sure why it should not though, even in the future

All code that is not run regularly will rot and break

composer.json
ste93cry: I know, but still we want to track it and not let it slip in the codebase as part of another PR

Jean85: Already done in master in 8fea3e3

tests/Benchmark/SpanBench.php
Jean85: I'm not sure we can. It may give space for optimizing out the instantiation of the class, making the bench unuseful.

.github/workflows/ci.yaml
Jean85: > wouldn't make more sense to have a separate workflow that can run every now and then?

That would create a great overhead: separate CI job, separate setup, separate `composer install`... And for a step that [runs in less than a quarter of a second](https://github.com/getsentry/sentry-php/runs/4777675503?check_suite_focus=true#step:12:10), it seems an overkill.

tests/Benchmark/SpanBench.php
ste93cry: I doubt that there is this kind of optimization as PHP cannot know what the constructor of such class do (it may set some global variables for example). Anyway, if you feel safer, let's keep it as is

.github/workflows/ci.yaml
ste93cry: My suggestion was not related to the time spent running the step, but rather to the fact that if this step is here just to ensure that it runs, then it doesn't make sense to have it as part of the CI workflow

Ocramius: > then it doesn't make sense to have it as part of the CI workflow

It's to make sure that upgrades to PHP, dependencies, etc. do not break it.

If we don't have it in CI to prevent rot, then having the benchmarks as part of the repository doesn't make sense either, as they will just become abandonware in a random directory ;-)

ste93cry: With "CI workflow" I meant the workflow that runs for every push or for every PR. Do you think that running it once per week (just as an example) would still not be enough?

Ocramius: :shrug: it's 0.25s, just let it run 

If it was a long process (that takes 30~40 minutes) then I understand running it on weekly basis, but running with every PR (especially when considering @dependabot) is necessary

ste93cry: Ok then, let's keep it as-is

ste93cry (approved):

PR #1278, opened 2022-01-15
Change withScope to return callback's return value

HypeMC:
This PR supersedes #1263.

I've noticed that the previous PR has been inactive for some time, so I was hoping it'd be okay if I tried to finish it. I've applied the suggested changes from the previous PR & updated some tests.

All credits go to @vaibhavpandeyvpz

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,9 @@
 
 ## Unreleased
 
-- Update the Guzzle tracing middleware to meet the [expected standard](https://develop.sentry.dev/sdk/features/#http-client-integrations) (#1234)
+- Update Guzzle tracing middleware to meet the [expected standard](https://develop.sentry.dev/sdk/features/#http-client-integrations) (#1234)
 - Add `toArray` public method in `PayloadSerializer` to be able to re-use Event serialization
+- The `withScope` methods now return the callback's return value (#1263)
 
 ## 3.3.5 (2021-12-27)
 
--- a/src/functions.php
+++ b/src/functions.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Sentry;
 
+use Sentry\State\Scope;
 use Sentry\Tracing\Transaction;
 use Sentry\Tracing\TransactionContext;
 
@@ -91,10 +92,18 @@ function configureScope(callable $callback): void
  * is automatically removed once the operation finishes or throws.
  *
  * @param callable $callback The callback to be executed
+ *
+ * @return mixed|void The callback's return value, upon successful execution
+ *
+ * @psalm-template T
+ *
+ * @psalm-param callable(Scope): T $callback
+ *
+ * @psalm-return T
  */
-function withScope(callable $callback): void
+function withScope(callable $callback)
 {
-    SentrySdk::getCurrentHub()->withScope($callback);
+    return SentrySdk::getCurrentHub()->withScope($callback);
 }
 
 /**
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -2,8 +2,9 @@
 
 ## Unreleased
 
-- Update the Guzzle tracing middleware to meet the [expected standard](https://develop.sentry.dev/sdk/features/#http-client-integrations) (#1234)
+- Update Guzzle tracing middleware to meet the [expected standard](https://develop.sentry.dev/sdk/features/#http-client-integrations) (#1234)
 - Add `toArray` public method in `PayloadSerializer` to be able to re-use Event serialization
+- The `withScope` methods now return the callback's return value (#1263)
```
Jean85, line 7:
```suggestion
- The `withScope` methods now returns the callback's return value (#1263)
```

Jean85 (changes requested):

tests/State/HubTest.php
```diff
@@ -135,6 +135,30 @@ public function testPopScope(): void
...
+            $this->assertNotSame($scope, $scopeArg);
+
+            return 'foobarbaz';
+        });
+
+        $this->assertSame('foobarbaz', $callbackReturn);
+
+        $callbackInvoked = false;
+
+        $hub->configureScope(function (Scope $scopeArg) use (&$callbackInvoked, $scope): void {
```
ste93cry, line 152: Why are we testing `configureScope()` here? It looks like this code is duplicate from `testConfigureScope()`

tests/State/HubTest.php
```diff
@@ -135,6 +135,30 @@ public function testPopScope(): void
     }
 
     public function testWithScope(): void
+    {
+        $scope = new Scope();
+        $hub = new Hub($this->createMock(ClientInterface::class), $scope);
```
ste93cry, line 140: Since we don't care about the client we can pass `nulll`

src/functions.php
```diff
@@ -91,10 +92,16 @@ function configureScope(callable $callback): void
  * is automatically removed once the operation finishes or throws.
  *
  * @param callable $callback The callback to be executed
+ *
+ * @return mixed|void The callback's return value, upon successful execution
+ *
+ * @phpstan-template T
```
ste93cry, line 98: To keep consistency with what it was done in the rest of the codebase, can you please put a blank line between each `@phpstan-*` annotation? Also, please use `@psalm-*` annotations

ste93cry (changes requested):

CHANGELOG.md
HypeMC: @Jean85 Methods is plural since there are three implementations, so return should be singular, but I can change it if you want.

tests/State/HubTest.php
HypeMC: @ste93cry Yeah, I thought so too, but most test cases have this. If I understood correctly, this checks that the scope gets popped.

src/functions.php
HypeMC: Done

tests/State/HubTest.php
HypeMC: Done

tests/State/HubTest.php
ste93cry: Right, I didn't think about it

ste93cry (approved):
Code looks fine for me, great work to both @HypeMC and @vaibhavpandeyvpz

CHANGELOG.md
Jean85: No you're right.

Jean85 (approved):

PR #1281, opened 2022-01-17
Swap order of Retry and Error http client plugins

### LINKED ISSUE #1204: Sentry ratelimiting increases application response time to at least 3.5 seconds
### Environment

PHP 7.4.13
Sentry SaaS (sentry.io)

We are using an in-house framework, so no laravel or symphony and we don't use promises in the application itself (not sure if that's relevant but I thought I'd mention it).

Which SDK and version?

`$ composer show | grep sentry`
```
sentry/sdk                           3.1.0     This is a metapackage shipping sentry/sentry with a recommended HTTP client.
sentry/sentry                        3.1.5     A PHP SDK for Sentry (http://sentry.io)
```

### Steps to Reproduce

For debugging purposes we wrote a small python script that responds with a http 429 code to emulate being rate limited by sentry locally. We then made sure that in the `HttpTransport` we pointed to the local server that returns a 429 instead of the sentry store api endpoint. From there we were able to debug what happens when sentry rate limits our servers. Please let me know if you think we did something wrong in this debugging setup.

While stepping through the code we found that when sentry sends a 429 indicating we are rate limited the sentry SDK with the recommended HTTP client will retry the request 3 times. Each time it does a `usleep()` with a time that comes from `RetryPlugin::defaultDelay` in `vendor/php-http/client-common/src/Plugin/RetryPlugin.php`. This call to `usleep` causes the application to wait for a total of 3.5 seconds doing absolutely nothing, including not sending finishing a response while sentry is doing it's thing.

```
#!/usr/bin/env python
import string,cgi,time, datetime
from http.server import BaseHTTPRequestHandler, HTTPServer

class MyHandler(BaseHTTPRequestHandler):
  def do_GET(self):
    self.send_response(503) # let bots know whats up
    self.send_header('Content-type','text/html')
    self.send_header('Retry-After','0')
    self.end_headers()
    self.send_error(429)

class TimeoutingHTTPServer(HTTPServer):
  def finish_request(self, request, client_address):
    request.settimeout(5) # Really short timeout as there is only 1 thread
    HTTPServer.finish_request(self, request, client_address)

def main():
  try:
    server = TimeoutingHTTPServer(('localhost',83), MyHandler)
    server.serve_forever()
  except KeyboardInterrupt:
    server.socket.close()

if __name__ == '__main__':
  main()
```

### Expected Result

We expected sentry to handle the sending of events on a shutdown handler, as is indicated in the doc comment on `HttpTransport`: This transport sends the events using a syncronous HTTP client that will delay sending of the requests until the shutdown of the application.

### Actual Result

Last time we encountered this rate limiting our application slowed down significantly because each backend request took at least 3.5 seconds to complete.

blackwolf12333:
The ErrorPlugin is responsible for filtering out requests that returned http
error codes. In the original order this filtering would happen only after a
failing request to the sentry endpoint was retried some number of times, as
configured by `getSendAttempts`.

This meant that when sentry rate limits the requests from this client the
requests would be retried before resulting in a ClientErrorException.

This patch reorders the ErrorPlugin and the RetryPlugin so the
ClientErrorException is thrown before the sdk attempts to retry the request.

Fixes #1204 

Some extra context: this only happens when you have `php-http/client-common`
version 1.x installed.

### DIFF
--- a/src/HttpClient/HttpClientFactory.php
+++ b/src/HttpClient/HttpClientFactory.php
@@ -147,8 +147,8 @@ final class HttpClientFactory implements HttpClientFactoryInterface
         $httpClientPlugins = [
             new HeaderSetPlugin(['User-Agent' => $this->sdkIdentifier . '/' . $this->sdkVersion]),
             new AuthenticationPlugin(new SentryAuthentication($options, $this->sdkIdentifier, $this->sdkVersion)),
-            new RetryPlugin(['retries' => $options->getSendAttempts()]),
             new ErrorPlugin(),
+            new RetryPlugin(['retries' => $options->getSendAttempts()]),
         ];
 
         if ($options->isCompressionEnabled()) {


Jean85 (changes requested):
Please add a test (if possible) and a changelog entry.

ste93cry:
As mentioned in the issue, this is not the right solution: the order of the plugins matter, so swapping them will make `php-http/client-common:<2.0` work as expected and will break the app of all people using newer versions. Instead, we should bump the minimum required version of the package to ensure that noone can use one not behaving as expected

Jean85:
> As mentioned in the issue, this is not the right solution: the order of the plugins matter, so swapping them will make `php-http/client-common:<2.0` work as expected and will break the app of all people using newer versions. Instead, we should bump the minimum required version of the package to ensure that noone can use one not behaving as expected

That should be done targeting `develop` though.

ste93cry:
I'm not sure, the bump is to solve an issue and as such it falls under a patch version

Jean85:
Yes but, if you have issues to fix on the previous patch, there's no way to fix them for people stuck on `http/client-common` 1.x

IMHO we could even do the reverse on the old minor, restricting to `http/client-common` `^1.x` and fixing it there as suggested here.

WDYT?

blackwolf12333:
Yeah I agree with @Jean85 here, I made this before the discussion on the issue happened. I don't really expect this to break people who use `client-common >= 2` but I can't be sure. Anyway, I'll close this PR until we have a good solution.

PR #1282, opened 2022-01-19
Support http_connect_timeout and http_timeout options

matthew-gill:
I created this issue in the Symfony repo: https://github.com/getsentry/sentry-symfony/issues/592
This PR modifies the `HttpClientFactory` to allow configuration of the http (connect) timeout values.
If no option is set, it will use the default.

If I've missed the mark, feel free to reject - I'd just love to be able to do this in my app, if it's possible already please let me know, thanks!

### DIFF
--- a/src/HttpClient/HttpClientFactory.php
+++ b/src/HttpClient/HttpClientFactory.php
@@ -123,7 +123,10 @@ final class HttpClientFactory implements HttpClientFactoryInterface
     {
         if (class_exists(SymfonyHttplugClient::class)) {
             $symfonyConfig = [
-                'max_duration' => $options->getHttpTimeout(),
+                /*
+                 * {@see https://symfony.com/doc/current/http_client.html#dealing-with-network-timeouts}
+                 */
+                'max_duration' => $options->getHttpConnectTimeout() + $options->getHttpTimeout(),
             ];
 
             if (null !== $options->getHttpProxy()) {
--- a/src/HttpClient/HttpClientFactory.php
+++ b/src/HttpClient/HttpClientFactory.php
@@ -34,13 +34,13 @@ final class HttpClientFactory implements HttpClientFactoryInterface
     /**
      * @var int The timeout of the request in seconds
      */
-    private const DEFAULT_HTTP_TIMEOUT = 5;
+    public const DEFAULT_HTTP_TIMEOUT = 5;
 
     /**
      * @var int The default number of seconds to wait while trying to connect
      *          to a server
      */
-    private const DEFAULT_HTTP_CONNECT_TIMEOUT = 2;
+    public const DEFAULT_HTTP_CONNECT_TIMEOUT = 2;
 
     /**
      * @var StreamFactoryInterface The PSR-17 stream factory
@@ -123,7 +123,7 @@ final class HttpClientFactory implements HttpClientFactoryInterface
     {
         if (class_exists(SymfonyHttplugClient::class)) {
             $symfonyConfig = [
-                'max_duration' => self::DEFAULT_HTTP_TIMEOUT,
+                'max_duration' => $options->getHttpTimeout(),
             ];
 
             if (null !== $options->getHttpProxy()) {
@@ -135,8 +135,8 @@ final class HttpClientFactory implements HttpClientFactoryInterface
 
         if (class_exists(GuzzleHttpClient::class)) {
             $guzzleConfig = [
-                GuzzleHttpClientOptions::TIMEOUT => self::DEFAULT_HTTP_TIMEOUT,
-                GuzzleHttpClientOptions::CONNECT_TIMEOUT => self::DEFAULT_HTTP_CONNECT_TIMEOUT,
+                GuzzleHttpClientOptions::TIMEOUT => $options->getHttpTimeout(),
+                GuzzleHttpClientOptions::CONNECT_TIMEOUT => $options->getHttpConnectTimeout(),
             ];
 
             if (null !== $options->getHttpProxy()) {
@@ -148,8 +148,8 @@ final class HttpClientFactory implements HttpClientFactoryInterface
 
         if (class_exists(CurlHttpClient::class)) {
             $curlConfig = [
-                \CURLOPT_TIMEOUT => self::DEFAULT_HTTP_TIMEOUT,
-                \CURLOPT_CONNECTTIMEOUT => self::DEFAULT_HTTP_CONNECT_TIMEOUT,
+                \CURLOPT_TIMEOUT => $options->getHttpTimeout(),
+                \CURLOPT_CONNECTTIMEOUT => $options->getHttpConnectTimeout(),
             ];
 
             if (null !== $options->getHttpProxy()) {
--- a/src/HttpClient/HttpClientFactory.php
+++ b/src/HttpClient/HttpClientFactory.php
@@ -32,17 +32,6 @@ use Symfony\Component\HttpClient\HttplugClient as SymfonyHttplugClient;
 final class HttpClientFactory implements HttpClientFactoryInterface
 {
     /**
-     * @var int The timeout of the request in seconds
-     */
-    private const DEFAULT_HTTP_TIMEOUT = 5;
-
-    /**
-     * @var int The default number of seconds to wait while trying to connect
-     *          to a server
-     */
-    private const DEFAULT_HTTP_CONNECT_TIMEOUT = 2;
-
-    /**
      * @var StreamFactoryInterface The PSR-17 stream factory
      */
     private $streamFactory;
@@ -121,13 +110,10 @@ final class HttpClientFactory implements HttpClientFactoryInterface
      */
     private function resolveClient(Options $options)
     {
-        $httpTimeout = $options->getHttpTimeout() ?? self::DEFAULT_HTTP_TIMEOUT;
-        $httpConnectTimeout = $options->getHttpConnectTimeout() ?? self::DEFAULT_HTTP_CONNECT_TIMEOUT;
-
         if (class_exists(SymfonyHttplugClient::class)) {
             $symfonyConfig = [
-                'timeout' => $httpConnectTimeout,
-                'max_duration' => $httpTimeout,
+                'timeout' => $options->getHttpConnectTimeout(),
+                'max_duration' => $options->getHttpTimeout(),
             ];
 
             if (null !== $options->getHttpProxy()) {
@@ -139,8 +125,8 @@ final class HttpClientFactory implements HttpClientFactoryInterface
 
         if (class_exists(GuzzleHttpClient::class)) {
             $guzzleConfig = [
-                GuzzleHttpClientOptions::TIMEOUT => $httpTimeout,
-                GuzzleHttpClientOptions::CONNECT_TIMEOUT => $httpConnectTimeout,
+                GuzzleHttpClientOptions::TIMEOUT => $options->getHttpTimeout(),
+                GuzzleHttpClientOptions::CONNECT_TIMEOUT => $options->getHttpConnectTimeout(),
             ];
 
             if (null !== $options->getHttpProxy()) {
@@ -152,8 +138,8 @@ final class HttpClientFactory implements HttpClientFactoryInterface
 
         if (class_exists(CurlHttpClient::class)) {
             $curlConfig = [
-                \CURLOPT_TIMEOUT => $httpTimeout,
-                \CURLOPT_CONNECTTIMEOUT => $httpConnectTimeout,
+                \CURLOPT_TIMEOUT => $options->getHttpTimeout(),
+                \CURLOPT_CONNECTTIMEOUT => $options->getHttpConnectTimeout(),
             ];
 
             if (null !== $options->getHttpProxy()) {
--- a/src/HttpClient/HttpClientFactory.php
+++ b/src/HttpClient/HttpClientFactory.php
@@ -126,7 +126,8 @@ final class HttpClientFactory implements HttpClientFactoryInterface
                 /*
                  * {@see https://symfony.com/doc/current/http_client.html#dealing-with-network-timeouts}
                  */
-                'max_duration' => $options->getHttpConnectTimeout() + $options->getHttpTimeout(),
+                'max_duration' => $options->getHttpConnectTimeout() ?? HttpClientFactory::DEFAULT_HTTP_CONNECT_TIMEOUT
+                    + $options->getHttpTimeout() ?? HttpClientFactory::DEFAULT_HTTP_TIMEOUT,
             ];
 
             if (null !== $options->getHttpProxy()) {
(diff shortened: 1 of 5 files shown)


src/HttpClient/HttpClientFactory.php
```diff
@@ -103,7 +103,7 @@ public function create(Options $options): HttpAsyncClientInterface
         if (null === $httpClient) {
             if (class_exists(SymfonyHttplugClient::class)) {
                 $symfonyConfig = [
-                    'max_duration' => self::DEFAULT_HTTP_TIMEOUT,
+                    'max_duration' => $options->getHttpTimeout(),
                 ];
```
Jean85, line 107: That's strange... The Symfony client has no connect timeout setting? Does this mean that it conflates the two values? Should we sum the two values from the config here? Should we leave it like that?

src/Options.php
```diff
@@ -729,6 +771,8 @@ private function configureOptions(OptionsResolver $resolver): void
             'capture_silenced_errors' => false,
             'max_request_body_size' => 'medium',
             'class_serializers' => [],
+            'http_connect_timeout' => HttpClientFactory::DEFAULT_HTTP_CONNECT_TIMEOUT,
+            'http_timeout' => HttpClientFactory::DEFAULT_HTTP_TIMEOUT,
```
Jean85, line 775: We can't put the defaults here, the default should be null. That's because otherwise you can't handle the situation where you don't have any concrete client available, but you have the option set, like with the HTTP proxy option.

Jean85, line 775: Also, I'm not sure about the option names. We should check if there's an unified name already used in other Sentry SDKs.

Jean85 (changes requested):
Apart from the explicit comments, this PR requires targeting develop, a changelog entry and some tests.

src/HttpClient/HttpClientFactory.php
matthew-gill: Agreed. I had a look through the Symfony documentation: https://symfony.com/doc/current/http_client.html#dealing-with-network-timeouts

This mentions both the `max_duration` and `timeout`, from what I can gather, `max_duration` is a governing variable, so summing, while it feels a bit weird is probably the right way to go. That's done in `e5f9627`.

src/Options.php
matthew-gill: I had a look at the python SDK and the only references I could find where `timeout`, here we're being a bit more explicit, happy to change it if you like.

Jean85: You'll need coalescing, yes. Because `null` is a valid value that indicates that the user hasn't provided any value.
This is needed due to the fact that you need to implement a check like this one: https://github.com/getsentry/sentry-php/blob/32e5415803ff0349ccb5e5b5e77b016320762786/src/HttpClient/HttpClientFactory.php#L138

matthew-gill:
Thanks @Jean85 for the suggestions/comments. I've made most of them.
Re tests, I was only able to write tests for the `Options` part.

I tried to do them for the `HttpClientFactory` but as the test harness doesn't have any _real_ clients e.g. curl client or guzzle, I wasn't able to write tests without including the additional dependencies... they all use `class_exists` within `HttpClientFactory::resolveClient`

matthew-gill:
I tried to write some tests by using `class_alias` in PHP (not something I've ever used before)! But in order to get it to work with the new `PluginClient` pieces, I'd either have to use reflection or add a getter to that method, all of which I think feel a bit wrong for this case.

src/HttpClient/HttpClientFactory.php
```diff
@@ -123,7 +123,11 @@ private function resolveClient(Options $options)
     {
         if (class_exists(SymfonyHttplugClient::class)) {
             $symfonyConfig = [
-                'max_duration' => self::DEFAULT_HTTP_TIMEOUT,
+                /*
+                 * {@see https://symfony.com/doc/current/http_client.html#dealing-with-network-timeouts}
+                 */
+                'max_duration' => $options->getHttpConnectTimeout() ?? HttpClientFactory::DEFAULT_HTTP_CONNECT_TIMEOUT
```
Jean85, line 129: PHPStan says that `getHttpConnectTimeout` is not nullable, but it is. Put it in the baseline with `phpstan --generate-baseline`, it's an issue on all the getters, we should fix it somehow, maybe with array shapes.

Also, if you want to avoid repeating the coalesce every time, I think it's fine to have a temporary variable to reuse on each case.

Jean85 (changes requested):
Code seems fine to me, just a small comment and fixing PHPStan baseline is required.

src/HttpClient/HttpClientFactory.php
matthew-gill: Thanks, all done!

src/HttpClient/HttpClientFactory.php
```diff
@@ -121,9 +121,15 @@ public function create(Options $options): HttpAsyncClientInterface
...
+        $httpTimeout = $options->getHttpTimeout() ?? self::DEFAULT_HTTP_TIMEOUT;
+        $httpConnectTimeout = $options->getHttpConnectTimeout() ?? self::DEFAULT_HTTP_CONNECT_TIMEOUT;
+
         if (class_exists(SymfonyHttplugClient::class)) {
             $symfonyConfig = [
-                'max_duration' => self::DEFAULT_HTTP_TIMEOUT,
+                /*
+                 * {@see https://symfony.com/doc/current/http_client.html#dealing-with-network-timeouts}
+                 */
+                'max_duration' => $httpConnectTimeout + $httpTimeout,
```
Jean85, line 132: @ste93cry do you agree with this change?

Jean85 (changes requested):
Looks fine to me :+1: great work!

In the meantime, I've checked with @bruno-garcia about the options naming, and he said that there's no standard, just Java SDK having sub-optimal names: 
https://github.com/getsentry/sentry-java/blob/01cda06a94805f9940229e07ec7f31747f70474c/sentry/src/main/java/io/sentry/SentryOptions.java
https://github.com/getsentry/sentry-java/blob/01cda06a94805f9940229e07ec7f31747f70474c/sentry/src/main/java/io/sentry/SentryOptions.java

Jean85:
For the CS issues, use `make cs-fix` to fix that automatically.

src/HttpClient/HttpClientFactory.php
ste93cry: We're changing a default here, so technically this is a breaking change. Other than that, the value is almost doubled so while I agree that if the underlying lib is not supporting splitted options this is the only way, I would leave this as it was before

src/Options.php
```diff
@@ -686,6 +686,46 @@ public function setTracesSampler(?callable $sampler): void
...
+    /**
+     * Gets the amount of time in seconds that http connections can take to connect.
+     */
+    public function getHttpConnectTimeout(): ?int
+    {
+        return $this->options['http_connect_timeout'];
+    }
+
+    /**
+     * Gets the amount of time in seconds that http connections can take to connect.
```
ste93cry, line 698:
```suggestion
     * Sets the amount of time in seconds that http connections can take to connect.
```

src/Options.php
```diff
@@ -686,6 +686,46 @@ public function setTracesSampler(?callable $sampler): void
...
+    /**
+     * Gets the amount of time in seconds to wait for the Sentry server to respond.
+     */
+    public function getHttpTimeout(): ?int
+    {
+        return $this->options['http_timeout'];
+    }
+
+    /**
+     * Gets the amount of time in seconds to wait for the Sentry server to respond.
```
ste93cry, line 718:
```suggestion
     * Sets the amount of time in seconds to wait for the Sentry server to respond.
```

ste93cry (changes requested):

src/Options.php
ste93cry: While the two options has been made `nullable`, the if above has not been changed to take this fact into account

src/HttpClient/HttpClientFactory.php
Jean85: Ok so we need this:
```suggestion
                'max_duration' => $httpTimeout,
```

mfb: I think it would be a good idea to set the SymfonyHttplugClient `timeout` option to `$httpConnectTimeout` here.

It's a bit confusing, because the Symfony HttpClient `timeout` option applies to any idle time, whether while connecting or while waiting for the response.  But despite the semantic difference, it's useful because setting this to e.g. the 2-second default for connect timeout allows the HttpClient to give up more quickly if there is no connection after 2 seconds, while still having 5 seconds max-duration configured for the full request and response.

Adding the `timeout` setting does cause an additional timeout to apply if there is a 2-second pause in data arriving during the response phase, but that seems fine to me and totally compatible with the default 5-second max duration (if you have huge network latency, then increasing the HttpConnectTimeout would give you more time for both connecting and waiting for the response).

In my testing w/ Sentry SDK + Symfony HttpClient, setting `timeout` to 2 seconds allowed Sentry SDK to give up several seconds more quickly when trying to connect to an unreachable Sentry server (and no side effects with working Sentry server).

Jean85: Thanks for digging through this @mfb! So your suggestion is:
```suggestion
                'timeout' => $httpConnectTimeout,
```

mfb: I was testing this actually:
```
                    'max_duration' => self::DEFAULT_HTTP_TIMEOUT,
                    'timeout' => self::DEFAULT_HTTP_CONNECT_TIMEOUT,
```
I think it makes sense to have 5 second end-to-end max duration, as we do now, and within that, a 2 second timeout for idle/connect time.  It seems like the two settings are less useful and more confusing if they get added together.

mfb: By the way I was playing with very low values of max_duration and timeout, and noticed that you can make the SDK send the event successfully but also give up before receiving the response and believe the send failed, which is actually great from a performance perspective; of course this would also increase the rate of actual failures assuming you have sporadic latency :)

mfb:
Looks like this didn't make it into 3.4.0 - but hopefully could make it into 3.5.0?

matthew-gill:
Sorry for the radio silence on this, we've decided to use relay to alleviate our concerns. I'm afraid I don't have the bandwidth to pick this up at the moment, any volunteers to see this through would be greatly appreciated 🙏

ste93cry:
>hopefully could make it into 3.5.0

Sure, why not. I will try to find some time in the next days/weeks to finish the work. In the meanwhile, thank you @matthew-gill for having done most of it

src/HttpClient/HttpClientFactory.php
rabauss: By the way the symfony docs say `Passing 0 as timeout allows monitoring responses in a non-blocking way.`
Or is there another way to configure the sentry SDK asynchronously?

rabauss:
How can we support this PR?

src/HttpClient/HttpClientFactory.php
ste93cry: >I think it would be a good idea to set the SymfonyHttpClient timeout option to $httpConnectTimeout here.

I updated the PR accordingly 👍

>Or is there another way to configure the sentry SDK asynchronously?

I think that this applies if you use something along the lines of `$client->stream()` to stream the responses, which is not the case with the transport we provide out-of-the-box

src/Options.php
ste93cry: @Jean85 I think we should re-evaluate the fact that these two new options are `nullable`. It doesn't really make sense, as the timeout is a concept that generally speaking applies to any transport, be it HTTP or whatever else, and so they should always be set to a sensible value by default. Moreover, in retrospective, I'm starting to believe that there isn't a real benefit in throwing an error for cases like this one, and we may just silently ignore the error or document the behaviour rather than breaking the application at runtime. WDYT?

mfb:
I'm curious what is the rationale of having http_timeout and http_connect_timeout options when the underlying Symfony HTTP client used by the SDK has max_duration and timeout options, which have different meanings.  Wouldn't just having the same options be easier to grok?

ste93cry:
>the underlying Symfony HTTP client used by the SDK

The core SDK has no binding to any specific HTTP client, or at least we try to not have it... We provide some code to ease the intialization of the HTTP transport using some common clients, but this doesn't mean that we want to use them specifically besides that

Jean85 (changes requested):
I discussed this privately with @ste93cry, but I'm still of the opinion that we should make these new options nullable and throw when we can't apply them, otherwise we could silently ignore them, and confuse the user.

So, I would change the check into something like:
```php
if (
    null !== $options->getHttpProxy()
    || null !== $options->getHttpConnectTimeout()
    || null !== $options->getHttpTimeout()
    ) {
            throw new \RuntimeException('The "http_connect_timeout", "http_timeout" and "http_proxy" options are natively supported ony with "php-http/curl-client", "symfony/http-client" and "php-http/guzzle6-adapter"; either use one of them, or switch to build your HTTP transport manually.');
}
```

ste93cry:
>we should make these new options nullable and throw when we can't apply them, otherwise we could silently ignore them, and confuse the user.

While it's true that we could potentially confuse the user, I think that documenting the fact that these options could not apply to all HTTP clients would be enough. Making them `nullable` would confuse even more, as the user would not know what `null` means: does it means that a default applies, or that no timeout is set at all? Moreover, making the `Options::getXXX()` methods nullable just for the sake of checking elsewhere if those options apply to a particular client feels a code smell

Jean85:
Well, I think we need a tie breaker here... @stayallive? @smeubank? @HazAT?

stayallive:
I think that making them nullable makes the most sense to be honest, because:

1. It's confusing when we have options that we can't apply, especially since they can have performance implications, so the user should know they are doing something that has no effect
2. The argument that making them nullable is confusing could be true, but we can easily document what `null` means here in the documentation, most people will never even configure these values anyway so having them nullable makes more sense I think. We've always had "hidden" default timeouts set so this is a great way to communicate that those are used and we can document where to find those values
3. Let's not try to get too hung up on perfectly architected code (whatever that definition is for you), in the end this client must report (potentially critical) errors to somewhere and must do it so with as little impact on the application as possible. It's goal is not to have perfectly architected code if that conflicts with the first 2 "goals" IMHO

HazAT:
I am with @stayallive. Let's just document what `null`  does (or doesn't do, meaning doesn't apply to all transports)
I wouldn't even throw an exception; if the http_client supports it, great; if it doesn't and you configured something manually, you probably anyway know what you are doing.

So I would say let's merge it.

More importantly, for me would be the change `send_attempts` to 0, not in this PR but referring to this comment https://github.com/getsentry/sentry-php/issues/1311#issuecomment-1119941034

ste93cry:
>I wouldn't even throw an exception; if the http_client supports it, great; if it doesn't and you configured something manually, you probably anyway know what you are doing.

If we don't want to throw the exception, it makes little sense to have those options nullable. The whole argument in making them like so is because we want to throw an exception when the option could not apply to a transport, but as you said it makes little sense

>More importantly, for me would be the change `send_attempts` to 0

This is a breaking change, people don't expect a default to change between minor versions

stayallive:
There is no need to change `send_attempts` in this PR. 

We do that separately and include that in a new major release of course instead of a minor.

ste93cry:
I didn't mean that we need to change `send_attempts` in this PR, it was just a comment answering a side-topic. The point is that the only reason we want to make the `http_*_timeout` options nullable is that we need to do it to be able to identify if the user set that option and throw an exception if he did and we can't autoconfigure the HTTP client. Since @HazAT agrees that it doesn't make sense to throw an error in that case, I cannot really understand why we want to keep going this way

stayallive:
Don't think we are, as @HazAT mentioned we are going to document how they work but not going to add the exception in that case the defaults can live in the options 👍 

This is how the PR currently is, I don't think there is anything to do, good to merge?

src/Options.php
stayallive: This has been concluded to not have the options be nullable or throw exception.

> https://github.com/getsentry/sentry-php/pull/1282#issuecomment-1123322201

stayallive (approved):

ste93cry (approved):

PR #1291, opened 2022-02-23
Support handling of the server rate limits

ste93cry:
As per title, this PR adds support for handling the rate limits when sending events to the Sentry server. There are obvious limitations of this implementation due to the fact that the status of the rate limits does not persist across handling multiple server requests, but at least within the lifecycle of a single request issues like keeping flooding the server should now be avoided

### DIFF
--- a/src/Transport/HttpTransport.php
+++ b/src/Transport/HttpTransport.php
@@ -59,6 +59,11 @@ final class HttpTransport implements TransportInterface
     private $logger;
 
     /**
+     * @var RateLimiter The rate limiter
+     */
+    private $rateLimiter;
+
+    /**
      * Constructor.
      *
      * @param Options                    $options           The Sentry client configuration
@@ -82,6 +87,7 @@ final class HttpTransport implements TransportInterface
         $this->requestFactory = $requestFactory;
         $this->payloadSerializer = $payloadSerializer;
         $this->logger = $logger ?? new NullLogger();
+        $this->rateLimiter = new RateLimiter($this->logger);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


tests/Transport/HttpTransportTest.php
```diff
@@ -250,6 +251,80 @@ public function testSendReturnsRejectedPromiseIfSendingFailedDueToHttpClientExce
...
+            $this->requestFactory,
+            $this->payloadSerializer,
+            $logger
+        );
+
+        // Event should be sent, but the server should reply with a HTTP 429
+        $promise = $transport->send($event);
+
+        try {
+            $promiseResult = $promise->wait();
```
Jean85, line 305:
```suggestion
            $promiseResult = $promise->wait();
            $this->fail('Promise should throw');
```

tests/Transport/HttpTransportTest.php
```diff
@@ -250,6 +251,80 @@ public function testSendReturnsRejectedPromiseIfSendingFailedDueToHttpClientExce
...
+
+        $this->assertSame(PromiseInterface::REJECTED, $promise->getState());
+        $this->assertSame(ResponseStatus::rateLimit(), $promiseResult->getStatus());
+        $this->assertSame($event, $promiseResult->getEvent());
+
+        // Event should not be sent at all because rate-limit is in effect
+        $promise = $transport->send($event);
+
+        try {
+            $promiseResult = $promise->wait();
```
Jean85, line 318:
```suggestion
            $promiseResult = $promise->wait();
            $this->fail('Promise should throw');
```

Jean85 (changes requested):

src/Transport/HttpTransport.php
```diff
@@ -95,7 +101,18 @@ public function send(Event $event): PromiseInterface
             throw new \RuntimeException(sprintf('The DSN option must be set to use the "%s" transport.', self::class));
         }
 
-        if (EventType::transaction() === $event->getType()) {
+        $eventType = $event->getType();
+
+        if ($this->rateLimiter->isRateLimited($eventType)) {
+            $this->logger->warning(
+                sprintf('Rate limit exceeded for sending requests of type "%s".', (string) $eventType),
+                ['event' => $event]
```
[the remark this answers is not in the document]
ste93cry: We already do this in all places where we log things, so I wouldn't be too much worried about this at this point

tests/Transport/HttpTransportTest.php
ste93cry: The asserts right after the `try`/`catch` block will catch the failure, there is no point to fail programmatically here

stayallive (approved):
Looks good to me! 💪

Jean85 (approved):

smeubank:
thanks for this @ste93cry ! 🚀

PR #1293, opened 2022-02-25
Add sentry-trace header to Guzzle

AAllport:
This PR allows developers to opt-in to attaching the sentry-trace header to requests made with this middleware

We use the same technique (by copy-pasting the middleware and adding the same changes) at work on one of our multi-service projects, and it works great!


Jean85 (changes requested):
Thank you for your contribution!
Can you please add a changelog entry and some tests?

I would consider this a new feature, so please target `develop`; do you agree @ste93cry?

AAllport:
Oh, I didn't realise there was an existing one.
Didn't see it in the SDK when we originally made the edits on our end a few months back

Happy to close this one off :)

PR #1294, opened 2022-02-27
Treat the project ID component of the DSN as a `string` rather than an `integer`

### LINKED ISSUE #1292: Project IDs should be strings
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io) or self-hosted/on-premise (which version?)

Which SDK and version?
PHP

### Steps to Reproduce

https://github.com/getsentry/sentry-php/blob/ea704bbe145657a507d73cd8bf6cc017d269fffa/src/Dsn.php#L102

### Expected Result

Project IDs should be cast as strings, not int

### Actual Result

Project IDs are int in from PHP sdks

we need project ID to not be forced to an integer because its plausible we will need it to be a UUID, or we will need it to be a 64 bit int (which breaks in JS, as an example)

ste93cry:
Fixes #1292. With this PR, the `Dsn::getProjectId()` method deprecates returning the project ID as an `integer`. In the next major version, we can reintroduce the return type and make it `string`. It looks like version `2.x` was correct, but this behaviour was lost when we introduced the `Dsn` class. I'm unsure whether we want to classify this as a bug or as a feature. In any case, we cannot avoid the code to handle both typehints as changing it would be a breaking change

### DIFF
--- a/src/Dsn.php
+++ b/src/Dsn.php
@@ -98,11 +98,6 @@ final class Dsn implements \Stringable
 
         $segmentPaths = explode('/', $parsedDsn['path']);
         $projectId = array_pop($segmentPaths);
-
-        if (!ctype_digit($projectId)) {
-            throw new \InvalidArgumentException('"%s" DSN must contain a valid project ID.');
-        }
-
         $lastSlashPosition = strrpos($parsedDsn['path'], '/');
         $path = $parsedDsn['path'];
 
--- a/src/Dsn.php
+++ b/src/Dsn.php
@@ -38,7 +38,7 @@ final class Dsn implements \Stringable
     private $secretKey;
 
     /**
-     * @var int The ID of the resource to access
+     * @var string The ID of the resource to access
      */
     private $projectId;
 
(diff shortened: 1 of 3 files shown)


Jean85 (approved):

src/Dsn.php
```diff
@@ -99,7 +99,7 @@ public static function createFromString(string $value): self
         $segmentPaths = explode('/', $parsedDsn['path']);
         $projectId = array_pop($segmentPaths);
 
-        if ((int) $projectId <= 0) {
+        if (!ctype_digit($projectId)) {
```
stayallive, line 102: This will still fail if the project ID is a non-number, so project ID cannot be a UUID for example with this change still.

Shouldn't we move this check to the getter when returning `int` because returning a non-number as `int` could cause unexpected results (which is probably why this check is still here)?

stayallive (changes requested):

src/Dsn.php
ste93cry: >This will still fail if the project ID is a non-number, so project ID cannot be a UUID for example with this change still.

You're right, but this is what other SDK still do, e.g. [here](https://github.com/getsentry/sentry-javascript/blob/f8aeeeab0006c9a0527fd8392bede8c534648325/packages/utils/src/dsn.ts#L50) and [here](https://github.com/getsentry/sentry-python/blob/b63b5ae32d254771cbd027009d0b6def7e2e07fb/sentry_sdk/utils.py#L188). I understand the concern that if you pass something that is not a number there will still be issues, but there are no public information that I'm aware of that the IDs will not be just digits also in the future

>returning a non-number as `int` could cause unexpected results

I wouldn't worry too much about the the cast in the getter because in the worst case the project ID will be `0`

stayallive: Check the original issue. It's stated that it needs to be opaque, and UUID is mentioned explicitly too 😬

ste93cry: Well, it's written that it could be an UUID or a 64bit number. I have no problems in removing this check, but since all other SDK seems to still require a number (stringified), I wonder if we should do it or not. @smeubank what do you suggest to do?

smeubank: @HazAT can i get a PHP sanity check here, so to comply with the original request it needs to be UUID or 64 bit int?

stayallive (approved):
👍

PR #1306, opened 2022-05-02
Fix PHPStan baseline

ste93cry:
Some recent release of PHPStan broke the build. With this PR I'm updating the baseline to keep ignoring the unresolvable errors

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -16,12 +16,12 @@ parameters:
 			path: src/ClientInterface.php
 
 		-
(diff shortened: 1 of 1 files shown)


Jean85 (approved):
LGTM :+1: 

Maybe we should discuss removing the caret from the PHPStan/Psalm constraints, to avoid breakages like this one...

ste93cry:
Yea, that's a possibility, but do we see a need for it? I mean, sometimes it happens that we have to update the baseline of course, but do we have to do that so many times that it justifies locking the version down to the patch level?

Jean85:
I we're ok with the amount of work required, I'm fine with leaving it as it is :+1:

Issue #1311, opened 2022-05-06
SDK performance negatively affects application if Sentry is offline.


bruno-garcia:
This came up on Twitter: https://twitter.com/JasnK/status/1522640769089249281

> Sentry PHP SDK 3.1.1 and the latest Sentry DSN shown in Client Keys.

Related issues: https://github.com/getsentry/mobile-team-meta/issues/3

mfb:
Being able to set the Symfony HTTP client max_duration and timeout options would help here - see PR at #1282

e.g. if there is no connection after _n_ seconds, the HTTP client's timeout would be reached and send could be aborted.

jason-klein:
Thank you Bruno! We are using Sentry PHP SDK 3.1.1 with Laravel 9.10.1 and PHP 8.0.18 on Apache 2.4.51 in a Docker Container hosted on AWS ECS/EC2.

We are using the latest Sentry DSN shown in Client Keys.  Our DSN is formatted like the following example:
`https://1234567890abcdef1234567890abcdef@o9999999.ingest.sentry.io/9999999`

**Summary**
The Sentry SDK had a negative impact on our web-based application performance. Our average page load times increased 30X-90X from 275-280 ms up to 8-25 seconds during the Sentry outage that began around 2022-05-06 10:00 UTC. When we realized Sentry was the issue and re-deployed our services with the Sentry SDK disabled, the problem immediately disappeared.

**Diagnosing Issue**
- External monitoring and our web load balancer both confirmed that TTFB (time to first byte) on our application servers had increased nearly 100x.
- Web traffic was normal. No unusual spikes in web requests or HTTP error codes.
- Server metrics were normal. (e.g. Database/Cache/Container CPU/Memory/Disk, etc)
- **Sentry** indicated that our PHP application response times were normal
- Cross-referencing the original request time in Apache logs with the HTTP response header time and TTFB in browser inspection confirmed that Apache was receiving the request immediately, but responding slowly.

The information above led us to believe there was a problem with a network-based service that our application depends on to process each request (e.g. DNS). We eventually determined Sentry was causing our issues, by keeping PHP requests running much longer than usual which depleted all available PHP workers across all of our Apache container instances. When a PHP request would finally timeout, that worker became available to service another request.

**Temporary Solution**
Disable Sentry SDK during Sentry outage

**Attempted Solutions**
[...]

PR #1312, opened 2022-05-11
Lower default `send_attempts` option to `0` and deprecate it

stayallive:
Following the discussion on #1311 we have decided to lower the default `send_attempts` to `0` effectively disabling retries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,7 @@
 
 ## Unreleased
 
-- Lower the default `send_attempts` option option to `0`, this disables retries. If you require retries you can increase the `send_attempts` options to the desired value (#1312)
-- Deprecate the `send_attempts` option (#1312)
+- Set the default `send_attempts` to `0` (this disables retries) and deprecate the option. If you require retries you can increase the `send_attempts` option to the desired value. (#1312)
 - Bump minimum version of `guzzlehttp/psr7` package to avoid [`CVE-2022-24775`](https://github.com/guzzle/psr7/security/advisories/GHSA-q7rv-6hp3-vh96) (#1305)
 
 ## 3.4.0 (2022-03-14)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Lower the default `send_attempts` option option to `0`, this disables retries. If you require retries you can increase the `send_attempts` options to the desired value (#1312)
+- Deprecate the `send_attempts` option (#1312)
```
ste93cry, line 6: Maybe it would be better to squash together these two lines? Otherwise, the user will first read that the default value was changed and then that the option has been deprecated and he may think that the two things are unrelated or he may be confused because it seems that they are in conflict

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Lower the default `send_attempts` option option to `0`, this disables retries. If you require retries you can increase the `send_attempts` options to the desired value (#1312)
```
ste93cry, line 5: There's a typo: the `option` word is repeated two times in a row

ste93cry (changes requested):

ste93cry (approved):

Jean85 (approved):
LGTM :+1: 

Since we will be dropping this in the next major, this means that we'll be dropping the `RetryPlugin`? Maybe we should add some way to easily add plugins as needed.

ste93cry:
>this means that we'll be dropping the `RetryPlugin`

Yes, indeed.

>Maybe we should add some way to easily add plugins as needed.

Yep, I think this has come out as a feature request in the past. Or, we could simply ditch HTTPlug. With PSR-18 I don't really see anymore the need for it

Jean85:
> Yep, I think this has come out as a feature request in the past. Or, we could simply ditch HTTPlug. With PSR-18 I don't really see anymore the need for it

That would require a change on how the default approach works... Can you open a dedicated issue so we can discuss that widely?

Issue #1314, opened 2022-05-12
(PHP 8) Stacktrace with frames referring to anonymous classes are not grouped together


toemat:
Hi! I found [this issue](https://github.com/getsentry/sentry-php/issues/955) while trying to figure out why my events have not been grouped lately. I think PHP 8 changed the syntax of these anonymous class lines from `{file}0x{mem}` to `{file}:{line}${mem}`

PHP 7.x: `class@anonymous/home/test.php0x7f76c7628065`
PHP 8.x: `class@anonymous/home/test.php:3$3e0a7`

This seems to be the regex to remove the 0x style address, maybe adjust to match either 0x{hex} OR ${hex}?

https://github.com/getsentry/sentry-php/blob/64b9696d965994e0bdbd35da8e6de90bc39a269f/src/FrameBuilder.php#L79

So `/0x[a-fA-F0-9]+$/` to `/(\$|0x)[a-fA-F0-9]+$/`

Does it seem like I'm on the right track? I can open a PR if that's best.

ste93cry:
>Does it seem like I'm on the right track?

You are, indeed. I found that the change of the string format was done in [`php/php-src#0f2cdbf`](https://github.com/php/php-src/commit/0f2cdbf214efd98b4bdaf5ca41728faf00e7c037#diff-85701127596aca0e597bd7961b5d59cdde4f6bb3e2a109a22be859ab7568b4d2L6350-L6361) to fix a bug

>I can open a PR if that's best.

PR #1315, opened 2022-05-12
Remove memory addresses from anonymous class frames

toemat:
Match the new `${addr}` syntax as well as the previous `0x{addr}` syntax when stripping out the memory address so that events can group correctly again.

For https://github.com/getsentry/sentry-php/issues/1314

### DIFF
--- a/tests/StacktraceTest.php
+++ b/tests/StacktraceTest.php
@@ -203,7 +203,7 @@ final class StacktraceTest extends TestCase
             ],
         ];
 
-        yield 'Backtrace with frame containing memory address 0x format' => [
+        yield 'Backtrace with frame containing memory address in PHP <7.4.2 format' => [
             new Options([
                 'prefixes' => ['/path-prefix'],
             ]),
@@ -242,7 +242,7 @@ final class StacktraceTest extends TestCase
             ],
         ];
 
-        yield 'Backtrace with frame containing memory address $ format' => [
+        yield 'Backtrace with frame containing memory address in PHP >=7.4.2 format' => [
             new Options([
                 'prefixes' => ['/path-prefix'],
             ]),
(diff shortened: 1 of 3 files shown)


stayallive:
Could you also add a test so we don't regress in the future?

toemat:
Added a new test and updated CHANGELOG, thank you for the guidance! 🙌

stayallive (approved):

tests/StacktraceTest.php
```diff
@@ -241,6 +241,45 @@ public function buildFromBacktraceDataProvider(): \Generator
                 ],
             ],
         ];
+
+        yield 'Backtrace with frame containing memory address $ format' => [
```
ste93cry, line 245: What about changing the description to something more meaningful, e.g. `Backtrace with frame containing memory address in PHP >=7.4.2 format`?

tests/StacktraceTest.php
```diff
@@ -203,7 +203,7 @@ public function buildFromBacktraceDataProvider(): \Generator
             ],
         ];
 
-        yield 'Backtrace with frame containing memory address' => [
+        yield 'Backtrace with frame containing memory address 0x format' => [
```
ste93cry, line 206: What about changing the description to something more meaningful, e.g. `Backtrace with frame containing memory address in PHP <7.4.2 format`?

ste93cry (changes requested):
Just some small changes and nitpicks, but good work overall

ste93cry:
I just realized that you are targeting the wrong branch. Since this is a bugfix, it should be merged into `master`

toemat:
Thank you @ste93cry, I really appreciate the review and guidance. I addressed the comments (should I respond to them, or 'resolve' them?). I'll switch the target to `master`.

ste93cry:
>should I respond to them, or 'resolve' them?

I don't think there is a right answer. Usually, I mark them as done if there's nothing more to add, and I leave a comment otherwise.

>I addressed the comments

Thank you for your work! A release will follow soon 🚢

PR #1324, opened 2022-05-25
Ability to set or remove single tag in the Event

Furgas:
For easier event processing instead of `getTags -> change -> setTags`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Add support for `monolog/monolog:^3.0` (#1321)
+- Add `setTag` and `removeTag` public methods to `Event` for easier manipulation of tags (#1324)
 
 ## 3.5.0 (2022-05-19)
 
--- a/tests/EventTest.php
+++ b/tests/EventTest.php
@@ -101,4 +101,19 @@ final class EventTest extends TestCase
             ['environment', 'foo'],
         ];
     }
+
+    public function testSetAndRemoveTag(): void
+    {
+        $tagName = 'tag';
+        $tagValue = 'value';
+
+        $event = Event::createEvent();
+        $event->setTag($tagName, $tagValue);
+
+        $this->assertSame([$tagName => $tagValue], $event->getTags());
+
+        $event->removeTag($tagName);
+
+        $this->assertEmpty($event->getTags());
+    }
 }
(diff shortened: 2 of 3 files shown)


ste93cry:
I have nothing cons, however IIRC there had been already some discussion about whether such method had to be exposed or not, and it was a decision that had to be made cross-SDK. @vladanpaunovic can you please chime in?

vladanpaunovic (approved):
Hey @Furgas, thanks for your contribution!

I did a small [check across SDKs](https://cs.github.com/?scope=org%3Agetsentry&scopeName=getsentry&q=remove_tag) and many of our other SDKs already have these properties so I would add it here.

There is a [similar issue on Python](https://github.com/getsentry/sentry-python/issues/1344) too where we will have to align this.

@ste93cry, code looks fine to me. Feel free to review it.

tests/EventTest.php
```diff
@@ -101,4 +101,19 @@ public function gettersAndSettersDataProvider(): array
...
+
+    public function testSetAndRemoveTag(): void
+    {
+        $tagName = 'tag';
+        $tagValue = 'value';
+
+        $event = Event::createEvent();
+        $event->setTag($tagName, $tagValue);
+
+        $this->assertEquals([$tagName => $tagValue], $event->getTags());
```
ste93cry, line 113: Please use `assertSame`

ste93cry (changes requested):
There are a few things off missing: first, an update of the CHANGELOG. Second, we should add these methods to the `Scope` class as well (which is where they are mostly useful). Last, please target the `develop` branch

Furgas:
> Second, we should add these methods to the `Scope` class as well

These methods are already in the `Scope` class. It was my motivation to add them to `Event` too, so I can use them in integrations (via `Scope::addGlobalEventProcessor`).

ste93cry:
You're right, I didn't check that before commenting and taken for granted that they weren't present already

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Add support for `monolog/monolog:^3.0` (#1321)
+- Add `setTag` and `removeTag` public methods to `Event` for easier manipulation of tags
```
ste93cry, line 6: Can you please add the reference to this PR at the end of the line?

ste93cry (changes requested):

PR #1326, opened 2022-05-28
Monolog3 support

olegpro:


Jean85:
Already done in #1321, without breaking support for older Monolog versions.

Issue #1313, opened 2022-05-12
Deploy Sidecar Manually


smeubank:
The first phase of our PoC we want to ensure there is a simple way to deploy relay as a sidecar with a PHP application manually

### Problem Statement
We need to first evaluate the practicality and value of deploying a server side application with a sidecar. The potential value add during this phase is reduce round trip latency to call the sidecar application hosted next to the PHP application. Thereby hopefully increasing the capacity for high throughput applications.

### Outcome

- simple set of instructions for a developer to deploy their app and the sidecar
    - need a new option to flag that the Sentry instrumented application is using a sidecar
        - can just be boolean flag

- Once this is done we begin to measure if we see performance gains
  - we can then evaluate investigating the value in finding ways to possibly deploy this more seamlessly and add more functionality to relay to support unique nature of PHP

**For extra consideration:**
- could add a test (kinda health check)
- extra tools to help users with setting it up
  - supervisor then copy these lines etc
  - docker OS level

smeubank:
after the last sync: we agreed that for now an option for using local relay will not be moved on yet

& we will focus on ensuring the documentation and process for deploying relay for PHP is simple and clearly defined

[PHP Performance: Getting Started](https://docs.sentry.io/platforms/php/performance/#improve-response-time)

first thoughts, are that we do already outline the limitations and nature of PHP, maybe we could link to some external documentation/wiki with more info if that would be helpful to users

then already describe what realy could enable in the future with offloading logic to relay and enabling more Sentry functionality for more developers
- client reports
- higher throughput
- 429 
- auto session 

maybe i am forgetting something in this list, and maybe something in list is incorrect, but it's my first thoughts

vladanpaunovic:
Given that in our docs we are currently asking users to [manually set up their own Relay](https://docs.sentry.io/platforms/php/performance/#improve-response-time), I am closing this issue until we figure out how to better proceed here.

PR #1328, opened 2022-06-07
Add `use_local_relay` option

stayallive:
Relates to #1313.

This adds a `use_local_relay` option to the SDK option you can supply like:

```php
Sentry\init([
    'dsn' => 'https://fa516ddb2fb44c44a2c4c9ce2638af99@o12345.ingest.sentry.io/1234',
    'use_local_relay' => true,
]);
```

Internally this will result in the DSN used by the SDK to look like:

```
http://fa516ddb2fb44c44a2c4c9ce2638af99@localhost:3000/1234
```

(changing the scheme to `http`, the host to `localhost` and the port to `3000`)

If relay is deployed elsewhere in the infrastructure the user can simply modify their own DSN to make the needed adjustments instead of us adding options for custom scheme, host and port when relay is used (hence the `local` in the option name). This is just a convenience option for when you deploy Relay next to your application.

This might need to be enhanced later to still supply the original scheme/domain/port to let Relay now where to relay to, but this is not supported by Relay yet so I've not added this here at this time.

### DIFF
--- a/src/Dsn.php
+++ b/src/Dsn.php
@@ -117,6 +117,18 @@ final class Dsn implements \Stringable
     }
 
     /**
+     * Overwrite parts of the DSN to point to a local Relay instance.
+     *
+     * @return void
+     */
+    public function useLocalRelay(): void
+    {
+        $this->scheme = 'http';
+        $this->host = 'localhost';
+        $this->port = 3000;
+    }
+
+    /**
      * Gets the protocol to be used to access the resource.
      */
     public function getScheme(): string
--- a/src/Options.php
+++ b/src/Options.php
@@ -795,6 +795,7 @@ final class Options
             'capture_silenced_errors' => false,
             'max_request_body_size' => 'medium',
             'class_serializers' => [],
+            'use_local_relay' => false,
         ]);
 
         $resolver->setAllowedTypes('send_attempts', 'int');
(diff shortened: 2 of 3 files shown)


Jean85 (changes requested):
I'm :-1: on this change: having an additional option that overrides/hijacks the meaning of another options seems a bad pattern to me.

I would prefer to leave the possibility of having that value explicitly passed to the DSN. Do we have any impediment there, like a validation conflict?

stayallive:
@Jean85 this is a draft PR meant as a discussion point and not gospel at this moment. Thanks for your feedback though!

There is nothing stopping the user from modifying their own DSN (see the PR text what to do in case it's not on localhost:3000) but this might not be the only change needed in the future so it might not be ideal now but might be less not-ideal in the future and a nice addition.

Again this is a PR to aid in the discussions in #1313 and is still a work in progress / ongoing discussion 👍

ste93cry:
I agree with @Jean85. Coding-wise, this PR looks perfect, but the usefulness of this change is a mistery for me. Why adding a new option, increasing the complexity of the software and the cognitive load for its users, when it would be enough to document how to get the DSN that points to Relay? As a user of Sentry, I know that everything I have to do to use it is more or less get the DSN, set it, and I should be ready to go. I don't care if the DSN is pointing to the server or to a proxy, it's the same for the end user.

stayallive:
Please, let's curb the discussion for now until we are a little further in the process :)

I understand and appreciate the concerns you both are raising and I agree it does raise more questions at this point than answers, but some context is missing I'm not able to convey (very good) at this point in the process.

Again, I/we will never merge this before discussing this at length but it's not useful going into it at this moment this is just a placeholder hence the "In Progress" and "Discussion" labels and it being a draft PR! Sorry for the noise and the unclearness at this point!

Issue #1318, opened 2022-05-18
Monolog handler is not compatible with monolog v3


W0rma:
[`Sentry\Monolog\Handler`](https://github.com/getsentry/sentry-php/blob/master/src/Monolog/Handler.php) extends [`Monolog\Handler\AbstractProcessingHandler`](https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/AbstractProcessingHandler.php). 

The signature of the constructor has changed in [monolog 3](https://github.com/Seldaek/monolog/blob/3.0.0/UPGRADE.md#300) which leads to a fatal error:

```
Fatal error: Declaration of Sentry\Monolog\Handler::write(array $record): void must be compatible with Monolog\Handler\AbstractProcessingHandler::write(Monolog\LogRecord $record): void in /var/www/html/vendor/sentry/sentry/src/Monolog/Handler.php on line 51
```

Jean85:
This is definitely something that we need to tackle, but I'm unsure how we should proceed: do we do a conditional declaration with an `if (/** detect monolog version **/)` at the top? or do we add a new, separate `HandlerV3`?

WDYT @ste93cry @stayallive? Any possible outside impact that you may think of?

simPod:
We should mind this line https://github.com/getsentry/sentry-php/blob/6e8d3e24a7045b7222f8ad7987b7ebc6f0faf394/src/Monolog/Handler.php#L68 where it now reads only 'context' but it should support 'extra' as well. See https://github.com/Seldaek/monolog/blob/main/doc/message-structure.md

jiripudil:
Quoting the [release notes](https://github.com/Seldaek/monolog/releases/tag/3.0.0),

> If you want to support multiple Monolog versions this should be possible by type-hinting nothing, or `array|LogRecord` if you support PHP 8.0+. You can then code against the $record using Monolog 2 style as LogRecord implements ArrayAccess for BC.

It seems that the BC code in `LogRecord` even handles [things such as `level`](https://github.com/Seldaek/monolog/blob/60ad5183b5e5d6c9d4047e9f3072d36071dcc161/src/Monolog/LogRecord.php#L79-L88) in a way that it returns the same integer value as in 2.0. So perhaps this is simply a matter of dropping the `array` type-hint – or widening it to `array|LogRecord` if you can afford to bump the minimum PHP version to 8.0.

ste93cry:
>Any possible outside impact that you may think of?

[...]

PR #1329, opened 2022-06-09
feat: set monolog.extra

simPod:
https://github.com/getsentry/sentry-php/issues/1318#issuecomment-1129865484

### DIFF
--- a/src/Monolog/Handler.php
+++ b/src/Monolog/Handler.php
@@ -71,6 +71,12 @@ final class Handler extends AbstractProcessingHandler
                 $scope->setExtra('monolog.context', $monologContextData);
             }
 
+            $monologExtraData = $this->getMonologExtraData($record['extra']);
+            if ($monologContextData !== []) {
+                $scope->setExtra('monolog.extra', $monologExtraData);
+            }
+
+
             $this->hub->captureEvent($event, $hint);
         });
     }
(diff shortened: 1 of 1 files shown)


Jean85:
Duplicate of #1244, already merged in develop.

simPod:
Hello @Jean85 I checked the PR and does not seem to read `$record['extra']`. Can you review?

ste93cry:
We are not reading `extra` purposefully, as it's meant to be used by processors and we absolutely want to avoid logging anything blindly as it may result in exceptionally big messages

>The main difference is that `context` can be supplied in user land (it is the 3rd parameter to `Psr\Log\LoggerInterface` methods) whereas `extra` is internal only and can be filled by processors. The reason processors write to `extra` and not to `context` is to prevent overriding any user-provided data in `context`.

JanTvrdik:
The `extra` values are provided by processors, but they are provided with the intention of being logged. See

* [ProcessIdProcessor](https://github.com/Seldaek/monolog/blob/d381140ca134f7ece43cd2c139d2207bb26dc329/src/Monolog/Processor/ProcessIdProcessor.php) or
* [HostnameProcessor](https://github.com/Seldaek/monolog/blob/d381140ca134f7ece43cd2c139d2207bb26dc329/src/Monolog/Processor/HostnameProcessor.php) or
* [MemoryUsageProcessor](https://github.com/Seldaek/monolog/blob/d381140ca134f7ece43cd2c139d2207bb26dc329/src/Monolog/Processor/MemoryUsageProcessor.php) or 
* [MemoryPeakUsageProcessor](https://github.com/Seldaek/monolog/blob/d381140ca134f7ece43cd2c139d2207bb26dc329/src/Monolog/Processor/MemoryPeakUsageProcessor.php)

ste93cry:
You are right. I have no opinion against implementing support at this point, as long as logging is dependent on the `fillExtraContext` flag, so that in the worst case the user can still turn it off. @ Jean85 WDYT?

simPod:
@ste93cry in that case the PR should be good as is

@JanTvrdik thanks for input

ste93cry:
Because the branch got deleted, I cannot reopen this PR. If you can't too, please open a new one

simPod:
mb. https://github.com/getsentry/sentry-php/pull/1330

PR #1331, opened 2022-06-14
bug: Set sentry-trace header when using tracing middleware

DAcodedBEAT:
According to [PSR-7](https://www.php-fig.org/psr/psr-7/), `withHeader` operates in the following manner:
```
Return an instance with the provided value replacing the specified header.
[...]
This method MUST be implemented in such a way as to retain the
immutability of the message, and MUST return an instance that has the
new and/or updated header and value.
```
(see: https://www.php-fig.org/psr/psr-7/)

Due to this, we must set `$request` to the modified request object (from the return value of
withHeader) in order to actually send the sentry-trace header.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Set the `sentry-trace` header when using the tracing middleware (#1331)
+
 ## 3.6.0 (2022-06-10)
 
 - Add support for `monolog/monolog:^3.0` (#1321)
--- a/tests/Tracing/GuzzleTracingMiddlewareTest.php
+++ b/tests/Tracing/GuzzleTracingMiddlewareTest.php
@@ -104,7 +104,8 @@ final class GuzzleTracingMiddlewareTest extends TestCase
         $hub->setSpan($transaction);
 
         $middleware = GuzzleTracingMiddleware::trace($hub);
-        $function = $middleware(static function () use ($expectedPromiseResult): PromiseInterface {
+        $function = $middleware(function (Request $request) use ($expectedPromiseResult): PromiseInterface {
+            $this->assertNotEmpty($request->getHeader('sentry-trace'));
             if ($expectedPromiseResult instanceof \Throwable) {
                 return new RejectedPromise($expectedPromiseResult);
             }
--- a/src/Tracing/GuzzleTracingMiddleware.php
+++ b/src/Tracing/GuzzleTracingMiddleware.php
@@ -34,7 +34,7 @@ final class GuzzleTracingMiddleware
 
                 $childSpan = $span->startChild($spanContext);
 
-                $request->withHeader('sentry-trace', $childSpan->toTraceparent());
+                $request = $request->withHeader('sentry-trace', $childSpan->toTraceparent());
 
                 $handlerPromiseCallback = static function ($responseOrException) use ($hub, $request, $childSpan) {
                     // We finish the span (which means setting the span end timestamp) first to ensure the measured time


DAcodedBEAT:
@ste93cry I updated the CHANGELOG and fixed up the tests (I had to rework the unit tests to properly test the request/response values from Guzzle.). Let me know what you think and thanks for the quick reply!

ste93cry:
I'm not sure I understand why the entire testing class had to be changed. The lambda that is returned from `GuzzleTracingMiddleware::trace()` must be invoked by passing to it an handler, which is then called here:

https://github.com/getsentry/sentry-php/blob/6d1a6ee29c558be373bfe08d454a3c116c02dd0d/src/Tracing/GuzzleTracingMiddleware.php#L83

What we want to assert is that the request contains the `sentry-trace` header. To do it, it's enough to add the `$request` parameter to the handler and add the missing assertion...

https://github.com/getsentry/sentry-php/blob/6d1a6ee29c558be373bfe08d454a3c116c02dd0d/tests/Tracing/GuzzleTracingMiddlewareTest.php#L107

DAcodedBEAT:
I wanted to make sure the middleware was fully-working from the Guzzle client's configuration perspective (with the added benefit of removing PHPStan violations in the baseline since Guzzle would've been introduced as a dev dependency by using their [MockHandler](https://docs.guzzlephp.org/en/stable/testing.html#mock-handler)) but you're right -- this can be just the assertion on the header.

I updated this in [72094da](https://github.com/getsentry/sentry-php/pull/1331/commits/72094da9ea6a80fc565b9abae0ba36120acb8aaf).  If this test update is acceptable, I can either squash and force-push to my branch or you can do squash these commits on merge. :+1:

Jean85 (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 ## Unreleased
+- Set the `sentry-trace` header when using the tracing middleware (#1331)
```
ste93cry, line 4: Nitpick: please add a blank line

```suggestion

- Set the `sentry-trace` header when using the tracing middleware (#1331)
```

tests/Tracing/GuzzleTracingMiddlewareTest.php
```diff
@@ -104,7 +104,8 @@ public function testTrace(Request $request, $expectedPromiseResult, array $expec
         $hub->setSpan($transaction);
 
         $middleware = GuzzleTracingMiddleware::trace($hub);
-        $function = $middleware(static function () use ($expectedPromiseResult): PromiseInterface {
+        $function = $middleware(static function (Request $request) use ($expectedPromiseResult): PromiseInterface {
+            self::assertNotEmpty($request->getHeader('sentry-trace'));
```
ste93cry, line 108: Just to keep consistency with how we write assertions usually, I think it would be better to invoke the `assertNotEmpty()` method on `$this`

```suggestion
        $function = $middleware(function (Request $request) use ($expectedPromiseResult): PromiseInterface {
            $this->assertNotEmpty($request->getHeader('sentry-trace'));
            
```

ste93cry (changes requested):

CHANGELOG.md
DAcodedBEAT: I squashed my commits and addressed this in 3e61f2c

ste93cry (approved):
Thank you!

DAcodedBEAT:
Thanks for merging this @ste93cry - can't wait to use the distributed tracing capability!

I noticed this was flagged for the `3.6` milestone but `3.6.0` has already been tagged off.  Is this going into a `3.6.1` or would this be in a new semver version?

ste93cry:
Yes, since it's a bugfix it will be released with the next patch version. We usually wait a few days/weeks to see if more issues comes and then we make the release

PR #1334, opened 2022-07-14
Fix `Scope::getTransaction()` so that it returns also unsampled transactions

### LINKED ISSUE #1333: Scope::getTransaction returns null when Transaction isn't sampled
### Environment

How do you use Sentry?
Sentry SaaS (sentry.io)

Which SDK and version?
sentry-php@3.4.0

### Steps to Reproduce

1. initialize sentry with a very low sample rate for easier repro
2. call `$transaction = startTransaction`
3. call `SentrySdk::getCurrentHub()->setSpan($transaction)`
4. call `SentrySdk::getCurrentHub()->getTransaction()`

### Expected Result

I expected to get the instance of the Transaction I just started.

### Actual Result

`getTransaction` returned null because when you call `startTransaction` and this transaction isn't getting sampled it does not initialize the SpanRecorder (Hub.php:256). However, `getTransaction` gets the transaction from `Scope` which checks for an active SpanRecorder and returns null otherwise.

This behavior is not documented anywhere, I think it would suffice to add a note to the `Hub::getTransaction` method that if it's not sampled it will return null.

ste93cry:
Fixes #1333: `Scope::getTransaction()` was using the `SpanRecorder` to find the active transaction to return, but if the transaction is unsampled the recorder doesn't keep track of it and so `null` is returned. By adding a reference to the parent transaction inside each span, we can bypass the recorder and directly retrieve it, regardless of its sampling status. The original fix was implemented in getsentry/sentry-javascript#2952, and this is just a porting of it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix `Scope::getTransaction()` so that it returns also unsampled transactions (#1334)
+
 ## 3.6.1 (2022-06-27)
 
 - Set the `sentry-trace` header when using the tracing middleware (#1331)
--- a/src/State/Scope.php
+++ b/src/State/Scope.php
@@ -408,12 +408,8 @@ final class Scope
      */
     public function getTransaction(): ?Transaction
     {
-        $span = $this->span;
-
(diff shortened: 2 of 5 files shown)


stayallive (approved):
Looks good to me 💪 (even a changelog entry 😎)

Jean85 (approved):

PR #1341, opened 2022-08-11
fix: Add back xsi:noNamespaceSchemaLocation

cleptric:
Add back `xsi:noNamespaceSchemaLocation` to re-enable IDE autocompletion for this file.


phpunit.xml.dist
```diff
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
ste93cry, line 3: Nitpick: do you mind aligning back all the attributes to where they were before your changes?

ste93cry (changes requested):
Please revert also the changes that removed the `<coverage>` section, otherwise PHPUnit will not generate the code coverage report and Codecov won't work. It's not really important that there is a warning about the failed validation of the config file, as it's expected due to us running two completely different versions of PHPUnit because we have to support old PHP versions. If you are really bothered by that warning, we could simply split the `phpunit.xml.dist` into two, one for each version of PHPUnit, but I don't think it's worth it

cleptric:
I'm wondering if it's necessary to upload the code coverage on each matrix run. Wouldn't one upload during Linux/8.1 suffice?

ste93cry:
No, because some tests run only for some PHP versions or some OS, so each coverage file gets uploaded and Codecov makes the sum of all of them and generates the final report

ste93cry (approved):

PR #1346, opened 2022-08-16
ref: Drop symfony/polyfill-uuid in favour of a standalone implementation

cleptric:
As we only use one single method of `symfony/polyfill-uuid`, I'm in favor of removing this as a dependency and implementing the method directly in the SDK.

The code is based on [symfony/polyfill-uuid](https://github.com/symfony/polyfill/blob/d45b58e9f527e408fb28c4b6431b7284f8c168ff/src/Uuid/Uuid.php#L356), hence the `@copyright` annotation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+Drop symfony/polyfill-uuid in favour of a standalone implementation (#1346)
+
 ## 3.7.0 (2022-07-18)
 
 - Fix `Scope::getTransaction()` so that it returns also unsampled transactions (#1334)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-Drop symfony/polyfill-uuid in favour of a standalone implementation (#1346)
+- Drop symfony/polyfill-uuid in favour of a standalone implementation (#1346)
 
 ## 3.7.0 (2022-07-18)
 
--- a/composer.json
+++ b/composer.json
@@ -35,8 +35,7 @@
         "psr/http-message-implementation": "^1.0",
         "psr/log": "^1.0|^2.0|^3.0",
         "symfony/options-resolver": "^3.4.43|^4.4.30|^5.0.11|^6.0",
-        "symfony/polyfill-php80": "^1.17",
-        "symfony/polyfill-uuid": "^1.13.1"
+        "symfony/polyfill-php80": "^1.17"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.19|3.4.*",
--- a/src/EventId.php
+++ b/src/EventId.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Sentry;
 
+use Sentry\Util\SentryUid;
+
 /**
  * This class represents an event ID.
  *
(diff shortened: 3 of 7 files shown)


ste93cry:
What's the benefit of this? Every code we have in this package is code that we have to maintain, and UUID implementations are one of those things that are likely to have bugs because of weak algorithms

stayallive:
I think that dropping some dependencies makes sense, maybe this is not the most obvious one but it's also yet another dependency for some very simple piece of code that generates random IDs. The SDK docs also play nice in on that: https://develop.sentry.dev/sdk/philosophy/#dependencies-cost. The chance is slim but this polyfil might clash with user defined global functions and constants so having our own implementation based on a maintained one makes sense here.

Looking at the actual ID format (from: https://develop.sentry.dev/sdk/envelopes/#prerequisites):

> UUIDs are declared as either 32 character hexadecimal strings without dashes ("12c2d058d58442709aa2eca08bf20986"), or 36 character strings with dashes ("12c2d058-d584-4270-9aa2-eca08bf20986"). It is recommended to omit dashes and use UUID v4 in all cases.

We might take the opportunity to generate the UUID without the dashes in the first place instead of stripping them later removing a little string stripping around the code.

ste93cry:
>I think that dropping some dependencies makes sense, maybe this is not the most obvious one but it's also yet another dependency for some very simple piece of code that generates random IDs.

Not all code that is a few lines long is also simple. In particular, generating UUIDs is something that if done incorrectly leads to bugs because there is not enough randomness or entrophy, and that's the reason why a lot of libs exist to help people generating them. Moreover, we are importing the code of this package but we are not testing it properly besides ensuring that it returns a string in the correct format, which honestly means nothing in this case, and this is one more thing that once we do we have to maintain. I think that we have greater problems to solve or features to add than changing things that honestly bring 0 real impact to the end-user, so it would be wiser to not reinvent the wheel every time, even more since we're not in like in the JS SDK where every byte count. :relaxed:

cleptric:
@ste93cry The main benefit here is getting rid of one dependency. In this specific case, we're using a single method from this library. The code in question wasn't updated in the last two years, so I consider this as a good sign that it works well and is stable.
As we're still using the exact same code as we did before, I do not worry about bugs introduced by this.
Given the complex nature of having to support multiple PHP versions at once, fewer dependencies are a good thing.

I also like the idea of @stayallive in terms of already providing the correctly formatted uuid.

Nothing major, but a small improvement at least.

ste93cry:
>Given the complex nature of having to support multiple PHP versions at once, fewer dependencies are a good thing

Sure, but there are libraries and libraries, and certainly a polyfill providing a few functions is not the one that will change this complexity.

>The code in question wasn't updated in the last two years, so I consider this as a good sign that it works well and is stable

Sure, but to reach this stability it took quite a few bug fixes and changes, and if a new bug gets discovered in the future we won't even know about it because we now have this code as part of our package. If you told me that it's time to ditch Httplug I would happily answer "yes!", but trying to drop a polyfill just for the sake of "having one less dependency" is something I can't get really see the benefit at all. This dependency is a polyfill for the `ext-uuid` extension which provides an optimized way (at work we are using it for example) to generate UUIDs, so if people have it installed they can take advantage of native UUID generation. Last, this package is so common that it's probably you will have it installed anyway on your system because some other dep is requiring it, especially if those deps are Symfony or Laravel

stayallive (approved):
Still optimized if `uuid_create` is available (with or without polyfill) and 1 dependency less, seems good to me 👍

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+Drop symfony/polyfill-uuid in favour of a standalone implementation (#1346)
```
stayallive, line 5: 😉 

```suggestion
- Drop symfony/polyfill-uuid in favour of a standalone implementation (#1346)
```

vladanpaunovic (approved):

PR #1347, opened 2022-08-17
feat: Add setter for type on ExceptionDataBag

### LINKED ISSUE #1338: Expose exception severity in data bag
Atm, the ExceptionDataBag does not include the exception's severity and it's not possible to get it either, since the exception itself is not in the DataBag too

It would be incredibly useful if either the original exception itself or at least the `$exception->getSeverity()` were available in the data bag for further, custom processing.

My use case:
in `setBeforeSendCallback` I'd like to overwrite the `type: "ErrorException",` with a custom string, depending on if the error is a PHP native error (`E_NOTICE`) or comes from a user triggered error (`E_USER_NOTICE`)
Atm I have to strpos the `context_line` for `USER NOTICE`, which is not very robust and leads to quite a few false positives unfortunately.

e.g. simplified:

```php
function ( \Sentry\Event $event ) {
	$exceptions = $event->getExceptions();
	$raw_exception = $exceptions[0]->getException(); // @todo this is what I woudl be ideal, a getSeverity() on it would be fine though
	if ( $raw_exception->getSeverity() === \E_USER_NOTICE ) {
		$exceptions[0]->setType( 'E_USER_NOTICE' ); // @todo ExceptionDataBag.php also needs a setType method added
	}
	
	return $event;	
}
```

On a side notee: since we use Sentry.io should I open a support ticket with them for faster processing of this request?

cleptric:
This will allow users to set the type of exception at a later stage.
As mentioned in #1100, I didn't add any tests here.

Closes #1338

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Add setter for type on the `ExceptionDataBag`
+
 ## 3.7.0 (2022-07-18)
 
 - Fix `Scope::getTransaction()` so that it returns also unsampled transactions (#1334)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Add setter for type on the `ExceptionDataBag`
+- Add setter for type on the `ExceptionDataBag` (#1347)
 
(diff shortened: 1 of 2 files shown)


stayallive (approved):
🚢

PR #1349, opened 2022-08-17
ref:	Do not register the error handler if the DSN is null

### LINKED ISSUE #1190: Do not register the error handler if the DSN is null
Related to https://github.com/getsentry/sentry-symfony/issues/46#issuecomment-780003387

> it should be possible to change the way the client is built during the container compilation so that if the DSN is null then we skip adding those integrations to the client.

cleptric:
Closes #1190

### DIFF
--- a/src/Integration/IntegrationRegistry.php
+++ b/src/Integration/IntegrationRegistry.php
@@ -126,28 +126,17 @@ final class IntegrationRegistry
             return [];
         }
 
-        /**
-         * If the DSN is null, we do not want to register any eror handler.
-         * Hence, we do not setup the ExceptionListenerIntegration, ErrorListenerIntegration and
-         * FatalErrorListenerIntegration in this case.
-         */
-        if (null === $options->getDsn()) {
-            return [
-                new RequestIntegration(),
-                new TransactionIntegration(),
-                new FrameContextifierIntegration(),
-                new EnvironmentIntegration(),
-            ];
-        }
-
-        return [
-            new ExceptionListenerIntegration(),
-            new ErrorListenerIntegration(),
-            new FatalErrorListenerIntegration(),
+        $integrations = [
             new RequestIntegration(),
             new TransactionIntegration(),
             new FrameContextifierIntegration(),
             new EnvironmentIntegration(),
         ];
+
+        if (null !== $options->getDsn()) {
+            array_unshift($integrations, new ExceptionListenerIntegration(), new ErrorListenerIntegration(), new FatalErrorListenerIntegration());
+        }
+
+        return $integrations;
     }
 }
--- a/src/Integration/IntegrationRegistry.php
+++ b/src/Integration/IntegrationRegistry.php
@@ -126,6 +126,20 @@ final class IntegrationRegistry
             return [];
         }
 
+        /**
+         * If the DSN is null, we do not want to register any eror handler.
+         * Hence, we do not setup the ExceptionListenerIntegration, ErrorListenerIntegration and
+         * FatalErrorListenerIntegration in this case.
+         */
+        if (null === $options->getDsn()) {
+            return [
+                new RequestIntegration(),
+                new TransactionIntegration(),
+                new FrameContextifierIntegration(),
+                new EnvironmentIntegration(),
+            ];
+        }
+
         return [
             new ExceptionListenerIntegration(),
             new ErrorListenerIntegration(),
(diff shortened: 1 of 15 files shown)


stayallive (approved):
Seems like a good change to me! 👍

tests/Integration/IntegrationRegistryTest.php
```diff
@@ -67,6 +67,7 @@ public function setupOnce(): void
 
         yield 'No default integrations and no user integrations' => [
             new Options([
+                'dsn' => 'http://public@example.com/sentry/1',
```
ste93cry, line 70: I think it would be better to invert all these test cases so that we test only once that the `*ListenerIntegration` integrations are installed and we limit as much as possible the amount of listeners that are added to the error handler

tests/phpt/error_handler_captures_errors_not_silencable_on_php_8_and_up.phpt
```diff
@@ -55,9 +55,11 @@ $transportFactory = new class implements TransportFactoryInterface {
...
 error_reporting(E_ALL & ~E_USER_ERROR);
 
-$client = ClientBuilder::create(['error_types' => E_ALL, 'capture_silenced_errors' => false])
-    ->setTransportFactory($transportFactory)
-    ->getClient();
+$client = ClientBuilder::create([
+    'dsn' => 'http://public@example.com/sentry/1',
+    'error_types' => E_ALL,
+    'capture_silenced_errors' => false
+])->setTransportFactory($transportFactory)->getClient();
```
ste93cry, line 62: Nitpicking: this looks a bit messy to read, what if we extracted the options into its own `$options` variable so that we can keep the formatting as before with each method call on its own line?

tests/phpt/error_handler_respects_current_error_reporting_level.phpt
```diff
@@ -21,14 +29,32 @@ while (!file_exists($vendor . '/vendor')) {
 
 require $vendor . '/vendor/autoload.php';
 
+$transportFactory = new class implements TransportFactoryInterface {
```
ste93cry, line 32: Why are you adding this? If there was no need before there should be no need now :thinking:

ste93cry (changes requested):

cleptric:
@ste93cry Thanks for the feedback. I only noticed your branch after I finished this, sadly.

tests/phpt/error_handler_captures_errors_not_silencable_on_php_8_and_up.phpt
cleptric: Yes, will do that.

tests/phpt/error_handler_respects_current_error_reporting_level.phpt
cleptric: With the `dsn` option set, the `DefaultTransportFactory` no longer returns the `NullTransport` in this case, so there is actually a `Http\Discovery\NotFoundException` thrown during running this test.

tests/Integration/IntegrationRegistryTest.php
cleptric: My reasoning was to test the 99% use case, which would be with a none empty DSN.

src/Integration/IntegrationRegistry.php
```diff
@@ -126,6 +126,20 @@ private function getDefaultIntegrations(Options $options): array
...
+         * Hence, we do not setup the ExceptionListenerIntegration, ErrorListenerIntegration and
+         * FatalErrorListenerIntegration in this case.
+         */
+        if (null === $options->getDsn()) {
+            return [
+                new RequestIntegration(),
+                new TransactionIntegration(),
+                new FrameContextifierIntegration(),
+                new EnvironmentIntegration(),
+            ];
```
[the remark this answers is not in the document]
cleptric: Mainly did this for better readability and less complexity.

tests/phpt/error_handler_respects_current_error_reporting_level.phpt
ste93cry: Makes sense, didn't think about this consequence 👍

src/Integration/IntegrationRegistry.php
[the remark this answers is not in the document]
ste93cry: I'm not convinced that having to maintain two separated lists instead of just splitting its creation is really changing so much the readability, but it's your call

ste93cry (approved):
If we want to treat it as a bugfix, we can merge in `master` and release it earlier. Please also set the labels and milestones so that we keep track of what ends up in which version

PR #1360, opened 2022-09-13
feat: Add support for Dynamic Sampling

cleptric:
### Add support for Dynamic Sampling

We decided not to access the request internally and will instead rely on people passing in a correctly populated `TransactionContext` to be able to use Dynamic Sampling.
This is in line with the implementation of the propagation of `sentry-trace`.

For our Laravel and Symfony SDKs, we can then set this up automatically.

Additional, the following required items for DS were added :

- The `TransactionSource` https://develop.sentry.dev/sdk/event-payloads/transaction/#transaction-annotations
- The `user_segment` https://develop.sentry.dev/sdk/event-payloads/user/#attributes & https://develop.sentry.dev/sdk/performance/dynamic-sampling-context/#envelope-header
- An adequately populated envelope header https://develop.sentry.dev/sdk/performance/dynamic-sampling-context/#envelope-header

Usage example:

```php
use Sentry\Tracing\TransactionContext;
use function Sentry\startTransaction;

$sentryTraceHeader = $request->getHeaderLine('sentry-trace');
$baggageHeader = $request->getHeaderLine('baggage');

$transactionContext = TransactionContext::fromHeaders($sentryTraceHeader, $baggageHeader);

...
startTransaction($transactionContext);
```

_Co-authored-by: Alex Bouma_


tests/Tracing/DynamicSamplingContextTest.php
```diff
@@ -0,0 +1,176 @@
...
+
+        $dsc = DynamicSamplingContext::fromTransaction($transaction, $hub);
+
+        $this->assertSame((string) $transaction->getTraceId(), $dsc->get('trace_id'));
+        $this->assertSame((string) $transaction->getMetaData()->getSamplingRate(), $dsc->get('sample_rate'));
+        $this->assertSame('foo', $dsc->get('transaction'));
+        $this->assertSame('public', $dsc->get('public_key'));
+        $this->assertSame('1.0.0', $dsc->get('release'));
+        $this->assertSame('test', $dsc->get('environment'));
+        $this->assertSame('my_segment', $dsc->get('user_segment'));
```
cleptric, line 106: I'm not sure how to properly mock the scope in this test. Any help is kindly appreciated.

src/Tracing/Transaction.php
```diff
@@ -59,6 +65,29 @@ public function setName(string $name): void
...
+    /**
+     * Gets the transaction dynamic sampling context.
+     */
+    public function getDynamicSamplingContext(): DynamicSamplingContext
+    {
+        if (null !== $this->getMetadata()->getDynamicSamplingContext()) {
+            return $this->getMetadata()->getDynamicSamplingContext();
+        }
+
+        $dsc = DynamicSamplingContext::fromTransaction($this->transaction, $this->hub);
```
cleptric, line 85: Given the `$this->metadata = $context->getMetadata();` call in the constructor, do we might need to add some annotation to please phpstan?

cleptric: As the `Span::class` contains a property `Transaction|null`, I added it to the Transaction as well, annotating it as `Transaction`. Not sure if this is a good approach, though.

stayallive: This is what I would propose to do here, it cannot be null on the transaction class so makes sense to override with a new annotation.

cleptric: Great. I already added the new property and will add a template annotation to please psalm as well.

src/Event.php
```diff
@@ -671,6 +679,31 @@ public function getType(): EventType
         return $this->type;
     }
 
+    /**
+     * Sets the SDK metadata with the given name.
+     *
+     * @param string $name The name that uniquely identifies the SDK metadata
+     * @param mixed  $data The data of the SDK metadata
```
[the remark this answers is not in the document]
cleptric: This is by design 😅 Our style checker formats the comments to be aligned to each other.

src/State/Hub.php
```diff
@@ -234,16 +234,23 @@ public function startTransaction(TransactionContext $context, array $customSampl
         $tracesSampler = $options->getTracesSampler();
 
         if (null === $transaction->getSampled()) {
-            $sampleRate = null !== $tracesSampler
-                ? $tracesSampler($samplingContext)
-                : $this->getSampleRate($samplingContext->getParentSampled(), $options->getTracesSampleRate());
+            if (null !== $tracesSampler) {
```
vladanpaunovic, line 237: nit: it is always easier to read if you compare a variable against a constant
```php
if ($tracesSampler !== null) {
```

Same could be applied to a line above

src/Tracing/DynamicSamplingContext.php
```diff
@@ -0,0 +1,201 @@
...
+     * @var array<string, string>
+     */
+    private $entries;
+
+    /**
+     * Indicates that the dsc is frozen and cannot be mutated.
+     *
+     * @var bool
+     */
+    private $frozen = false;
```
vladanpaunovic, line 31: Why is this `false`?

vladanpaunovic: Also, for boolean variable names, it is always better to prefix them with `is`, `has` or something similar, where possible. This could be easier to read as `isFrozen`.

src/Tracing/DynamicSamplingContext.php
```diff
@@ -0,0 +1,201 @@
...
+    /**
+     * Set a new key value pair on the dsc.
+     *
+     * @param string $key   the list member key
+     * @param string $value the list member value
+     */
+    public function set(string $key, string $value): void
+    {
+        if ($this->frozen) {
+            return;
```
vladanpaunovic, line 50: nit: I don't know what would be the use case where people try to set this when frozen but I think it may be valuable to print something in the logs so users are aware why the context is not set.

src/State/Hub.php
stayallive: I agree, but style fixer defaults Yoda style since we use the Symfony code style.

Not sure if worth changing it since that would create a significant diff.

src/Tracing/DynamicSamplingContext.php
```diff
@@ -0,0 +1,201 @@
...
+    /**
+     * Mark the dsc as frozen.
+     */
+    public function freeze(): void
+    {
+        $this->frozen = true;
+    }
+
+    /**
+     * Indicates that the dsc is frozen and cannot be mutated.
```
vladanpaunovic, line 86: nit(again): This comment is not really describing the variable.

It can be something like:
Identigies weather the dynamic sampling context is mutable or not

src/Tracing/DynamicSamplingContext.php
```diff
@@ -0,0 +1,201 @@
...
+    {
+        $dsc = new self();
+
+        $dsc->set('trace_id', (string) $transaction->getTraceId());
+        $dsc->set('sample_rate', (string) $transaction->getMetaData()->getSamplingRate());
+        $dsc->set('transaction', $transaction->getName());
+
+        $client = $hub->getClient();
+
+        if (null !== $client && null !== $options = $client->getOptions()) {
```
vladanpaunovic, line 163: nit: maybe extract `$options` in a variable above to make this easier to read

src/State/Hub.php
cleptric: I personally don't like it either, bur for the sake of the current codebase, I would not change it 🙂

src/Event.php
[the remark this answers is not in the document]
vladanpaunovic: 🤣

src/State/Hub.php
vladanpaunovic: TIL! Thanks both!

@cleptric, let's think about this (out of the scope of this PR). It may be worth doing this and I am quite confident there is an automated way to cover most of the cases.

tests/Serializer/PayloadSerializerTest.php
```diff
@@ -37,7 +40,10 @@ final class PayloadSerializerTest extends TestCase
 
     protected function setUp(): void
     {
-        $this->serializer = new PayloadSerializer();
+        $options = new Options([
```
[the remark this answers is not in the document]
cleptric: As we didn't need a `tearDown()` before, I don't think we need one now.

src/Tracing/DynamicSamplingContext.php
cleptric: Once the DSC is populated for the first time, we must make it immutable. This implements this.
As people should not mess around with the DSC from the outside, this API won't be documented in the docs as well.

src/Tracing/DynamicSamplingContext.php
cleptric: By default, a DSC is mutable until you add values to it. This refelts that.

src/State/Hub.php
cleptric: This is something I'm happy to do in our next major, v4.0.0 😄

src/Tracing/DynamicSamplingContext.php
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+     * Serialize the dsc as a string.
+     */
+    public function __toString(): string
+    {
+        $string = '';
+
+        foreach ($this->entries as $key => $value) {
+            $string .= urlencode(self::SENTRY_ENTRY_PREFIX . $key) . '=' . urlencode($value);
```
mattjohnsonpint, line 198: `urlencode` here won't meet the baggage spec requirements

cleptric: Great catch!
I replaced `urlencode/urldecode` with  `rawurlencode/rawurldecode`. Thanks 🙂

src/Tracing/DynamicSamplingContext.php
```diff
@@ -0,0 +1,205 @@
...
+            $keyValue = trim($keyValueAndProperties[0]);
+
+            if (empty($keyValue) || !str_contains($keyValue, '=')) {
+                continue;
+            }
+
+            [$key, $value] = explode('=', $keyValue, 2);
+
+            if (0 === strpos($key, self::SENTRY_ENTRY_PREFIX)) {
+                $dsc->set(rawurldecode(str_replace(self::SENTRY_ENTRY_PREFIX, '', $key)), rawurldecode($value));
```
ste93cry, line 136: Shouldn't we trim just the prefix and not all occurrences of this constant? I would have expected using `substr` to play safe

src/Tracing/DynamicSamplingContext.php
```diff
@@ -0,0 +1,205 @@
...
+        $dsc->set('trace_id', (string) $transaction->getTraceId());
+        $dsc->set('sample_rate', (string) $transaction->getMetaData()->getSamplingRate());
+        $dsc->set('transaction', $transaction->getName());
+
+        $client = $hub->getClient();
+
+        if (null !== $client) {
+            $options = $client->getOptions();
+
+            if (null !== $options) {
```
ste93cry, line 166: A client always has the options, so this statement is useless and can be removed

src/Tracing/DynamicSamplingContext.php
```diff
@@ -0,0 +1,205 @@
...
+
+        return $dsc;
+    }
+
+    /**
+     * Serialize the dsc as a string.
+     */
+    public function __toString(): string
+    {
+        $string = '';
```
ste93cry, line 195: Instead of building the string ourselves, wouldn't be easier to collect the parts into an array and then simply call `implode`? This would mean not bothering about the handling of the commas

src/Tracing/TransactionContext.php
```diff
@@ -97,4 +127,57 @@ public static function fromSentryTrace(string $header)
...
+            if (isset($matches['sampled'])) {
+                $context->parentSampled = '1' === $matches['sampled'];
+                $hasSentryTrace = true;
+            }
+        }
+
+        $dsc = DynamicSamplingContext::fromHeader($baggageHeader);
+
+        if (true === $hasSentryTrace && false === $dsc->hasEntries()) {
+            /**
```
ste93cry, line 165: :pray: use `//` for comments and `/**` for docblocks. There are a few other occurrencies to change across the PR

src/Tracing/TransactionContext.php
```diff
@@ -97,4 +127,57 @@ public static function fromSentryTrace(string $header)
...
+
+            if (isset($matches['sampled'])) {
+                $context->parentSampled = '1' === $matches['sampled'];
+                $hasSentryTrace = true;
+            }
+        }
+
+        $dsc = DynamicSamplingContext::fromHeader($baggageHeader);
+
+        if (true === $hasSentryTrace && false === $dsc->hasEntries()) {
```
ste93cry, line 164: There's no need to strictly compare against `true` or `false` when the other side can only be a `boolean`

src/Tracing/TransactionContext.php
```diff
@@ -97,4 +127,57 @@ public static function fromSentryTrace(string $header)
...
+            /**
+             * The request comes from an old SDK which does not support Dynamic Sampling.
+             * Propagate the Dynamic Sampling Context as is, but frozen, even without sentry-* entries.
+             */
+            $dsc->freeze();
+            $context->getMetadata()->setDynamicSamplingContext($dsc);
+        }
+
+        if (true === $hasSentryTrace && true === $dsc->hasEntries()) {
```
ste93cry, line 173: There's no need to strictly compare against `true` or `false` when the other side can only be a `boolean`

src/Tracing/TransactionContext.php
```diff
@@ -97,4 +127,57 @@ public static function fromSentryTrace(string $header)
...
         return $context;
     }
+
+    /**
+     * Returns a context populated with the data of the given headers.
+     *
+     * @param string $sentryTraceHeader The sentry-trace header from an incoming request
+     * @param string $baggageHeader     The baggage header from an incoming request
+     *
+     * @return self
```
ste93cry, line 137: Please add the return typehint to the function and remove it from here

tests/Tracing/TransactionContextTest.php
```diff
@@ -80,4 +85,98 @@ public function fromSentryTraceDataProvider(): iterable
...
     }
+
+    /**
+     * @dataProvider fromHeadersDataProvider
+     */
+    public function testFromHeaders(string $sentryTraceHeader, string $baggageHeader, ?SpanId $expectedSpanId, ?TraceId $expectedTraceId, ?bool $expectedParentSampled, ?string $expectedDynamicSamplingContextClass, ?bool $expectedDynamicSamplingContextFrozen)
+    {
+        $spanContext = TransactionContext::fromHeaders($sentryTraceHeader, $baggageHeader);
+
+        if (null !== $expectedSpanId) {
```
ste93cry, line 96: Calling `assertEquals()` should be enough, no matter what's the expected value is, because it covers also the case where `$expectedSpanId === null`

tests/Tracing/DynamicSamplingContextTest.php
```diff
@@ -0,0 +1,181 @@
...
+                'public_key' => 'public',
+                'sample_rate' => '1',
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider toStringDataProvider
+     */
+    public function testToString(string $header, string $expectedString): void
```
ste93cry, line 153: I suggest to return the `DynamicSamplingContext` object instance directly from the data provider so that it makes it clear that we're not interested in how it is constructed but just that it gets represented correctly when casted to a `string`

tests/Tracing/DynamicSamplingContextTest.php
```diff
@@ -0,0 +1,181 @@
...
+    /**
+     * @dataProvider getEntriesDataProvider
+     */
+    public function testGetEntries(string $header, array $expectedDynamicSamplingContext): void
+    {
+        $dsc = DynamicSamplingContext::fromHeader($header);
+        $this->assertSame($expectedDynamicSamplingContext, $dsc->getEntries());
+    }
+
+    public function getEntriesDataProvider()
```
ste93cry, line 124:
```suggestion
    public function getEntriesDataProvider(): \Generator
```

tests/Tracing/DynamicSamplingContextTest.php
```diff
@@ -0,0 +1,181 @@
...
+    /**
+     * @dataProvider toStringDataProvider
+     */
+    public function testToString(string $header, string $expectedString): void
+    {
+        $dsc = DynamicSamplingContext::fromHeader($header);
+        $this->assertSame($expectedString, (string) $dsc);
+    }
+
+    public function toStringDataProvider()
```
ste93cry, line 159:
```suggestion
    public function toStringDataProvider(): \Generator
```

tests/Tracing/DynamicSamplingContextTest.php
```diff
@@ -0,0 +1,181 @@
...
+        $this->assertSame($expectedTraceId, $dsc->get('trace_id'));
+        $this->assertSame($expectedPublicKey, $dsc->get('public_key'));
+        $this->assertSame($expectedSampleRate, $dsc->get('sample_rate'));
+        $this->assertSame($expectedRelease, $dsc->get('release'));
+        $this->assertSame($expectedEnvironment, $dsc->get('environment'));
+        $this->assertSame($expectedUserSegment, $dsc->get('user_segment'));
+        $this->assertSame($expectedTransaction, $dsc->get('transaction'));
+    }
+
+    public function fromHeaderDataProvider()
```
ste93cry, line 43:
```suggestion
    public function fromHeaderDataProvider(): \Generator
```

tests/Serializer/PayloadSerializerTest.php
```diff
@@ -418,12 +426,16 @@ public function serializeDataProvider(): iterable
         $event = Event::createTransaction(new EventId('fc9442f5aef34234bb22b9a615e30ccd'));
         $event->setSpans([$span1, $span2]);
 
+        $dsc = DynamicSamplingContext::fromHeader('sentry-public_key=public,sentry-trace_id=d49d9bf66f13450b81f65bc51cf49c03,sentry-sample_rate=1');
+        $event->setSdkMetadata('dynamic_sampling_context', $dsc);
```
ste93cry, line 430: Shouldn't this be a new test case so that we keep covering a transaction event that does not have the SDK metadata set?

ste93cry (changes requested):

tests/Serializer/PayloadSerializerTest.php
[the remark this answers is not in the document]
ste93cry: There is no need to reset variables, `setUp()` is called before each test and so their value gets overwritten anyway

src/Tracing/DynamicSamplingContext.php
ste93cry: I wonder if we should throw an exception rather than silently doing nothing... I mean, if changing the context after it has been frozen is an error then it should be reported as such

src/Event.php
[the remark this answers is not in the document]
ste93cry: Aligning the annotations like this is common standard in PHP 😉

src/State/Hub.php
ste93cry: For experience, I can tell you that when you get used to it it's not a big deal, and many projects use such style. There is a reason we decided long ago to use Yoda style: it prevents misassignation of variables. For example, a bug could be introduced if you do this:

```php
if ($this->foo == 'bar') {}
```

While using Yoda style would prevent it entirely because you can't assign to a constant or literal value

src/Tracing/TransactionSource.php
```diff
@@ -0,0 +1,90 @@
...
+declare(strict_types=1);
+
+namespace Sentry\Tracing;
+
+/**
+ * This enum represents all the possible types of transaction sources.
+ *
+ * @see https://develop.sentry.dev/sdk/event-payloads/transaction/#transaction-annotations
+ */
+final class TransactionSource implements \Stringable
```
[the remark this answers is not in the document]
ste93cry: It's a mistake that the other classes are missing those protections: if it's a singleton, then it makes no sense to allow it to be cloned nor serialized as you may have back a different instance object.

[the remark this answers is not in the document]
cleptric: We can follow up on this at a later point. This is nothing critical.

src/UserDataBag.php
```diff
@@ -159,6 +175,24 @@ public function setEmail(?string $email): void
...
+     */
+    public function getSegment(): ?string
+    {
+        return $this->segment;
+    }
+
+    /**
+     * Sets the segment of the user.
+     *
+     * @param string|null $segment The email
```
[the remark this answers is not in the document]
cleptric: Every setter in this class has an argument annotation. I do agree that it's mostly boilerplate, but I would leave it for consistency.

src/Tracing/DynamicSamplingContext.php
cleptric: We should be a bit careful when to throw/trigger errors.

src/Serializer/PayloadSerializer.php
```diff
@@ -175,17 +194,34 @@ public function toArray(Event $event): array
 
     private function serializeAsEnvelope(Event $event): string
     {
-        $envelopeHeader = JSON::encode([
+        /**
+         * @see https://develop.sentry.dev/sdk/envelopes/#envelope-headers
+         */
```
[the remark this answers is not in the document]
cleptric: I would expect our code-style checks to catch this.

src/Tracing/TransactionSource.php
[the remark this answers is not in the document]
ste93cry: It's a bug, definitely. I don't mind if we don't fix the other classes here as it would be out of scope, but the singleton must be implemented properly for the class we're introducing here at least.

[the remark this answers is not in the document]
cleptric: We didn't run into this yet, so it's not a bug. Please create an issue so we can tackle this at a later stage.

src/Tracing/DynamicSamplingContext.php
```diff
@@ -0,0 +1,201 @@
...
+
+            $keyValue = trim($keyValueAndProperties[0]);
+
+            if (empty($keyValue) || !str_contains($keyValue, '=')) {
+                continue;
+            }
+
+            [$key, $value] = explode('=', $keyValue, 2);
+
+            if (0 === strpos($key, self::SENTRY_ENTRY_PREFIX)) {
```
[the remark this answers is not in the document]
ste93cry: There is nothing wrong in using polyfills, especially if they enable access to functions that simplify the reading of the codebase. `str_starts_with()` is way clearer than `0 === strpos()` for the reader at a first glance, and it allows a code running on PHP 8 to be faster because it will use the native function. Last, as I said there are other places where we are already using the polyfilled functions, so it really doesn't make sense to retract now.

cleptric:
Thanks for the feedback everyone. I'm planning to merge this tomorrow.

ste93cry (approved):

cleptric:
@ste93cry Next time, please don’t update my branches without consulting me first.
You are always welcome to branch off from my branch and propose your changes in a separate PR.

While I appreciate the updates to the test suite, I’ll have to revert your last commit partially, as my team and I agree on not adding further deprecation warnings to the code base.
We added an @deprecated annotation, we will point this out more clearly in the changelog and will cover this as well in the migration guide for v4.0.

ste93cry:
>We added an @deprecated annotation, we will point this out more clearly in the changelog and will cover this as well in the migration guide for v4.0.

People don't care about docblocks, they don't even see them in their day-to-day work and there is no way for them to know that something got deprecated if they don't look specifically to each doblock each time. `@trigger_error` is the standard way nowadays in PHP projects to let them know that some method is gonna be deprecated and removed. 

>as my team and I agree on not adding further deprecation warnings to the code base

I would love to hear the reasons behind this choice. Furthermore, I would like to add that, while we're at it, we're all in the same team, and so if such a decision was taken, at least it should have been communicated to the most active contributors of this project :sweat_smile:

cleptric:
Nothing is wrong if people continue using `TransactionContext::fromSentryTrace`, if they don't want to use Dynamic Sampling.
We will likely release this before Symfony and Laravel, so I rather not add deprecation warnings to people's apps, without a way for them to fix it.

We discussed this in our daily, the team is @getsentry/team-web-sdk-backend.

mfb:
Having just a @deprecated annotation gets a thumbs up from me, as static analysis tools can easily pick up on these annotations and warn the developer.  A deprecation could always be "escalated" later on by adding trigger_error() call, when it's more actionable.

vladanpaunovic:
I wonder how many notifications `https://github.com/deprecated` gets because of `@deprecated` :D

PR #1365, opened 2022-09-19
 edit code style :) 

YasinAshorian:


ste93cry:
To be honest, I don't understand the purpose of this PR, nor its changes. Using `is_null` makes the code less pleasent to read and also less performant because you need to call a function... The rest of the commits are from `master`, and will be ported to `develop` with the next branch merge we do.

YasinAshorian (commented):
Excellent :)

PR #1366, opened 2022-09-19
Code simplification

YasinAshorian:


cleptric:
@YasinAshorian You can run our coding style checks with `composer phpcs`.
You may notice that it recommends reverting your changes back to `null` checks.

https://github.com/getsentry/sentry-php/actions/runs/3080739864/jobs/4979133543

Issue #1348, opened 2022-08-17
Improve release flow


cleptric:
A few ideas on how we can make the whole release process a bit more straightforward and align it better with other SDKs.

### Generate changelogs

Like on the Python SDK, we should enable `changelogPolicy: auto` so changelogs will be updated and pushed automatically. This will require us to be stricter about the commit messages, but this is already encouraged.

### Set version number

~~Like on the Python SDK, we should expose the SDK version number in the code and update it automatically via a new release.~~ done in https://github.com/getsentry/sentry-php/pull/1367

Jean85:
> This will require us to be stricter about the commit messages, but this is already encouraged.

Since we do squash-merge on PRs, that shouldn't be too hard.

> Like on the Python SDK, we should expose the SDK version number in the code and update it automatically via a new release.

This is already possible in a dynamic way using `jean85/pretty-package-versions`, which is already a dependency, and it exposes the version reading from Composer. It's already used to show it in events.

PR #1367, opened 2022-09-19
ref: Use constant for the SDK version

cleptric:
As we do make quite a few assumptions in our ingestion pipeline about the SDK version, I added a new constant instead of relying on the composer version.

During DS (#1360) development, we noticed that my composer alias (dev-dsc) broke quite a few things in the UI, so this will solve such issues in the future.
This is also in line with our other SDKs, such as [Python](https://github.com/getsentry/sentry-python/commit/59dea5254506770b3d53fd4e8496516704489611) 

I added the necessary `bump-version.sh` script, which will update and commit the version once we run the `publish-release` workflow.
https://github.com/getsentry/craft#pre-release-version-bumping-script-conventions

refs: https://github.com/getsentry/sentry-php/issues/1348

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -7,7 +7,6 @@ statusProvider:
   name: github
 artifactProvider:
   name: none
-preReleaseCommand: ""
 targets:
   - name: github
   - name: registry
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Use constant for the SDK version (#1367)
+
 ## 3.8.0 (2022-09-05)
 
 - Add `Sentry\Monolog\BreadcrumbHandler`, a Monolog handler to allow registration of logs as breadcrumbs (#1199)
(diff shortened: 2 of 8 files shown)


stayallive (approved):
Only thought that popped up is that the script needs to be known in the `.craft.yml` but maybe it works out of the box by convention (as this also seems to work in the Python repo without explicitely added to the Craft config). At the very least we could remove the empty `preReleaseCommand` maybe by now? But not sure, the Craft wizards will have to advise here. 

Otherwise this looks good!

cleptric:
@stayallive Good catch! I removed the empty `preReleaseCommand` for now and pinged our craft mastermind about this.

cleptric:
According to https://github.com/getsentry/craft#pre-release-command, `preReleaseCommand` is set to `preReleaseCommand: bash scripts/bump-version.sh` by default.

stayallive:
Ahh, nice, okay then it should work 😎

Issue #1368, opened 2022-09-19
Tracing/SpanContext->setTags() should work the same was as Tracing/Span->setTags()


Allieway:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

latest

### Steps to Reproduce

```
$context->setTags(['test' => 'blah']);
$context->getTags(); // ['test' => 'blah']
$context->setTags(['test1' => 'blah1']);
$context->getTags(); // ['test1' => 'blah1']
```

VS

```
$span->setTags(['test' => 'blah']);
$span->getTags(); // ['test' => 'blah']
$span->setTags(['test1' => 'blah1']);
$span->getTags(); // ['test' => 'blah', 'test1' => 'blah1']
```
This applies to Transactions and Spans and their corresponding contexts, since transactions extend spans in both situations.

It would just require changing this line: 
https://github.com/getsentry/sentry-php/blob/c0037a571336722a8853068d56e2e6ecfdf1ee1f/src/Tracing/SpanContext.php#L152
to match this line:
https://github.com/getsentry/sentry-php/blob/c0037a571336722a8853068d56e2e6ecfdf1ee1f/src/Tracing/Span.php#L270

From what I can tell this wouldn't break anything in the SDK or anyone's existing implementation of the SDK.

array_merge() replaces the first array's value with the second if the keys match. This ensures the last call takes priority.

For SDK: 
no duplicate keys to break anything and if context is passed to startTransaction/Span or new Span/Transaction it will pass on its tags. 

For implementers currently:
1. They don't have any problem currently. 
2. They figured out the same problem I did and are manually using array_merge before passing to setTags()
3. They have no idea anything is wrong, they are calling setTags() twice, and have tags being lost.
For implementers after change:
1. No change
2. No change
3. Fixed without them noticing!

### Expected Result

Merge existing tags with setTags()

### Actual Result

Replace existing tags with setTags()

cleptric:
Once you pass the `SpanContext` into the `Span` constructor, you should not modify the context anymore, and instead, mutate the span. The `SpanContext::setTags()` method reflects this design decision.

Allieway:
Thanks for the comment @cleptric 
I understand not wanting to modify context once it's passed. I just am setting tags multiple times in the context before passing it to span. I don't see any logic preventing someone from modifying context after starting a span with that context. So I'm unsure how merging tag arrays was designed with that in mind.
It just makes sense to me to have the logic work the same way for both objects in public methods that have the same name.

PR #1371, opened 2022-09-20
Span context set tags update

Allieway:
Information is here: https://github.com/getsentry/sentry-php/issues/1368


cleptric (changes requested):
This changes the behaviour of the setter in a BC-breaking way.
I still don't get the whole picture of what this would solve, so I'm reluctant of merging it.

Issue #1362, opened 2022-09-15
Transactions not showing up in sentry when using manual instrumentation.


Allieway:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

newest

### Steps to Reproduce
I might be doing something wrong but my code looks correct, this is a lot of info, see next bolded text for TL;DR

I have a trace that starts with a transaction and has 13 spans within in, some span children of the transaction, and other span children of those children. 
I close spans in order, each span has the correct tags thanks to my array_merge before setting context tags. 
I set the description of the parent transaction to the url path I'm implementing manual instrumentation on, and each other spans description to it's index in the full trace.
But the description of the top level span is set to "Index 13" -- but so is the last child span "Index 13"
The tags in the main section of the event details are the same ones as the child "Index 13", as are all the other attributes, including "parentSpanId". 
The root parent has no tags when I click on it in event details, but it does have the same "parentSpanId" and other default filled fields. 
When I go to the "Trace View" page it doesn't show children, only the root span. That root span is "Index 13" and does not have a "parentSpanId" property on that page view, but the "spanId" property matches the last child span with the same description(and the root span that I believe might be working incorrectly). 
I var_dumped as far as the eye can see. Before and After I modified any context(transaction and span), before and after I modified the transaction, before and after I modified all spans, and before and after I called ->finish() on spans and transactions. 
After ->finish() is called on the transaction object, the var_dump of that object is the very last output on the page. It has the correct tags, properties, name, and ID. It has an endTimeStamp and sampled is true.
The only reference to that ID on the event details page is in the parentSpanId field. There are no properties, name, or tags that match my output of the root transaction object after->finish(); is called(or anytime that transaction is output) on that same page. 

**TL;DR**:
Something is happening that causes the root span(transaction?) to have the properties of the last child span on the transaction summary page. Even the "Trace Details" section of the event contain the last spans properties, including parentSpanId. 
Keep in mind, I touched PHP, Sentry, and APM in general all for the first time 2 weeks ago, so it's possible I'm hugely misunderstanding what's going on. Below is the transaction output with details scrubbed, I may contact sentry support as well? Honestly I'm not sure what I should do at this point. 

```
object(Sentry\Tracing\Transaction)
private 'name' => string 'Nowhere to be found on event details page'
protected 'spanId' => 
    object(Sentry\Tracing\SpanId)
      private 'value' => string 'only visible as parentSpanId property'
protected 'traceId' => 
[...]

PR #1372, opened 2022-09-20
Transaction context priority over scope in event payload.

Allieway:
Information here: https://github.com/getsentry/sentry-php/issues/1362


cleptric (changes requested):
Please provide us with a minimal reproducible example or a concise explanation about what this would solve.

PR #1374, opened 2022-09-21
ref: Use constant for the SDK version

cleptric:
As master is currently [failing](https://github.com/getsentry/sentry-php/actions/runs/3096549882/jobs/5012498359), I decided to promote this to a bugfix.

That might be a bit awkward when merging it back into `develop`, but I think we can make it work.

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -7,7 +7,6 @@ statusProvider:
   name: github
 artifactProvider:
   name: none
-preReleaseCommand: ""
 targets:
   - name: github
   - name: registry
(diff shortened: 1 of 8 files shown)


stayallive (approved):

rvanlaak:
Were there still other usages of `PrettyVersion`, or could the dependency on that package get removed as well?

stayallive:
Yes, we have a [`ModulesIntegration`](https://github.com/getsentry/sentry-php/blob/e5921546c9894b9f79d9b00acc704bad126b43c6/src/Integration/ModulesIntegration.php) still depending on it 😉

cleptric:
@rvanlaak It's currently still used in our optional `ModulesIntegration`.
I'll work on that topic during #1335

PR #1378, opened 2022-09-24
Expose a function to retrieve the URL of the CSP endpoint

### LINKED ISSUE #1363: Add end-user function to generate CSP report-uri
### Problem Statement

It is possible to use Sentry as endpoint for collecting CSP violations, see https://docs.sentry.io/product/security-policy-reporting/. In order to do so, it requires passing a specific report-uri value to the CSP header in the response of your application. It would be great if the SDK would be able to generate such an URI.

### Solution Brainstorm

The [Raven: Sentry Integration](https://www.drupal.org/project/raven) module for Drupal has an implentation of this, see https://git.drupalcode.org/project/raven/-/blob/4.x/src/Plugin/CspReportingHandler/Raven.php. It would be great to have a public function for this URI in any application using Sentry PHP SDK.

ste93cry:
Fixes #1363. This is just a backport of what the Ruby SDK does, nothing more nothing less.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Expose a function to retrieve the URL of the CSP endpoint (#1378)
+
 ## 3.8.1 (2022-09-21)
 
 - fix: Use constant for the SDK version (#1374)
--- a/src/Client.php
+++ b/src/Client.php
@@ -120,6 +120,30 @@ final class Client implements ClientInterface
     /**
      * {@inheritdoc}
      */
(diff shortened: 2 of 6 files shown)


stayallive (approved):

cleptric:
@ste93cry Could you please also open a PR to update our docs?

cleptric (approved):

cleptric (commented):
There should also be an update to the `CHANGELOG.md`.

ste93cry:
Well, after years it seems I still forget to do it :sweat_smile:

PR #1379, opened 2022-09-26
test: Check if the correct trace_id is applied to the event context

cleptric:
Follow up on #1362 & #1372

While our current implementation is correct, we did not have any test case that covers it.

Setting the correct `trace_id` to the context is important to be able to search for transactions with a `trace_id` inside Discover.


stayallive (approved):
👌

PR #1380, opened 2022-09-27
fix: Merge hint extra's from captureMessage in event extra's when pre…

Celine68:
…paring event

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- fix: Merge hint extra's from captureMessage in event extra's when preparing event (#1380)
+
 ## 3.8.1 (2022-09-21)
 
 - fix: Use constant for the SDK version (#1374)
--- a/src/Client.php
+++ b/src/Client.php
@@ -230,6 +230,8 @@ final class Client implements ClientInterface
             if (null !== $hint->stacktrace && null === $event->getStacktrace()) {
                 $event->setStacktrace($hint->stacktrace);
             }
+
+            $event->setExtra(array_merge($event->getExtra(), $hint->extra));
         }
 
         $this->addMissingStacktraceToEvent($event);


cleptric (changes requested):
This will need a test case.

Mind sharing why you're proposing this change as well? 
Is there something not working on the product (https://sentry.io)?

Gwemox:
Yes, the captureMessage method does not process the data provided in EventHint->extra : 
https://forum.sentry.io/t/eventhint-extra-does-nothing-when-using-capturemessage/15836

cleptric:
@Gwemox Looks like the link is not working.
From our develop docs, https://develop.sentry.dev/sdk/unified-api/#terminology

> extra: Truly arbitrary data attached by client users. This is a deprecated feature but will continue to be supported for the foreseeable future. Users are encouraged to use contexts instead.

Celine68:
@cleptric  It's impossible to add event context through the captureMessage function. Is the captureMessage function deprectaed then ? Because it's not indicated.

stayallive:
This data should not be in the `EventHint` but in the scope, and also it shouldn't be in the `extra` anymore: https://docs.sentry.io/platforms/php/enriching-events/context/#additional-data.

You can set "extra" structured context like this: https://docs.sentry.io/platforms/php/enriching-events/context/#structured-context

In your case the second example might be exactly what you are looking for:

```php
\Sentry\withScope(function (\Sentry\State\Scope $scope) use ($e): void {
    $scope->setContext('character', [
        'name' => 'Mighty Fighter',
        'age' => 19,
        'attack_type' => 'melee'
    ]);

    \Sentry\captureMessage('The fighter is out of energy!');
    // or: \Sentry\captureException($e);
});
```

`withScope` only applies to the callback so it doesn't modify the global scope.

Celine68:
A huge thanks @stayallive  ! It does work this way !

PR #1382, opened 2022-09-28
feat: Add setSource to TransactionContext

cleptric:
The current API of setting the `TransactionSource` is pretty clunky...

```
$transactionContext->getMetadata()->setSource(TransactionSource::route());
```

... so I added a new setter.

```
$transactionContext->setSource(TransactionSource::route());
```

Do we need a getter as well?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 - feat: Expose a function to retrieve the URL of the CSP endpoint (#1378)
 - feat: Add support for Dynamic Sampling (#1360)
   - Add `segment` to `UserDataBag`
-  - Add `TransactionSource`, to set information about the transaction name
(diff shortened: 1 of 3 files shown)


stayallive:
Maybe removing the `get` helps to not need a getter too.

```php
$source = $transactionContext->metadata()->getSource();

$transactionContext->metadata()->setSource(TransactionSource::route());
```

Somehow this also feels less clunky, but +1 on this change anyway since I suspect this could be a commonly used setter.

stayallive (approved):

cleptric:
I still remember the time when we replaced hundreds of mixed setters/getters over at CakePHP 😅
I think having a unified naming with `get` & `set` prefixes works well.
If we feel the need arises to have `getSource()` as well, we can iterate on this.

PR #1386, opened 2022-09-30
Cancel in progress ci runs when there is a new commit pushed.

antonpirker:
Cancel in progress CI runs when a new commit is pushed to the PR. This should save some CI minutes (and concurrency)


cleptric (approved):
Thanks Anton!

PR #1389, opened 2022-09-30
meta: Check for github-actions updates

cleptric:
Let @dependabot do the work, so we don't have to 😄

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule: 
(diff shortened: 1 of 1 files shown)


YasinAshorian (approved):

stayallive (approved):

Jean85 (changes requested):
What's the behavior of this? Does it change the `composer.json`? What about the `.lock`? What does it updates everything? Or just against CVEs?

stayallive:
@Jean85 this does only `package-ecosystem: "github-actions"`.

Composer is not touched or looked at.

Jean85 (approved):

PR #1390, opened 2022-09-30
Allow a formatted message with empty params to be sent to Sentry

mfb:
Currently the PayloadSerializer logic will only send the formatted message to Sentry if the message params are not empty.

However, I would like to have the option to send both the uninterpolated message (which is used to group events into issues) and the formatted message (which is displayed in the Sentry UI) without params to Sentry.

Example use cases for this capability would be that I want to reduce the size of event payloads, or if I simply don't want the message params to be rendered in the Sentry UI.

I'm not aware of any reason for the SDK to have such logic restricting which components of the message are sent to Sentry.

### DIFF
--- a/src/Serializer/PayloadSerializer.php
+++ b/src/Serializer/PayloadSerializer.php
@@ -155,13 +155,13 @@ final class PayloadSerializer implements PayloadSerializerInterface
         }
 
         if (null !== $event->getMessage()) {
-            if (empty($event->getMessageParams())) {
+            if (empty($event->getMessageParams()) && is_null($event->getMessageFormatted())) {
                 $result['message'] = $event->getMessage();
             } else {
                 $result['message'] = [
                     'message' => $event->getMessage(),
                     'params' => $event->getMessageParams(),
-                    'formatted' => $event->getMessageFormatted() ?? vsprintf($event->getMessage(), $event->getMessageParams()),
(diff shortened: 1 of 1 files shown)


YasinAshorian (commented):
I don't understand the reason for not using php features !!

mfb:
> I don't understand the reason for not using php features !!

If you want to propose changing the coding standards, this PR isn't the place and exclamation points aren't helpful. I'm generally agnostic about what they are as long as they are enforced, since they are different for every project and it's hard to keep track :p

Also for the record, I was able to workaround the issue that inspired this PR, so I no longer really *need* it, but it seems like a nice-to-have.

YasinAshorian:
> > I don't understand the reason for not using php features !!
> 
> If you want to propose changing the coding standards, this PR isn't the place and exclamation points aren't helpful. I'm generally agnostic about what they are as long as they are enforced, since they are different for every project and it's hard to keep track :p
> 
> Also for the record, I was able to workaround the issue that inspired this PR, so I no longer really _need_ it, but it seems like a nice-to-have.

Maybe you are right

Jean85 (changes requested):
I can understand this, but we can't merge this as it is: a changelog entry and some tests are required.

Also, the combination of `??` and a ternary is not very readable..

mfb:
I won't have time to fix up this PR, so if anyone else is interested please take it over, or we close it. I'd agree that refactoring for readability is needed.

cleptric:
Closing this for now. Happy to re-open if someone wants to tackle this.

PR #1401, opened 2022-10-11
fix: Suppress errors on is_callable

cleptric:
Reproduction case https://github.com/NotGeri/yii-test

The `is_callable` is triggering a warning in this case. As we already put it into a try/catch, I'm ok with suppressing eventual errors here as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


NotGeri:
Thank you again, just wanted to confirm that the fix seems to work fine. Really appreciate the quick help!

PR #1388, opened 2022-09-30
Feat: Allow passing an eventId as context

jwissels:
We would like to pass an `eventId` as a parameter in the `$record` so that we can communicate the `eventId` to other handlers in the logger.

PR #1403, opened 2022-10-12
Feat: Allow passing an eventId as context

jwissels:
Had to open a new PR because i couldn't change source branch in the original PR ( https://github.com/getsentry/sentry-php/pull/1388 ) 

We would like to pass an `eventId` as a parameter in the `$record` so that we can communicate the `eventId` to other handlers in the logger.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -51,10 +51,13 @@
       <code>$record['level']</code>
       <code>$record['message']</code>
     </PossiblyInvalidArgument>
-    <PossiblyInvalidArrayOffset occurrences="1">
+    <PossiblyInvalidArrayOffset occurrences="2">
+      <code>$record['context']['event_id']</code>
       <code>$record['context']['exception']</code>
     </PossiblyInvalidArrayOffset>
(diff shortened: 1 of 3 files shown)


jwissels:
@cleptric is there a way i could get this on the backlog again? Was away for the holidays and missed the bot. 
Would really appreciate this feature

cleptric:
Could you extend the use case a bit? I'm a bit hesitant to allow creating Event IDs based on abatiary context data.

jwissels:
We use https://github.com/Seldaek/monolog to log to multiple destinations, sentry and slack most notably. 
We'd like to provide the `id` so that in the message in slack we can link straight to the exception in sentry.

cleptric:
Could you not use Sentry's Slack integration to do this?

jwissels:
We use business logic on when we want to log to slack which is hard to replicate in sentry. We also enrich the exception differently. 

In previous versions ( with the `RavenHandler` ) it was possible to pass an `id` via context data. Is there maybe a more elegant way then context data to get the `id`? I'm happy to refactor.

PR #1407, opened 2022-10-18
Fix sampling decision when parent is sampled

### LINKED ISSUE #1405: `Hub::startTransaction` ignores `sentry-trace` sampling decision
### Problem Statement

From https://github.com/getsentry/sentry-laravel/issues/595

https://github.com/getsentry/sentry-php/blob/27c8b04c5e1e56ced4ccd28bf518f1265413153c/src/State/Hub.php#L225-L229

If the SDK is not configured for tracing, we currently falsely ignore the sampling decision made by the head SDK.

This might have an impact on existing applications, but given the fact that the `sentry-trace` header has to be manually enabled via `tracingOrigins`, I think we can add the new behaviour as a bug fix release.

stayallive:
The current sampling decision always set the transaction to not sampled if sampling is disabled in the SDK not taking in account the parent sampling decision which is considered a bug.

This resolved that by also looking at the parent sampling decision when starting a transaction.

Fixes #1405.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- fix: Sampling now correctly takes in account the parent sampling decision if available instead of always being `false` when tracing is disabled (#1407)
+
 ## 3.9.1 (2022-10-11)
 
 - fix: Suppress errors on is_callable (#1401)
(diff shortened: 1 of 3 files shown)


cleptric (approved):
Changelog and 🚀

PR #1415, opened 2022-10-21
ref: Enable the ModulesIntegration by default

### LINKED ISSUE #1355: Add auto-discovery of popular PHP frameworks
### Problem Statement

I would like to know which PHP frameworks are people using with our [sentry/sdk](https://packagist.org/packages/sentry/sdk) so we can make a better informed decisions on which frameworks to support more and when.

We already have a [modules integration](https://docs.sentry.io/platforms/php/integrations/#modulesintegration) which is opt in and very few people know about it, hence few of them are actually using it. So the information we get from there are not sufficient enough to drive any meaningful decisions.

### Solution Brainstorm

1. create a list of top 10 php frameworks
2. on `sentry.init()` check whether users framework is in one of top 10
    -  ideally, we should find alternative/better methods to discover a framework and fallback to `composer.json` as a last resort in order to minimise the impact on customers overhead.
3. utilise the same `modules` object on [event payload](https://develop.sentry.dev/sdk/event-payloads/) to store the framework like:
```php
'modules' => [
  'some/framework' => '1.2.3', // if version is unknown to us, use unknown
  ...
]
```

cleptric:
Closes #1355

### DIFF
--- a/src/Integration/IntegrationRegistry.php
+++ b/src/Integration/IntegrationRegistry.php
@@ -131,6 +131,7 @@ final class IntegrationRegistry
             new TransactionIntegration(),
             new FrameContextifierIntegration(),
             new EnvironmentIntegration(),
+            new ModulesIntegration(),
         ];
 
         if (null !== $options->getDsn()) {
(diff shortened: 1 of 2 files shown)


stayallive (approved):
👍

PR #1418, opened 2022-10-24
Decouple from "guzzlehttp/psr7" and "php-http/discovery", replace by "friendsofphp/well-known-implementations"

nicolas-grekas:
Related to https://github.com/getsentry/sentry-php/discussions/1411 and https://github.com/FriendsOfPHP/well-known-implementations/issues/2

In this PR, I propose to rely on [FriendsOfPHP/well-known-implementations](https://github.com/FriendsOfPHP/well-known-implementations) to ensure a working out-of-the-box experience. This package is a composer-plugin that will install the most suited `async-http-implementation` depending on the existing dependencies of the projects that require sentry-php.

The package also provides a discovery mechanism that covers roughly the same purpose as `php-http/discovery`, but that plays well with the plugin part of the package.

The attached patch removes `php-http/discovery` from the mandatory dependencies, but also removes the dependency on `guzzlehttp/psr7`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,12 +23,11 @@
         "php": "^7.2|^8.0",
         "ext-json": "*",
         "ext-mbstring": "*",
+        "friendsofphp/well-known-implementations": "^1",
         "guzzlehttp/promises": "^1.4",
-        "guzzlehttp/psr7": "^1.8.4|^2.1.1",
         "jean85/pretty-package-versions": "^1.5|^2.0.4",
         "php-http/async-client-implementation": "^1.0",
         "php-http/client-common": "^1.5|^2.0",
-        "php-http/discovery": "^1.11",
         "php-http/httplug": "^1.1|^2.0",
         "php-http/message": "^1.5",
         "psr/http-factory": "^1.0",
@@ -39,6 +38,7 @@
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.19|3.4.*",
+        "guzzlehttp/psr7": "^1.8.4|^2.1.1",
         "http-interop/http-factory-guzzle": "^1.0",
         "monolog/monolog": "^1.6|^2.0|^3.0",
         "nikic/php-parser": "^4.10.3",
@@ -89,6 +89,7 @@
         "sort-packages": true,
         "allow-plugins": {
             "composer/package-versions-deprecated": true,
+            "friendsofphp/well-known-implementations": false,
             "phpstan/extension-installer": true
         }
     },
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace Sentry;
 
-use Http\Discovery\Psr17FactoryDiscovery;
+use FriendsOfPHP\WellKnownImplementations\WellKnownPsr17Factory;
 use Psr\Log\LoggerInterface;
 use Sentry\HttpClient\HttpClientFactory;
 use Sentry\Serializer\RepresentationSerializerInterface;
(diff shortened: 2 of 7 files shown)


composer.json
```diff
@@ -89,6 +89,7 @@
         "sort-packages": true,
         "allow-plugins": {
             "composer/package-versions-deprecated": true,
+            "friendsofphp/well-known-implementations": false,
```
nicolas-grekas, line 92: Running the plugin is not needed for tests. Only the "lib" part of the package is used when running the CI.

src/ClientBuilder.php
```diff
@@ -175,19 +175,19 @@ private function createTransportInstance(): TransportInterface
...
     private function createDefaultTransportFactory(): DefaultTransportFactory
     {
-        $streamFactory = Psr17FactoryDiscovery::findStreamFactory();
+        $psr17Factory = new WellKnownPsr17Factory();
         $httpClientFactory = new HttpClientFactory(
-            Psr17FactoryDiscovery::findUriFactory(),
-            Psr17FactoryDiscovery::findResponseFactory(),
-            $streamFactory,
+            null,
+            null,
```
nicolas-grekas, line 181: The first two arguments of `HttpClientFactory` are never used. We can thus make them nullable and skip injecting and factory there.

src/HttpClient/HttpClientFactory.php
```diff
@@ -54,16 +55,16 @@ final class HttpClientFactory implements HttpClientFactoryInterface
...
      * @param StreamFactoryInterface        $streamFactory   The PSR-17 stream factory
      * @param HttpAsyncClientInterface|null $httpClient      The HTTP client
      * @param string                        $sdkIdentifier   The SDK identifier
      * @param string                        $sdkVersion      The SDK version
      */
     public function __construct(
-        UriFactoryInterface $uriFactory,
-        ResponseFactoryInterface $responseFactory,
+        ?UriFactoryInterface $uriFactory,
+        ?ResponseFactoryInterface $responseFactory,
```
nicolas-grekas, line 67: Those arguments are never used, no need to force passing anything.

src/HttpClient/HttpClientFactory.php
```diff
@@ -110,7 +111,7 @@ public function create(Options $options): HttpAsyncClientInterface
      */
     private function resolveClient(Options $options)
     {
-        if (class_exists(SymfonyHttplugClient::class)) {
+        if (ConcreteImplementation::VENDOR_SYMFONY === ConcreteImplementation::HTTPLUG_VENDOR) {
```
nicolas-grekas, line 114: Using these consts gives more control over which implementation should be used (the consts are generated by the plugin xor computed once at autoload-time when the plugin is disabled)

src/Integration/RequestFetcher.php
```diff
@@ -22,6 +22,6 @@ public function fetchRequest(): ?ServerRequestInterface
             return null;
         }
 
-        return ServerRequest::fromGlobals();
+        return (new WellKnownPsr17Factory())->createServerRequestFromGlobals();
```
nicolas-grekas, line 25: This removes the "require" coupling to "guzzlehttp/psr7": "well-known-implem" provides the same feature, but compatible with any PSR-7 implementations.

tests/bootstrap.php
```diff
@@ -12,8 +10,6 @@
 
 require_once __DIR__ . '/../vendor/autoload.php';
 
-ClassDiscovery::appendStrategy(MockClientStrategy::class);
```
nicolas-grekas, line 15: "well-known-implem" will discover and auto-load "php-http/mock-client" when no other implem is found.
This decouples from "php-http/discovery".

cleptric:
Thanks for proposing this @nicolas-grekas 

This is definitely something we will consider when making a decision on the path forward in the next major of the SDK.

nicolas-grekas:
@cleptric here is the result of what we talked about at SymfonyCon: https://github.com/php-http/discovery/pull/208

Issue #1427, opened 2022-11-10
Traces sample rate cannot be disabled


koenhoeijmakers:
### Environment
sentry/sentry-laravel: 3.0.1
laravel/framework: 9.39.0

### Steps to Reproduce

1. Use the default config file containing: `'traces_sample_rate' => (float) (env('SENTRY_TRACES_SAMPLE_RATE', 0.0)),`
2. 
![image](https://user-images.githubusercontent.com/2232776/201041212-8979a9a2-fca0-4d58-877e-0be93220dd8c.png)

### Expected Result

I would expect Performance tracing to be disabled

### Actual Result

It is not disabled and we're almost at the point where we have to upgrade our sentry plan because we're running out of events rapidly

cleptric:
I moved this to the core SDK, as the issue needs to be fixed there.

I think we are missing the following behavior:

- By default, `traces_sample_rate` should be `null`, so we do exit early and do not take anything else into account.
- If a user applies `traces_sample_rate` `0` or higher, we check in the order of `traces_sampler` -> `parent_sampled` -> `traces_sample_rate`.

@stayallive How does this sound to you?

mfb:
So zero doesn't really mean zero I guess? I will have to document this carefully in my downstream package :)   Having a toggle + sample rate seems potentially more intuitive.

stayallive:
@cleptric this sounds good to me, as long as we document this correctly. Are you aware if other SDKs handle it a similar way?

Currently the docs aren't a 100% correct either too: https://docs.sentry.io/platforms/php/configuration/options/#tracing-options

cleptric:
Yes, our Python SDK uses `None` https://github.com/getsentry/sentry-python/blob/e6238d828e11d63833b9a1400aaf8286b05d1c02/sentry_sdk/tracing_utils.py#L113##

cleptric:
Fixed in `3.12.0` https://github.com/getsentry/sentry-php/releases/tag/3.12.0

PR #1428, opened 2022-11-11
fix: Set `traces_sample_rate` to `null` by default

cleptric:
The new default value for `traces_sample_rate` is now `null`.

If neither `traces_sample_rate: 0` nor a `traces_sampler` is set in the options, we do not enable tracing.

Closes https://github.com/getsentry/sentry-php/issues/1427

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


sl0thentr0py (approved):

stayallive (approved):
👍

PR #1436, opened 2022-11-27
Feature: attachments

### LINKED ISSUE #996: how to attach uploaded files?
people will upload excels to our app but the excel contents may not fit to our requirements and sometimes our app will crash... therefore, it could be easier for us if we can have the file to test.

jiripudil:
closes #996

### DIFF
--- a/src/Attachment.php
+++ b/src/Attachment.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry;
+
+final class Attachment
+{
+    public const CONTENT_TYPE_DEFAULT = 'application/octet-stream';
+
+    /**
+     * @var string
+     */
+    private $filename;
+
+    /**
+     * @var string
+     */
+    private $contentType;
+
+    /**
+     * @var string
+     */
+    private $data;
+
+    public function __construct(string $data, string $filename, string $contentType = self::CONTENT_TYPE_DEFAULT)
+    {
+        $this->data = $data;
+        $this->filename = $filename;
+        $this->contentType = $contentType;
+    }
+
+    public function getFilename(): string
+    {
+        return $this->filename;
+    }
+
+    public function getContentType(): string
+    {
+        return $this->contentType;
+    }
+
+    public function getData(): string
+    {
+        return $this->data;
+    }
+
+    /**
+     * Helper method to create an Attachment from a file in the filesystem.
+     */
+    public static function fromFile(string $filename, string $contentType = self::CONTENT_TYPE_DEFAULT): self
+    {
+        $data = file_get_contents($filename);
+        if (false === $data) {
+            throw new \InvalidArgumentException("File $filename is not readable.");
+        }
+
+        return new self(
+            $data,
+            basename($filename),
+            $contentType
+        );
+    }
+}
--- a/src/Event.php
+++ b/src/Event.php
@@ -167,6 +167,11 @@ final class Event
      */
     private $type;
 
+    /**
+     * @var Attachment[] The attachments associated with this event
+     */
+    private $attachments = [];
+
     private function __construct(?EventId $eventId, EventType $eventType)
     {
         $this->id = $eventId ?? EventId::generate();
@@ -746,5 +751,25 @@ final class Event
     public function setSpans(array $spans): void
     {
         $this->spans = $spans;
+    }
+
+    public function hasAttachments(): bool
+    {
+        return [] !== $this->attachments;
+    }
+
+    /**
+     * A list of attachments associated with this event.
+     *
+     * @return Attachment[]
+     */
+    public function getAttachments(): array
+    {
+        return $this->attachments;
+    }
+
+    public function addAttachment(Attachment $attachment): void
+    {
+        $this->attachments[] = $attachment;
     }
 }
--- a/src/Serializer/PayloadSerializer.php
+++ b/src/Serializer/PayloadSerializer.php
@@ -38,7 +38,7 @@ final class PayloadSerializer implements PayloadSerializerInterface
      */
     public function serialize(Event $event): string
     {
-        if (EventType::transaction() === $event->getType()) {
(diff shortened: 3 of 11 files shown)


src/Serializer/PayloadSerializer.php
```diff
@@ -38,7 +38,7 @@ public function __construct(Options $options)
      */
     public function serialize(Event $event): string
     {
-        if (EventType::transaction() === $event->getType()) {
+        if (EventType::transaction() === $event->getType() || $event->hasAttachments()) {
```
jiripudil, line 41: This same condition is duplicated in the test and, more importantly, in HttpTransport. Perhaps it could be extracted into a separate helper method, something like `requiresEnvelope()`?

src/Event.php
```diff
@@ -167,6 +167,11 @@ final class Event
      */
     private $type;
 
+    /**
+     * @var Attachment[] The attachments associated with this event
+     */
+    private $attachments = [];
```
jiripudil, line 173: This is the main diversion from the Javascript SDK which I drew inspiration from. It puts this field into `EventHint` instead. Dunno if it makes more sense that way, but it would require passing the hint along with the event to the transport layer, and thus a huge BC break in the `TransportInterface`

src/Attachment.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        return $this->data;
+    }
+
+    /**
+     * Helper method to create an Attachment from a file in the filesystem.
+     */
+    public static function fromFile(string $filename, string $contentType = self::CONTENT_TYPE_DEFAULT): self
+    {
+        $data = file_get_contents($filename);
```
stayallive, line 53: Since we are dealing with PHP here... we are reading the full file into memory without checking if it's within limits of something we can sent to Sentry (Sentry allows at most 20MB for a compressed request, and at most 100MB of uncompressed attachments per event). Although we might let the server bounce our request instead of adding client side validation.

I wonder if it's possible to use streams to move the files instead of needing them in memory twice (once here and once when the payload is constructed). At the very least we should try to keep the memory footprint as low as possible until we are constructing the evelope.

src/Attachment.php
```diff
@@ -0,0 +1,64 @@
...
+
+    public function getData(): string
+    {
+        return $this->data;
+    }
+
+    /**
+     * Helper method to create an Attachment from a file in the filesystem.
+     */
+    public static function fromFile(string $filename, string $contentType = self::CONTENT_TYPE_DEFAULT): self
```
stayallive, line 51: It might be nice to have a `$path` and a (optional) `$filename` so you can set a more descriptive filename (different from the `basename`)?

src/Event.php
```diff
@@ -747,4 +752,24 @@ public function setSpans(array $spans): void
     {
         $this->spans = $spans;
     }
+
+    public function hasAttachments(): bool
```
stayallive, line 756: Considering every method in this file has (maybe useless) comments it is nice to also add a docbloc here.

src/Event.php
```diff
@@ -747,4 +752,24 @@ public function setSpans(array $spans): void
...
+     * A list of attachments associated with this event.
+     *
+     * @return Attachment[]
+     */
+    public function getAttachments(): array
+    {
+        return $this->attachments;
+    }
+
+    public function addAttachment(Attachment $attachment): void
```
stayallive, line 771: Considering every method in this file has (maybe useless) comments it is nice to also add a docbloc here.

src/Serializer/PayloadSerializer.php
```diff
@@ -215,12 +215,32 @@ private function serializeAsEnvelope(Event $event): string
...
+        ];
+
+        foreach ($event->getAttachments() as $attachment) {
+            $data = $attachment->getData();
+
+            $attachmentItemHeader = [
+                'type' => 'attachment',
+                'filename' => $attachment->getFilename(),
+                'content_type' => $attachment->getContentType(),
+                'length' => \strlen($data),
```
stayallive, line 236: We could replace this with `filesize()` on the file path (to not read the file multiple times into memory to count it's length).

src/Serializer/PayloadSerializer.php
```diff
@@ -215,12 +215,32 @@ private function serializeAsEnvelope(Event $event): string
...
+            ];
+
+            $items[] = JSON::encode($attachmentItemHeader);
+            $items[] = $data;
```
stayallive, line 240: As discussed earlier it might be better to only read the file here once. Otherwise we have it at least twice in memory.

src/Event.php
stayallive: This is fine to me.

stayallive (changes requested):
Overal this looks good, my biggest concerns are with the memory usage but that is in part because of how the attachment has to be added to the envelope, but we can minimize the memory usage a but by lazy reading the file instead of copying it's contents around.

tests/Serializer/PayloadSerializerTest.php
```diff
@@ -54,7 +55,7 @@ public function testSerialize(Event $event, string $expectedResult, bool $isOutp
 
         $result = $this->serializer->serialize($event);
 
-        if (EventType::transaction() !== $event->getType()) {
+        if (EventType::transaction() !== $event->getType() && !$event->hasAttachments()) {
```
cleptric, line 58: This will break old self-hosted Sentry installations without Relay when using attachments. This might be odd in terms of UX, as we can't make assumptions about the destination we sent the events to.

stayallive:
It might be that this feature PR came a little too early. The next planned major version of the SDK is going to overhaul the HTTP transport layer (and also drop support for non-envelope payloads) which could make it possible to add the attachments to the event hint and implement more memory conservative options of dealing with the attachment files.

The memory issue is especially important since when we are handling events (errors usually) we might actually already be memory constraint and this feature could cause out of memory issues without users realizing making them possibly miss events and I feel like missing attachments instead of the event is the better option.

So we might want to hold of on this feature for the 3.x series of the SDK.

PR #1439, opened 2022-12-11
Add error handler integrations option

HypeMC:
Sometimes you might want to exclude the error handler integrations from the default ones in case you are using some other error handler implementation, eg a one provided by your framework (see https://github.com/getsentry/sentry-symfony/issues/421).

One problem that having multiple error handlers may cause is having duplicate events in Sentry for the same event.

This PR introduces a new `error_handler_integrations` option which can be used to skip enabling the error handler integrations without the need to disable all default ones. When set to `null`, which is the default value, it'll fallback to using the `default_integrations` option. This is done to preserve backward compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- feat: Add `error_handler_integrations` option (#1439)
+
 ## 3.12.0 (2022-11-22)
 
 - feat: Add `before_send_transaction` option (#1424)
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -241,6 +241,11 @@ parameters:
 			path: src/Options.php
 
 		-
+			message: "#^Method Sentry\\\\Options\\:\\:hasErrorHandlerIntegrations\\(\\) should return bool but returns mixed\\.$#"
+			count: 1
+			path: src/Options.php
+
+		-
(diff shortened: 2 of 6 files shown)


cleptric:
This should already be possible using [`Options::integrations`](https://docs.sentry.io/platforms/php/integrations/#custom-integrations), so I rather not add another option to achieve the same.

HypeMC:
@cleptric Hi, while it is possible to do it the way you've suggested, it would require [coping some of the code](https://github.com/getsentry/sentry-php/blob/510dd816bf33f5f1322d564b6a562bfc422b265f/src/Integration/IntegrationRegistry.php#L89-L108) to mimic the default behavior and then also keeping that code in sync with any future changes. I was hoping to avoid that, but if you think it'd be better to solve it on the sentry bundle's side using this option, I'm ok with that.

cleptric:
We're doing similar things in our [Laravel SDK](https://github.com/getsentry/sentry-laravel/blob/master/src/Sentry/Laravel/ServiceProvider.php#L176-L210).
While I agree that this adds a bit of boilerplate to your application, I think this is also a use case not many people have.

There is definitely room for improvement on the Symfony SDK when it comes to the way we report errors, which is tracked via https://github.com/getsentry/sentry-symfony/issues/337

HypeMC:
@cleptric Thank you for the info, I'll close this one then.

PR #1441, opened 2022-12-12
Fix nullability in getters and setters for traces_sample_rate in Options

jnoordsij:
In #1428 the default value for `traces_sample_rate` was set to `null`. When calling `getTracesSampleRate()` on a new `Options` instance, this results in a PHP runtime exception as this method has a `float` return type. Similarly, passing `null` to `setTracesSampleRate()` is not allowed.

This PR adjusts the parameter and return type for these methods to allow for `null`.

I've also slightly refactored `isTracingEnabled` to use functions for accessing configuration and added a test for getting/setting `null`.

#skip-changelog

### DIFF
--- a/src/State/Hub.php
+++ b/src/State/Hub.php
@@ -239,7 +239,7 @@ final class Hub implements HubInterface
             } else {
                 $sampleRate = $this->getSampleRate(
                     $samplingContext->getParentSampled(),
-                    $options->getTracesSampleRate()
+                    $options->getTracesSampleRate() ?? 0
                 );
             }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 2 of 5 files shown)


src/State/Hub.php
```diff
@@ -239,7 +239,7 @@ public function startTransaction(TransactionContext $context, array $customSampl
             } else {
                 $sampleRate = $this->getSampleRate(
                     $samplingContext->getParentSampled(),
-                    $options->getTracesSampleRate()
+                    $options->getTracesSampleRate() ?? 0
```
stayallive, line 242: For future reference: technically we should never be here with a `null` in the sample rate.

stayallive (approved):
💪 good one, since we internally always checked the values before we have never reached this but I can see how userland code might use this and trigger the issue. Thanks!

cleptric:
@jnoordsij fixed in https://github.com/getsentry/sentry-php/releases/tag/3.12.1

PR #1443, opened 2022-12-20
feat: add object ID to serialized value

annuh:
This change will add the object ID to the serialized value, if it exists.

For example: `Object App\Entity\Order(#123)`

### DIFF
--- a/src/Serializer/AbstractSerializer.php
+++ b/src/Serializer/AbstractSerializer.php
@@ -244,7 +244,7 @@ abstract class AbstractSerializer
             $objectId = null;
             if (isset($value->id)) {
                 $objectId = $value->id;
-            } elseif (method_exists($value, 'getId')) {
+            } elseif (is_callable([$value, 'id']) && method_exists($value, 'getId')) {
                 $objectId = $value->getId();
             }
 
--- a/src/Serializer/AbstractSerializer.php
+++ b/src/Serializer/AbstractSerializer.php
@@ -248,7 +248,7 @@ abstract class AbstractSerializer
                 $objectId = $value->getId();
             }
 
-            return 'Object ' . \get_class($value) . (($objectId !== null) ? ' (#' . $objectId . ')'  : '');
+            return 'Object ' . \get_class($value) . (($objectId !== null) ? '(#' . $objectId . ')'  : '');
         }
 
         if (\is_resource($value)) {
--- a/src/Serializer/AbstractSerializer.php
+++ b/src/Serializer/AbstractSerializer.php
@@ -244,7 +244,7 @@ abstract class AbstractSerializer
             $objectId = null;
             if (isset($value->id)) {
                 $objectId = $value->id;
-            } elseif (\is_callable([$value, 'id']) && method_exists($value, 'getId')) {
+            } elseif (\is_callable([$value, 'getId']) && method_exists($value, 'getId')) {
                 $objectId = $value->getId();
             }
 
--- a/src/Serializer/AbstractSerializer.php
+++ b/src/Serializer/AbstractSerializer.php
@@ -241,7 +241,7 @@ abstract class AbstractSerializer
         }
 
         if (\is_object($value)) {
-            return 'Object ' . \get_class($value);
+            return 'Object ' . \get_class($value) . (method_exists($value, 'getId') ? ' (#' . $value->getId() . ')'  : '');
         }
 
         if (\is_resource($value)) {
--- a/src/Serializer/AbstractSerializer.php
+++ b/src/Serializer/AbstractSerializer.php
@@ -248,7 +248,7 @@ abstract class AbstractSerializer
                 $objectId = $value->getId();
             }
 
-            return 'Object ' . \get_class($value) . (($objectId !== null) ? '(#' . $objectId . ')'  : '');
+            return 'Object ' . \get_class($value) . ((null !== $objectId) ? '(#' . $objectId . ')'  : '');
         }
 
         if (\is_resource($value)) {
--- a/src/Serializer/AbstractSerializer.php
+++ b/src/Serializer/AbstractSerializer.php
@@ -248,7 +248,7 @@ abstract class AbstractSerializer
                 $objectId = $value->getId();
             }
 
-            return 'Object ' . \get_class($value) . ((null !== $objectId) ? '(#' . $objectId . ')'  : '');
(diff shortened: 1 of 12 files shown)


src/Serializer/AbstractSerializer.php
```diff
@@ -241,7 +241,7 @@ protected function serializeValue($value)
         }
 
         if (\is_object($value)) {
-            return 'Object ' . \get_class($value);
+            return 'Object ' . \get_class($value) . (method_exists($value, 'getId') ? ' (#' . $value->getId() . ')'  : '');
```
cleptric, line 244: Using `property_exists` on `id` would be a bit more generic IMO.
Having some sort of test case that covers this would be appreciated as well, so we do not introduce regressions down the line.

annuh: Good suggestion! I think it would be good to keep the `$object->getId()` fallback, since this a common way to expose an object's ID (ie. when using Doctrine entities)

src/Serializer/AbstractSerializer.php
```diff
@@ -241,7 +241,14 @@ protected function serializeValue($value)
         }
 
         if (\is_object($value)) {
-            return 'Object ' . \get_class($value);
+            $objectId = null;
+            if (isset($value->id)) {
+                $objectId = $value->id;
+            } elseif (method_exists($value, 'getId')) {
```
mfb, line 247: If you're trying to call an arbitrary method on unknown object I think you'd have to use `is_callable()` to ensure you have access to the method?

src/Serializer/AbstractSerializer.php
```diff
@@ -241,7 +241,14 @@ protected function serializeValue($value)
...
         if (\is_object($value)) {
-            return 'Object ' . \get_class($value);
+            $objectId = null;
+            if (isset($value->id)) {
+                $objectId = $value->id;
+            } elseif (\is_callable([$value, 'getId']) && method_exists($value, 'getId')) {
+                $objectId = $value->getId();
+            }
+
+            return 'Object ' . \get_class($value) . ((null !== $objectId) ? '(#' . $objectId . ')' : '');
```
cleptric, line 251:
```suggestion
            return 'Object ' . \get_class($value) . (null !== $objectId) ? '(#' . $objectId . ')' : '';
```

tests/Serializer/AbstractSerializerTest.php
```diff
@@ -576,6 +594,23 @@ public static function testy(): void
     }
 }
 
+#[\AllowDynamicProperties]
```
cleptric, line 597: Do we need this attribute here?

tests/Serializer/AbstractSerializerTest.php
```diff
@@ -576,6 +594,23 @@ public static function testy(): void
...
+class SerializerTestObjectWithIdProperty extends SerializerTestObject
+{
+    public $id = 'bar';
+}
+
+#[\AllowDynamicProperties]
```
cleptric, line 603: Do we need this attribute here?

cleptric (approved):
Thanks! Besides the comments, LGTM :shipit:

tests/Serializer/AbstractSerializerTest.php
annuh: No, you are right 🙂  Will remove them.

src/Serializer/AbstractSerializer.php
annuh: I think the extra parentheses makes it more clear.

src/Serializer/AbstractSerializer.php
```diff
@@ -241,7 +241,14 @@ protected function serializeValue($value)
         }
 
         if (\is_object($value)) {
-            return 'Object ' . \get_class($value);
+            $objectId = null;
+            if (isset($value->id)) {
+                $objectId = $value->id;
```
mfb, line 246: I believe isset() returns TRUE for a protected property, and then an Error will be thrown when you try to access it.

src/Serializer/AbstractSerializer.php
```diff
@@ -241,7 +241,14 @@ protected function serializeValue($value)
...
+            } elseif (\is_callable([$value, 'getId']) && method_exists($value, 'getId')) {
```
mfb, line 247: Not sure there is any reason to call both is_callable() and method_exists() - if it's callable it must also exist? One hopes.

src/Serializer/AbstractSerializer.php
mfb: Ah looks like I am wrong here, isset returns FALSE for inaccessible property.

src/Serializer/AbstractSerializer.php
annuh: According to https://www.php.net/manual/en/function.is-callable.php

> Notes:
> An object is always callable if it implements [__invoke()](https://www.php.net/manual/en/language.oop5.magic.php#object.invoke), and that method is visible in the current scope.
A class name is callable if it implements [__callStatic()](https://www.php.net/manual/en/language.oop5.overloading.php#object.callstatic).
If an object implements [__call()](https://www.php.net/manual/en/language.oop5.overloading.php#object.call), then this function will return true for any method on that object, even if the method is not defined.

So it makes sense to use both functions IMO: 
- `method_exists()` to check if the method `getId()` exists (it doesn't check if it's public or not)
- `is_callable()` to check if getId()` can be called

mfb: Ok today I learned :)

src/Serializer/AbstractSerializer.php
smuuf: This also invokes magic `__isset()` (and then `__get()` ) method for objects, which can be problematic if these magic methods have side effects. Which is exactly what I had to investigate and deal with in our own codebase.

I'd propose that this behavior (magically getting some kind of ID of objects present within traceback) should check if the property _actually_ exists _(via `property_exists()`, which doesn't invoke magic methods)_.

Or at least this magic feature should be optional - i,e. there should be an option to disable it.

cleptric: Sorry you ran into issues because of this!

The approach via `property_exists` sounds reasonable to me. Mind opening a dedicated issue or PR for this? Thanks!

cleptric: https://github.com/getsentry/sentry-php/issues/1476

PR #1449, opened 2023-01-20
Minor cleanups

nicolas-grekas:
Extracted from https://github.com/getsentry/sentry-php/pull/1418

`HttpClientFactory` doesn't use its 2 first arguments.

#skip-changelog


cleptric (approved):
Thank you!

PR #1451, opened 2023-01-21
Log exception's code

### LINKED ISSUE #1450: Log exception's code
### Problem Statement

I work a lot with the [Twilio SDK](https://github.com/twilio/twilio-php). In case of an error, a [RestException](https://github.com/twilio/twilio-php/blob/0739f62e6280d9a20925d2f30a71bfd2bcac48f5/src/Twilio/Version.php#L84-L88) is thrown with a text description of the error, as well as a code in [their internal error system](https://www.twilio.com/docs/api/errors). In the current implementation, I can't see this code.

### Solution Brainstorm

According to Sentry's documentation, there is a [exception's meta information](https://develop.sentry.dev/sdk/event-payloads/exception/#meta-information), which seems to be the way to provide the wanted information.

I have [tried to implement it](https://github.com/getsentry/sentry-php/compare/master...erickskrauch:sentry-php:32257d37f6fd405004b0c018534a8408d25130bc), but it seems like Sentry filters out this field on the service side (I checked that this field presented just before JSON encoding when a request is about to be sent).

erickskrauch:
Fixes #1450.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # CHANGELOG
 
 ## Unreleased
+- feat: Add `data` param to `ExceptionMechanism` class (#1450)
+- feat: Log exception's `code` to the `ExceptionMechanism::data` (#1450)
 
 - feat: add object ID to serialized value (#1443)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,9 @@
 
 ## Unreleased
 
+- feat: add object ID to serialized value (#1443)
 - feat: Add support for sending arbitrary extra data as part of the exception mechanism (#1450)
 - feat: Log exception's `code` to the `ExceptionMechanism::data` (#1450)
-
-- feat: add object ID to serialized value (#1443)
 
 ## 3.12.1 (2023-01-12)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # CHANGELOG
 
 ## Unreleased
-- feat: Add `data` param to `ExceptionMechanism` class (#1450)
+- feat: Add support for sending arbitrary extra data as part of the exception mechanism (#1450)
 - feat: Log exception's `code` to the `ExceptionMechanism::data` (#1450)
 
 - feat: add object ID to serialized value (#1443)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 ## Unreleased
+
 - feat: Add support for sending arbitrary extra data as part of the exception mechanism (#1450)
 - feat: Log exception's `code` to the `ExceptionMechanism::data` (#1450)
 
--- a/src/ExceptionMechanism.php
+++ b/src/ExceptionMechanism.php
@@ -77,4 +77,14 @@ final class ExceptionMechanism
     {
         return $this->data;
     }
+
+    /**
+     * Sets the arbitrary extra data.
+     *
+     * @param array<string, mixed> $data
+     */
+    public function setData(array $data): void
+    {
+        $this->data = $data;
+    }
 }
--- a/src/ExceptionMechanism.php
+++ b/src/ExceptionMechanism.php
@@ -27,17 +27,25 @@ final class ExceptionMechanism
     private $handled;
 
     /**
+     * @var array|null Optional information from the operating system or runtime
+     */
+    private $meta;
+
+    /**
      * Class constructor.
      *
-     * @param string $type    Unique identifier of this mechanism determining
-     *                        rendering and processing of the mechanism data
-     * @param bool   $handled Flag indicating whether the exception has been
-     *                        handled by the user (e.g. via try..catch)
+     * @param string     $type    Unique identifier of this mechanism determining
+     *                            rendering and processing of the mechanism data
+     * @param bool       $handled Flag indicating whether the exception has been
+     *                            handled by the user (e.g. via try..catch)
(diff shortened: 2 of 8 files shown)


src/ExceptionMechanism.php
```diff
@@ -26,18 +26,28 @@ final class ExceptionMechanism
...
+     * @param string     $type    Unique identifier of this mechanism determining
+     *                            rendering and processing of the mechanism data
+     * @param bool       $handled Flag indicating whether the exception has been
+     *                            handled by the user (e.g. via try..catch)
+     * @param array|null $data    Arbitrary extra data that might help the user
+     *                            understand the error thrown by this mechanism
+     * @psalm-param array<string, mixed>|null $data
      */
-    public function __construct(string $type, bool $handled)
+    public function __construct(string $type, bool $handled, array $data = null)
```
ste93cry, line 46: Why having this argument to be `null` by default instead of an empty `array`? I think that if we want to leave it as-is, then we should make the argument `nullable` for consistency: it doesn't make sense that an argument cannot accept `null`, but its default value is `null`

src/ExceptionMechanism.php
```diff
@@ -26,18 +26,28 @@ final class ExceptionMechanism
...
-     *                        rendering and processing of the mechanism data
-     * @param bool   $handled Flag indicating whether the exception has been
-     *                        handled by the user (e.g. via try..catch)
+     * @param string     $type    Unique identifier of this mechanism determining
+     *                            rendering and processing of the mechanism data
+     * @param bool       $handled Flag indicating whether the exception has been
+     *                            handled by the user (e.g. via try..catch)
+     * @param array|null $data    Arbitrary extra data that might help the user
+     *                            understand the error thrown by this mechanism
+     * @psalm-param array<string, mixed>|null $data
```
ste93cry, line 44: We can directly use `@param array<string, mixed>` instead of adding `@psalm-param`, as we did it already in a few other places of the codebase

src/ExceptionMechanism.php
```diff
@@ -26,18 +26,28 @@ final class ExceptionMechanism
      */
     private $handled;
 
+    /**
+     * @var array|null Arbitrary extra data that might help the user understand
```
ste93cry, line 30: If we accept `array<string, mixed>` in the constructor, this variable should be typehinted the same

ste93cry (changes requested):

erickskrauch:
Thanks for the quick review. I made the requested changes.

cleptric (commented):
Overall I'm ok with adding the new `data` property to `ExceptionMechansim`.
I'm not sure yet if we want to add the code by default, but I have to think about this a bit more.

Please also add a `setData()` method, as we might need to overwrite the data set in an upcoming feature concerning Pii/sensitive data.
Additionally, there is currently no test coverage for the `PayloadSerializer`.

erickskrauch:
Should I introduce tests for the `PayloadSerializer` or it's on your side?

cleptric:
Codecov is pointing out the lines you added in `PayloadSerializer.php` that are not covered here https://github.com/getsentry/sentry-php/pull/1451/files. So you should add tests for that, please 🙂

erickskrauch:
I wasn't sure if I needed to create a separate case to serialize arbitrary data from `ExceptionMechanism` if it was submitted, so I added it to the existing case. Tell me if a separate case is needed.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # CHANGELOG
 
 ## Unreleased
+- feat: Add `data` param to `ExceptionMechanism` class (#1450)
```
ste93cry, line 4: I suggest to change this description to `Add support for sending arbitrary extra data as part of the exception mechanism`. I find this to be more helpful for readers than knowing the details of how this was implemented.

tests/ExceptionMechanismTest.php
```diff
@@ -0,0 +1,61 @@
...
+
+declare(strict_types=1);
+
+namespace Sentry\Tests;
+
+use PHPUnit\Framework\TestCase;
+use Sentry\ExceptionMechanism;
+
+/**
+ * @covers \Sentry\ExceptionMechanism
```
ste93cry, line 11: There is no need for this kind of annotations, please remove it :broom:

ste93cry:
> I added it to the existing case

For me it's fine as-is 👍

ste93cry (changes requested):
Some small changes, but for me it looks perfect!

CHANGELOG.md
erickskrauch: No problem. I'm not a native speaker, so my descriptions might be weird :)

ste93cry:
You need to add a blank line in the `CHANGELOG` between the title of the section and the lines you added, otherwise Danger will keep complaining. Also, please target the `develop` branch as this is a new feature and so it cannot be released as part of the next patch version.

ste93cry:
There’s a commit that has not been merged yet into develop and that is part of your branch now. Do you mind rebasing the PR to drop it?

erickskrauch:
@ste93cry, I hope I understood correctly which commit you were talking about 😅

ste93cry (approved):
Thanks for the contribution, it looks fine to me!

cleptric:
The data set to the `ExceptionMechansim` is currently overwritten, in case the global exception handler handles the exception https://github.com/getsentry/sentry-php/blob/bfa3193944fe385af7d6b6b063962fc45956a817/src/Integration/AbstractErrorListenerIntegration.php#L39

erickskrauch:
Yes, thank you for pointing that out. I fixed it.

PR #1452, opened 2023-01-25
WIP: Fix #1426 ignore previous exeptions not being ignored

### LINKED ISSUE #1426: Improve 'ignore_exceptions' configuration to allow to ignore even 'previous' exceptions
### Context
Currently we have a problem, where a specific exception is caught and wrapped as "BadRequestHttpException". Including it in `ignore_exceptions` list is too general, and including the specific exception (which is included as previous exception) does not work for `ignore_exceptions` configuration, since it only checks the main exception and does not iterate previous exceptions `Sentry\Integration\IgnoreErrorsIntegration:102`. 

### Expected Result
Ability to ignore exceptions which are wrapped in general exceptions as previous exceptions. 

### Actual Result
`ignore_exceptions` list allows only top level exceptions to be ignored.

rjd22:


cleptric:
I fear that this might introduce a regression for users that are using the integration today. It could happen that Exceptions will be ignored that are currently sent to Sentry.

I would suggest you either add a [custom integration](https://docs.sentry.io/platforms/php/integrations/#custom-integrations) or handle this in [before_send](https://docs.sentry.io/platforms/php/configuration/options/#before-send).

This might be worth revisiting for v.4.0.

rjd22:
@cleptric I understand. Wouldn't it be sensible then to keep this PR open and merge it with v4.0? Most likely it will still be mergable then.

cleptric:
We still have an open issue for this, https://github.com/getsentry/sentry-php/issues/1426
As long as you do not delete your fork, we can reopen the PR at any time.

One possible path forward would be the introduction of a new top-level option called `ignore_exceptions`:

```php
Sentry\init([
    'ignore_exceptions' => [\Exception::class],
]);
```

We could then drop the event in the client.
I actually prefer this over the integration, as it makes the API a bit cleaner and easier to use.
Would you be willing to propose this in a new PR?

rjd22:
I'm willing to do the work. Would that mean adding an Option and making supporting code to ignore exceptions here?:

https://github.com/getsentry/sentry-php/blob/8cf074301c6e7ee094255e8eb114c16d74356521/src/Client.php#L161

cleptric:
`Hub::captureException()` looks like a good place to apply the ignore logic.
Exceptions by the global exception error handler will also be passed through here.

Just to be extra clear, this would deprecate the current `IgnoreErrorsIntegration` and add support for previous exceptions as well.

Issue #1429, opened 2022-11-15
GuzzleTracingMiddleware: Sanitise span description


cleptric:
### Problem Statement

https://github.com/getsentry/sentry-php/blob/68ee84af06adacd2b86496892d737f28c329b0fa/src/Tracing/GuzzleTracingMiddleware.php#L34

This might contain sensitive data or Pii that should be removed. 

### Solution Brainstorm

Sanitise all GET parameters and remove HTTP basic credentials if present in the URI.

See https://github.com/getsentry/develop/pull/773

PR #1453, opened 2023-01-31
fix: Sanatize HTTP client spans & breadcrumbs

cleptric:
Closes https://github.com/getsentry/sentry-php/issues/1429

See https://develop.sentry.dev/sdk/data-handling/#spans & https://develop.sentry.dev/sdk/data-handling/#breadcrumbs.

refs: https://github.com/getsentry/team-webplatform-meta/issues/2

### DIFF
--- a/src/Tracing/GuzzleTracingMiddleware.php
+++ b/src/Tracing/GuzzleTracingMiddleware.php
@@ -6,6 +6,7 @@ namespace Sentry\Tracing;
 
 use Closure;
 use GuzzleHttp\Exception\RequestException as GuzzleRequestException;
+use GuzzleHttp\Psr7\Uri;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use Sentry\Breadcrumb;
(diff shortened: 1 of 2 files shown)


DAcodedBEAT (approved):

src/Tracing/GuzzleTracingMiddleware.php
```diff
@@ -29,9 +30,20 @@ public static function trace(?HubInterface $hub = null): Closure
                     return $handler($request, $options);
                 }
 
+                $partialUri = Uri::fromParts([
+                    'scheme' => $request->getUri()->getScheme(),
+                    'host' => $request->getUri()->getHost(),
+                    'port' => $request->getUri()->getPort(),
+                    'path' => $request->getUri()->getPath(),
+                ]);
```
stayallive, line 38: `l`: The only "worry" I have here is that a `MalformedUriException` is thrown for some reason. Although this might be totally unfounded since a URI was already there but maybe we want to try/catch anyway just to be 100% sure since not all PSR-7's might be created equal.

stayallive (approved):

src/Tracing/GuzzleTracingMiddleware.php
cleptric: Yeah, this might be a safer choice in the end. During my tests, the request normally fails with a cURL exception before this code runs.

stayallive: Yeah that's why I figured this might be overkill, so it might not be 100% needed. I also wouldn't know what to fallback too, because dropping the breadcrumb also seems wrong... so maybe leave as is and see what edge cases fall out. Don't think this could every happen outside synthetic tests anyway.

cleptric:
@retnek The changes here are unrelated, the transaction name changes were made in our Symfony SDK https://github.com/getsentry/sentry-symfony/pull/692.

Nonetheless, I'm sorry that you ran into this. We had to make these changes, as we got a lot of problems with people only seeing `<<unparameterized>>` transaction names on the performance page.
Please send an email to michi [at] sentry.io including your org name, so we can get you some transaction quota back.

Issue #1457, opened 2023-02-06
Add enableTracing option


smeubank:
A new simplified options for configuring to turn tracing & performance on should be added to all SDKs. For each SDK the default sample rate should be agreed upon.  

Docs do not need to be updated yet, nor the onboarding wizard. This can wait until the top level options are added to a number of SDKs and we decide if we want to configure the default onboarding experience, for all new users.

See [develop docs](https://develop.sentry.dev/sdk/performance/#enabletracing) for more information. This is a non-breaking change to be clear. Any of the three options should enable performance, alone or in combination

### Open point:
recommend default it 1.0 or 100%. Can discuss if it should be less.

cleptric:
Holding off on this for the moment, as we cannot set a default `traces_sample_rate: 1.0` due to overhead concerns. I also don't have a good answer when it comes to a lower default value, as this heavily depends on the application and other factors like using a local Relay.

stayallive:
Tracing is still opt-in right? 

So won't this extra option make no difference at all from what it is now (because it defaults to `false`? I mean if we tell users to set `traces_sample_rate: 1.0` or `enable_tracing: true` to enable tracing makes no difference for the end-result (and the implicated perf issues related to PHP) and the same warnings (using relay etc.) apply for both "methods". 

It's just a little nicer to tell users to `enable_tracing: true`, little syntax sugar if I'm reading it correctly?

cleptric:
Yes, correct. I'm just afraid of hiding `traces_sample_rate: 1.0` as the default behind `enable_tracing: true`. If these concerns are not justified or too far-fetched, we can of course add the new option.

PR #1458, opened 2023-02-08
feat: Add enableTracing Option

HazAT:
Fixes https://github.com/getsentry/sentry-php/issues/1457

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -64,8 +64,6 @@ final class Options
         }
     }
 
-    
-
     /**
      * Gets the number of attempts to resend an event that failed to be sent.
      *
@@ -152,7 +150,7 @@ final class Options
     }
 
     /**
-     * Sets if tracing should be enabled or not. If null tracesSampleRate takes 
+     * Sets if tracing should be enabled or not. If null tracesSampleRate takes
      * precedence.
      *
      * @param ?bool $enableTracing Boolean if tracing should be enabled or not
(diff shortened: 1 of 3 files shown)


src/Options.php
```diff
@@ -143,6 +149,27 @@ public function getTracesSampleRate(): ?float
...
+
+        $this->options = $this->resolver->resolve($options);
+    }
+
+    /**
+     * Gets if tracing is enabled or not.
+     */
+    public function getEnableTracing(): ?bool
+    {
+        return $this->options['enable_tracing'];
```
cleptric, line 170: You can run `vendor/bin/phpstan -b` to update the phpstan baseline file.

src/Options.php
```diff
@@ -143,6 +149,27 @@ public function getTracesSampleRate(): ?float
         return $this->options['traces_sample_rate'];
     }
 
+    /**
+     * Sets if tracing should be enabled or not. If null tracesSampleRate takes
+     * precedence.
+     *
+     * @param ?bool $enableTracing Boolean if tracing should be enabled or not
+     */
+    public function setEnableTracing(?bool $enableTracing): void
```
cleptric, line 158: We should also test this method inside `OptionsTest::testGettersAndSetters`

cleptric (approved):
Thanks @HazAT :shipit: 🚀

src/Options.php
HazAT: Yeah, maybe true - it is a bit confusing.
The thing is the logic behind how `isTracingEnabled` behaves works a bit differently than just a simple boolean.
This getter however really returns the plain content.

PR #1459, opened 2023-02-09
Leverage php-http/discovery v1.15 and decouple from guzzlehttp/psr7

nicolas-grekas:
This is a follow up of #1429 

Since this PR and our discussion at SymfonyCon, I've been working on moving those ideas to "php-http/discovery", and [v1.15.0](https://github.com/php-http/discovery/releases/tag/1.15.0) has just been tagged :tada: 

By bumping to v1.15.0, we keep compatibility with PHP >= 7.1 and composer v1 & v2, but we also benefit from:
- https://github.com/php-http/discovery/pull/209
- https://github.com/php-http/discovery/pull/208

This allows simplifying the dependencies of sentry-php since we can now move guzzlehttp/psr7 from require to require-dev.


cleptric:
Thanks, I'll take a look.

nicolas-grekas:
Composer [v2.2.20](https://github.com/composer/composer/releases/tag/2.2.20) and v2.5.3 have now been released.
Together with php-http/discovery [v1.5.1](https://github.com/php-http/discovery/releases/tag/1.15.1) this PR should be doubly ready :)

PR #1464, opened 2023-02-09
fix: pin php-http/discovery to < 1.15

stayallive:
From #1462, we decided it might be better to pin the dependency to a non-problematic version for the time being (php-http/discovery#213).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "jean85/pretty-package-versions": "^1.5|^2.0.4",
         "php-http/async-client-implementation": "^1.0",
         "php-http/client-common": "^1.5|^2.0",
-        "php-http/discovery": "^1.11",
+        "php-http/discovery": "^1.11, <1.15",
         "php-http/httplug": "^1.1|^2.0",
         "php-http/message": "^1.5",
         "psr/http-factory": "^1.0",


cleptric:
I don't want to prompt our users with `php-http/discovery contains a Composer plugin which is currently not in your allow-plugins config` when they update their dependencies. Before we ship https://github.com/getsentry/sentry-php/pull/1459, there are currently no benefits of using 1.15.

cleptric (approved):

nicolas-grekas:
I wouldn't have merged this because it's not sentry's responsibility to force an upper bound to another dep.
But anyway, https://github.com/getsentry/sentry-php/pull/1459 should be the next step to :shrug: 
See https://github.com/php-http/discovery/issues/213#issuecomment-1424503490 also

HypeMC:
Won't this block any project using sentry from using v1.15. Sentry is not the only library that uses `php-http/discovery`, there might be libraries/projects that can already benefit from the new version, but now they're blocked from doing so.

mfb:
A new plugin feels like a BC break for library maintainers such as myself. E.g., my library depends on sentry/sdk and is in turn used by thousands of actual end-user projects. I obviously don't have a way to enable/disable plugins in my end users' projects. I could update my README and roll a new release to attempt to educate users about how they suddenly need to enable or disable the new plugin, but that still feels like a BC break.

nicolas-grekas:
@mfb please have a look at the discussion in https://github.com/php-http/discovery/issues/213 it might give a few useful hint about the topic.

mfb:
@nicolas-grekas I've read that but seems like the takeaway is simply that the plugin needs to be disabled or enabled. And as a library maintainer, I cannot disable or enable the plugin, all I can do is ask the end users to disable or enable the plugin for their projects. I'm happy to add that instruction to my README, but I also wanted to point out that ideally, users wouldn't see such breakage happen without a major release somewhere in the chain of dependencies.

PR #1467, opened 2023-02-13
feat: Add support for Cron Monitoring

### LINKED ISSUE #1466: Allow to create & send Envelopes
### Problem Statement

The majority of event ingestion these days is done via sending [envelopes](https://develop.sentry.dev/sdk/envelopes/) to any Relay.
The SDK currently lacks a proper abstraction for this, making it hard to implement upcoming features like Cron Monitoring.

### Solution Brainstorm

Add a new abstraction around envelopes. The goal should be an API to create a valid envelope, including headers, items, and the payload.
It might be tricky to properly expose the `Transport::send` method, as it is currently part of the `TransportInterface` and expects an `Event` instead of an `Envelope` instance being passed.

However, this should be implemented in a none BC breaking way.

cleptric:
This adds support to send [Cron Monitor](https://docs.sentry.io/product/crons/) check-ins as envelopes to Relay.

```php
<?php
$checkIn = new CheckIn(
    monitorSlug: getenv('MONITOR_SLUG'),
    status: CheckInStatus::inProgress(),
);

$event = Event::createCheckIn();
$event->setCheckIn($checkIn);

$this->hub->captureEvent($event);

// do stuff

$event = Event::createCheckIn();
$event->setCheckIn(new CheckIn(
    id: $checkIn->getId(),
    monitorSlug: getenv('MONITOR_SLUG'),
    status: CheckInStatus::ok(),
));

$this->hub->captureEvent($event);
```

### DIFF
--- a/src/CheckIn.php
+++ b/src/CheckIn.php
@@ -1,0 +1,117 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry;
+
+use Sentry\Util\SentryUid;
+
+final class CheckIn
+{
+    /**
+     * @var string The check-in ID
+     */
+    private $id;
+
+    /**
+     * @var string The monitor slug
+     */
+    private $monitorSlug;
+
+    /**
+     * @var \Sentry\CheckInStatus The status of the check-in
+     */
+    private $status;
+
+    /**
+     * @var string|null The release
+     */
+    private $release;
+
+    /**
+     * @var string|null The environment
+     */
+    private $environment;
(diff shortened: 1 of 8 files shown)


antonpirker (approved):
looks good to me (but "me" is no php expert) :-)

stayallive (approved):
This makes it obvious we need a refactor at some point to stop dealing with non-envelopes, but other than that, no comments 💪

PR #1484, opened 2023-03-01
fix(serializer): serializeValue uses only scalar object IDs

AdamLutka:
Handle object IDs that aren't scalars. It prevents php warning 'Array to string conversion' and similar.


cleptric:
Can you please target this against `develop`? We merged #1483 yesterday which changed a few things.

PR #1490, opened 2023-03-10
Introduce `trace` helper function

stayallive:
This adds the `trace` helper function that reduces a lot of boilerplate of adding spans for user code, allowing users to quickly wrap existing calls with a span and it looks something like this:

```php
use Sentry\State\Scope;
use Sentry\Tracing\SpanContext;
use function Sentry\trace;

$args = [...];

$result = trace(
    function (Scope $scope) use ($args) {
        $scope->setTag('some-tag', 'some-value');

        return $this->doSomethingThatNeedsToBeTraced($args);
    },
    tap(new SpanContext, function (SpanContext $context) use ($args) {
        $context->setOp('something.measurable');
        $context->setData(['context' => $args]);
        $context->setDescription('Something measurable');
    }),
);
```

_The `SpanContext` is a unfortunate object in our API at the moment and doesn't have a fluent interface so I've used the `tap()` helper from Laravel to keep it a little more readable in the example._

This was originally coined in Laravel (getsentry/sentry-laravel#598) but is general enough for all of PHP to benefit from.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Sentry;
 
 use Sentry\State\Scope;
+use Sentry\Tracing\SpanContext;
 use Sentry\Tracing\Transaction;
 use Sentry\Tracing\TransactionContext;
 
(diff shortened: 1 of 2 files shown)


cleptric:
```php
$client = new Client();

$spanContext = new SpanContext();
$spanContext->setOp('http.client');
$spanContext->setDescription('GET http://google.com');

trace(
    function (Scope $scope) use ($client) {
        return $client->get('http://google.com');
    },
    $spanContext,
);
```

```php
$spanContext = new SpanContext();
$spanContext->setOp('http.client');
$spanContext->setDescription('GET http://google.com');

$transaction = SentrySdk::getCurrentHub()->getTransaction();
$span = $transaction->startChild($spanContext);

$client = new Client();
$client->get('http://google.com');

$span->finish();
```

A small improvement, but it an improvement!

cleptric (approved):

stayallive:
Yeah, the context thing is unfortunate. If it had a fluent syntax we could so some really cool things like:

```php
trace(...)
    ->op('some.op')
    ->description('Something to trace')
    ->run()
```

Which honestly might be attainable actually... without breaking changes. Worth it?

cleptric:
Not sure tbh. We now have a way better span creator, called profiling 😛

I still think it's worth going this route in v4, but would be ok with the current approach for now.

PR #1502, opened 2023-03-21
Fix wrong parameter was passed on `debug_backtrace()`

kayw-geek:
https://www.php.net/manual/en/function.debug-backtrace.php

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -352,7 +352,7 @@ final class Client implements ClientInterface
         }
 
         $event->setStacktrace($this->stacktraceBuilder->buildFromBacktrace(
-            debug_backtrace(0),
+            debug_backtrace(),
             __FILE__,
             __LINE__ - 3
         ));


kayw-geek:
> Do you mind explaining what are you trying to achieve with this change? Passing `0` should disable `DEBUG_BACKTRACE_PROVIDE_OBJECT`, which is the default value for the parameter

The corresponding value of this constant is `1` , so I think I should change it to `DEBUG_BACKTRACE_PROVIDE_OBJECT` ?

ste93cry:
As explained above, we're passing `0` to avoid PHP to fill the object in the backtrace. What use-case do you have for needing such piece of info?

kayw-geek:
> As explained above, we're passing `0` to avoid PHP to fill the object in the backtrace. What use-case do you have for needing such piece of info?

I looked at the definition of PHP built-in constant, and this `DEBUG_BACKTRACE_PROVIDE_OBJECT` actually has a value of `1` .
```PHP
define('DEBUG_BACKTRACE_PROVIDE_OBJECT', 1);
```

ste93cry:
The `$options` parameter is a bitmask, so you can combine the various options which are `DEBUG_BACKTRACE_PROVIDE_OBJECT` and `DEBUG_BACKTRACE_IGNORE_ARGS`. If you don't want to use any of them, you have to pass `0` which effectively disables all those options

kayw-geek:
> As explained above, we're passing `0` to avoid PHP to fill the object in the backtrace. What use-case do you have for needing such piece of info?

https://github.com/php/php-src/blob/b698108133f5403bb440bc94dbebe92ff487e667/Zend/zend.h#L412

kayw-geek:
I'm sorry, I understand your intention.

PR #1503, opened 2023-03-22
feat: Add `ignore_exceptions` & `ignore_transactions`

cleptric:
This adds two new options, namely `ignore_exceptions` and `ignore_transactions`.

### ignore_exceptions

Make it dead simple to ignore an Exception, without the need to deal with any integrations.
I actually marked `IgnoreErrorsIntegration` as deprecated.

Additionally, we take `$exception->getPrevious()` into account, when applying the option.

```php
\Sentry\init([
    'ignore_exceptions' => [BadThingsHappenedException::class],
]);
```

closes https://github.com/getsentry/sentry-php/issues/1426

### ignore_transactions

Same story, make it dead simple to ignore a transaction, without the need to fiddle around with the `traces_sampler` or any `before_send_transaction` stuff.

```php
\Sentry\init([
    'ignore_transactions' => ['GET /health'],
]);
```

Got inspired after reading https://stevenwoodson.com/blog/conserving-sentry-transactions-by-ignoring-laravel-routes/.


src/Client.php
```diff
@@ -311,6 +316,47 @@ private function prepareEvent(Event $event, ?EventHint $hint = null, ?Scope $sco
...
+        if ($event->getType() === EventType::transaction()) {
+            $transactionName = $event->getTransaction();
+
+            if (null === $transactionName) {
+                return $event;
+            }
+
+            if (\in_array($transactionName, $this->options->getIgnoreTransactions(), true)) {
+                $this->logger->info(
+                    'The event will be discarded because it matches a entry "ignore_transactions".',
```
stayallive, line 349:
```suggestion
                    'The event will be discarded because it matches a entry in "ignore_transactions".',
```

src/Client.php
```diff
@@ -311,6 +316,47 @@ private function prepareEvent(Event $event, ?EventHint $hint = null, ?Scope $sco
...
+    {
+        if ($event->getType() === EventType::event()) {
+            $exceptions = $event->getExceptions();
+
+            if (empty($exceptions)) {
+                return $event;
+            }
+
+            foreach ($exceptions as $exception) {
+                if (\in_array($exception->getType(), $this->options->getIgnoreExceptions(), true)) {
```
stayallive, line 329: The only thing is that this works slight different from the now deprecated `IgnoreErrorsIntegration` because that uses `is_a`, not sure if intentional but it might be nice to say you want to ignore `HttpException::class` instead of `[ClientHttpException::class, ServerHttpException::class, ...]` on the other hand this is more "expected" and you can still make that logic work with `before_send` of course.

Just a little rambling, I have no hard feelings either way so fine with the current impl.

tests/ClientTest.php
```diff
@@ -587,6 +587,53 @@ public function processEventDiscardsEventWhenItIsSampledDueToSampleRateOptionDat
...
+    public function testProcessEventDiscardsEventWhenIgnoreTransactionsMatches(): void
+    {
+        $event = Event::createTransaction();
+        $event->setTransaction('GET /foo');
+
+        /** @var LoggerInterface&MockObject $logger */
+        $logger = $this->createMock(LoggerInterface::class);
+        $logger->expects($this->once())
+            ->method('info')
+            ->with('The event will be discarded because it matches a entry "ignore_transactions".', $this->callback(static function (array $context): bool {
```
stayallive, line 622:
```suggestion
            ->with('The event will be discarded because it matches a entry in "ignore_transactions".', $this->callback(static function (array $context): bool {
```

stayallive (approved):

src/Options.php
```diff
@@ -416,6 +416,50 @@ public function setServerName(string $serverName): void
...
+     */
+    public function getIgnoreTransactions(): array
+    {
+        return $this->options['ignore_transactions'];
+    }
+
+    /**
+     * Sets a list of transaction names to be ignored and not sent to Sentry.
+     *
+     * @param int[] $ignoreTransaction The list of transaction names to be ignored
```
ste93cry, line 454: The param type should be `string[]`

src/Options.php
```diff
@@ -416,6 +416,50 @@ public function setServerName(string $serverName): void
...
+    {
+        $options = array_merge($this->options, ['ignore_exceptions' => $ignoreErrors]);
+
+        $this->options = $this->resolver->resolve($options);
+    }
+
+    /**
+     * Gets a list of transaction names to be ignored and not sent to Sentry.
+     *
+     * @return int[]
```
ste93cry, line 444: The return type should be `string[]`

src/Client.php
```diff
@@ -311,6 +316,47 @@ private function prepareEvent(Event $event, ?EventHint $hint = null, ?Scope $sco
         return $event;
     }
 
+    private function applyIgnoreOptions(Event $event, ?EventHint $hint): ?Event
```
ste93cry, line 319: It seems that you are not using `$hint`, so it can be removed

src/Client.php
```diff
@@ -281,6 +281,11 @@ private function prepareEvent(Event $event, ?EventHint $hint = null, ?Scope $sco
             return null;
         }
 
+        $event = $this->applyIgnoreOptions($event, $hint);
+        if (null === $event) {
```
ste93cry, line 285: Nitpick, I find this to be more readable and consistent with the rest of the codebase but feel free to ignore

```suggestion
        $event = $this->applyIgnoreOptions($event, $hint);
        
        if (null === $event) {
```

ste93cry (changes requested):

ste93cry:
While I understand the reasoning behind these changes, I feel that the client is getting too much responsibility imho. The solution involving an integration was taken from the JS SDK and allows us to decouple the client from all these features, which improves the maintenability of the project. I also remember that we already had a discussion in the past about whether to consider "previous" exceptions, and I think that a real improvement would be to make this configurable: in fact, I may want to ignore all the exceptions, but only if they are the topmost.

cleptric:
> While I understand the reasoning behind these changes, I feel that the client is getting too much responsibility imho. The solution involving an integration was taken from the [JS](https://github.com/getsentry/sentry-javascript/blob/86b89b9a271ba4bfe66dd74284992a07e1bc9c72/packages/core/src/integrations/inboundfilters.ts#L17) and [Go](https://github.com/getsentry/sentry-go/blob/85b380d192353dc9ca3df14fc4f8fa727a33cb2c/integrations.go) SDKs and allows us to decouple the client from all these features, which improves the maintenability of the project. I also remember that we already had a discussion in the past about whether to consider "previous" exceptions, and I think that a real improvement would be to make this configurable: in fact, I may want to ignore all the exceptions, but only if they are the topmost.

We tend to move away from event processors, as they are very opaque and confusing for people to use. Once https://github.com/getsentry/rfcs/pull/34 lands, we yet again will move more things into the Client, which is fine.

ste93cry (approved):

marandaneto:
@cleptric we also have that for some SDKs as well.
https://docs.sentry.io/platforms/java/configuration/options/#ignored-exceptions-for-type
It's a very handy utility.

PR #1519, opened 2023-04-20
feat: Add `TransactionContext::fromEnvironment`

cleptric:
`TransactionContext::fromEnvironment` has the same behaviour as `TransactionContext::fromHeaders`.

A use-case for this is the `sentry-cli`, where the CLI wraps a PHP script or framework console command to be cron monitored and exposes `SENTRY_TRACE` & `SENTRY_BAGGAGE` when executing it.

### DIFF
--- a/src/Tracing/TransactionContext.php
+++ b/src/Tracing/TransactionContext.php
@@ -137,6 +137,17 @@ final class TransactionContext extends SpanContext
     }
 
     /**
+     * Returns a context populated with the data of the given environment variables.
+     *
+     * @param string $sentryTrace The sentry-trace value from the environment
+     * @param string $baggage     The baggage header value from the environment
+     */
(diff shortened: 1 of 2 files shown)


src/Tracing/TransactionContext.php
```diff
@@ -136,18 +136,42 @@ public static function fromSentryTrace(string $header): self
...
+     * @param string $baggage     The baggage header value from the environment
+     */
+    public static function fromEnvironment(?string $sentryTrace = null, ?string $baggage = null): self
+    {
+        if (null === $sentryTrace) {
+            $sentryTrace = getenv('SENTRY_TRACE');
+        }
+        if (null === $baggage) {
+            $baggage = getenv('SENTRY_BAGGAGE');
+        }
```
cleptric, line 153: @stayallive We can also eliminate these default values and pass in the env vars explicitly in Laravel/Symfony.

Issue #1533, opened 2023-05-17
Issues with SymfonyHttplugClient while resolving the HttpClient (Symfony 6.2)


gjuric:
### How do you use Sentry?

Self-hosted/on-premise

### Version

3.3.0

### Steps to Reproduce

I've updated a Symfony 6.2 project that also uses Httplug. This upgraded `php-http/client-common` to v2.7.0 that dropped the dependency on `php-http/message-factory` since it is not used directly by this project.

The issue is that `Sentry\HttpClient\HttpClientFactory::resolveClient()` only checks if `class_exists(SymfonyHttplugClient::class)` before trying to provision an instance of SymfonyHttplugClient, but as can bee seen here (Symfony 6.2), it is not possible to provision this client without other dependencies:

https://github.com/symfony/symfony/blob/752378c568ce11bf3439dabd147d9e8ec004585c/src/Symfony/Component/HttpClient/HttplugClient.php#L44-L54

Dependencies are:

* php-http/httplug
* php-http/message-factory (deprecated)
* psr/http-factory

Since the deprecated dependency is not needed in Symfony 6.3 I propose that resolving of SymfonyHttplugClient should be wrapped in a try/catch block, and continue with other possible clients if an error is caught.

### Expected Result

No exception about instantiating a HttpPlug client is thrown.

### Actual Result

Exception is thrown:

> You cannot use "Symfony\Component\HttpClient\HttplugClient" as the "php-http/message-factory" package is not installed. Try running "composer require nyholm/psr7".

PR #1534, opened 2023-05-17
fix: Require php-http/message-factory

cleptric:
closes https://github.com/getsentry/sentry-php/issues/1533 & https://github.com/getsentry/sentry-laravel/issues/694

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # CHANGELOG
 
+## 3.18.2
+
+The Sentry SDK team is happy to announce the immediate availability of Sentry PHP SDK v3.18.2.
+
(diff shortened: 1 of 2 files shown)


stayallive (approved):
For readers, this is a temporary fix to fix the issues we are seeing when trying to use the SDK. We will have a more permanent and correct fix (without needing to require an abandoned package) soon but for now this solves the installation and usage issues we are seeing with the Sentry SDK.

jacow:
I got a warning "Package php-http/message-factory is abandoned, you should avoid using it. Use psr/http-factory instead." on Laravel 9.2 . Maybe it would be cooler if sentry using psr/http-factory instead.

PR #1542, opened 2023-05-24
Use HTTP/1.1 when compression is enabled

stayallive:
curl 8.1+ is more strict in it's handling for http/2 requests and this is causing issues with the current state of our compression. Since there is no good way to detect from userland if we are going to make a http/2 request we can't omit the problematic `TE` header dynamically. Removing the compression altogether is also not really an option since payloads might be rejected by the Sentry server because uncompressed they would be too large.

So I landed on the option to force http/1.1 requests if compression is enabled until we can figure out a better solution to the problem where we can support both 1.1 and 2.0 requests. However, it's not a major issue since http/2 doesn't offer many if any benefits for the type of communication the Sentry SDK does.

I've also taken the opportunity to add support for Guzzle 7 timeout and proxy options since those were not implemented from the looks of it.

Note that I didn't force the http version on the guzzle client since 1.1 is [their default](https://docs.guzzlephp.org/en/stable/request-options.html#version).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # CHANGELOG
 
+## 3.19.1
+
+The Sentry SDK team is happy to announce the immediate availability of Sentry PHP SDK v3.19.1.
+
+### Bug Fixes
+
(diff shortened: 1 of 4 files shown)


cleptric (approved):
🚀

PR #1559, opened 2023-06-30
Compat support for PHPUnit `10`

ste93cry:
Now that PHPUnit `10` has been released, a few things that changed must be addressed before even considering updating to it. The biggest change that affects us is that the data providers are now expected to be `static` functions. While updating things here and there, I also found some little typos or mistakes in the code that I fixed in separate commits.

### DIFF
--- a/src/Integration/FrameContextifierIntegration.php
+++ b/src/Integration/FrameContextifierIntegration.php
@@ -27,7 +27,7 @@ final class FrameContextifierIntegration implements IntegrationInterface
     /**
      * Creates a new instance of this integration.
      *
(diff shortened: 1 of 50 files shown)


Jean85 (approved):
Very nice cleanup! Wow do we use a lot of data providers :smile: 

BTW `static` is not a requirement for PHPUnit 10, just a deprecation... but it's nice to be already on top of it!

ste93cry:
> BTW `static` is not a requirement for PHPUnit 10, just a deprecation... but it's nice to be already on top of it!

Right, non-`static` providers will be unsupported in PHPUnit `11`

PR #1568, opened 2023-07-13
Strip path prefixes on profile frames

stayallive:
This was an open TODO that is resolved!

### DIFF
--- a/tests/Profiling/ProfileTest.php
+++ b/tests/Profiling/ProfileTest.php
@@ -9,6 +9,7 @@ use Sentry\Context\OsContext;
 use Sentry\Context\RuntimeContext;
 use Sentry\Event;
 use Sentry\EventId;
+use Sentry\Options;
 use Sentry\Profiling\Profile;
 
 final class ProfileTest extends TestCase
@@ -16,9 +17,9 @@ final class ProfileTest extends TestCase
     /**
      * @dataProvider formattedDataDataProvider
      */
-    public function testGetFormattedData(Event $event, array $excimerLog, $expectedData): void
+    public function testGetFormattedData(Event $event, array $excimerLog, $expectedData, ?Options $options = null): void
     {
-        $profile = new Profile();
+        $profile = new Profile($options);
         // 2022-02-28T09:41:00Z
         $profile->setStartTimeStamp(1677573660.0000);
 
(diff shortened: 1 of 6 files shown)


ste93cry:
Wouldn't be better to make `PrefixStripper` a `final class` with a `static` method accepting the list of prefixes instead of using a `trait`?

stayallive:
That could also work yes. I don't think there is a technical reason why either way is "better" so potato, potahto really in this case 😄

tests/Profiling/ProfileTest.php
```diff
@@ -217,6 +218,176 @@ public static function formattedDataDataProvider(): \Generator
             ],
         ];
 
+        yield [
+            $event,
+            [
+                [
+                    'trace' => [
+                        [
+                            'file' => '/index.php',
```
cleptric, line 227:
```suggestion
                            'file' => '/var/www/html/index.php',
```

and below, to actually test it 😬

cleptric (approved):
A quick test run with some app that reports to https://sentry.io would be appreciated, besides that, :shipit:

ste93cry:
> That could also work yes. I don't think there is a technical reason why either way is "better"

Some of the reasons I can think of is that:

- Even if not impossible, they are more difficult to test than a `class` or a `method`
- They introduce horizontal inheritance, so understanding a code base can be significantly harder. Injecting a dependency is a lot cleaner in this aspect, as you make it clear what a class needs to work. But in this case, a `static` helper method would probably be enough
- Since a `trait` is basically a way to copy-paste some code in multiple places, you are effectively coupling it with the classes it will be used in, risking to make some change in one that doesn't suit the others. In this case the method does not access anything on $this, but you never know in the future what could happen.

Lately I've seen a lot of discussions around traits and why people think they are either good or bad. Honestly I don't have a real strong opinion on this, and in this case I agree that it doesn't really matter probably, but if it was me I would have created an helper class. Also, I don't like passing around all the client config when the function really needs just a plain `array` of strings.

If you really want to leave it as-is, I would at least add the `Trait` suffix to the trait to make it explicit what is it.

tests/Profiling/ProfileTest.php
stayallive: This one I don't get, I've duplicated the existing test case, to have one with and one without prefixes. Isn't that good enough for a test or did I misinterpret your comment?

(test case to sentry.io coming up!)

stayallive:
For the requested sentry.io test (using a Laravel application):

[Before](https://devalex.sentry.io/profiling/profile/tny/e1d323e9f2984415b8633d8b17612418/flamechart/?colorCoding=by+system+vs+application+frame&query=&sorting=call+order&tid=0&view=top+down):

image

[After](https://devalex.sentry.io/profiling/profile/tny/2ec2959c321e496e8eaf1995701ed0b6/flamechart/?colorCoding=by+system+vs+application+frame&query=&sorting=call+order&tid=0&view=top+down):

image

Looks a lot better to me 😎

tests/Profiling/ProfileTest.php
stayallive: Oh I got it, 🤦 yes you are totally right... made the input the same for both test cases now 😅

PR #1573, opened 2023-07-26
Add captureCheckIn()

will2877:
This PR comes from the discussion related to this topic in #https://github.com/getsentry/sentry-symfony/pull/719 and adds the upsert `MonitorConfig` to the `HubInterface` as well as the descending `Hub` and `HubAdapter`.

This way we don't need to add the `environment` and `release` details multiple times but will take this information from the `Options` as provided by the `ClientInterface` in the Hub.


src/State/HubInterface.php
```diff
@@ -112,6 +115,18 @@ public function captureEvent(Event $event, ?EventHint $hint = null): ?EventId;
...
+     * Captures a CheckIn to the configured Monitor.
+     *
+     * @param string $slug Identifier of the Monitor
+     * @param MonitorConfig $upsertMonitorConfig Configuration of the Monitor
+     * @param CheckInStatus $status The status of the Monitor
+     * @param CheckIn|null $previos A CheckIn that may have preceded the current CheckIn
+     *
+     * @return CheckIn The created CheckIn
+     */
+    public function captureCheckIn(string $slug, MonitorConfig $upsertMonitorConfig, CheckInStatus $status, ?CheckIn $previous = null): CheckIn;
```
cleptric, line 128: This is a breaking change, please remove it.

cleptric (changes requested):
Thank you for your contribution! 🚀

Please also add `captureCheckIn` to `functions.php`.

I'll give this a quick spin and review this further later this week.

src/State/HubInterface.php
ste93cry: More precisely, we should add the method definition to the PHPDoc of the class using the `@method` tag

will2877: This was a typo.

ste93cry: The typo had nothing to do with the request: a new method cannot be added to an interface without causing a breaking change because that would force all implementors to add it, otherwise their code will not run. That's why I asked to add the method to the PHPDoc: this way we document it and in the next major version of the SDK we will add the method for real.

src/functions.php
```diff
@@ -66,6 +66,21 @@ function captureLastError(?EventHint $hint = null): ?EventId
...
+ * Captures a CheckIn to the configured Monitor.
+ *
+ * @param string        $slug                Identifier of the Monitor
+ * @param MonitorConfig $upsertMonitorConfig Configuration of the Monitor
+ * @param CheckInStatus $status              The status of the Monitor
+ * @param CheckIn|null  $previous            A CheckIn that may have preceded the current CheckIn
+ *
+ * @return CheckIn The created CheckIn
+ */
+function captureCheckIn(string $slug, MonitorConfig $upsertMonitorConfig, CheckInStatus $status, ?CheckIn $previous = null): ?CheckIn
```
cleptric, line 79: Commenting here, but this applies down the call chain.

- A `MonitorConfig` should be optional. People might create their monitor in the UI; hence no need to require the config object. Would also call the parameter `monitorConfig`.
- We still want to support so-called Heartbeats, a single check-in with status `ok` or `error` and an optional duration set.
- We only need the prior check-in ID, so not sure if we should pass in the whole object. Similar to `captureEvent`, we might return the check-in ID only. We could create a concrete `CheckInId` type.

Some examples of how the API could be used:

```php
$checkInId = Sentry\captureCheckIn(
    slug: 'slug',
    status: CheckInStatus::inProgress()
);


// do stuff

Sentry\captureCheckIn(
    checkInId: $checkInId,
    slug: 'slug',
    status: CheckInStatus::ok()
);
```

```php
$checkInId = Sentry\captureCheckIn(
    slug: 'slug',
    status: CheckInStatus::inProgress(),
    monitorConfig: $monitorConfig,
);

// do stuff

Sentry\captureCheckIn(
    checkInId: $checkInId,
    slug: 'slug',
    status: CheckInStatus::ok()
);
```

```php
Sentry\captureCheckIn(
    slug: 'slug',
    status: CheckInStatus::ok(),
    duration: 10,
);
```

src/State/Hub.php
```diff
@@ -169,6 +172,31 @@ public function captureLastError(?EventHint $hint = null): ?EventId
         return null;
     }
 
+    public function captureCheckIn(string $slug, MonitorConfig $upsertMonitorConfig, CheckInStatus $status, ?CheckIn $previous = null): ?CheckIn
+    {
+        $client = $this->getClient();
+
+        if (!$client instanceof ClientInterface) {
```
ste93cry, line 179: I would find more appropriate to check that the client is not `null` because the the real meaning of this line is to ensure that that instance is not `null` and not that it implements the interface. It is a small change indeed, but imo it conveyes to the reader the expectation in a clearer way.

src/State/Hub.php
```diff
@@ -169,6 +172,31 @@ public function captureLastError(?EventHint $hint = null): ?EventId
...
+            return null;
+        }
+
+        $options = $client->getOptions();
+        $event = Event::createCheckIn();
+        $checkIn = new CheckIn(
+            $slug,
+            $status,
+            $previous ? $previous->getId() : null,
```
ste93cry, line 188: As per above, I think it's always better to be explicit with checks: in this case, we really want to make it clear to the reader that `$previous` must not be `null`, so I would change the code accordingly

tests/FunctionsTest.php
```diff
@@ -185,6 +191,65 @@ public static function captureLastErrorDataProvider(): \Generator
...
+            'environment' => Event::DEFAULT_ENVIRONMENT,
+            'release' => '1.1.8',
+        ]);
+        /** @var ClientInterface&MockObject $client */
+        $client = $this->createMock(ClientInterface::class);
+        $client->expects($this->once())
+            ->method('getOptions')
+            ->willReturn($options);
+
+        SentrySdk::setCurrentHub($hub);
```
ste93cry, line 210: We should avoid calls like this in the tests as it affects the singleton instance of the SDK, which is shared across all tests. While it's true that we have a PHPUnit listener that resets the state after a test ran, it's better to proactively avoid issues. Instead, if possible, prefer mocking the hub like we did in similar tests

tests/FunctionsTest.php
```diff
@@ -185,6 +191,65 @@ public static function captureLastErrorDataProvider(): \Generator
...
+        $this->assertNull(captureCheckIn(...$expectedFunctionCallArgs));
+        $hub->bindClient($client);
+
+        $checkIn = captureCheckIn(...$expectedFunctionCallArgs);
+        $this->assertInstanceOf(CheckIn::class, $checkIn);
+        $this->assertSame(Event::DEFAULT_ENVIRONMENT, $checkIn->getEnvironment());
+        $this->assertSame('1.1.8', $checkIn->getRelease());
+    }
+
+    public static function captureCheckInProvider(): \Generator
```
ste93cry, line 221: I'm not sure I understand why this data provider is needed: the only difference that I see is that one test case returns a monitor for a crontab schedule while the other returns a monitor for an interval schedule. However, I fail to see how this could affect the public API: at this level of unit testing, we should only care about the fact that the call is proxied to the hub instance. This comment applies to the rest of the tests as well.

ste93cry (changes requested):

tests/FunctionsTest.php
cleptric: This is fine, we do this in other tests as well. We can fix it once it becomes a real issue.

cleptric:
@will2877 Thanks again for opening this PR. I pushed some changes to your branch and we'll get this merged soon 🙂

ste93cry (approved):

cleptric (approved):

PR #1574, opened 2023-07-27
[4.x] Profiler frame processor

stayallive:
A very pragmatic way to set a frame processor. This is intended for internal use in framework SDKs to do further processing on the frame. For example, in Laravel we want to show original view paths instead of the cached view path for frames inside blade views.


stayallive:
See getsentry/sentry-laravel#747 for an example on how I'd use it.

Issue #1541, opened 2023-05-23
Replace php-http/message-factory


mrsombre:
### How do you use Sentry?

Sentry Saas (sentry.io)

### Version

3.19.0

### Steps to Reproduce

Execute `composer require sentry/sentry`.

### Expected Result

I expect `psr/http-factory` is used and no warning appears.

### Actual Result

```
./composer.json has been updated
Running composer update sentry/sentry
Loading composer repositories with package information
Updating dependencies
Nothing to modify in lock file
Writing lock file
Installing dependencies from lock file (including require-dev)
Nothing to install, update or remove
Package php-http/message-factory is abandoned, you should avoid using it. Use psr/http-factory instead.
Generating autoload files
59 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
No security vulnerability advisories found
Using version ^3.19 for sentry/sentry
```

cleptric:
We currently rely on `php-http/message-factory` and can't easily remove this dependency without a likely breaking change. Closing as won't fix.

nicogominet:
> Package php-http/message-factory is abandoned, you should avoid using it. Use psr/http-factory instead.
I also would like to get this fixed.

It looks like this is [a temporary solution](https://github.com/getsentry/sentry-php/pull/1534) anyway as @stayallive said, so I don't understand why the "won't fix" end-of-discussion answer here. Unless I misunderstood the other PR, if anything this **will** be fix "soon".

cleptric:
For some more contex:

The Symonfy HTTP Client version 6.2 and below will throw an exception if `php-http/message-factory` is absent. See https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/HttpClient/HttplugClient.php#L48-L50

While this behaviour was removed in 6.3 and up, we can't force people to upgrade to the latest version via the SDK. We already ran into this issue in https://github.com/getsentry/sentry-php/issues/1533, which was caused by `php-http/client-common` removing this dependency.

nicogominet:
Thank you @cleptric for the explanation, I think I understand why this is happening now.

In my case, I believe this will be fixed when `6.3` will be released (currently latest is `RC2`).

rintdev:
Hmm, can't we keep this issue open to track this? I'm trying to avoid using abandoned package dependencies.

gndk:
> While this behaviour was removed in 6.3 and up, we can't force people to upgrade to the latest version via the SDK.

[...]

PR #1576, opened 2023-07-31
Support Symfony `PSR-18` HTTP client

ste93cry:
To work around issue #1541, we could support the built-in Symfony `PSR-18` HTTP client and make it the preferred one to use. `Http\Client\Common\PluginClient` ensures that a non-async client gets wrapped by [`EmulatedHttpAsyncClient`](https://github.com/php-http/client-common/blob/2.x/src/EmulatedHttpAsyncClient.php) and this allows us to not have to change anything in the rest of the codebase. There is no version of Symfony that doesn't have both the `PSR-18` and Httplug client, so the possibilities of instantiating the `SymfonyHttplugClient` class should be none.

Since the `HttpTransport` is already [syncronous](https://github.com/getsentry/sentry-php/blob/bf0bdd367d56a40a47a369715fcec3688a00769b/src/Transport/HttpTransport.php#L131), I don't expect any performance penalty with this change. However, the dependency on the `php-http/message-factory` package can now be removed safely.

### DIFF
--- a/src/HttpClient/HttpClientFactory.php
+++ b/src/HttpClient/HttpClientFactory.php
@@ -25,6 +25,7 @@ use Sentry\HttpClient\Plugin\GzipEncoderPlugin;
 use Sentry\Options;
 use Symfony\Component\HttpClient\HttpClient as SymfonyHttpClient;
 use Symfony\Component\HttpClient\HttplugClient as SymfonyHttplugClient;
+use Symfony\Component\HttpClient\Psr18Client as SymfonyPsr18Client;
 
 /**
  * Default implementation of the {@HttpClientFactoryInterface} interface that uses
@@ -111,7 +112,7 @@ final class HttpClientFactory implements HttpClientFactoryInterface
      */
     private function resolveClient(Options $options)
     {
-        if (class_exists(SymfonyHttplugClient::class)) {
+        if (class_exists(SymfonyPsr18Client::class) || class_exists(SymfonyHttplugClient::class)) {
             $symfonyConfig = [
                 'timeout' => $options->getHttpConnectTimeout(),
                 'max_duration' => $options->getHttpTimeout(),
@@ -120,6 +121,10 @@ final class HttpClientFactory implements HttpClientFactoryInterface
 
             if (null !== $options->getHttpProxy()) {
                 $symfonyConfig['proxy'] = $options->getHttpProxy();
+            }
+
+            if (class_exists(SymfonyPsr18Client::class)) {
+                return new SymfonyPsr18Client(SymfonyHttpClient::create($symfonyConfig));
             }
 
             return new SymfonyHttplugClient(SymfonyHttpClient::create($symfonyConfig));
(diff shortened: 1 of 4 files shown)


cleptric:
Can you push the branch to this repo? This will make it easier for folks giving this a try.

ste93cry:
Sure, [here](https://github.com/getsentry/sentry-php/tree/support-symfony-psr18-http-client) it is. I will keep it in sync with the one in my fork repo so that I don't have to open a new PR.

src/HttpClient/HttpClientFactory.php
```diff
@@ -111,7 +112,7 @@ public function create(Options $options): HttpAsyncClientInterface
      */
     private function resolveClient(Options $options)
     {
-        if (class_exists(SymfonyHttplugClient::class)) {
+        if (class_exists(SymfonyPsr18Client::class) || class_exists(SymfonyHttplugClient::class)) {
```
cleptric, line 115: While this might be more of a hypothetical concern if, for any reason, we do end up with the `SymfonyHttplugClient` or if we fall through and `php-http/discovery defaults to `SymfonyHttplugClient`, we run into the same issue we're trying to solve.

Again, our stance is that we can accept living with an abandoned package for a bit longer. Causing exceptions, even if they do not stem from the SDK, is not ok though.

ste93cry: I can easily remove everything related to the Httplug client, the PSR-18 client is more than enough as explained above. Do you feel safer this way?

cleptric: Not for the `php-http/discovery` case. Let's keep it for the moment. Maybe we get some feedback in #1541

ste93cry: >Not for the php-http/discovery case. 

We can technically replace that case with the PSR-18 discovery. The only thing that would change is that we have to require `psr/http-client-implementation` instead of `php-http/async-client-implementation`. I may think about an unlikely case where you will not be able to update quietly to the new version because you are missing installed a PSR-18 compatible client, but as far as I see every single supported HTTP client by the autodiscovery that would be resolved now is also PSR-18 compatible, so there should be nothing to do anything: if it worked before, it should work next.

ste93cry:
@cleptric do you want to move forward with this? One week should be enough to gather for feedback considering the activity of the original issue

cleptric:
@ste93cry please update the PR and move the branch to this repo. I can run some tests this week.

ste93cry:
>move the branch to this repo

I've already pushed the branch to this repo too. However I would prefer to keep working on my fork so that I don't need to close this PR, resulting in a loss of the context and conversations we had. In the end, the source branch should not matter much...

ste93cry:
@cleptric do you have any news? If you don't want to proceed, I will close the PR because there's no point in keeping it open for so long

ste93cry:
Closing this because it's evident that there is a lack of interest in merging this PR, especially now that the new SDK is out.

PR #1577, opened 2023-08-01
Improve tests of `captureCheckIn()`

ste93cry:
Round of small improvements to the tests of the `captureCheckIn()` function. I also noticed that the method definition added to `HubInterface` was incorrect, so I fixed it.

### DIFF
--- a/src/State/HubInterface.php
+++ b/src/State/HubInterface.php
@@ -23,7 +23,7 @@ use Sentry\Tracing\TransactionContext;
  * stack of pairs of clients and scopes. It is the main entry point to talk
  * with the Sentry client.
  *
- * @method string|null captureCheckIn(string $slug, CheckInStatus $status, $duration = null, ?MonitorConfig $upsertMonitorConfig = null, ?string $checkInId = null) Captures a check-in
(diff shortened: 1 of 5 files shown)


src/State/HubInterface.php
```diff
@@ -23,7 +23,7 @@
  * stack of pairs of clients and scopes. It is the main entry point to talk
  * with the Sentry client.
  *
- * @method string|null captureCheckIn(string $slug, CheckInStatus $status, $duration = null, ?MonitorConfig $upsertMonitorConfig = null, ?string $checkInId = null) Captures a check-in
+ * @method string|null captureCheckIn(string $slug, CheckInStatus $status, int|float|null $duration = null, ?MonitorConfig $monitorConfig = null, ?string $checkInId = null) Captures a check-in
```
cleptric, line 26: Does this make phpstan happier now? Maybe we can update the baseline file.

ste93cry: The baseline hasn't changed since 2 months, why should this change affect it?

cleptric: Ahh, I meant we might be able to update the doc blocks to `{@inheritdoc}`.

ste93cry: It doesn't work, but we can use `{@inheritdoc}` and just add `@param int|float|null $duration` next to it to make PHPStan happier, which is better than the current situation anyway

PR #1578, opened 2023-08-03
Try PCOV coverage driver

stayallive:
It's suggested the PCOV driver is more performant and might have more stable output.


Jean85:
The performance side may not be true anymore with Xdebug 3 and its `mode`; but I'm not sure if that's correctly used in CI.

stayallive:
Yeah this was mostly testing for stability since xdebug didn't always give stable results, but that was solved another way so for now there is not much reason to go forward with this 😄

PR #1582, opened 2023-08-14
Merge EventHint extra parameter into Event

Eredost:
When using capture methods, the 'extra' parameter inside EventHint is ignored and so does not appear in Sentry reports.

```php
captureException($exception, EventHint::fromArray([
    'extra' => ['foo' => 'bar'],
]));
```

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -248,6 +248,10 @@ final class Client implements ClientInterface
             if (null !== $hint->stacktrace && null === $event->getStacktrace()) {
                 $event->setStacktrace($hint->stacktrace);
             }
+
+            if (!empty($hint->extra)) {
+                $event->setExtra(array_merge($event->getExtra(), $hint->extra));
+            }
         }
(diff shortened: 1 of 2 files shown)


stayallive:
Hey @Eredost we've had this PR before: #1380.

We ended up not implementing it since the `extra` data is deprecated and (confusingly) the `extra` on the `EventHint` is not 1:1 the same as the `extra` data on an `Event`.

More context: https://github.com/getsentry/sentry-php/pull/1380#issuecomment-1259623842 https://github.com/getsentry/sentry-php/pull/1380#issuecomment-1259689864

Let me know if that helps and if there is any way we can help you put the data on the scope instead of in the hint instead: https://docs.sentry.io/platforms/php/enriching-events/context/

Eredost:
Hi @stayallive 

I didn't know that we could add context this way, it will do the job, thanks a lot!

PR #1587, opened 2023-09-13
Rollback to is_a() to ignore exceptions instead of in_array()

fmata:
Hello,

https://github.com/getsentry/sentry-php/pull/1503 reintroduced `ignore_exceptions` in config and deprecated `IgnoreErrorsIntegration`. As reported in https://github.com/getsentry/sentry-php/pull/1503#discussion_r1144412115 the behavior changed : before we could list interfaces or class hierarchy to ignore, after we must list all children exceptions (dozens or more in a decent project as stated by @stayallive).

It has been merged and released in the minor release 3.17. But deprecates a feature whithout upgrade path to have the exact 
same feature in a minor release is not sem ver compliant.

Can you merge my PR and release a new patch version to rollback as before 3.17 ?

Thanks

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -327,13 +327,15 @@ final class Client implements ClientInterface
             }
 
             foreach ($exceptions as $exception) {
-                if (\in_array($exception->getType(), $this->options->getIgnoreExceptions(), true)) {
-                    $this->logger->info(
-                        'The event will be discarded because it matches an entry in "ignore_exceptions".',
-                        ['event' => $event]
-                    );
+                foreach ($this->options->getIgnoreExceptions() as $ignoredException) {
+                    if (is_a($exception->getType(), $ignoredException, true)) {
+                        $this->logger->info(
+                            'The event will be discarded because it matches an entry in "ignore_exceptions".',
+                            ['event' => $event]
+                        );
 
-                    return null;
+                        return null;
+                    }
                 }
             }
         }
--- a/tests/ClientTest.php
+++ b/tests/ClientTest.php
@@ -26,6 +26,7 @@ use Sentry\Serializer\SerializerInterface;
 use Sentry\Severity;
 use Sentry\Stacktrace;
 use Sentry\State\Scope;
+use Sentry\Tests\Fixtures\code\CustomException;
 use Sentry\Transport\TransportFactoryInterface;
 use Sentry\Transport\TransportInterface;
 use Symfony\Bridge\PhpUnit\ExpectDeprecationTrait;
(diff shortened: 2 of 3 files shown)


cleptric:
While I agree that `is_a` might be beneficial, making these changes now can have worse effects on people who already use the new option. Hence, I'm not moving forward with this for now.

We can address this in the next major version of the SDK, which we already started working on.

fmata:
@cleptric beneficial or not, it was the behavior until a **minor** release (3.17) which changed it without warning or upgrade note. The mistake is here, 3.17 already breaks many apps by alerting with false positives. Revert a breaking change of a documented feature is sem-ver compliant.

Mistake can happens :) but do not respect sem-ver is a bad signal for your users.

Or maybe there is a solution not deprecated. Can you give me an hint ?

cleptric:
Can you clarify what you mean by `breaks many apps by alerting with false positives`?
I also don't see any breaking change introduced in 3.17. The `IgnoreErrorsIntegration` works exactly as to prior versions. Hence, I find the claim us not following semver a bit exaggerated.

fmata:
Before 3.17 : 
```yaml
sentry:
    dsn: '%env(SENTRY_DSN)%'
    options:
        environment: '%app.deploy_env%'
        integrations:
            - Sentry\Integration\IgnoreErrorsIntegration
        send_default_pii: true
    register_error_listener: false
    register_error_handler: false

monolog:
    handlers:
        sentry:
            type: sentry
            level: !php/const Monolog\Logger::ERROR
            hub_id: Sentry\State\HubInterface

services:
    Sentry\Integration\IgnoreErrorsIntegration:
        arguments:
            $options:
                ignore_exceptions:
                    - ApiPlatform\Symfony\Validator\Exception\ValidationException
                    - Symfony\Component\HttpKernel\Exception\HttpExceptionInterface
                    - Symfony\Component\Security\Core\Exception\AccessDeniedException
```
After : 
```yaml
sentry:
    dsn: '%env(SENTRY_DSN)%'
    options:
        environment: '%app.deploy_env%'
        ignore_exceptions:
            - ApiPlatform\Symfony\Validator\Exception\ValidationException
            - Symfony\Component\HttpKernel\Exception\HttpExceptionInterface
            - Symfony\Component\Security\Core\Exception\AccessDeniedException
        send_default_pii: true
    register_error_listener: false
    register_error_handler: false

monolog:
    handlers:
        sentry:
            type: sentry
            level: !php/const Monolog\Logger::ERROR
            hub_id: Sentry\State\HubInterface
```
All children of HttpExceptionInterface (NotFoundHttpException, BadRequestHttpException and so on...) are ignored.

cleptric:
I recommend reverting back to `IgnoreErrorsIntegration` if you rely on the `is_a` check. In the next major version, we'll add this to `ignore_exceptions` as well as RegEx support.

stayallive (approved):
Excellent change. Going to make it much easier to ignore a certain class of exceptions 👍

cleptric:
Will fix CI later.

Issue #1586, opened 2023-09-05
PHP SDK 4.0.0


stephanie-anderson:
This outlines the work we plan for the next major version of the PHP SDK.

- [x] https://github.com/getsentry/sentry-php/pull/1589

## Remove deprecated methods

Most of the current methods that are marked as deprecated will be removed. An extensive upgrade guide will be provided.

## Ship a default HTTP client

After internal discussions, we decided to ship a default HTTP client based on cURL. A Sentry-provided interface will still support the use of third-party HTTP clients. For our Symfony SDK, we'll rely on the Symfony HTTP Client instead. The Laravel SDK will use the default HTTP client.

## Remove most dependencies

Once we rely on our own HTTP client, we can remove the majority of dependencies of the SDK. 

## Remove the use of the `/store` endpoint

We'll remove the use of the `/store` endpoint, which currently is only used for error reporting if performance is disabled in the SDK. This will bump the self-hosted version requirement to 20.6.0 or higher, released on June 16, 2020.

## Simplified performance APIs

We are currently working on removing transactions from the public API surface in all SDKs and will therefore revamp the performance API of the PHP SDK with an emphasis on spans.

## Other topics

- We'll aim to provide a better API than the current `PayloadSeralizer`, making it easier for our users to provide custom serializers.
- Most classes will be annotated as `@final` instead of using the  `final` keyword.
- Most container objects will provide a fluent API.
- Based on RFC [#34](https://github.com/getsentry/rfcs/pull/43), we'll introduce lifecycle hooks to the SDK.
- Based on RFC [#62](https://github.com/getsentry/rfcs/pull/62), we'll add an event scrubber to exhaustively control both PII (Personally Identifiable Information) and security credentials that are sent to Sentry.
- The SDK won't actively throw any exceptions moving forward, necessitating a better `debug: true` experience for local development.

ste93cry:
I beg you, please do not ship your own HTTP client. I can live with the decision of shipping a well-known HTTP client by default (ignoring all the negative feedback you got, btw), but not with developing yet another client, which was also the way SDK `1.x` worked and that had a lot of issues for this reason.

Jean85:
I definitely agree with @ste93cry request. Nearly all clients in the PHP ecosystem use Curl underneath, and Guzzle or Symfony are both two very stable clients.
Yes, we had some issues recently, but those were mainly related to the HTTPlug stuff. We can ditch that and stay on a stable client.

[...]

PR #1589, opened 2023-09-20
[4.x] Add a default cURL HTTP client

cleptric:
**_Work in Progress_**

This adds a cURL-based HTTP client and simplifies the setup when using a custom `Transport` or `HttpClient`.
As mentioned in #1586, this removes the `/store` endpoint entirely. All events are now sent to the `/envelope` endpoint.


src/ClientBuilder.php
```diff
@@ -136,60 +143,35 @@ public function setSdkVersion(string $sdkVersion): ClientBuilderInterface
...
-        return $transportFactory->create($this->options);
+        return $this;
     }
 
     /**
-     * Creates a new instance of the {@see DefaultTransportFactory} factory.
+     * {@inheritdoc}
      */
-    private function createDefaultTransportFactory(): DefaultTransportFactory
+    public function getClient(): ClientInterface
```
Jean85, line 173: Having `getClient` and `getHttpClient` may be confusing... What's the difference? Can we use another name?

We also have Transport too!

src/HttpClient/HttpClient.php
```diff
@@ -0,0 +1,158 @@
...
+        $this->sdkIdentifier = $sdkIdentifier;
+        $this->sdkVersion = $sdkVersion;
+    }
+
+    public function sendRequest(string $requestData): Response
+    {
+        $dsn = $this->options->getDsn();
+        if (null === $dsn) {
+            throw new \RuntimeException('The DSN option must be set to use the HttpClient.');
+        }
```
Jean85, line 39: What happens when we set the DSN to `null` to disable the client? This may be a big BC, for framework libraries too, we need to propagate this outside.

src/HttpClient/HttpClient.php
```diff
@@ -0,0 +1,158 @@
...
+
+        $curlHandle = curl_init();
+        curl_setopt($curlHandle, \CURLOPT_URL, $dsn->getEnvelopeApiEndpointUrl());
+        curl_setopt($curlHandle, \CURLOPT_HTTPHEADER, $this->getRequestHeaders());
+        curl_setopt($curlHandle, \CURLOPT_USERAGENT, $this->sdkIdentifier . '/' . $this->sdkVersion);
+        curl_setopt($curlHandle, \CURLOPT_TIMEOUT, $this->options->getHttpTimeout());
+        curl_setopt($curlHandle, \CURLOPT_CONNECTTIMEOUT, $this->options->getHttpConnectTimeout());
+        curl_setopt($curlHandle, \CURLOPT_SSL_VERIFYPEER, true);
```
Jean85, line 47: What if I want to disable this? i.e. if I have a local, self-hosted, not exposed, not behind TLS Sentry server?

[edit] I saw the TODO below, this row is a duplicate.

src/HttpClient/HttpClient.php
```diff
@@ -0,0 +1,158 @@
...
+        curl_setopt($curlHandle, \CURLOPT_ENCODING, '');
```
Jean85, line 48: Isn't this the option to have compression? We may skip that in this PR (gzip caused issues in the past) but we may require it as a feature later.

src/HttpClient/HttpClient.php
```diff
@@ -0,0 +1,158 @@
...
+        ];
+
+        $dsn = $this->options->getDsn();
+        if (null === $dsn) {
+            return $headers;
+        }
+
+        $data = [
+            'sentry_version' => Client::PROTOCOL_VERSION,
+            'sentry_client' => $this->sdkIdentifier . '/' . $this->sdkVersion,
```
Jean85, line 116: These two values always go together... Maybe they need a dedicated value object?

src/HttpClient/HttpClient.php
```diff
@@ -0,0 +1,158 @@
...
+        ]);
+    }
+
+    /**
+     * @TODO(michi) This might need a bit more love,
+     * but we only really care about X-Sentry-Rate-Limits and Retry-After
+     *
+     * @return string[]
+     */
+    protected function getResponseHeaders(?int $headerSize, string $body): array
```
Jean85, line 140: This is a BIG hot spot: making some mistakes here would open the door for CVEs.

src/HttpClient/Response.php
```diff
@@ -0,0 +1,63 @@
...
+
+    public function getError(): string
+    {
+        return $this->error;
+    }
+
+    public function hasError(): bool
+    {
+        return '' !== $this->error;
+    }
```
Jean85, line 62: This is fragile: what if we can't fetch the error message?

Jean85 (commented):
As I've already said in https://github.com/getsentry/sentry-php/issues/1586#issuecomment-1721864613, I'm against writing our own HTTP client but, since I recognize I don't have a final say in this, I'm still giving my review.

I would at least suggest leveraging PSR-18 (HTTP client) and so PSR-7 (request/response) with it, but we could ditch PSR-17 (factories) to keep it simple.

ste93cry:
I agree with everything already said by @Jean85: first of all, I don’t understand the reason to have our own `HttpClientInterface` instead of just using the PSR-18 one. There is no benefit I can think of, it would be just another layer that needs to be implemented by the end user. Second, and more importantly, I don’t understand why we are reimplementing a full cURL client, which we have to maintain and for which there is no benefit because it’s a lot simpler and less customizable than any other out of here. Moreover, what’s happening here is that we are reinventing the wheel: I can understand that this SDK has to ship with a default client, even though I don’t agree, but at least choose one reliable client instead of reinventing the wheel. We had our own HTTP client in the past, and it was full of issues. Given that the SDK is integrated with both Laravel and Symfony, it would be wise to choose the `symfony/http-client` which is common for both

src/Transport/Result.php
```diff
@@ -2,16 +2,18 @@
...
-namespace Sentry;
+namespace Sentry\Transport;
+
+use Sentry\Event;
 
 /**
  * This class contains the details of the sending operation of an event, e.g.
  * if it was sent successfully or if it was skipped because of some reason.
  */
-final class Response
```
ste93cry, line 11: What's the point of removing `final` here? This class is a value object and it doesn't need to be extended by third parties

src/Transport/ResultStatus.php
```diff
@@ -2,13 +2,13 @@
...
 
-namespace Sentry;
+namespace Sentry\Transport;
 
 /**
  * This enum represents all possible reasons an event sending operation succeeded
  * or failed.
  */
-final class ResponseStatus implements \Stringable
+class ResultStatus implements \Stringable
```
ste93cry, line 11: What's the point of removing `final` here? This class represents an `enum`, which by definition has a fixed set of values, so it doesn't have to be extended or it would be pointless

src/Transport/ResultStatus.php
```diff
@@ -2,13 +2,13 @@
...
 declare(strict_types=1);
 
-namespace Sentry;
+namespace Sentry\Transport;
 
 /**
  * This enum represents all possible reasons an event sending operation succeeded
  * or failed.
  */
-final class ResponseStatus implements \Stringable
```
ste93cry, line 11: What's the point of removing `final` here? This class represents an enum, which by definition has a fixed set of values, so it doesn't have to be extended or it would be pointless

src/HttpClient/Response.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sentry\HttpClient;
+
+class Response
```
ste93cry, line 7: What's the point of removing `final` here? This class is a value object and it doesn't need to be extended by third parties

src/HttpClient/HttpClient.php
cleptric: This flag only handles automatic decompression of HTTP downloads. Setting it to `''` enables all supported built-in compressions.

src/HttpClient/HttpClient.php
cleptric: Please expand on this. The only two header values currently used are `X-Sentry-Rate-Limits` and `Retry-After` and I fail to see how this could lead to a vulnerability.

src/HttpClient/HttpClient.php
cleptric: An upcoming feature will require us always to construct an envelope in certain scenarios. I agree that this is not the best solution. I have to think about improving it.

src/HttpClient/Response.php
cleptric: https://github.com/getsentry/sentry-php/blob/fa5145d9d7f0531291769b2296272ca4c57cf412/src/HttpClient/HttpClient.php#L86 currently sets this. All other cases pass `''`. We could make it optional, but I don't see a scenario were this would not work.

src/HttpClient/HttpClient.php
cleptric: I added `http_ssl_verify_peer`. Your use case is the same one I had in mind.

src/ClientBuilder.php
Jean85: Should we mark that as `@internal` then?

cleptric: Totally, this is all still WIP, as mentioned in the PR description. 😃
Docs and annotation will follow once the implementation and tests are in place.

src/HttpClient/HttpClient.php
Jean85: Just today I got a notification about a Guzzle CVE that hit them a few months ago with a 7.5 score about headers: https://security.snyk.io/vuln/SNYK-PHP-GUZZLEHTTPPSR7-5426094

Handling headers could open us to CVE reports.

cleptric: Yep, but I would argue that this heavily depends on the context of how the header values are used. We could limit the parsing on the mentioned two headers. In the context of Guzzle, I agree that this is much more critical than in our case.

Most SDKs will talk to a somewhat "trusted" server, be it Sentry SaaS or self-hosted, and even if someone sneaks in some tampered headers, the worst I can see right now is that the rate-limiting won't work.

src/ClientBuilder.php
```diff
@@ -26,14 +26,14 @@ final class ClientBuilder implements ClientBuilderInterface
...
     /**
-     * @var TransportFactoryInterface|null The transport factory
+     * @var TransportInterface The transport
      */
-    private $transportFactory;
+    private $transport;
 
     /**
-     * @var TransportInterface|null The transport
+     * @var HttpClientInterface The HTTP client
```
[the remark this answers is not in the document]
cleptric:
```php
$httpClient = new class() implements HttpClientInterface {
    public function sendRequest(string $requestData, Options $options): Response
    {
        $client = HttpClient::create([
            'headers' => [
                'Content-Type' => 'application/x-sentry-envelope',
                'User-Agent' => Client::SDK_IDENTIFIER . '/' . Client::SDK_VERSION,
            ],
        ]);

        $response = $client->request('POST', $options->getDsn()->getEnvelopeApiEndpointUrl(), [
            'body' => $requestData,
        ]);

        return new Response($response->getStatusCode(), $response->getHeaders(), '');
    }
};

Sentry\init([
    'dsn' => env('SENTRY_DSN'),
    'http_client' => $httpClient,
]);
```
I like the UX of this approach much more than the boilerplate required today. https://docs.sentry.io/platforms/php/configuration/transports/

We could even go the extra mile and add two sample implementations for Symfony and Guzzle to the SDK.

src/HttpClient/HttpClient.php
Jean85: That's not true. My was just an example of what security implications we would be opening to if we start writing our own HTTP client.

That specific vulnerability allows the attacker to slip through a custom, malicious body in the response, but I could easily imagine a similar vulnerability in the opposite direction; so you would have the SDK as a vector attack toward our end user or toward Sentry.io or a self-hosted server.

And I'm just limiting myself to the headers. The attack surface may be reduced due to the specific usage of the client (for Sentry only) but it's still there, and it comprises headers, body, format/compression in both requests and responses.

src/ClientBuilder.php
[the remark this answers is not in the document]
mfb: If it becomes super easy for SDK users to setup the Symfony or Guzzle HTTP client (without boilerplate) then this could be a net improvement :)  I do share concerns that there will be bugs (as well as missing features) when implementing your own HTTP client - we spent a lot of time working on HTTP client issues in the 1.x branch (although to be fair that branch was trying to do more than the 3.x branch, like sending multiple events in parallel at the end of the request).

src/HttpClient/HttpClient.php
```diff
@@ -0,0 +1,89 @@
...
+            curl_close($curlHandle);
+
+            $message = 'cURL Error (' . $errorCode . ') ' . $error;
+
+            return new Response(0, [], $message);
+        }
+
+        $statusCode = curl_getinfo($curlHandle, \CURLINFO_HTTP_CODE);
+        $headerSize = curl_getinfo($curlHandle, \CURLINFO_HEADER_SIZE);
+        $headers = Http::getResponseHeaders($headerSize, (string) $body);
```
stayallive, line 83: ~~I think we should use `\CURLOPT_HEADERFUNCTION` instead. This gives us a callback for each headerline already normalized to lowercase names: https://curl.se/libcurl/c/CURLOPT_HEADERFUNCTION.html. We would also want to set `CURLOPT_SUPPRESS_CONNECT_HEADERS` for the proxy requests.~~

Realized this is a PHP 7.3+ option, so not possible for now but should make header parseing more reliable.

src/HttpClient/HttpClient.php
```diff
@@ -0,0 +1,89 @@
...
+        curl_setopt($curlHandle, \CURLOPT_HTTP_VERSION, \CURL_HTTP_VERSION_1_1);
+
+        $httpSslVerifyPeer = $options->getHttpSslVerifyPeer();
+        if ($httpSslVerifyPeer) {
+            curl_setopt($curlHandle, \CURLOPT_SSL_VERIFYPEER, true);
+        }
+
+        $httpProxy = $options->getHttpProxy();
+        if (null !== $httpProxy) {
+            curl_setopt($curlHandle, \CURLOPT_PROXY, $httpProxy);
```
stayallive, line 58: We might also want to set `CURLOPT_HEADEROPT` to `CURLHEADER_SEPARATE` to be safe. Before cURL `7.42.1` `CURLOPT_HTTPHEADER` was also sent to the proxy which might cause issues, in later versions `CURLHEADER_SEPARATE` became the standard but since we don't know the cURL version we might want to set this ourselfs.

src/Client.php
```diff
@@ -173,14 +173,18 @@ public function captureEvent(Event $event, ?EventHint $hint = null, ?Scope $scop
         }
 
         try {
-            /** @var Response $response */
-            $response = $this->transport->send($event)->wait();
-            $event = $response->getEvent();
+            /** @var Result $result */
+            $result = $this->transport->send($event);
+            $event = $result->getEvent();
```
stayallive, line 178: We might also want to wrap this in a `trace()` so requests to Sentry are logged as a span in case errors are transmitted inside a performance traced request (or maybe that should be done inside the Transport or somewhere else but it would be nice to have a span for it.

src/ClientBuilder.php
```diff
@@ -68,127 +68,97 @@ final class ClientBuilder implements ClientBuilderInterface
...
     {
         $this->options = $options ?? new Options();
+
+        $this->httpClient = $this->options->getHttpClient() ?? new HttpClient($this->sdkIdentifier, $this->sdkVersion);
+        $this->transport = $this->options->getTransport() ?? new HttpTransport(
+            $this->options,
+            $this->httpClient,
+            new PayloadSerializer($this->options),
+            $this->logger
+        );
```
stayallive, line 78: We should move this down to the `getClient()` which is the actual builder method. And only create a client or use the one from the options if the transport was not set. Also saves needing the `httpClient` instance variable.

src/HttpClient/HttpClient.php
```diff
@@ -0,0 +1,89 @@
...
+        $curlHandle = curl_init();
+        curl_setopt($curlHandle, \CURLOPT_URL, $dsn->getEnvelopeApiEndpointUrl());
+        curl_setopt($curlHandle, \CURLOPT_HTTPHEADER, Http::getRequestHeaders($dsn, $this->sdkIdentifier, $this->sdkVersion));
+        curl_setopt($curlHandle, \CURLOPT_USERAGENT, $this->sdkIdentifier . '/' . $this->sdkVersion);
+        curl_setopt($curlHandle, \CURLOPT_TIMEOUT, $options->getHttpTimeout());
+        curl_setopt($curlHandle, \CURLOPT_CONNECTTIMEOUT, $options->getHttpConnectTimeout());
+        curl_setopt($curlHandle, \CURLOPT_ENCODING, '');
+        curl_setopt($curlHandle, \CURLOPT_POST, true);
+        curl_setopt($curlHandle, \CURLOPT_POSTFIELDS, $requestData);
+        curl_setopt($curlHandle, \CURLOPT_RETURNTRANSFER, true);
```
stayallive, line 47: Since we don't actually use the body anywhere we should disable this to save on memory. Headers should still be returned (untested).

src/Options.php
```diff
@@ -648,6 +650,30 @@ public function getIntegrations()
...
+    {
+        $options = array_merge($this->options, ['http_client' => $httpClient]);
+
+        $this->options = $this->resolver->resolve($options);
+    }
+
+    public function getHttpClient(): ?HttpClientInterface
+    {
+        return $this->options['http_client'];
+    }
```
stayallive, line 675: I'm wondering if we should only allow the `http_client` option. Since right now setting a `transport` overshadows the client. If you need a custom transport I don't think it's bad to use the client builder directly but setting the `http_client` seems like a reasonable thing to have especially since the `transport` has quite a few more dependencies needed to work in almost all cases, WDYT?

src/Options.php
```diff
@@ -925,9 +975,13 @@ private function configureOptions(OptionsResolver $resolver): void
             'in_app_include' => [],
             'send_default_pii' => false,
             'max_value_length' => 1024,
+            'transport' => null,
+            'http_client' => null,
             'http_proxy' => null,
+            'http_proxy_authentication' => null,
```
stayallive, line 981: Didn't we support proxy authentication before? We might otherwise need to make sure you couldn't pass the `http_proxy` as `user:pass@proxyhost.example` or we have a BC, not a big one, but keeping it simpler might be worth splitting `http_proxy` on the `@` in the client instead of adding another option for it.

src/Util/Http.php
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        $authHeader = [];
+        foreach ($data as $headerKey => $headerValue) {
+            $authHeader[] = $headerKey . '=' . $headerValue;
+        }
+
+        return array_merge($headers, [
+            'X-Sentry-Auth' => 'Sentry ' . implode(', ', $authHeader),
+        ]);
```
stayallive, line 41: Micro-optim: maybe move the `$headers` down and remove the `array_merge`.

src/Util/Http.php
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * @return string[][]
+     */
+    public static function getResponseHeaders(?int $headerSize, string $body): array
```
stayallive, line 47: This might be the most fragile part from what I'm reading online. Not saying I have a better option currently but this requires some more scrutiny than I can give at this moment.

stayallive (commented):
Overal this is starting to look great, mostly because it's a heck of a lot simpler.

src/HttpClient/HttpClient.php
Jean85: Why? Do we still want to support PHP 7.2 and below even in the next major?

stayallive: I made a mistake `CURLOPT_HEADERFUNCTION` is available way longer, not starting in PHP 7.3, I've changed my comment to reflect that 🤦

src/Client.php
```diff
@@ -216,7 +220,7 @@ public function getIntegration(string $className): ?IntegrationInterface
     /**
      * {@inheritdoc}
      */
-    public function flush(?int $timeout = null): PromiseInterface
+    public function flush(?int $timeout = null): Result
     {
         return $this->transport->close($timeout);
     }
```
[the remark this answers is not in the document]
cleptric: This could become handy if we ever invest more into async runtimes. I would keep it.

src/HttpClient/HttpClient.php
cleptric: > **CURLOPT_RETURNTRANSFER** `true` to return the transfer as a string of the return value of `curl_exec()` instead of outputting it directly.

If we remove this, we will just echo the response string. Better to keep it like that 😬

src/Options.php
cleptric: I'm aware of some people who added custom retry logic and such. I would like to keep this as is, to ease swapping out the transport. The current logic in `ClientBuilder::__construct` should work well, or am I overlooking something?

src/Options.php
cleptric: Imo, we did not. Referring to https://curl.se/libcurl/c/CURLOPT_PROXY.html

src/HttpClient/HttpClient.php
cleptric: I like this a lot, thanks Alex! Have a look at https://github.com/getsentry/sentry-php/pull/1589/commits/c9b3a2da444767b2793c5994a228e92c1675d4c1

src/Options.php
stayallive: Huh, interesting, okay!

src/HttpClient/HttpClient.php
stayallive: Yikes! Good call 😉

src/Client.php
stayallive: Okay that would make sense, although it also doesn't 😅 Interestring to know why since it could account for a large-ish gap in the spans if the app sends an error and the connection is terrible for some reason. Would be great to know (and nice insights maybe for Sentry data-miners).

stayallive (approved):
👍🚀

cleptric:
Merging now to unblock other features. The remaining ToDos are documented in https://github.com/getsentry/sentry-php/issues/1595

src/ClientBuilder.php
cleptric: Moved to https://github.com/getsentry/sentry-php/issues/1595

src/Options.php
cleptric: Moved to https://github.com/getsentry/sentry-php/issues/1595

src/Client.php
cleptric: TBD, will get back later.

PR #1339, opened 2022-08-11
ci: Run test suite against PHP 8.2

cleptric:
Closes #1337

PR #1591, opened 2023-10-07
Run CI tests for PHP 8.3

jnoordsij:
With a few RC's released and the final release coming up in november, it might be a good thing to start running tests on the new PHP.

I did not encounter any issues or deprecations while testing, so everything seems to be all ready 😄

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,10 +65,6 @@ jobs:
         run: composer update --no-progress --no-interaction --prefer-dist
         if: ${{ matrix.dependencies == 'highest' }}
 
-      - name: Restrict lowest Symfony version on PHP 8.1 & 8.2
-        run: composer require symfony/options-resolver:^4.4.30 --no-update
-        if: ${{ matrix.dependencies == 'lowest' && matrix.php == '8.1' || matrix.php == '8.2' }}
-
       - name: Install lowest dependencies
         run: composer update --no-progress --no-interaction --prefer-dist --prefer-lowest
         if: ${{ matrix.dependencies == 'lowest' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,7 @@ jobs:
           - '8.0'
           - '8.1'
           - '8.2'
+          - '8.3'
         dependencies:
           - lowest
           - highest


jnoordsij:
Note: when checking the related workflow code and previous PR's (#1245 and #1339), I noticed that adding the new version to the `Restrict lowest Symfony version on PHP 8.1 & 8.2` matcher was not needed. I found out (`composer update "symfony/options-resolver:3.4.43" --dry-run`) this is due to the addition of `phpbench` as dev dependency (see #1274), which requires `symfony/options-resolver ^4.2 || ^5.0`, so that the v3 version is invalid anyways. 

While for recent PHP versions this is perfectly fine, this does mean the lowest versions dependency tests don't test this particular `symfony/options-resolver` version anymore at all (see e.g. https://github.com/getsentry/sentry-php/actions/runs/6198920906/job/16830371050#step:9:85). I'm not sure if this is a real issue, if so, a follow-up issue should be opened.

Jean85:
IMHO we should drop support for Symfony 3.4: it's EOL for a long time, so long that next month 4.4 will be EOL too!

ste93cry:
>we should drop support for Symfony 3.4

I agree, but as usual this is a decision that will require a discussion with the guys at Sentry. Also, with version `4.0` of this SDK in development, I find it unlikely that we will stop supporting Symfony `3.4` now, but I think it's feasible in the next major

> I noticed that adding the new version to the `Restrict lowest Symfony version on PHP 8.1 & 8.2` matcher was not needed. I found out (`composer update "symfony/options-resolver:3.4.43" --dry-run`) this is due to the addition of `phpbench` as dev dependency (see #1274), which requires `symfony/options-resolver ^4.2 || ^5.0`, so that the v3 version is invalid anyways. 

This was surely unwanted, but there is no version of `phpbench` that is compatible with Symfony `3.4` unless we go way back to its version `0.16`. The question is if we're fine with the missing test of this version of the framework or not. If the answer is the latter, we should relax the constraint in the `composer.json` accordingly

cleptric (approved):

cleptric:
We can remove `"symfony/options-resolver": "^3.4.43"` in version 4.0 of the SDK.
I'm more concerned about PHP version support and `^4.4.30` supports PHP >= 7.1.

PR #1599, opened 2023-10-24
Update PHPCS

cleptric:


.php-cs-fixer.dist.php
```diff
@@ -10,9 +10,16 @@
             'imports_order' => ['class', 'function', 'const'],
         ],
         'declare_strict_types' => true,
+        'get_class_to_class_keyword' => false,
         'yoda_style' => true,
         'self_accessor' => false,
-        'phpdoc_no_useless_inheritdoc' => false,
+        'modernize_strpos' => false,
```
cleptric, line 16: `str_contains` is only available from PHP 8.0

.php-cs-fixer.dist.php
```diff
@@ -10,9 +10,16 @@
             'imports_order' => ['class', 'function', 'const'],
         ],
         'declare_strict_types' => true,
+        'get_class_to_class_keyword' => false,
```
cleptric, line 13: `PHP Fatal error:  Cannot use ::class with dynamic class name` in PHP 7.4 and below

.php-cs-fixer.dist.php
```diff
@@ -10,9 +10,16 @@
...
         'yoda_style' => true,
         'self_accessor' => false,
-        'phpdoc_no_useless_inheritdoc' => false,
+        'modernize_strpos' => false,
+        'nullable_type_declaration_for_default_null_value' => [
+            'use_nullable_type_declaration' => true,
+        ],
+        'no_superfluous_phpdoc_tags' => [
+            'allow_mixed' => true,
```
cleptric, line 21: This would remove all `@mixed` annotations, which makes phpstan unhappy.

stayallive (approved):
🥳

PR #1603, opened 2023-10-24
Simplify breadcrumb API

stayallive:
Currently the breadcrumb creation looks something like this:

```php
\Sentry\addBreadcrumb(
    new \Sentry\Breadcrumb(
        \Sentry\Breadcrumb::LEVEL_INFO,
        \Sentry\Breadcrumb::TYPE_DEFAULT,
        'auth',                // category
        'User authenticated',  // message (optional)
        ['user_id' => $userId] // data (optional)
    )
);
```

I would like to be able to write this shorter, something like:

```php
\Sentry\addBreadcrumb('auth', 'User authenticated', ['user_id' => $userId]);

// or with PHP 8:
\Sentry\addBreadcrumb(
    category: 'auth', 
    message: 'User authenticated',
    metadata: ['user_id' => $userId],
);
```

This removed the need for creating new objects and setting the level and type so you only have to pass in a category (could maybe also be defaulted) and a message and optional metadata.

It was also suggested to have a global `sentry()` function (not namespaced) so it could be just `sentry()->addBreadcrumb(...)` but not sure if we want to go that far.

This is more of a discussion PR, curious to hear your thoughts! My main goal is to be able to add context and leave breadcrumbs as short and easy as possible to have users do it more and more easily.


cleptric:
No objections. I do like that we don't break for existing users, and the new way of adding breadcrumbs seems much easier, which is great.

cleptric (approved):

PR #1605, opened 2023-10-25
Fix request fetcher

cleptric:
While we wanted to use https://github.com/Nyholm/psr7, their server package https://github.com/Nyholm/psr7-server seems not to be maintained anymore, hence opting for https://github.com/guzzle/psr7.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,7 @@
         "ext-json": "*",
         "ext-mbstring": "*",
         "ext-curl": "*",
+        "guzzlehttp/psr7": "^1.8.4|^2.1.1",
         "jean85/pretty-package-versions": "^1.5|^2.0.4",
         "psr/log": "^1.0|^2.0|^3.0",
         "symfony/options-resolver": "^4.4.30|^5.0.11|^6.0|^7.0"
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -24,14 +24,14 @@
         "ext-json": "*",
         "ext-mbstring": "*",
         "ext-curl": "*",
+        "guzzlehttp/psr7": "^1.0|^2.0",
```
ste93cry, line 27: Please require version `^1.8.4|^2.1.1` as before: it was that way because previous versions had security issues, and in this way you are letting users install them which is bad

HazAT (approved):

ste93cry (changes requested):

composer.json
HazAT: fair point, we had this discussion internally how we should go about external dependencies and if we should dictate which version users should have. We decided to make it on to the user to pump deps accordingly.

However in this specific case, since we already had it, we might as well leave it like this.
If this becomes ever a problem here we can reconsider.

ste93cry (approved):

composer.json
ste93cry: >fair point, we had this discussion internally how we should go about external dependencies and if we should dictate which version users should have. We decided to make it on to the user to pump deps accordingly.

I understand, but you should definitely dictate what version users should have in case the dependency is used directly in your code and isn't just transient. In this case, the code in this SDK actually uses a package affected by [`CVE-2022-24775`](https://cve.mitre.org/cgi-bin/cvename.cgi?name=2022-24775), and in the end it is your package that is vulnerable, not the project of the user.

PR #1620, opened 2023-10-31
Fix setting cURL request headers

cleptric:
It turns out cURL does not expect an associative array when setting custom request headers.


stayallive (approved):
Ahh, good gotcha 👌

PR #1626, opened 2023-11-06
E2E HTTP client tests

stayallive:
This is an attempt on creating E2E HTTP client tests to make sure we send the correct data/headers and also parse a response correctly. We don't need to test with HTTPS and be super fancy since we rely on cURL, we just have to validate we hold cURL correctly basically.


cleptric (approved):
🪨

PR #1629, opened 2023-11-06
FOR-10907: Backport of explicit protocol set

j03k64:
Backport of explicit protocol setting: https://github.com/ShootProof/sentry-php/commit/6ce65abab960ddbd7b22355afe3088261fab4775#diff-e2b9681244f52c28911e89834f05a2953654ccb96e1d353b42cae64a753bab63

### DIFF
--- a/src/HttpClient/HttpClientFactory.php
+++ b/src/HttpClient/HttpClientFactory.php
@@ -140,6 +140,8 @@ final class HttpClientFactory implements HttpClientFactoryInterface
             $curlConfig = [
                 \CURLOPT_TIMEOUT => $options->getHttpTimeout(),
                 \CURLOPT_CONNECTTIMEOUT => $options->getHttpConnectTimeout(),
+                // Backport of explicit protocol setting: https://github.com/ShootProof/sentry-php/commit/6ce65abab960ddbd7b22355afe3088261fab4775#diff-e2b9681244f52c28911e89834f05a2953654ccb96e1d353b42cae64a753bab63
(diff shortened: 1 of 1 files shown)


cleptric:
Did you accidentally open the PR against the wrong fork? 😄

j03k64:
Yes! Sorry! Doing a quick bandaid until I can get through dependency hell.

Jean85:
@cleptric isn't this a backport of https://github.com/getsentry/sentry-php/pull/1542?
Are we sure that we don't need it anyway?

j03k64:
> @cleptric isn't this a backport of #1542? Are we sure that we don't need it anyway?

I'd say it should be a full backport (guzzle and curl) and not just the `curl` portion that we use. I got around the dependency issue and ended up upgrading to 3.22 instead of this route anyway.

Jean85:
Sorry, my worded my reply poorly. I don't think we're interested in backporting the fix onto an old unsupported version. We may be interested in a forward port in 4.x though.

stayallive:
No need, `4.x` already has it: https://github.com/getsentry/sentry-php/blob/198828374e21e25b3e1c66ad956fee61b19cccd5/src/HttpClient/HttpClient.php#L74.

cleptric:
Just for completeness sake.

We could, in theory, bump to HTTP2 anytime, the bug in cURL does not happen in our HTTP client.

The issue was that the Decoder Plugin sets a `TE` header with values being invalid in a HTTP2 context. https://github.com/php-http/client-common/blob/676f98fb3785d2c9e1b16aa24d9fcc39066f3f06/src/Plugin/DecoderPlugin.php#L54-L67

See https://github.com/php-http/client-common/issues/230

PR #1630, opened 2023-11-06
Drupal integration supports the 4.x SDK

mfb:
Updated the drupal integration today to support the 4.x SDK

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,11 +55,14 @@ The following integrations are fully supported and maintained by the Sentry team
 - [Symfony](https://github.com/getsentry/sentry-symfony)
 - [Laravel](https://github.com/getsentry/sentry-laravel)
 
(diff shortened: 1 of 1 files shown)


ste93cry (approved):

cleptric:
Nice, that was fast! 👏 
Did the changelog/upgrade guide help, or were there things not mentioned?

mfb:
> Nice, that was fast! 👏 Did the changelog/upgrade guide help, or were there things not mentioned?

@cleptric The one thing I would suggest documenting, in addition to "Added ext-curl as a composer requirement," would be that if curl is missing at runtime, Sentry SDK does not throw an error or exception.  It does log an error if you attached a logger, and otherwise fails gracefully.  So I used drupal requirements API to surface the error if curl is missing.

PR #1631, opened 2023-11-07
WordPress integration supports the 4.x SDK

stayallive:
Updated the WordPress integration today to support the 4.x SDK.

Very smoothly, even able to cleanup some code with the `tags` option being back 💪


cleptric (approved):
Nice!

PR #1632, opened 2023-11-07
Search for curl_init() only in root namespace

mfb:
If curl is missing at runtime, the curl_init() function call searches the Sentry\HttpClient namespace, which results in a confusing error message: "Call to undefined function Sentry\HttpClient\curl_init()". We only want to search the root namespace.

### DIFF
--- a/src/HttpClient/HttpClient.php
+++ b/src/HttpClient/HttpClient.php
@@ -40,6 +40,10 @@ class HttpClient implements HttpClientInterface
             throw new \RuntimeException('The request data is empty.');
         }
 
+        if (!\extension_loaded('curl')) {
+            throw new \RuntimeException('The cURL PHP extension must be enabled to use the HttpClient.');
+        }
+
         $curlHandle = curl_init();
 
         $requestHeaders = Http::getRequestHeaders($dsn, $this->sdkIdentifier, $this->sdkVersion);


mfb:
This triggered a code style rule that functions shouldn't have the leading backslash, but it's quite useful at least here IMHO.

stayallive:
Hmm. In theory we require `ext-curl` so this should never happen. 

I do understand the error is unfortunate here. Might be better to throw a `RuntimeException('The cURL extension is missing.')` or something along those lines which might even be more descriptive yet?

mfb:
> Hmm. In theory we require `ext-curl` so this should never happen.

Those platform requirements are for composer only.  Meanwhile, the app, let's say it's a WordPress site, could have a long lifespan after composer runs, with PHP extensions being enabled and disabled in a control panel, code being moved around, etc. 

I updated the branch to allow the leading backslash on curl_init(), so at least the existing error message is improved.

cleptric:
We catch all exceptions thrown in the client here https://github.com/getsentry/sentry-php/blob/master/src/Transport/HttpTransport.php#L86-L95

I'm also not in favour of checking the requirement during runtime, this can cause a bad day for people deploying stuff when we suddenly throw exceptions because cURL is not installed.

Before we add hacks to the code style, let's add a new logger entry, please.

cleptric:
```php
if (!function_exists('curl_init')) {
    throw new \RuntimeException('The cURL extension must be enabled to use the HttpClient.');
}
```

Put this below the DSN check and we're good.

cleptric (approved):
Thanks!

mfb:
I was assuming that extension_loaded is faster than function_exists()  - it has less to search, but idk :)

cleptric:
Yeah, makes sense, good change!

stayallive:
A few years ago it made [negligible difference](https://www.drupal.org/project/drupal/issues/1942432#comment-10006777) in speed so `extension_loaded` seems more descriptive here 👍 

Thanks!

PR #1633, opened 2023-11-08
Fix capturing OOM errors when very memory constrained

stayallive:
We currently reserve 10 KiB of memory which we free when handling a OOM error, however this is in many cases way to small. This is especially try when extremely memory constrained and the OOM is triggered by allocating just a tiny bit of extra memory from PHP.

For example when testing with:

```php
$foo = str_repeat('x', 1024 * 1024 * 1024);
```

This doesn't need the reserved memory since a huge block of memory is being allocated and fails which results in the full memory never being allocated.

In contrary when testing with:

```php
$array = [];
for ($i = 0; $i < 100000000; ++$i) {
    $array[] = str_repeat('a', 100);
}
```

Here the memory is being allocated in tiny chunks, so when PHP runs out of memory it becomes prominent that the 10 KiB of reserved memory is way to little. From my testing we'd need 30-60x that amount to handle sending a relatively simple payload.

So the solution is to increase the memory limit (once) and use that extra headroom to send the OOM event, I've chosen for 2 MiB, from my testing I think 2 MiB is usually just enough but to be a little on the safe side I've opted for 5 MiB. This value is configurable if needed and this behavior can also be completely disabled if required.

To configure this behavior you can do the following before running `\Sentry\init(...)`:

```php
// The amount of reserved memory every request (in bytes), this has to be a positive integer amount of bytes. Defaults to 16 KiB
$errorHandler = \Sentry\ErrorHandler::registerOnceFatalErrorHandler(16 * 1024);
// The amount of memory in bytes to increase the `memory_limit` to when handling a out of memory error, can also be set to `null` to disable increasing the `memory_limit`. Defaults to 5 MiB
$errorHandler->setMemoryLimitIncreaseOnOutOfMemoryErrorInBytes(5 * 1024 * 1024);
```

~~I think we would also be able to shrink the reserved memory we are currently holding but since it's already very little I think it's safe to leave the value like it is.~~ As it turns out we actually need a little more for the regex matching to work correctly on some configurations so I've upped it very slightly by default.

Once accepted I will also port this to the new 4.x branch which suffers from the same problems.

I've also had to ignore some phpstan errors caused by a dependency update.


src/ErrorHandler.php
```diff
@@ -22,7 +22,12 @@ final class ErrorHandler
      *
      * @internal
      */
-    public const DEFAULT_RESERVED_MEMORY_SIZE = 10240;
+    public const DEFAULT_RESERVED_MEMORY_SIZE = 16 * 1024; // 16 KiB
+
+    /**
+     * The regular expression used to match the message of an out of memory error.
+     */
+    private const OOM_MESSAGE_MATCHER = '/^Allowed memory size of (?<memory_limit>\d+) bytes exhausted \(tried to allocate \d+ bytes\)/';
```
Jean85, line 30: Are we sure that the message matches on different PHP versions? We should test for that.

Jean85 (commented):
This is a very risky approach...

In containerized environment, we may risk to make the kernel unhappy, which in turn can decide to kill the whole process (and our event with it). But the amount (and time) we're risking is low, and the opposite risk is to lose the message due to the limit, so I think it may be worth it.

stayallive:
Yeah. It's definitely not ideal. However I felt like the relatively small memory increase for a process that is being killed anyway (because of the fatal error) seemed like the best of the worst options. I'd be open to other but the only other one would be to take the 5 MiB beforehand but that seems even worse and unnecessary. 

I also made it possible to configure / disable (just like you can with the current memory reserve) to allow users that cannot risk the increase to disable it if they wish.

src/ErrorHandler.php
stayallive: Well, we are indirectly testing it, because disabling the memory increase (which is the same as a incorrect matcher) fails the added `phpt` test which I think is good enough.

.github/workflows/ci.yml
```diff
@@ -69,8 +69,10 @@ jobs:
         run: composer update --no-progress --no-interaction --prefer-dist --prefer-lowest
         if: ${{ matrix.dependencies == 'lowest' }}
 
-      - name: Run tests
-        run: vendor/bin/phpunit --coverage-clover=coverage.xml
+      - name: Run unit tests
+        run: vendor/bin/phpunit --testsuite unit --coverage-clover=coverage.xml
```
ste93cry, line 73: I don't understand the advantages of splitting the test suite in two, except perhaps being able to run them independently (but that is rare) 🤔

.github/workflows/ci.yml
```diff
@@ -69,8 +69,10 @@ jobs:
...
+      - name: Run phpt tests
+        run: vendor/bin/phpunit --testsuite phpt --no-coverage
```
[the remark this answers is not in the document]
stayallive: The reason is that the OOM tests OOM because of the coverage information being collected. I'm working on moving the OOM tests to their own folder to only run those without coverage.

src/ErrorHandler.php
```diff
@@ -315,16 +341,21 @@ private function handleError(int $level, string $message, string $file, int $lin
...
-        }
-
+        // Free the reserved memory that allows us to potentially handle OOM errors
         self::$reservedMemory = null;
+
         $error = error_get_last();
 
         if (!empty($error) && $error['type'] & (\E_ERROR | \E_PARSE | \E_CORE_ERROR | \E_CORE_WARNING | \E_COMPILE_ERROR | \E_COMPILE_WARNING)) {
+            // If we did not do so already and we are allowed to increase the memory limit, we do so when we detect an OOM error
+            if (false === self::$didIncreaseMemoryLimit && null !== $this->memoryLimitIncreaseValue && 1 === preg_match(self::OOM_MESSAGE_MATCHER, $error['message'], $matches)) {
```
jarstelfox, line 351: It's totally reasonable to limit it to just OOMs. However, I wonder if it is always acceptable to bump the memory, and then drop it back down after the handler runs.

Since this increase can be disabled/configured, it seems sane enough to me.

The main reason for suggesting this is to reduce the maintenance burden. You would not need regex matching.

I also understand if you'd rather not bump memory for all fatal errors.

stayallive: Yeah I thought about that, but only doing it when absolutely needed seemed like the best approach. If the regex turns out to be a mistake at a later point because the error message is changes we might change our stance for this. I think the fatal error always exits the process so it's probably safe to do it always, but I'm also not sure that is actually true :)

.github/workflows/ci.yml
stayallive: This is because the OOM tests do not run correctly because the code coverage code causes them to OOM again. So I disabled the coverage for the OOM tests specifically now and added all other `phpt` tests back to the normal test suite with coverage.

When running `phpunit` locally all tests in all suites are executed by default so there is no difference for local development.

src/ErrorHandler.php
jarstelfox: It does look like [this string is relatively safe to assume](https://github.com/search?q=repo%3Aphp%2Fphp-src%20%22Allowed%20memory%20size%20of%22&type=code).

[Here is a slightly better regex](https://github.com/php/php-src/blob/524b13460752fba908f88e3c4428b91fa66c083a/Zend/tests/new_oom.phpt#L15) from PHP's source tests.

You may want to use this regex instead.

stayallive: I've updated the regex, only the optional newline was different (`\d+` is functionally identical to `[0-9]+` in this case) and I don't need to match the full string, the first part is enough confirmation it's a OOM error IMHO.

jarstelfox:
> This is a very risky approach...
> 
> In containerized environment, we may risk to make the kernel unhappy, which in turn can decide to kill the whole process (and our event with it).

The only other option I see is to reserve more memory (the str_repeat) such that we always have enough space for Sentry to complete its task. i.e. keep 5MB for all processes.

This also has trade-offs. Mainly, 99.9% of processes will never need this extra overhead. Only the few processes that OOM would.

src/ErrorHandler.php
```diff
@@ -315,16 +343,21 @@ private function handleError(int $level, string $message, string $file, int $lin
      */
     private function handleFatalError(): void
     {
-        // If there is not enough memory that can be used to handle the error
-        // do nothing
-        if (null === self::$reservedMemory) {
-            return;
-        }
```
ste93cry, line 322: There is a [comment](https://github.com/getsentry/sentry-php/blob/198828374e21e25b3e1c66ad956fee61b19cccd5/src/ErrorHandler.php#L372) in the code that explains the reason `self::$reservedMemory` is set to null at some point, so removing this check is gonna cause trouble, likely.

stayallive: The check is removed. However it is still set to `null`. This just prevents handling multiple fatal errors (if that's even possible) which doesn't seem logical to me so it was simplified to always setting it to `null` without the check.

ste93cry: >This just prevents handling multiple fatal errors

No, this check prevents an exception from being logged twice. See below for a code that reproduces the problem:

```php
$errorHandler = ErrorHandler::registerOnceFatalErrorHandler();
$errorHandler->addFatalErrorHandlerListener(static function (): void {
    echo 'Fatal error listener called' . PHP_EOL;
});

$errorHandler = ErrorHandler::registerOnceExceptionHandler();
$errorHandler->addExceptionHandlerListener(static function (): void {
    echo 'Exception listener called' . PHP_EOL;
});

throw new \Exception('Hello World!');

// Exception listener called
// 
// Fatal error: Uncaught Exception: hello world in [...]
// Stack trace:
// #0 {main}
//   thrown in [...] on line [...]
// Fatal error listener called
// PHP Fatal error:  Uncaught Exception: hello world in [...]
// Stack trace:
// #0 {main}
//   thrown in [...] on line [...]
```

stayallive: We should probably have a test for that 😬

ste93cry: I agree, it was forgotten but you're welcome to contribute it 😀

stayallive: I got tripped up by the comment above the `if` statement, because that says another thing. So I've added a more descriptive boolean flag instead on relying on the presence of the reserved memory which reads better to me and should have the same functionality.

tests/phpt/error_handler_handles_fatal_error_only_once.phpt
```diff
@@ -0,0 +1,36 @@
+--TEST--
+Test that fatal errors are only handled once
```
ste93cry, line 2: It's the exception that can be handled twice, not the fatal error. Please also change the file name to reflect it :pray:

src/ErrorHandler.php
ste93cry: The boolean flag is clearly more descriptive, I'm just worried that we're not forgetting anything else, e.g. the fact that before the memory was released and now it is not. I don't know if this can constitute a problem, but to be on the safe side it would be better to keep doing that. Also, since the property contains a `bool`, we can avoid checking it using `===`

stayallive: If we forgetting anything else I'd rather find out honestly so we can add a test for it and we know why it's needed. 

I can't think of a case where it would help at this point. Also the very tiny memory reserve is not very useful for anything and only applicable for OOM situations that are not handled by the `handleException` anyway, hence this PR in the first place. I have been through the blame-view but this has been added in a massive PR so there is no context for this check.

src/ErrorHandler.php
```diff
@@ -315,16 +348,25 @@ private function handleError(int $level, string $message, string $file, int $lin
...
         }
 
+        // Free the reserved memory that allows us to potentially handle OOM errors
         self::$reservedMemory = null;
+
         $error = error_get_last();
 
         if (!empty($error) && $error['type'] & (\E_ERROR | \E_PARSE | \E_CORE_ERROR | \E_CORE_WARNING | \E_COMPILE_ERROR | \E_COMPILE_WARNING)) {
+            // If we did not do so already and we are allowed to increase the memory limit, we do so when we detect an OOM error
+            if (false === self::$didIncreaseMemoryLimit && null !== $this->memoryLimitIncreaseValue && 1 === preg_match(self::OOM_MESSAGE_MATCHER, $error['message'], $matches)) {
```
cleptric, line 362:
```suggestion
            if (
                false === self::$didIncreaseMemoryLimit &&
                null !== $this->memoryLimitIncreaseValue &&
                1 === preg_match(self::OOM_MESSAGE_MATCHER, $error['message'], $matches)
            ) {
```

cleptric (approved):
We should update the docs prior merging, https://docs.sentry.io/platforms/php/integrations/#fatalerrorlistenerintegration.

Thanks for investigating this!

PR #1648, opened 2023-11-23
Forbid serializing the HubAdapter in the first place

Nemo64:
The symfony cache takes the hint when something isn't serializable. But it will fail if something successfully serializes and then fails to deserialize.

This small change makes the HubAdapter still work, when it accidentally get's caught in the Metadata Cache of Doctrine. This happens, for example, when you use a logger in doctrine events.


Jean85 (approved):
LGTM, but a test would be lovely!

cleptric (changes requested):
CI is failing

cleptric:
Added a test in https://github.com/getsentry/sentry-php/pull/1663

PR #1650, opened 2023-11-25
Handle variadic and not declared arguments in stack traces

### LINKED ISSUE #1645: Stack traces are missing variadic and not declared arguments
### How do you use Sentry?

Self-hosted / on-premises

### SDK version

4.0.1

### Steps to reproduce

1. Create function with variadic parameters or without parameters. Ex.:
```php
function variadic(...$params) {...}
function noParams() {...}
```
2. Call this function with some arguments.
3. Throw an exception inside that function.

### Expected result

Sentry shows all passed arguments in stack trace.

### Actual result

Sentry shows only first argument for variadic function and no arguments for function without parameters.
With this code:
```php
function variadic(...$params)
{
    noParams($params);
}

function noParams()
{
    throw new \Exception();
}

Artisan::command('debug', function () {
    variadic('a', 1, [123]);
});
```

I get this stack trace in Sentry:
CleanShot 2023-11-20 at 18 51 49@2x

xwillq:
This pull request fixes #1645.
Short summary: SDK wasn't sending some function arguments to Sentry, namely default, variadic, not declared arguments and arguments with value `null`. This pull request reworks function responsible for handling parameters in order to send all of the above.

I tested my changes with all inputs variations I could think of, it seems to be working. However, I don't quite understand how to add tests for it. Testing this method requires using stack trace with real function or real class and method. Where do I create them? Current `FrameBuilder` tests don't do anything similar, so I'm not sure what would be the right way to do it.

Also, PHPStan gives me `Unreachable statement - code above always terminates.` error on lines 254 and 267. Those statements are definitely reachable, and I confirmed it with manual testing. I tried changing method parameter from `mixed[]` to `array<array-key, mixed>` but it didn't help. Do you have an idea what could be causing this?

### DIFF
--- a/src/FrameBuilder.php
+++ b/src/FrameBuilder.php
@@ -36,6 +36,12 @@ final class FrameBuilder
     private $representationSerializer;
 
     /**
+     * @var string The default name for function parameters when we cannot get
+     *             them from function definition
+     */
+    private const DEFAULT_PARAMETER_NAME = 'param';
+
+    /**
      * Constructor.
      *
      * @param Options                           $options                  The SDK client options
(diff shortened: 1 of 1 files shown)


xwillq:
Creating this as draft before all mentioned issues are addressed.

Jean85 (changes requested):
Thanks for the contribution!

To create tests for this, you can create some stub (anonimous?) classes or methods that have the kind of parameters that you need to test, and call the SOT on that.

PR #1654, opened 2023-12-04
Update UPGRADE-4.0.md

### LINKED ISSUE #1651: Logger tag is missing
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

v4

### Steps to reproduce

After upgrading to v4 SDK the tag `logger` (along with its default value `php`) is missing.
This would not be anything note worthy if it was mentioned in the upgrading guide.

### Expected result

The removal of the tag-value pair `logger:php` mentioned in the migration guide.

### Actual result

The removal is not mentioned and the actual tag is no longer added by default, but needs to be explicitly added in the config options.

cleptric:
Closes #1651.

### DIFF
--- a/UPGRADE-4.0.md
+++ b/UPGRADE-4.0.md
@@ -59,6 +59,7 @@
 - Removed `Dsn::getSecretKey()`.
 - Removed `Dsn::setSecretKey()`.
 - Removed `EventType::default()`.
+- Removed adding the value of the `logger` option as a tag on the event. If you rely on this behaviour, add the tag manually.
 
 - Added return type to `Dsn::getProjectId(): string`.
 - Changed return type to `Options::getLogger(): ?LoggerInterface`.


Jean85 (approved):

UPGRADE-4.0.md
```diff
@@ -59,6 +59,7 @@
 - Removed `Dsn::getSecretKey()`.
 - Removed `Dsn::setSecretKey()`.
 - Removed `EventType::default()`.
+- Removed adding the value of the `logger` option as a tag on the event. If you rely on this behaviour, add the tag manually.
```
mfb, line 62: As far as I can tell, we can still add logger on the event via the Event::setLogger() method, I think it's just the adding it via the client Client::setLogger() method that was changed to instead add a debug logger?

So I would suggest removing this sentence and fleshing out the above sentence "The logger option now accepts a Psr\Log\LoggerInterface instance instead of string" to explain that the logger tag can now be added only via the Event::setLogger() method.

mfb: Seems I was slightly wrong about this, the Event::setLogger() method still exists, but it doesn't do anything?  So I filed PR #1657

This is important to fix because e.g. the Monolog handler has `$event->setLogger(sprintf('monolog.%s', $record['channel']));`

PR #1657, opened 2023-12-08
Restore event logger to payload

mfb:
The Event::setLogger() and Event::getLogger() methods still exist, but it seems they no longer add the logger to the event payload. As a result, when an event is sent to Sentry, the JSON shows `"logger":""` - empty string, rather than the value that was passed to Event::setLogger().

As far as how Event::setLogger() is used, it's not just me.  The Monolog handler has `$event->setLogger(sprintf('monolog.%s', $record['channel']));`


cleptric (commented):
Should add the logger to the tests, https://github.com/getsentry/sentry-php/blob/master/tests/Serializer/PayloadSerializerTest.php#L73

mfb:
It looks like there was a test for this, but it was removed in 6ba2a2c228c689c7d87b4490f006c0ab05151f77 

Feel free to work on this if anyone has time/interest, otherwise it's on my list.. :)

mfb:
@cleptric ok, I restored the test to PayloadSerializerTest

cleptric (approved):
Thanks!

cleptric:
Shipped in [4.2.0](https://github.com/getsentry/sentry-php/releases/tag/4.2.0)

PR #1663, opened 2023-12-19
Disallow to seralize the `HubAdapter`

cleptric:
Replaces https://github.com/getsentry/sentry-php/pull/1648

### DIFF
--- a/tests/State/HubAdapterTest.php
+++ b/tests/State/HubAdapterTest.php
@@ -42,12 +42,20 @@ final class HubAdapterTest extends TestCase
         clone HubAdapter::getInstance();
     }
 
-    public function testGetInstanceReturnsUnserializableInstance(): void
+    public function testHubAdapterThrowsExceptionOnSerialization(): void
+    {
+        $this->expectException(\BadMethodCallException::class);
+        $this->expectExceptionMessage('Serializing instances of this class is forbidden.');
+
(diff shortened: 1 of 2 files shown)


tests/State/HubAdapterTest.php
```diff
@@ -47,6 +47,9 @@ public function testGetInstanceReturnsUnserializableInstance(): void
         $this->expectException(\BadMethodCallException::class);
         $this->expectExceptionMessage('Unserializing instances of this class is forbidden.');
 
+        $this->expectException(\BadMethodCallException::class);
```
ste93cry, line 50: You should split the test into two, because only one exception can be thrown at time. Furthermore, `expectException()` is in no case cumulative. I suggest to rename this test method to `testHubAdapterThrowsExceptionOnSerialization`, and add `testHubAdapterThrowsExceptionOnSerialization`

ste93cry (changes requested):

tests/State/HubAdapterTest.php
cleptric: How should we test `unserialize` when we need to call `serialize` first?

ste93cry: We could hardcode the argument passed to `unserialize()` to be `O:10:"HubAdapter":0:{}`

ste93cry (approved):

PR #1679, opened 2024-01-16
Add cron monitor wrapper

cleptric:


src/functions.php
```diff
@@ -81,6 +81,35 @@ function captureCheckIn(string $slug, CheckInStatus $status, $duration = null, ?
...
+
+    try {
+        $start = microtime(true);
+        $result = $callback();
+        $duration = microtime(true) - $start;
+
+        return $result;
+    } catch (\Throwable $e) {
+        $status = CheckInStatus::error();
+    } finally {
```
stayallive, line 108: I believe that with a catch you do need to rethrow the error otherwise it's swallowed.

```suggestion
    } catch (\Throwable $e) {
        $status = CheckInStatus::error();
        
        throw $e;
    } finally {
```

stayallive (approved):

PR #1680, opened 2024-01-17
Add support for W3C traceparent header

cleptric:
https://github.com/getsentry/team-sdks/issues/41

### DIFF
--- a/src/Tracing/PropagationContext.php
+++ b/src/Tracing/PropagationContext.php
@@ -9,7 +9,9 @@ use Sentry\State\Scope;
 
 final class PropagationContext
 {
-    private const TRACEPARENT_HEADER_REGEX = '/^[ \\t]*(?<trace_id>[0-9a-f]{32})?-?(?<span_id>[0-9a-f]{16})?-?(?<sampled>[01])?[ \\t]*$/i';
+    private const SENTRY_TRACEPARENT_HEADER_REGEX = '/^[ \\t]*(?<trace_id>[0-9a-f]{32})?-?(?<span_id>[0-9a-f]{16})?-?(?<sampled>[01])?[ \\t]*$/i';
+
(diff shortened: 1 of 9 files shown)


src/Tracing/TransactionContext.php
```diff
@@ -156,6 +158,21 @@ private static function parseTraceAndBaggage(string $sentryTrace, string $baggag
...
+                $hasSentryTrace = true;
+            }
+
+            if (!empty($matches['span_id'])) {
+                $context->parentSpanId = new SpanId($matches['span_id']);
+                $hasSentryTrace = true;
+            }
+
+            if (isset($matches['sampled'])) {
+                $context->parentSampled = $matches['sampled'] === '01';
```
cleptric, line 173: https://www.w3.org/TR/trace-context/#trace-flags

While this is a hex-encoded bit field, I think we can get away with just checking for `01`, as the other values are either `00` or omitting it altogether.

src/Tracing/PropagationContext.php
```diff
@@ -65,6 +67,14 @@ public function toTraceparent(): string
         return sprintf('%s-%s', (string) $this->traceId, (string) $this->spanId);
     }
 
+    /**
+     * Returns a string that can be used for the W3C `traceparent` header & meta tag.
+     */
+    public function toW3CTraceparent(): string
+    {
+        return sprintf('00-%s-%s', (string) $this->traceId, (string) $this->spanId);
```
cleptric, line 75: https://www.w3.org/TR/trace-context/#version-format

Hardcoding the version to `00` seems to be okay.

src/Tracing/Span.php
```diff
@@ -567,6 +567,20 @@ public function toTraceparent(): string
...
+     */
+    public function toW3CTraceparent(): string
+    {
+        $sampled = '';
+
+        if ($this->sampled !== null) {
+            $sampled = $this->sampled ? '-01' : '-00';
+        }
+
+        return sprintf('00-%s-%s%s', (string) $this->traceId, (string) $this->spanId, $sampled);
```
cleptric, line 581: https://www.w3.org/TR/trace-context/#version-format

Hardcoding the version to 00 seems to be okay.

PR #1683, opened 2024-01-18
Rename `$callable` to `$callback`

cleptric:
We call this parameter `$callback` on all other occurrences in the codebase.
Mainly doing it so named arguments stay consistent. It's not BC, but we haven't released #1670 yet.


stayallive (approved):
👍 good catch!

PR #1686, opened 2024-01-19
http_ssl_verify_peer sets CURLOPT_SSL_VERIFYPEER correctly

tanenbaum:
CURLOPT_SSL_VERIFYPEER is by default true. Set to false when http_ssl_verify_peer is false instead to override the default.

### DIFF
--- a/src/HttpClient/HttpClient.php
+++ b/src/HttpClient/HttpClient.php
@@ -74,8 +74,8 @@ class HttpClient implements HttpClientInterface
         curl_setopt($curlHandle, \CURLOPT_HTTP_VERSION, \CURL_HTTP_VERSION_1_1);
 
         $httpSslVerifyPeer = $options->getHttpSslVerifyPeer();
-        if ($httpSslVerifyPeer) {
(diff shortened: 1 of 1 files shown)


stayallive (approved):
Oops, that makes sense 😅

Curious about the use you had for needing to use this option!

tanenbaum:
> Curious about the use you had for needing to use this option!

I'm testing against a server with a self-signed cert - I've just upgraded the php dep to 4.x. Previously on 3.x I had a custom transport factory that allowed setting those curl options, although I may need to do that again with `CURLOPT_SSL_VERIFYHOST` too.

The next problem I have is the php 8 alpine image I'm using doesn't seem to obey any of these curl settings :sob: - but that's nothing to do with this library, not even the `curl --insecure` command works :man_shrugging:

tanenbaum:
Hey @stayallive what's the typical timeline around getting fixes merged and released?

I'm trying to decide if I should drop in my own http client implementation (duplicating this one essentially) or wait for a release. Thanks!

cleptric:
We'll get this released today/tomorrow.

tanenbaum:
Thanks!

PR #1687, opened 2024-01-19
Allow fluent use of Transaction::setName() method

mfb:
For better developer experience, I'd say Transaction::setName() should also be a fluent method

### DIFF
--- a/src/Tracing/Transaction.php
+++ b/src/Tracing/Transaction.php
@@ -70,10 +70,14 @@ final class Transaction extends Span
      * Sets the name of this transaction.
      *
      * @param string $name The name
+     *
+     * @return $this
      */
-    public function setName(string $name): void
+    public function setName(string $name): self
     {
         $this->name = $name;
+
+        return $this;
     }
 
     /**
--- a/tests/Tracing/TransactionTest.php
+++ b/tests/Tracing/TransactionTest.php
@@ -86,6 +86,18 @@ final class TransactionTest extends TestCase
         $transaction->finish();
     }
 
+    public function testFluentApi(): void
+    {
+        $transaction = new Transaction(TransactionContext::make());
+        $tags = ['foo' => 'bar'];
+        $name = 'baz';
+        $transaction->setTags($tags)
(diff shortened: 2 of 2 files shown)


cleptric:
Any use case in mind? You normally set the name via the context.

mfb:
I set the name via the context, and then later in the request life cycle, set it to the route name; Sentry Laravel SDK also calls Transaction::setName() in the same way.

cleptric:
Yep, but do you have any code example in mind where the fluent API matters?

mfb:
I don't think the fluent API matters per se, it's just to make the methods consistent for better developer experience. 

The other set* methods (inherited from the Span class) return $this, so a developer might reasonably expect this method to also return $this (until they execute the code or run phpstan and notice it doesn't work of course :)

mfb:
The goal would be for something like this to work
```php
$transaction->setTags(['bar' => 'baz'])
  ->setName('foo')
  ->finish();
```

Currently something like this is required:
```php
$transaction->setTags(['bar' => 'baz'])
  ->setName('foo');
$transaction->finish();
```

cleptric:
I think this won't work. `setTag` returns a `Span` and calling `setName` on it throws :/

mfb:
> I think this won't work. `setTags` returns a `Span` and calling `setName` on it throws :/

This code works fine in my testing; setTags() returns the Transaction and nothing was thrown.  What is thrown for you?

cleptric:
hmm, then maybe I misremembered, we had a similar issue before. Anyways, a test please and this is good to go.

stayallive:
We had a self return type hint on span methods which didn't work (because it returned the Span instead of Transaction). Should work fine in the transaction methods though 👍

mfb:
I'm unclear what wouldn't work with return types on Span. If a Transaction method is supposed to return a Span, I thought it work fine because a Transaction is a Span.

But in any case yeah that issue isn't relevant for this issue.

cleptric (approved):

PR #1694, opened 2024-02-01
feat(profiling): Made OsContext and RuntimeContext optional

idsulik:
Hi! Within the PR I made `OsContext` and `RuntimeContext` optional because it was a surprise for me when I tried to profile my application and I didn't receive profiling events so after debugging it turned out that I needed to enable EnvironmentIntegration to make it work.

p.s. I'll add test if it's ok to make them optional(who knows maybe the server requires these fields)

### DIFF
--- a/src/Profiling/Profile.php
+++ b/src/Profiling/Profile.php
@@ -26,14 +26,14 @@ use Sentry\Util\SentryUid;
  *     lineno: int|null,
  * }
  * @phpstan-type SentryProfile array{
- *    device: array{
+ *    device?: array{
  *        architecture: string,
  *    },
  *    event_id: string,
- *    os: array{
+ *    os?: array{
  *       name: string,
- *       version: string,
- *       build_number: string,
+ *       version?: string,
+ *       build_number?: string,
  *    },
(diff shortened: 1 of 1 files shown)


stayallive:
This actually has a reason, it is required by the Sentry server so we cannot make them optional at this moment. The documentation for the profiles payload lives here: https://develop.sentry.dev/sdk/profiles/ in case you are interested.

idsulik:
@stayallive thank you, could you please add information about EnvironmentIntegration to the doc? Because I've spent a lot of time to find the reason why I don't see the profiling event

cleptric:
It's already there https://docs.sentry.io/platforms/php/profiling/troubleshooting/.

idsulik:
oh, I've missed it, I think it should be on this page https://docs.sentry.io/platforms/php/profiling/ because without it you won't see profiling data.

stayallive:
The `EnvironmentIntegration` is a default integration, so it makes sense it is in the troubleshooting section right since "normally" the integration would be enabled?

idsulik:
okay, thank you!

PR #1699, opened 2024-02-07
Lazy initialize default values in client builder

stayallive:
There are a few reasons to do this, but my main change is so the custom values actually take effect since they were not actually used if they were not set on the options. ~~I quickly looked into writing tests for it but realized why we are not doing that already since it a major hassle.~~ We have tests for the setters and getters on the builder but they were not actually working as expected since setting values on the builder like the logger and http transport/client didn't apply unless they were set on the `Options`, setting them on the client builder had no effect, added tests for that now! 😅

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -230,6 +230,22 @@ class Client implements ClientInterface
     }
 
     /**
+     * @internal
+     */
+    public function getLogger(): LoggerInterface
+    {
+        return $this->logger;
+    }
+
+    /**
+     * @internal
+     */
+    public function getTransport(): TransportInterface
+    {
+        return $this->transport;
+    }
+
+    /**
      * Assembles an event and prepares it to be sent of to Sentry.
      *
      * @param Event          $event The payload that will be converted to an Event
--- a/src/ClientBuilder.php
+++ b/src/ClientBuilder.php
@@ -23,12 +23,12 @@ final class ClientBuilder
     private $options;
 
     /**
-     * @var TransportInterface The transport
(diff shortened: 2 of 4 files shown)


Jean85 (commented):
This makes sense, but we may need to check if the framework integrations invalidate this by calling `getClient` early.

I would also suggest to put those lazy fallbacks on newlines, putting `??` at the start of the newline, for better readability (and alignment with PER-CS 2.0)

stayallive:
As far as I understand this is not an issue for Symfony (https://github.com/getsentry/sentry-symfony/blob/24d29681ef395afaf40fb37941fa14c67836a785/src/DependencyInjection/SentryExtension.php#L149-L161) and it certainly is not an issue for Laravel.

I'll make the CS changes, that sounds fine to me!

stayallive:
@Jean85 can you point me to the section of the [PER-CS 2.0](https://www.php-fig.org/per/coding-style/) where this is mentioned? I wanted to check what the spacing should be but the only `??` I can find is the binary operators section which mentions no newlines.

cleptric (approved):

Jean85:
> @Jean85 can you point me to the section of the [PER-CS 2.0](https://www.php-fig.org/per/coding-style/) where this is mentioned? I wanted to check what the spacing should be but the only `??` I can find is the binary operators section which mentions no newlines.

I can't find it, I'm probably misremembering. I personally prefer to put the operator at the start of the line, so it's better aligned and more clear about what's happening, regardless from the length of the line itself.

PR #1701, opened 2024-02-12
short circuit ignore_exceptions

jspaetzel:
It might be a nice optimization for ignored exceptions to short circuit in captureException? Something like this

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -150,6 +150,11 @@ class Client implements ClientInterface
     {
         $hint = $hint ?? new EventHint();
 
+        if ($this->isIgnoredException(\get_class($exception))) {
+            // short circuit to avoid processing into an event
+            return null;
+        }
+
         if ($hint->exception === null) {
             $hint->exception = $exception;
         }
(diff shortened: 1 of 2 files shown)


jspaetzel:
Somewhat related to this it might be nice to provide a way to short circuit using a configured callback. I personally would rather implement a callback that hooks into here instead of beforeSend

cleptric:
Your changes look reasonable, thank you!

> Somewhat related to this it might be nice to provide a way to short circuit using a configured callback. I personally would rather implement a callback that hooks into here instead of beforeSend

We could make `ignore_exceptions` accept `callable`, too. I do see the benefit, so why not? 🙂 

`before_send` still has its valid use cases, as it's meant to mutate the event before it is transmitted to Sentry.

jspaetzel:
Agreed! I'll keep this pr small but may open another to have ignore_exceptions accept a callable, I like the sound of that!

stayallive (approved):
Looks like a good little optimization change to me!

cleptric (approved):

cleptric:
Released in [v4.6.0](https://github.com/getsentry/sentry-php/releases/tag/4.6.0)

PR #1707, opened 2024-02-27
adds `JSON_ERROR_NON_BACKED_ENUM` to allowed `JSON::encode()` errors.

ju1ius:
The `JSON_ERROR_NON_BACKED_ENUM` error code is not yet documented (see https://github.com/php/doc-en/issues/2747), but it is returned when trying to encode encode a non-backed enum to JSON.

Since this error code is semantically equivalent to `JSON_ERROR_UNSUPPORTED_TYPE` which is already in the allow list, it makes sense to also add it.

For context, I was adding a Monolog record as a Breadcrumd metadata item, which looked like this:
```php
[
    'type' => 'default',
    'category' => 'doctrine',
    'level' => 'debug',
    'timestamp' => 1709047886.0,
    'message' => 'Executing statement: {sql} (parameters: {params}, types: {types})',
    'data' => [
        'sql' => 'SELECT id, key FROM ... WHERE id = ? AND key = ?',
        'params' => [42, 'somekey'],
        'types' => [
            \Doctrine\DBAL\ParameterType::INTEGER,
            \Doctrine\DBAL\ParameterType::STRING,
        ],
    ],
],
```

The encoding error is triggered by `Doctrine\DBAL\ParameterType`, which is a unit enum in Doctrine DBAL v4.
Since the `JSON_ERROR_NON_BACKED_ENUM` is not handled by `JSON::encode()`, this caused an exception to be thrown in `PayloadSerializerInterface::serialize()` and as a result the whole event was ignored (nothing was sent to the sentry endpoint).

### DIFF
--- a/tests/Util/Fixtures/NonBackedEnum.php
+++ b/tests/Util/Fixtures/NonBackedEnum.php
@@ -1,4 +1,6 @@
-<?php declare(strict_types=1);
+<?php
+
+declare(strict_types=1);
 
 namespace Sentry\Tests\Util\Fixtures;
 
--- a/tests/Util/Fixtures/NonBackedEnum.php
+++ b/tests/Util/Fixtures/NonBackedEnum.php
@@ -1,0 +1,9 @@
+<?php declare(strict_types=1);
+
+namespace Sentry\Tests\Util\Fixtures;
+
+enum NonBackedEnum
+{
+    case None;
+    case Some;
+}
--- a/src/Util/JSON.php
+++ b/src/Util/JSON.php
@@ -34,7 +34,7 @@ final class JSON
 
         $encodedData = json_encode($data, $options, $maxDepth);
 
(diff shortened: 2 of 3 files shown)


cleptric (commented):
Please add a test for this. You can skip it on PHP versions that do not support `JSON_ERROR_NON_BACKED_ENUM`.

cleptric (approved):
Thanks

tests/Util/Fixtures/NonBackedEnum.php
```diff
@@ -0,0 +1,9 @@
+<?php declare(strict_types=1);
+
+namespace Sentry\Tests\Util\Fixtures;
```
cleptric, line 3:
```suggestion
<?php

declare(strict_types=1);

namespace Sentry\Tests\Util\Fixtures;
```

ju1ius:
> Thanks

You're welcome

cleptric:
This was released with [4.6.1](https://github.com/getsentry/sentry-php/releases/tag/4.6.1).

Issue #1715, opened 2024-03-08
`TransactionIntegration` should always set a transaction


cleptric:
https://github.com/getsentry/sentry-php/blob/1b9e3b2f9304f777b5b0eb98e3b770d820ecb888/src/Integration/TransactionIntegration.php#L39-L43

We should make sure we always end up writing something into `$event->transaction`, to offer a better UX in the product. With my local setup, `$_SERVER['PATH_INFO']` seems to be not populated, using Apache.

PR #1716, opened 2024-03-08
Set the active transaction name on the event

cleptric:
Brainstorming a fix for https://github.com/getsentry/sentry-php/issues/1715

The scenario here is that an error should contain the active transaction name when the error is sent to Sentry. The only other place besides the integration where we set this property on the event is in [`Transaction::finish()`](https://github.com/getsentry/sentry-php/blob/master/src/Tracing/Transaction.php#L181).

### DIFF
--- a/src/Integration/TransactionIntegration.php
+++ b/src/Integration/TransactionIntegration.php
@@ -36,6 +36,13 @@ final class TransactionIntegration implements IntegrationInterface
                 return $event;
             }
 
+            $transaction = SentrySdk::getCurrentHub()->getTransaction();
+            if ($transaction !== null) {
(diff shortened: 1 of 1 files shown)


src/Integration/TransactionIntegration.php
```diff
@@ -36,6 +36,13 @@ public function setupOnce(): void
                 return $event;
             }
 
+            $transaction = SentrySdk::getCurrentHub()->getTransaction();
+            if ($transaction !== null) {
+                $event->setTransaction($transaction->getName());
+
+                return $event;
+            }
+
```
cleptric, line 45: We should likely move this after the hint was applied, as this should take precedence.

PR #1717, opened 2024-03-08
Add timing span when emiting a timing metric

cleptric:


src/Metrics/Metrics.php
```diff
@@ -144,7 +147,12 @@ public function timing(
     ) {
         $startTimestamp = microtime(true);
 
-        $result = $callback();
+        $result = trace(function () use ($callback) {
+            return $callback();
```
cleptric, line 151: Move timing emission in here after the callback was called.

stayallive (approved):
I think this looks good!

PR #1720, opened 2024-03-17
Fix: typo

Lisio:


Jean85 (changes requested):
This is a typo, correct, but changing the name like this would be a breaking change; maybe we can duplicate the method, and deprecate the one with the typo, while it just forwards the call to the correct one?

cleptric:
Closing in favor of https://github.com/getsentry/sentry-php/pull/1725.

PR #1722, opened 2024-03-25
Add error message to failed requests

iwex:
Sometimes we don't log useful info about an error.
In my case, it was hard to debug that `{"detail":"envelope exceeded size limits for type 'event' (https://develop.sentry.dev/sdk/envelopes/#size-limits)"}`

This PR will add an error message for some unsuccessful requests.

I can't find an API for possible response - https://develop.sentry.dev/sdk/envelopes/
So I guess we can just use general info about error.

### DIFF
--- a/src/HttpClient/HttpClient.php
+++ b/src/HttpClient/HttpClient.php
@@ -107,7 +107,7 @@ class HttpClient implements HttpClientInterface
 
         $error = '';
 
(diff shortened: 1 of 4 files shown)


stayallive (approved):
I think this is a good change, modified it a bit but looks good 👍

Issue #1723, opened 2024-03-28
Metrics Rate Limits


stephanie-anderson:
Project description:
https://github.com/getsentry/team-sdks/issues/77

PR #1726, opened 2024-04-02
Handle `metric_bucket` rate limits

cleptric:
This adds handling rate limits for metrics.
The namespace will not be customizable from the SDK, hence we do not need to check for the namespace, as `custom` will always be inferred.

closes https://github.com/getsentry/sentry-php/issues/1723


stayallive (approved):
Noice, this was simpler than I anticipated. 💪

PR #1728, opened 2024-04-03
Handle namespaces for `metric_bucket` rate limits

cleptric:
https://github.com/getsentry/sentry-php/pull/1726 seemed to simple to be true, and it wasn't 😬

This adds proper handling of metric namespaces. While the SDK does not support adding a namespace to a metric, Relay could reply with a `X-Sentry-Rate-Limits` header that could contain none `custom` namespaces, in which case we're still good sending our, inferred `custom` namespaces metrics.

Refs https://github.com/getsentry/team-sdks/issues/77

### DIFF
--- a/src/Serializer/EnvelopItems/MetricsItem.php
+++ b/src/Serializer/EnvelopItems/MetricsItem.php
@@ -38,7 +38,7 @@ class MetricsItem implements EnvelopeItemInterface
 
         foreach ($metrics as $metric) {
             /**
-             * In case of us adding support for setting a namespace for metrics,
+             * In case of us adding support for emitting metrics from other namespaces,
              * we have to alter the RateLimiter::class to properly handle these
              * namespaces.
              */
--- a/src/Serializer/EnvelopItems/MetricsItem.php
+++ b/src/Serializer/EnvelopItems/MetricsItem.php
@@ -37,6 +37,12 @@ class MetricsItem implements EnvelopeItemInterface
         $metricMetaPayload = [];
 
         foreach ($metrics as $metric) {
+            /**
+             * In case of us adding support for setting a namespace for metrics,
+             * we have to alter the RateLimiter::class to properly handle these
+             * namespaces.
+             */
+
             // key - my.metric
             $line = preg_replace(self::KEY_PATTERN, '_', $metric->getKey());
 
--- a/src/Transport/HttpTransport.php
+++ b/src/Transport/HttpTransport.php
@@ -59,7 +59,7 @@ class HttpTransport implements TransportInterface
         $this->httpClient = $httpClient;
         $this->payloadSerializer = $payloadSerializer;
         $this->logger = $logger ?? new NullLogger();
-        $this->rateLimiter = new RateLimiter();
+        $this->rateLimiter = new RateLimiter($this->logger);
     }
 
     /**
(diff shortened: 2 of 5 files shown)


src/Transport/HttpTransport.php
```diff
@@ -123,15 +123,7 @@ public function send(Event $event): Result
...
 
-        if ($this->rateLimiter->handleResponse($response)) {
-            $eventType = $event->getType();
-            $disabledUntil = $this->rateLimiter->getDisabledUntil($eventType);
-
-            $this->logger->warning(
-                sprintf('Rate limited exceeded for requests of type "%s", backing off until "%s".', (string) $eventType, gmdate(\DATE_ATOM, $disabledUntil)),
-                ['event' => $event]
-            );
-        }
```
cleptric, line 134: I move this into the `RateLimiter`, felt like a better fit.

src/Transport/RateLimiter.php
```diff
@@ -24,36 +36,80 @@ final class RateLimiter
...
+                $retryAfter = $now + (ctype_digit($parameters[0]) ? (int) $parameters[0] : self::DEFAULT_RETRY_AFTER_SECONDS);
 
                 foreach (explode(';', $parameters[1]) as $category) {
-                    $this->rateLimits[$category ?: 'all'] = $now + $delay;
+                    switch ($category) {
+                        case self::DATA_CATEGORY_METRIC_BUCKET:
+                            $namespaces = [];
+                            if (isset($parameters[4])) {
+                                $namespaces = explode(';', $parameters[4]);
+                            }
```
cleptric, line 80: @jan-auer Elvis did not make it. Ended up being `[0 => '']` foe the none set case.

src/Serializer/EnvelopItems/MetricsItem.php
```diff
@@ -37,6 +37,12 @@ public static function toEnvelopeItem(Event $event): string
         $metricMetaPayload = [];
 
         foreach ($metrics as $metric) {
+            /**
+             * In case of us adding support for setting a namespace for metrics,
```
jan-auer, line 41: Tiny nitpick: Not necessarily just setting namespaces, but simply the SDK emitting metrics from other namespaces. That could also be built-in functionality without an explicit user-facing API, such as client reports.

src/Transport/RateLimiter.php
```diff
@@ -24,36 +36,80 @@ final class RateLimiter
...
-                $parameters = array_splice($parameters, 0, 2);
-                $delay = ctype_digit($parameters[0]) ? (int) $parameters[0] : self::DEFAULT_RETRY_AFTER_DELAY_SECONDS;
+                /**
+                 * $parameters[0] - retry_after
+                 * $parameters[1] - categories
+                 * $parameters[2] - scope (not used)
+                 * $parameters[3] - reason_code (not used)
+                 * $parameters[4] - namespaces (only returned if categories contains "metric_bucket").
+                 */
+                $parameters = explode(':', $limit, 5);
```
jan-auer, line 70: Can you use something like `list($retry_after, $categories, ...)` here to translate the comment into code? Note that the last few components are optional, though.

jan-auer (approved):

src/Transport/RateLimiter.php
cleptric: This will fail miserably, if not set 😬

stayallive: You can append `str_repeat(':', 5)` to the `$limit` so the explode always returns at least 5 parts 😉

stayallive (approved):
That is a lot more complex yeah, but looks good from a distance!

PR #1729, opened 2024-04-03
Update metric normalization

cleptric:
https://develop-git-statsd-unicode-escapes.sentry.dev/sdk/metrics/#normalization

closes https://github.com/getsentry/sentry-php/issues/1731

### DIFF
--- a/src/Serializer/EnvelopItems/MetricsItem.php
+++ b/src/Serializer/EnvelopItems/MetricsItem.php
@@ -120,11 +120,17 @@ class MetricsItem implements EnvelopeItemInterface
     {
         $tagValue = str_replace(
             [
+                "\n",
+                "\r",
+                "\t",
                 '\\',
                 '|',
                 ',',
             ],
             [
+                '\n',
+                '\r',
+                '\t',
                 '\\\\',
                 '\u{7c}',
                 '\u{2c}',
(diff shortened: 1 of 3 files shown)


cleptric:
@stayallive When you look at the linked docs, regarding Line Feed, Carriage Return and Tab.
I don't think you can use these characters in PHP. Or can you?

stayallive:
You can right... `"\n\r\t"` (only inside `"`)?

So technically you can, but it's a stretch anyone would do it on purpose. But should also be easy to replace right? Something like `str_replace(["\n", "\r", ...], ['\n', '\r', ...], $item)` I'm thinking?

src/Serializer/EnvelopItems/MetricsItem.php
```diff
@@ -110,4 +115,27 @@ public static function toEnvelopeItem(Event $event): string
             $statsdPayload
         );
     }
+
+    private static function replaceTagValueCharacters(string $tagValue): string
```
markushi, line 119: Not sure if relevant for PHP, but only relay control characters are replaced as well: https://github.com/getsentry/relay/blob/3208e3ce5b1fe4d147aa44e0e966807c256993de/relay-metrics/src/protocol.rs#L62C7-L62C20

cleptric: Does not apply to any SDKs, we do this in Relay only.

stayallive (approved):
Code wise this looks good!

PR #1738, opened 2024-05-06
Add array shape for init function

PrinsFrank:
To make looking for configuration options easier, and to help static analysis tools to check for incorrect array key types, this PR introduces an array shape for the init function.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -4,6 +4,9 @@ declare(strict_types=1);
 
 namespace Sentry;
 
+use Psr\Log\LoggerInterface;
+use Sentry\HttpClient\HttpClientInterface;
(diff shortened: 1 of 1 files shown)


cleptric (approved):
Thanks!

BenjaminEllisSo:
Useful feature, but I think there's an error:
            'profiler_sample_rate' or 'profiles_sample_rate' ?
            
            The array shape wants profiler_sample_rate BUT the Options class / client builder wants profiles_sample_rate 
            
  So if you follow the array shape, it throws an error

Jean85:
That's really strange, SA should pick this up in our code too, unless there's no use of that option directly.

Can you file a PR to reproduce this?

PR #1739, opened 2024-05-14
`class_serializers` supports closures

tw2066:
`class_serializers` supports closures

eg.

```php
    'class_serializers' => [
        'callable' => static function (object $object) {
            if(\Hyperf\Stringable\Str::startsWith($object::class,'HyperfExample')){
                return json_decode(json_encode($object),true);
            }
            return $object;
        }
    ],

```

### DIFF
--- a/src/Serializer/AbstractSerializer.php
+++ b/src/Serializer/AbstractSerializer.php
@@ -167,8 +167,7 @@ abstract class AbstractSerializer
         foreach ($this->options->getClassSerializers() as $type => $serializer) {
             if ($object instanceof $type) {
                 $serializers[] = $serializer;
-            }
-            elseif ('callable' == $type && is_callable($serializer)) {
+            } elseif ($type == 'callable' && \is_callable($serializer)) {
                 $serializers[] = $serializer;
             }
         }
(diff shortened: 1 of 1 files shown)


stayallive:
This doesn't do what you think it does, have you tested your change?

https://github.com/getsentry/sentry-php/blob/d12ed37fcfaddaa041f221fe165bd0d435dbef0f/tests/Serializer/SerializerTest.php#L111-L119

The serializer already can be a closure just like we do in our test cases.

The key in `class_serializers` must match the class your callback is serializing, so "function" in your example probably won't work but needs to be some (base) class available in your application.

tw2066:
Consider making a special judgment on type

cleptric (changes requested):
This needs tests and I'm currently not convenient why this change is needed in the first place.

stayallive:
When using our `ClientBuilder` directly you can set your own serializer (that extends the SDK one potentially) in which you can do whatever you want: 

https://github.com/getsentry/sentry-php/blob/d12ed37fcfaddaa041f221fe165bd0d435dbef0f/src/ClientBuilder.php#L78-L83

This current approach doesn't make sense to me, `class_serializers` with a magic key where the key is not a class name is going to cause much confusion. So either a custom serializer is the way to go here or creating a whole new option (`object_serializer` maybe) to support a catch-all serializer for objects could make sense but then we might want to create a more generic option to serialize more types.

Anyway, using the client builder seems like the best bet here for your specific needs?

PR #1740, opened 2024-05-16
Test span sampled status before creating child spans

stayallive:
As reported in getsentry/sentry-laravel#894 we have some logic issues.

There are only 2 here in the base SDK, both solved by this PR:

- Fixes `trace` helper to check parent span sampled status
- Fixes Guzzle tracker to check parent span sampled status
- Fixes Guzzle tracker not always recording a breadcrumb (a bonus fix)


cleptric (approved):
We should add more tests to cover all cases.

PR #1741, opened 2024-05-18
Represent callable strings as strings

### LINKED ISSUE #1734: Strings that happen to be a php function are reported as a callable
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

4.7.0

### Steps to reproduce

If a function calling argument is an array containing various strings, any strings that happen to be a php function will be reported as a callable.

For example this array: `['scope' => 'header', 'type' => 'file']` will be reported as:
```
{
  scope: Callable void header [string header; bool [replace]; int [response_code]],
  type: Callable file [string filename; int [flags]; mixed|null [context]]
}
```

because header() and file() happen to be functions.

My 2 cents is that is_callable() should not be called on strings, they can be reported simply as strings.

### Expected result

```
{
  scope: header,
  type: file
}
```

### Actual result

```
{
  scope: Callable void header [string header; bool [replace]; int [response_code]],
  type: Callable file [string filename; int [flags]; mixed|null [context]]
}
```

mfb:
I'm honestly not sure what the logic should be here. A simple callable string like "file" I don't think should be assumed to be a callable.  But seems a lot more reasonable to represent a callable string containing "::" as callable.  Here I also represent callable strings containing an underscore as callables.

Fixes #1734

### DIFF
--- a/src/Serializer/AbstractSerializer.php
+++ b/src/Serializer/AbstractSerializer.php
@@ -280,7 +280,7 @@ abstract class AbstractSerializer
      */
     protected function serializeCallable($callable): string
     {
-        if (\is_string($callable) && !\function_exists($callable)) {
+        if (\is_string($callable)) {
             return $callable;
         }
 
--- a/src/Serializer/AbstractSerializer.php
+++ b/src/Serializer/AbstractSerializer.php
@@ -292,7 +292,7 @@ abstract class AbstractSerializer
             if (\is_array($callable)) {
                 $reflection = new \ReflectionMethod($callable[0], $callable[1]);
                 $class = $reflection->getDeclaringClass();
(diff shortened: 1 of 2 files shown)


stayallive:
Honestly, how about we consider `is_string() && function_exists()` as a non-callable?

I feel like we can't do it 100% right anyway (without some very advanced reflection and docblock parsing and even then...) and the false-positive rate is going to be a lot smaller than the current situation.

mfb:
Ok, another take with the simpler logic, representing all callable strings as strings.

cleptric:
Please add a test.

stayallive (approved):
Thanks, this will reduce false-positives! 👍

PR #1743, opened 2024-05-24
Deprecate `enable_tracing` option

stayallive:


src/functions.php
```diff
@@ -28,7 +28,6 @@
  *     context_lines?: int|null,
  *     default_integrations?: bool,
  *     dsn?: string|bool|null|Dsn,
- *     enable_tracing?: bool|null,
```
stayallive, line 31: Already removed to discourage it's usage

PR #1768, opened 2024-08-05
Update functions.php to correct options

BenjaminEllisSo:
fix `profiler_sample_rate` to `profiles_sample_rate` to match what Options.php expects


cleptric (approved):
Thank you!

PR #1773, opened 2024-08-14
Set `http` breadcrumb level based on response code

cleptric:
4XX response codes are set to `warning` and 5XX are set to `error`.

Part of https://github.com/getsentry/team-sdks/issues/100


stayallive (approved):
👍

PR #1777, opened 2024-09-05
CS

stayallive:
- Run php-cs-fixer on codebase
- Remove deprecated option (no behaviour change): https://cs.symfony.com/doc/rules/function_notation/nullable_type_declaration_for_default_null_value.html
- Add missing type hints
- Fix phpstan complaints
- Fix php-cs-fixer option that default to an incompatible value in `@Symfony` for out PHP constraint

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -18,9 +18,6 @@ return (new PhpCsFixer\Config())
         ],
         'self_accessor' => false,
         'modernize_strpos' => false,
-        'nullable_type_declaration_for_default_null_value' => [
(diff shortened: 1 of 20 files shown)


cleptric (approved):
Thanks! Next major we either pin phpcs or look for an alternative. These constant changes are super annoying.

Jean85:
> Thanks! Next major we either pin phpcs or look for an alternative. These constant changes are super annoying.

I would suggest pinning, since we have Dependabot already that would take care of keeping it up to date.

cleptric:
I don’t need more dependabot notifications either. We can update it once a year, more than sufficient for our use case.

PR #1778, opened 2024-09-05
Allow profiler to start before SDK and transaction is initialized

stayallive:
In plain PHP this is less interesting but in frameworks it can be very.

This allow you to drop a `\Sentry\Profiling\Profiler::startProfiling()` right after the composer autoloader which enable you to start profiling as early as possible capturing the bootstrap of the framework too.

If we later know we are going to discard the transaction we discard the profiling data, if we are sampling the transaction we continue to use the already started (excimer) profiler as normal.

_CI failure because: #1777_

### DIFF
--- a/src/Profiling/Profiler.php
+++ b/src/Profiling/Profiler.php
@@ -16,7 +16,12 @@ final class Profiler
     /**
      * @var \ExcimerProfiler|null
      */
-    private $profiler;
+    private static $profiler;
+
+    /**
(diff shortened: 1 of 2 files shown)


cleptric (commented):
With continues profiling now being in beta, let's talk about how we implement this in PHP first, as this has some overlap.

cleptric:
We'll have this features available with continues profiling.

PR #1779, opened 2024-09-06
Fix Scope extra data setup by Event extra data

devicebusy:
Without this fix API doesn't log extra data to Sentry

### DIFF
--- a/src/State/Scope.php
+++ b/src/State/Scope.php
@@ -359,7 +359,7 @@ class Scope
             $event->setTags(array_merge($this->tags, $event->getTags()));
         }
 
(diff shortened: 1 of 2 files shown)


stayallive:
I'm not sure we want to fix this, because extra data is deprecated: https://docs.sentry.io/platforms/php/enriching-events/context/#additional-data

Can't you use [context](https://docs.sentry.io/platforms/php/enriching-events/context/) instead?

devicebusy:
@stayallive Thank you, Context works fine. Probably Sentry::setExtras() should be marked as deprecated?

PR #1781, opened 2024-09-30
[Monolog] Only skip real exceptions

wjzijderveld:
When adding a `exception` property to the monolog context, it would get skipped regardless if the exception was used as hint or not.


cleptric (commented):
Looks like something for the next major. Not keen of changing this behavior in a minor.

wjzijderveld:
I can  understand that, in a weird way it's a BC break.
Personally consider it a bugfix, as [PSR-3 states](https://www.php-fig.org/psr/psr-3/#13-context) that the context can contain anything and would need to be verified if it's actually an Exception as it can be anything. And currently this data is completely dropped when send to Sentry, which is super confusing.

Litarnus:
@wjzijderveld we added [ExceptionToSentryIssueHandler](https://github.com/getsentry/sentry-php/blob/master/src/Monolog/ExceptionToSentryIssueHandler.php#L18) and [LogToSentryIssueHandler](https://github.com/getsentry/sentry-php/blob/master/src/Monolog/LogToSentryIssueHandler.php), which split up capturing monolog exceptions and converting log messages to issues into separate handler instead of having both functionalities combined in one handler. 

Those handlers handle the behaviour as described in this PR, so I will close this since we deprecated `Monolog\Handler` anyways

PR #1783, opened 2024-10-26
feat: Add support timeouts in milliseconds for HTTP client

nikolicaleksa:
Adds support for having timeout and connection timeout in milliseconds for HTTP client


cleptric:
Why do you see the need to reduce the timeout below 1s?

cleptric (approved):

mfb:
So is CURLOPT_TIMEOUT only an integer? Meaning this CURLOPT_TIMEOUT_MS would also be needed to configure 1.5 seconds?

PR #1798, opened 2025-01-22
Add date range to MIT licence

philipphofmann:
To align with all other SDK licenses.

#skip-changelog


cleptric:
We do not bump the license year. This was discussed a while ago, can't find a link though.

PR #1803, opened 2025-01-30
Serialize date

### LINKED ISSUE #1799: Serialize `DateTimeInterface`
### Problem Statement

I want to serialize date objects (`DateTimeInterface` to be concrete).

They are used by nearly every PHP application. Serializing them is actually very simple, and sending them doesn't cause much overhead.

This is how this would look like in Sentry:
Image

Compare it to how these are currently displayed:
Image

### Solution Brainstorm

In `AbstractSerializer::serializeRecursively()` we could add the following:

```PHP
            if (\is_object($value)) {

                foreach ($classSerializers as $classSerializer) {
                   ...
                }

                if ($value instanceof \DateTimeInterface) {
                    return $this->serializeDate($value);
                }
```

```PHP
    protected function serializeDate(\DateTimeInterface $date): string
    {
      $hasMicroseconds = $date->format("u") !== "000000";
      $formatted = $hasMicroseconds ? $date->format("Y-m-d H:i:s.u") : $date->format("Y-m-d H:i:s");
      $className = get_class($date);

      if ($date->getTimezone()) {
        // Also add timezone information
      }

      return "$className($formatted)";
    }

martinstuecklschwaiger:
Fixes #1799.

DateTime becomes readable on Sentry. 

Before they were serialized like this: `Object DateTimeImmutable`.
They now are serialized as `DateTime(2001-02-03 13:37:42)`.

### DIFF
--- a/src/Serializer/AbstractSerializer.php
+++ b/src/Serializer/AbstractSerializer.php
@@ -135,6 +135,10 @@ abstract class AbstractSerializer
                     }
                 }
 
+                if ($value instanceof \DateTimeInterface) {
+                    return $this->formatDate($value);
+                }
+
                 if ($this->serializeAllObjects || ($value instanceof \stdClass)) {
                     return $this->serializeObject($value, $_depth);
                 }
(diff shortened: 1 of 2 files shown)


martinstuecklschwaiger:
If we allowed class serializers to return scalar data as well (not only arrays), we could also make this a default serializer.

cleptric (approved):
Cool, thank you!

PR #1820, opened 2025-04-08
Improve handling of prefix stripping for anonymous frames

### LINKED ISSUE #1817: Grouping by Stacktrace with anonymous classes/functions
### How do you use Sentry?

Self-hosted

### SDK version

4.10.0

### Steps to reproduce

The anonymous class in the stack trace contains the full path of the file. Example:
`Namespace\Class\Subclass@anonymous/path/to/the/file/Class.php::method`

We have a very specific deployment scheme, and each deployment is in a new directory containing the date of deployment, so the file path is different with each deployment, e.g.: `Namespace\Class\Subclass@anonymous/path/2025-03-27T07-20-22Z/to/the/file/Class.php::method`

Unfortunately, it doesn't work well with grouping `By Exception Stack-trace`, as every release, the same error is treated as different, thus it is not grouped properly.

### Expected result

Grouping `By Exception Stack-trace` ignores the path of anonymous classes/functions. Or there is another way to group by stack trace, ignoring file paths.

### Actual result

Grouping `By Exception Stack-trace` ignores the path of anonymous classes/functions, so the same issues are not grouped every release:

Image

It doesn't work for PHP 8.2, and since PHP 8.4, it's even worse, because all anonymous functions have paths in the stack trace.

stayallive:
Improve handling of prefix stripping for anonymous frames. As reported in #1817 we did not handle the case where the class name is known opposed to being "class". Added a test and ensure we handle that case now. Also added a (micro) optimization to never even run the code if there is no benefit (no prefixes defined).

Fixes #1817

### DIFF
--- a/src/FrameBuilder.php
+++ b/src/FrameBuilder.php
@@ -76,9 +76,13 @@ final class FrameBuilder
 
             // Optimization: skip doing regex if we don't have prefixes to strip
             if ($this->options->getPrefixes()) {
-                $functionName = preg_replace_callback('/@anonymous\\x00([^:]+)(:.*)?/', function (array $matches) {
+                $prefixStrippedFunctionName = preg_replace_callback('/@anonymous\\x00([^:]+)(:.*)?/', function (array $matches) {
                     return "@anonymous\x00" . $this->stripPrefixFromFilePath($this->options, $matches[1]) . ($matches[2] ?? '');
                 }, $functionName);
+
+                if ($prefixStrippedFunctionName) {
+                    $functionName = $prefixStrippedFunctionName;
+                }
             }
 
             $rawFunctionName = \sprintf('%s::%s', $backtraceFrame['class'], $backtraceFrame['function']);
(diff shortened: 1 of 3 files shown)


src/FrameBuilder.php
```diff
@@ -74,8 +74,15 @@ public function buildFromBacktraceFrame(string $file, int $line, array $backtrac
         if (isset($backtraceFrame['class']) && isset($backtraceFrame['function'])) {
             $functionName = $backtraceFrame['class'];
 
-            if (mb_substr($functionName, 0, mb_strlen(Frame::ANONYMOUS_CLASS_PREFIX)) === Frame::ANONYMOUS_CLASS_PREFIX) {
-                $functionName = Frame::ANONYMOUS_CLASS_PREFIX . $this->stripPrefixFromFilePath($this->options, substr($backtraceFrame['class'], \strlen(Frame::ANONYMOUS_CLASS_PREFIX)));
+            // Optimization: skip doing regex if we don't have prefixes to strip
```
cleptric, line 77:
```suggestion
            // Skip if no prefixes are set
```

cleptric (approved):

PR #1823, opened 2025-04-15
Add package info to SDK payload

### LINKED ISSUE #1822: Add `packages` context
If users disable the modules integration, we loose which base PHP SDK version they use.
Inject the version used into the package context.

```json
"packages": [
    {
        "name": "composer:sentry/sentry",
        "version": "4.11.0"
    }
]
```

stayallive:
~~I have opted to implement this using a registration. This gives us the option to replace this integration in framework SDKs so we can also inject the framework SDK version by providing a specific `SDKModuleIntegration` there. This integration serves as a fallback is therefore registered at the end and only does something in case there is are no modules set by for example the default modules integration.~~

I misunderstood the assignment completely. So it's now rewritten to set information on the `packages` key of the [SDK payload](https://develop.sentry.dev/sdk/data-model/event-payloads/sdk/). Framework SDK are expected to add a integration like:

```php
class LaravelPackageIntegration implements IntegrationInterface
{
    {
        Scope::addGlobalEventProcessor(static function (Event $event, ?EventHint $hint = null): Event {
            $self = SentrySdk::getCurrentHub()->getIntegration(self::class);

            if (!$self instanceof self) {
                return $event;
            }

            $event->appendSdkPackage([
                'name' => 'composer:sentry/sentry-laravel',
                'version' => Version::SDK_VERSION,
            ]);

            return $event;
        });
    }
}
```

This will append a package to the payload.

Fixes #1822


src/Integration/SDKModuleIntegration.php
```diff
@@ -0,0 +1,37 @@
...
+        Scope::addGlobalEventProcessor(static function (Event $event): Event {
+            $integration = SentrySdk::getCurrentHub()->getIntegration(self::class);
+
+            // The integration could be bound to a client that is not the one
+            // attached to the current hub. If this is the case, bail out
+            if ($integration !== null && $event->getModules() === []) {
+                $event->setModules([
+                    'sentry/sentry' => Client::SDK_VERSION,
+                ]);
+            }
```
cleptric, line 32: tbh, we can just always attach this.

cleptric (approved):

src/Integration/SDKModuleIntegration.php
stayallive: Without the check if `getModules` is empty? This is done to not override the normal ModulesIntegraion in case it's also registered (like it would be by default).

cleptric (approved):

PR #1852, opened 2025-06-07
Fix `Options::setEnableLogs` and add `Options:: setBeforeSendLogCallback()`

cleptric:
`Options::setEnableLogs()` was toggling `enable_tracing`. Also added the missing `Options::setBeforeSendLogCallback()` method.


stayallive (approved):
🫣

PR #1862, opened 2025-06-12
Fix handling of backtrace frames

### LINKED ISSUE #1861: Since updating to version 4.13.0, I’m encountering the following error during CI tests:
Since updating to version 4.13.0, I’m encountering the following error during CI tests:

```
  Undefined array key 0
  at vendor/sentry/sentry/src/FrameBuilder.php:226
    222▕             if ($reflectionParameter->isVariadic()) {
    223▕                 // For variadic parameters, collect all remaining arguments into an array
    224▕                 $variadicArgs = [];
    225▕                 for ($i = $parameterPosition; $i < \count($backtraceFrameArgs); ++$i) {
  ➜ 226▕                     $variadicArgs[] = $backtraceFrameArgs[$i];
    227▕                 }
    228▕                 $argumentValues[$reflectionParameter->getName()] = $variadicArgs;
    229▕                 // Variadic parameter is always the last one, so we can break
    230▕                 break;
      +15 vendor frames 
```

Interestingly, I’m unable to reproduce this issue locally.

_Originally posted by @poldixd in https://github.com/getsentry/sentry-php/issues/1849#issuecomment-2966008000_

cleptric:
Fixes #1861

### DIFF
--- a/src/FrameBuilder.php
+++ b/src/FrameBuilder.php
@@ -221,7 +221,7 @@ final class FrameBuilder
 
             if ($reflectionParameter->isVariadic()) {
                 // For variadic parameters, collect all remaining arguments into an array
-                $variadicArgs = array_slice($backtraceFrameArgs, $parameterPosition);
+                $variadicArgs = \array_slice($backtraceFrameArgs, $parameterPosition);
                 $argumentValues[$reflectionParameter->getName()] = array_values($variadicArgs);
                 // Variadic parameter is always the last one, so we can break
                 break;
--- a/src/FrameBuilder.php
+++ b/src/FrameBuilder.php
@@ -221,11 +221,8 @@ final class FrameBuilder
 
             if ($reflectionParameter->isVariadic()) {
                 // For variadic parameters, collect all remaining arguments into an array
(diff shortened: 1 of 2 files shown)


cleptric:
@poldixd could you give this a try and lmk if it fixes your issue? I'm not sure in which cases this problem can actually happen, hence the fix is purely based on a synthetic test.
If you can share a repro, even better.

poldixd:
Thanks! The issue is now resolved on my end.

My original code looked like this and I created a test for this exception:

```php
try {
    $contactPerson->addMediaFromUrl('foo.bar')
        ->preservingOriginal()
        ->toMediaCollection(Attachment::MEDIALIBRARY_COLLECTIONNAME);
} catch (UnreachableUrl $th) {
    report($th);
}
```

It’s possible that the exception `UnreachableUrl` from the spatie/laravel-medialibrary package played a role in this:
https://github.com/spatie/laravel-medialibrary/blob/main/src/MediaCollections/Exceptions/UnreachableUrl.php

So far, everything seems to work fine with your fix—thanks again!

PR #1867, opened 2025-06-25
Add Monolog Sentry Logs handler

stayallive:


stayallive:
I don't think cursor understands the monstrosity that the `CompatibilityLogLevelTrait` file is 🤣

I believe the things it says are correct (for one version) but also wrong (for the other) because of the switch we make based on the Monolog API version.

cleptric (approved):

PR #1870, opened 2025-07-21
Update the DSC’s sample_rate if a traces_sampler is used

cleptric:


cleptric:
@stayallive added a test case as well

PR #1873, opened 2025-08-08
Fix trait name typos

cleptric:
@stayallive though it's marked `@internal` we might want to wait for v5? Not sure, not a big deal in any case 😄


stayallive:
Since it's internal I don't mind this. We can just document this as a misc to let people know. Needs a minor version though just in case. We aren't using these in Symfony/Laravel.

PR #1876, opened 2025-08-11
Refactor trace header parsing logic

cleptric:
A bit overdue to clean this up.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -311,6 +311,11 @@ parameters:
 			path: src/Tracing/GuzzleTracingMiddleware.php
 
 		-
+			message: "#^Property Sentry\\\\Tracing\\\\PropagationContext\\:\\:\\$parentSampled is never read, only written\\.$#"
+			count: 1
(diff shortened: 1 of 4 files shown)


phpstan-baseline.neon
```diff
@@ -310,6 +310,11 @@ parameters:
 			count: 1
 			path: src/Tracing/GuzzleTracingMiddleware.php
 
+		-
+			message: "#^Property Sentry\\\\Tracing\\\\PropagationContext\\:\\:\\$parentSampled is never read, only written\\.$#"
+			count: 1
+			path: src/Tracing/PropagationContext.php
+
```
cleptric, line 317: Didn't find an elegant way to fix this

Litarnus (approved):

Litarnus:
Can we use it somewhere else too? If not it might be more explicit to have a regular class instead of a Trait

cleptric:
We use it in two places. Kinda want to avoid sub-classing. Though we could also make it static tbh.

cleptric:
Ahh, nvm, we can't.

PR #1890, opened 2025-09-04
Remove `max_breadcrumbs` limit

FeBe95:
## Description

This PR **removes** the limit for the `max_breadcrumbs` option, which was previously set to `100`.

## Reasoning
The [documentation](https://docs.sentry.io/platforms/php/configuration/options/#max_breadcrumbs) clearly states there isn't a hard upper limit, but rather a soft limit imposed due to the maximum payload size.

> ### max_breadcrumbs
> | Type | int |
> | --- | --- |
> | Default | 100 |
>
> This variable controls the total amount of breadcrumbs that should be captured. This defaults to `100`, but you can set this to any number. However, you should be aware that Sentry has a [maximum payload size](https://develop.sentry.dev/sdk/data-model/envelopes/#size-limits) and any events exceeding that payload size will be dropped.

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -1336,16 +1336,9 @@ final class Options
 
         $resolver->setAllowedValues('max_request_body_size', ['none', 'never', 'small', 'medium', 'always']);
         $resolver->setAllowedValues('dsn', \Closure::fromCallable([$this, 'validateDsnOption']));
+        $resolver->setAllowedValues('max_breadcrumbs', \Closure::fromCallable([$this, 'validateMaxBreadcrumbsOptions']));
         $resolver->setAllowedValues('class_serializers', \Closure::fromCallable([$this, 'validateClassSerializersOption']));
-        
-        $resolver->setAllowedValues('sample_rate', fn ($value) => ($value >= 0 && $value <= 1));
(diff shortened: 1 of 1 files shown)


Jean85:
Not putting a limit on breadcrumbs led to memory leaks in the past, especially in long running processes (like a queue consumer, i.e. Symfony Messenger' `message:consume` command).

I would advise against not having a default limit.

FeBe95:
I haven't thought about memory issues yet. Thanks for bringing that up! However, I believe that "blocking" users from increasing this value to `100+` is not ideal. I am maintaining a project that heavily uses Sentry's breadcrumbs feature. We are hitting the default limit surprisingly often.

I guess, we could warn about potential memory issues with an increased limit in the docs. What do you think?

--- 

P.S. The "bug" that the cursor bot mentioned is an intentional change, to keep it consistent with other similar options, e.g. `max_value_length`. They also don't have a range check.
https://github.com/getsentry/sentry-php/blob/82dca2ee361cd442afe363721e9e4bc8830ee801/src/Options.php#L1323

cleptric:
I think allowing to send more breadcrumbs should be mostly fine. I would however recommend to maybe consider switching to [logs](https://docs.sentry.io/platforms/php/) in the long run.

Issue #1902, opened 2025-09-14
LogsAggregator tries to log to the SDK logger using the Sentry log level rather than a PSR-3 log level


mfb:
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

4.15.2

### Steps to reproduce

This line of code appears in LogsAggregator.php:

```php
$sdkLogger->log((string) $log->getLevel(), "Logs item: {$log->getBody()}", $log->attributes()->toSimpleArray());
```

This causes the Sentry logs item log level, which might be something like "fatal", "trace" or "warn", to be passed to the SDK logger, which on the other hand would be expecting a PSR-3 log level (i.e. not "fatal", "trace" or "warn", although "debug", "error" and "info" do overlap).

If you want to use the Sentry log level with the SDK logger, you would have to translate it to a PSR-3 log level. So far, there might only be a translation mapping going in the opposite direction, from PSR-3 log level to Sentry log level, so you might have to create a new mapping. Or you could just log all items with a generic level such as "info".

I noticed this issue because unfortunately, there are some loggers out there that actually throw an exception when an unrecognized log level is received, such as https://github.com/consolidation/log/blob/3.x/src/Logger.php#L158   But even if non-standard log levels can be logged, it's good practice to use the PSR-3 log levels with PSR-3 loggers.

### Expected result

Any logs item can be logged with an SDK logger active.

### Actual result

"debug", "error" and "info" logs items can be logged with an SDK logger active, but "fatal", "trace" and "warn" logs items are logged to the SDK logger with a non-standard log level, which could even throw an exception, depending on the particular logger in use.

Litarnus:
Hi @mfb! Thanks for reporting this, I will have a look

PR #1907, opened 2025-09-15
fix(logs): use PSR log level when using PSR-3 logger

Litarnus:
Fixes an issue where we use a PSR-3 logger in the `LogsAggregator` but pass Sentry `LogLevel` into it, which might lead to errors in different loggers as they expect PSR-3 levels.

Closes https://github.com/getsentry/sentry-php/issues/1902

### DIFF
--- a/src/Logs/LogsAggregator.php
+++ b/src/Logs/LogsAggregator.php
@@ -147,7 +147,6 @@ final class LogsAggregator
             return;
         }
 
-        // We check if it's a `LogsLogger` to avoid a infinite loop where the logger is logging the logs it's writing
         if ($sdkLogger !== null) {
             $sdkLogger->log($log->getPsrLevel(), "Logs item: {$log->getBody()}", $log->attributes()->toSimpleArray());
         }


src/Logs/LogLevel.php
```diff
@@ -70,6 +70,24 @@ public function getPriority(): int
...
+                return \Psr\Log\LogLevel::DEBUG;
+            case 'warn':
+                return \Psr\Log\LogLevel::WARNING;
+            case 'error':
+                return \Psr\Log\LogLevel::ERROR;
+            case 'fatal':
+                return \Psr\Log\LogLevel::CRITICAL;
+            case 'info':
+            default:
+                return \Psr\Log\LogLevel::INFO;
```
Litarnus, line 87: Not sure what's a good default level here but also this shouldn't really occur

mfb:
The one existential question I have is what does it mean if Sentry is logging a fatal error successfully? Does that mean Sentry is working normally - i.e. it's not really e.g. critical? :)

cleptric:
@mfb could you clarify your message above, as it raises some questions we do not fully understand.

mfb:
If you look at HttpTransport, it logs at info level when an item is sent (whether debug or fatal); warning if rate limit is hit; error if sending fails.  I was referring to how you're using a different philosophical approach with SDK logging for log items, by choosing a level based on the payload rather than how the SDK is functioning.

src/Logs/LogsAggregator.php
```diff
@@ -149,7 +149,7 @@ public function add(
 
         // We check if it's a `LogsLogger` to avoid a infinite loop where the logger is logging the logs it's writing
         if ($sdkLogger !== null) {
-            $sdkLogger->log((string) $log->getLevel(), "Logs item: {$log->getBody()}", $log->attributes()->toSimpleArray());
+            $sdkLogger->log($log->getPsrLevel(), "Logs item: {$log->getBody()}", $log->attributes()->toSimpleArray());
```
stayallive, line 152: What (I think) @mfb points out is that we copy the log level from the log item instead of logging always as "DEBUG" or "INFO" and mentioning the log level of the log item in the log message.

This was introduced by me I think, this maybe should be updated to mention the log item level in the log message and set our log level to debug (?).

In addition, the comment above this `if` is outdated and incorrect :) The `LogsLogger` never made it in but the comment remains 🙈

stayallive: In that case I did this very by the book 😎 

Then maybe the comment could be updated but no other changes are needed.

mfb: Ok, great - as long as it's by design :)

It's not 100% clear to me but I guess "appropriate" is saying the Sentry log level should be translated to an appropriate level, e.g. log a "fatal" to the console as "critical".  I'd also find it reasonable to just print the "fatal" level as part of an "info" log message - either way works for me.

stayallive (approved):

Issue #1896, opened 2025-09-11
[5.x] Remove `symfony/options-resolver`


cleptric:
In order to further reduce dependencies in the SDK, we'll remove the `symfony/options-resolver` package and replace it with our own, lightweight implementation.

- [ ] Allow to set defaults for an option
- [ ] Do not throw exceptions for none-existing options but instead emit debug logs
- [ ] Make sure we do not need to ignore as many phpstan errors as today

PR #1914, opened 2025-09-17
ref: replace symfony options resolver with custom resolver

Litarnus:
### Description
This PR replaces `symfony/options-resolver` with a custom option resolver that acts almost as a drop in replacement, but comes with a reduced feature set.
Notable changes between symfony and our resolver:
* Invalid values will no longer throw exceptions. Instead, if will fall back to the default value and produce a log message.
* The defaults will act as a schema, meaning that added values that have no default value will be ignored
* Values will be validated before and after normalization. The reason is so that normalizers will not be called with invalid values, removing the need to validate within normalizers.
* Defaults should be specified after declaring validation in order to get validated and normalized.

~~I also had to replace usages of PHP 8+ features since they used a polyfill which was pulled in through the `symfony/options-resolver`.~~ Was done in https://github.com/getsentry/sentry-php/pull/1922

#### Issues
resolves https://github.com/getsentry/sentry-php/issues/1896
closes PHP-24

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,6 @@
         "phpbench/phpbench": "^1.0",
         "phpstan/phpstan": "^1.3",
         "phpunit/phpunit": "^8.5|^9.6",
-        "symfony/phpunit-bridge": "^5.2|^6.0|^7.0",
         "vimeo/psalm": "^4.17"
     },
     "suggest": {
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -10,6 +10,7 @@ parameters:
     excludePaths:
         - tests/resources
         - tests/Fixtures
+        - src/Util/ClockMock.php
     dynamicConstantNames:
         - Monolog\Logger::API
     bootstrapFiles:
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -34,10 +34,6 @@
         </whitelist>
     </filter>
 
-    <listeners>
-        <listener class="\Symfony\Bridge\PhpUnit\SymfonyTestsListener" />
-    </listeners>
-
     <extensions>
         <extension class="Sentry\Tests\SentrySdkExtension" />
     </extensions>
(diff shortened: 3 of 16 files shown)


src/Options.php
```diff
@@ -1342,4 +1166,37 @@ private function validateContextLinesOption(?int $contextLines): bool
...
+        return $this->resolver->resolve($options, $logger);
+    }
+
+    /**
+     * Merges the passed options with the current options and resolves them.
+     * The result is stored back onto the class field.
+     *
+     * @param array<string, mixed> $override
+     */
+    private function updateOptions(array $override = []): self
```
stayallive, line 1194: We might need to make this public (and `@internal`) if we end up having a use for it in the SDKs. But `private` is fine for now.

src/Options.php
```diff
@@ -1342,4 +1166,37 @@ private function validateContextLinesOption(?int $contextLines): bool
...
+    {
+        $options = array_merge($this->options, $override);
+
+        $this->options = $this->resolveWithLogger($options);
```
stayallive, line 1198: My only concern is that this is fairly expensive to do for updating 2 or 3 options. Can we pass the overriden options to the resolver and have it only resolve those (new) keys and merge that with `$this->options` instead? Like:


```suggestion
        $this->options = array_merge($this->options, $this->resolver->resolveOnly($options));
```

I don't know if we use setters internally anywhere but I know users do and keeping this as cheap as possible seems like a good idea.

src/OptionsResolver.php
```diff
@@ -0,0 +1,266 @@
...
+    {
+        $processed = [];
+
+        foreach ($defaults as $option => $defaultValue) {
+            [$isValid, $normalized] = $this->normalizeAndValidate($option, $defaultValue);
+
+            if (!$isValid) {
+                // Since defaults are used as fallback values if passed options are invalid, we want to
+                // get hard errors here to make sure we have something to fall back to.
+                throw new \InvalidArgumentException(\sprintf('Invalid default for option "%s"', $option));
```
stayallive, line 53: Good use case for an exception since this is an "us" problem and not caused by user input.

stayallive (commented):
This looks pretty good to me, I have a few nits but nothing major. Code looks simple enough but still powerful validations present.

PR #1929, opened 2025-09-29
ref(logs): extract attribute compilation logic in Monolog LogsHandler

huangdijia:
## Summary

• Extracted the attribute compilation logic from the `handle()` method into a dedicated `compileAttributes()` method
• Improves code readability and maintainability by separating concerns
• Makes the attribute compilation logic reusable for future enhancements

## Changes

- Added protected method `compileAttributes(LogRecord $record): array`
- Refactored `handle()` method to use the new method instead of inline array_merge
- No functional changes - this is purely a code organization improvement

## Test plan

- [x] Existing tests should continue to pass as this is a refactoring with no functional changes
- [x] The Monolog handler should continue to work exactly as before
- [x] The `sentry.origin` attribute should still be properly added to log records

### DIFF
--- a/src/Monolog/LogsHandler.php
+++ b/src/Monolog/LogsHandler.php
@@ -66,7 +66,7 @@ class LogsHandler implements HandlerInterface
             self::getSentryLogLevelFromMonologLevel($record['level']),
             $record['message'],
             [],
-            array_merge($record['context'], $record['extra'], ['sentry.origin' => 'auto.logger.monolog'])
+            $this->compileAttributes($record)
         );
 
         return $this->bubble === false;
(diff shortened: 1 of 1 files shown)


cleptric:
Thanks, but the code is fine as is.

huangdijia:
> Thanks, but the code is fine as is.

I hope to support custom attributes combination, like: https://github.com/friendsofhyperf/components/blob/9b223569b9ade662f3c811138deb635746f0538c/src/sentry/src/Monolog/LogsHandler.php#L53-L64

cleptric:
It might be worth adding such "important" information in the PR description then.

huangdijia:
> It might be worth adding such "important" information in the PR description then.

That’s alright, I can handle it by extending a subclass.

huangdijia:
In fact, the current design is risky. For example:


```php
$record['context'] = ['foo' => 'foo'];
$record['extra'] = ['foo' => 'bar'];
$result = array_merge($record['context'], $record['extra']);

output: ['foo' => 'bar']
```

cleptric:
Would you mind restoring the PR? With the added context, it makes sense, and we can get this in.

huangdijia:
> Would you mind restoring the PR? With the added context, it makes sense, and we can get this in.


But I have already deleted the branch. Can this PR be restored?

PR #1930, opened 2025-09-29
ref: remove phpunit bridge

Litarnus:
> [!NOTE]
> Replaces Symfony PHPUnit Bridge/ClockMock with a new internal `Sentry\Util\ClockMock`, updates tests accordingly, and adjusts static analysis and PHPUnit config.
> 
> - **Testing**:
>   - Replace `Symfony\Bridge\PhpUnit\ClockMock` with internal `Sentry\Util\ClockMock` across tests and `tests/bootstrap.php` (adds additional `ClockMock::register(...)`, incl. `Sentry\Serializer\PayloadSerializer`).
>   - Remove `Symfony\Bridge\PhpUnit\ExpectDeprecationTrait` usages from tests.
> - **New Utility**:
>   - Add `src/Util/ClockMock.php` (ported clock mocking functions) used by tests.
> - **Config**:
>   - Remove `symfony/phpunit-bridge` from `require-dev` and drop PHPUnit `<listeners>` from `phpunit.xml.dist`.
>   - Exclude `src/Util/ClockMock.php` from analyzers (`phpstan.neon` excludePaths, `psalm.xml.dist` ignoreFiles).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ee9440ab4a6d31931665b0c1aabe14302a6d8736. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

Removes the `symfony/phpunit-bridge` dependency in favour of a vendored `ClockMock` class from said dependency. The motivation behind this is that a change in the bridge broke our pipeline and we only really use the `ClockMock` from the package.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,6 @@
         "phpbench/phpbench": "^1.0",
         "phpstan/phpstan": "^1.3",
         "phpunit/phpunit": "^8.5|^9.6",
-        "symfony/phpunit-bridge": "^5.2|^6.0|^7.0",
         "vimeo/psalm": "^4.17"
     },
     "suggest": {
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -10,6 +10,7 @@ parameters:
     excludePaths:
         - tests/resources
         - tests/Fixtures
+        - src/Util/ClockMock.php
     dynamicConstantNames:
         - Monolog\Logger::API
     bootstrapFiles:
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -34,9 +34,9 @@
         </whitelist>
     </filter>
 
-    <listeners>
-        <listener class="\Symfony\Bridge\PhpUnit\SymfonyTestsListener" />
-    </listeners>
+<!--    <listeners>-->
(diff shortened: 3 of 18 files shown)


Jean85:
The bridge is needed for the ClockMock. Also, it's useful to detect deprecations, but in newer PHPUnit versions (10+) that could be handled natively.

cleptric:
Our entire CI broke because of changes in the bridge. We can vendor in the ClockMock and move on.

Litarnus:
> Also, it's useful to detect deprecations

As far as I can see we don't use this

Jean85:
> > Also, it's useful to detect deprecations
> 
> As far as I can see we don't use this

You do: https://github.com/getsentry/sentry-php/blob/1895ab6525292cc75fefbb37680f5b693e5af49a/phpunit.xml.dist#L10-L12
See docs for usage: https://symfony.com/doc/current/components/phpunit_bridge.html#making-tests-fail

Since you still need to support older PHP versions, doing otherwise with PHPUnit 10+ is not a viable solution.

cleptric (approved):

Issue #1946, opened 2025-10-23
Function curl_close() is deprecated since 8.5, as it has no effect since PHP 8.0


ruudk:
### How do you use Sentry?

Sentry SaaS (sentry.io)

### SDK version

latest

### Steps to reproduce

Run on PHP 8.5 and you see this deprecation.

### Expected result

Nothing.

### Actual result

```
Function curl_close() is deprecated since 8.5, as it has no effect since PHP 8.0
```

PR #1947, opened 2025-10-23
fix: Only call curl_close() in PHP < 8

jamieburchell:
### Description
`curl_close()` has had no effect since PHP 8.0 and is deprecated in PHP 8.5. This patch ensures that `curl_close()` is only called in PHP < 8

#### Issues
resolves: https://github.com/getsentry/sentry-php/issues/1946


ruudk:
Thank you ❤️ 

Will this be automatically tagged ahead of the PHP 8.5 release? 🙏

Litarnus:
Hi @ruudk! We just released version `4.17.1` that includes this fix

PR #1950, opened 2025-10-28
feat: try to represent any logs attribute as string

stayallive:
We should try (harder) to represent values the user provides us as log attributes. So we now as last resort JSON encode the value which results in almost any value to be serialized and correctly logged. Also `null` values are now serialized as string instead of discarded.

### DIFF
--- a/tests/Attributes/AttributeTest.php
+++ b/tests/Attributes/AttributeTest.php
@@ -77,6 +77,14 @@ final class AttributeTest extends TestCase
         ];
 
         yield [
+            null,
+            [
+                'type' => 'string',
+                'value' => 'null',
+            ],
+        ];
+
+        yield [
             new class {
                 public function __toString(): string
                 {


src/Attributes/Attribute.php
```diff
@@ -83,6 +87,14 @@ public static function tryFromValue($value): ?self
             return new self($value, 'double');
         }
 
+        if ($value instanceof SerializableInterface) {
+            try {
+                return new self(JSON::encode($value->toSentry()), 'string');
+            } catch (\Throwable $e) {
+                // Ignore the exception and continue trying other methods
+            }
+        }
```
stayallive: Because we have it, I prefered this over the __toString or just raw JSON encode.

For safety we do ignore any errors to prevent crashing because serializing fails.

tests/Attributes/AttributeTest.php
```diff
@@ -112,6 +135,7 @@ public function testFromValueFactoryMethod(): void
     {
         $this->expectException(\InvalidArgumentException::class);
 
-        Attribute::fromValue([]);
+        // Since we support almost any type, we use a resource to trigger the exception
+        Attribute::fromValue(fopen(__FILE__, 'r'));
```
stayallive: It's now a challenge to find something we can't represent as an attribute.

Litarnus (approved):
nice! :shipit:

PR #1968, opened 2025-11-19
feat(metrics): introduce metrics

Litarnus:
Introducing metrics for PHP!

Metrics have 3 supported types
* Counter
* Gauge
* Distribution

Each of those metrics can be submitted by calling
`\Sentry\metrics()->count('test-counter', 1, ['tag' => 'foo'])`

It's also possible to attach units:
`\Sentry\metrics()->gauge('test-gauge', 10, ['tag' => 'foo'], \Sentry\Metrics\Unit::second())`

Up to 1000 metrics can be stored per request. Older metrics are dropped when the limit is reached.

Metrics need to be flushed manually by calling `\Sentry\metrics()->flush()`.

### DIFF
--- a/src/Metrics/Metrics.php
+++ b/src/Metrics/Metrics.php
@@ -6,8 +6,8 @@ namespace Sentry\Metrics;
 
 use Sentry\EventId;
 use Sentry\Tracing\SpanContext;
-
 use Sentry\Unit;
+
 use function Sentry\trace;
 
 class_alias(Unit::class, '\Sentry\Metrics\MetricsUnit');
--- a/src/Metrics/MetricsAggregator.php
+++ b/src/Metrics/MetricsAggregator.php
@@ -13,6 +13,7 @@ use Sentry\Metrics\Types\GaugeMetric;
 use Sentry\Metrics\Types\Metric;
 use Sentry\SentrySdk;
 use Sentry\State\Scope;
+use Sentry\Unit;
 use Sentry\Util\RingBuffer;
 
 /**
--- a/src/Metrics/Types/DistributionMetric.php
+++ b/src/Metrics/Types/DistributionMetric.php
@@ -4,9 +4,9 @@ declare(strict_types=1);
 
 namespace Sentry\Metrics\Types;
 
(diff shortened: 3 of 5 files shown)


src/Metrics/Metrics.php
```diff
@@ -28,58 +41,82 @@ public static function getInstance(): self
...
      * @param array<string, string> $tags
      *
-     * @deprecated Metrics are no longer supported. Metrics API is a no-op and will be removed in 5.x.
+     * @deprecated Use Metrics::count() instead. To be removed in 5.x.
      */
     public function increment(
         string $key,
         float $value,
-        ?MetricsUnit $unit = null,
+        ?Unit $unit = null,
```
cleptric:
```suggestion
        ?MetricsUnit $unit = null,
```

stayallive (commented):
Code wise the looks good (especially much love for the RingBuffer). 

The only thing I'm worried about a bit is that we modify the existing code, namely `MetricsUnit` and some parameters which is a backwards incompatible break for people that did not yet cleanup their code after we deprecated the previous metrics iteration. Do we feel like that is an acceptable risk? If this is going in 5.x this is a non-issue but the target is master so hence my ask.

Litarnus:
> Code wise the looks good (especially much love for the RingBuffer).
> 
> The only thing I'm worried about a bit is that we modify the existing code, namely `MetricsUnit` and some parameters which is a backwards incompatible break for people that did not yet cleanup their code after we deprecated the previous metrics iteration. Do we feel like that is an acceptable risk? If this is going in 5.x this is a non-issue but the target is master so hence my ask.

Initially we wanted to take the risk but after a longer thought we decided against it. I will revert the `Metrics` class back to how it originally was and will introduce a `TraceMetrics` class with the new methods. 
I will also add a new alias for `Unit` -> `MetricsUnit` since we want to use the unit in different places too.

Thanks for your review, I make this a draft again and ping you once it's updated!

stayallive (approved):
Code wise this looks good to me, I haven't dug into the spec much yet but I assume you and Michi got that covered pretty well already!

PR #1970, opened 2025-11-19
ref(rate-limit): handle profile and check_in rate limiting

Litarnus:
Adds rate limiting for `profile` and adds mapping from `check_in` => `monitor` to properly handle `check_in` rate limits.

Definitions for data categories are located here: https://develop.sentry.dev/sdk/expected-features/rate-limiting/#definitions

closes PHP-54

### DIFF
--- a/src/Transport/HttpTransport.php
+++ b/src/Transport/HttpTransport.php
@@ -108,7 +108,7 @@ class HttpTransport implements TransportInterface
                 // Just remove profiling data so the normal transaction can be sent.
                 $event->setSdkMetadata('profile', null);
                 $this->logger->warning(
-                    'Rate limit exceeding for sending requests of type "profile".',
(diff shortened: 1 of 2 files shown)


stayallive (approved):
Also added this as TODO to the agent. Wondering if it's worth refactoring a bit so we can have a `isRateLimited($event)` instead of doing it like this so we can re-use this logic in the agent.

Litarnus:
> Also added this as TODO to the agent. Wondering if it's worth refactoring a bit so we can have a `isRateLimited($event)` instead of doing it like this so we can re-use this logic in the agent.

I was thinking about this but profiles need to be removed from transaction and I didn't want to introduce mutation in the `RateLimiter`. 
That said we should look into refactoring to use it in both places

stayallive:
Hmm, yes that does make sense (not mutating the event in the rate limiter). Okay then let's roll with this and update this later if we find a better solution for re-use.

PR #1978, opened 2025-12-01
feat(client-reports): Add support for client reports

Litarnus:
Adds support for ClientReports.

This PR only adds the possibility to send client reports, it does not collect them yet


Litarnus:
@cursor push e7cd5c73a3

PR #1990, opened 2025-12-24
feat(scope): add getGlobalEventProcessors method

huangdijia:
## Summary

- Add `getGlobalEventProcessors()` static method to `Scope` class
- Enables testing and introspection of registered global event processors
- Refactors `applyToEvent()` to use the new getter method for consistency

## Changes

- **`src/State/Scope.php`**:
  - New public static method `getGlobalEventProcessors()` that returns the array of global event processors
  - Updated `applyToEvent()` to call `self::getGlobalEventProcessors()` instead of directly accessing `self::$globalEventProcessors`

## Motivation

This change provides a public API for accessing the global event processors, which is useful for:
- Testing framework integrations that register global processors
- Debugging and introspection of registered processors
- Future enhancements that may need to enumerate or inspect processors

## Test plan

- [x] Code follows existing coding standards
- [ ] Unit tests added/updated (if applicable)
- [x] No breaking changes to public API

### DIFF
--- a/src/State/Scope.php
+++ b/src/State/Scope.php
@@ -360,7 +360,9 @@ class Scope
     }
 
     /**
-     * Get the global event processors {@see Scope::applyToEvent}.
+     * Gets the list of global event processors that are applied in {@see Scope::applyToEvent}.
+     *
+     * @return callable[]
      */
     public static function getGlobalEventProcessors(): array
     {


huangdijia:
Additional note for long-lived framework environments:

In long-running, persistent frameworks (such as Swoole, RoadRunner, or similar setups), it is possible to leverage **AOP** at appropriate lifecycle boundaries (e.g. request start/end, coroutine creation) to capture a snapshot of `$globalEventProcessors` and store it in a request- or coroutine-scoped **Context**.

This approach provides several benefits:
- Prevents unintended cross-request or cross-coroutine state sharing caused by global static data
- Allows safe, request-level inspection, debugging, or assertions of the active global event processors
- Gives framework integrators a clearer lifecycle boundary without directly relying on static properties

Exposing `getGlobalEventProcessors()` as a read-only accessor enables this Context-based management pattern while preserving existing behavior and event processing flow.

Litarnus:
Hi @huangdijia! Thanks for opening the PR. Can you tell us the concrete reason why you would like to expose event processors. We are moving away from them so I'm reluctant to get it merged without a proper reason

huangdijia:
> Additional note for long-lived framework environments:
> 
> 
> 
> In long-running, persistent frameworks (such as Swoole, RoadRunner, or similar setups), it is possible to leverage **AOP** at appropriate lifecycle boundaries (e.g. request start/end, coroutine creation) to capture a snapshot of `$globalEventProcessors` and store it in a request- or coroutine-scoped **Context**.
> 
> 
> 
> This approach provides several benefits:
> 
> - Prevents unintended cross-request or cross-coroutine state sharing caused by global static data
> 
> - Allows safe, request-level inspection, debugging, or assertions of the active global event processors
> 
> - Gives framework integrators a clearer lifecycle boundary without directly relying on static properties
> 
> 
> 
> Exposing `getGlobalEventProcessors()` as a read-only accessor enables this Context-based management pattern while preserving existing behavior and event processing flow.

@Litarnus Look at the content quoted above.

Litarnus:
Could you explain a bit further how you intent to use this? With the currently proposed change you can read what global processors are currently stored, but without a way to override existing ones (since we just have an add method), it will not be possible to properly isolate them and it will still leak.

huangdijia:
I will change the addGlobalProcess method through AOP to save global processes in the context, and also change the getGlobalProcesses method through AOP to obtain global processes from the context.

Litarnus:
@huangdijia Since global event processors are going to be removed in the next major, I would like to avoid exposing them now. It could risk that someone starts using it now and will have to change it in the near future

PR #2000, opened 2026-01-27
feat(scopes): replace Hub with new scopes

Litarnus:
This PR replaces the Hub based approach to the new scope based approach. The main difference is that instead of having one Hub with multiple Scope layers, data is now divided into different scopes, namely GlobalScope, IsolationScope and CurrentScope. Further details are documented here: https://develop.sentry.dev/sdk/telemetry/scopes/


Litarnus:
This PR is already very stale and we will move forward with a simpler scopes concept that only has the global scope + isolation scope

PR #1875, opened 2025-08-11
Implement strict trace continuation for Sentry PHP

cleptric:
Adds `strictTraceContinuation` option to prevent unwanted trace propagation from third-party services with mismatched organization IDs, addressing #1830.

---
<a href="https://cursor.com/background-agent?bcId=bc-5874fd54-ddaa-4df9-b80d-6debd9f6e96f">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-cursor-dark.svg">
    <source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-cursor-light.svg">
    Open in Cursor
  </picture>
</a>
<a href="https://cursor.com/agents?id=bc-5874fd54-ddaa-4df9-b80d-6debd9f6e96f">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/open-in-web-dark.svg">
    <source media="(prefers-color-scheme: light)" srcset="https://cursor.com/open-in-web-light.svg">
    Open in Web
  </picture>
</a>

PR #2016, opened 2026-02-25
feat: add `strict_trace_continuation` support

Litarnus:
Introduces the `strict_trace_continuation`, which can be used to make sure that traces are only continued if the org IDs are matching. By default, this is set to `false` to keep current behaviour intact.

supersedes https://github.com/getsentry/sentry-php/pull/1875

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -789,6 +789,7 @@ final class Options
          * @var bool $result
          */
         $result = $this->options['strict_trace_continuation'];
+
         return $result;
     }
 


src/Tracing/Traits/TraceHeaderParserTrait.php
```diff
@@ -102,4 +111,51 @@ protected static function parseTraceAndBaggageHeaders(string $sentryTrace, strin
...
+
+        if ($client === null) {
+            return true;
+        }
+
+        $options = $client->getOptions();
+        $clientOrgId = $options->getOrgId();
+        if ($clientOrgId === null && $options->getDsn() !== null) {
+            $clientOrgId = $options->getDsn()->getOrgId();
+        }
```
giortzisg: nit: The name here is a bit confusing if we get the orgID from the DSN. clientOrgID is the one on clientOptions and the other is the DSN one.

PR #2018, opened 2026-03-02
Use sub-second timestamps in logs

martinstuecklschwaiger:
Sentry UI shows timestamps as `Feb 18, 17:48:27.000 CET` (milliseconds as `000`).

### Description
Breadcrumb timestamps use floats. Monolog also supports sub-second timestamps from DateTimeImmutable. This pull request uses split-second timestamps for creating Breadcrumbs in the Monolog BreadcrumbHandler.

* `U`: Seconds since the Unix Epoch  https://www.php.net/manual/en/datetime.format.php#:~:text=01%3A07%20%2B0200-,U,Seconds%20since%20the%20Unix%20Epoch,-(January%201%201970
* `u`: Microseconds
https://www.php.net/manual/en/datetime.format.php#:~:text=u,Microseconds

### DIFF
--- a/tests/Monolog/BreadcrumbHandlerTest.php
+++ b/tests/Monolog/BreadcrumbHandlerTest.php
@@ -15,6 +15,8 @@ final class BreadcrumbHandlerTest extends TestCase
 {
     /**
      * @dataProvider handleDataProvider
+     *
+     * @param LogRecord|array<string, mixed> $record
      */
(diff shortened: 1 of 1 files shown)


Litarnus:
Hi @martinstuecklschwaiger! Thanks for opening the PR!

It looks good to me. Since you added a fix for pre-1970 timestamps, can you also add a quick test for that?

Issue #1898, opened 2025-09-11
[5.x] Remove Psalm


cleptric:
We constantly run into issues where Psalm is fighting with PHPStan. We will remove psalm and only use PHPStan.

PR #2020, opened 2026-03-04
feat(ci): add mago to CI pipeline

Litarnus:
Introducing [Mago](https://mago.carthage.software) to the PHP SDK. For the initial version, I added all current failures to the baseline file to keep the PR small regarding code changes.

closes PHP-69


stayallive:
Worth seeing this as an opportunity to remove psalm (#1898) to prevent running three tools and see how it works before we commit? Ideally we have a single tool instead of many to prevent them disagreeing with each other.

Litarnus:
> Worth seeing this as an opportunity to remove psalm (#1898) to prevent running three tools and see how it works before we commit? 

Good point, I make sure to remove psalm before this gets merged. I think we have to have Mago running for some time in the repo to see how well it works.

> Ideally we have a single tool instead of many to prevent them disagreeing with each other.

Yes, that is my goal as well. So far it seems quite powerful and I'm fairly optimistic it can replace all other tools (including PHP-CS-FIXER)

giortzisg (approved):

PR #2029, opened 2026-03-17
feat(logs): Add Monolog channel to compiled log attributes

melkamar:
### Description

Adds `channel` name to the log metadata. 

The channel is a relevant piece of metadata of a log entry. This change includes the channel in the recorded attributes, but remains backwards compatible and will not overwrite existing extra or context fields named 'channel'.

I considered making this an opt-in configuration of the Handler in the constructor, but since this is backwards compatible, I figured that's not necessary.

### DIFF
--- a/src/Monolog/LogsHandler.php
+++ b/src/Monolog/LogsHandler.php
@@ -30,15 +30,22 @@ class LogsHandler implements HandlerInterface
     private $bubble;
 
     /**
+     * Whether to include the channel name as an attribute in the Sentry logs.
+     */
+    private bool $includeChannel;
+
+    /**
      * Creates a new Monolog handler that converts Monolog logs to Sentry logs.
      *
      * @param LogLevel|\Monolog\Level|int|null $logLevel the minimum logging level at which this handler will be triggered and collects the logs
      * @param bool                             $bubble   whether the messages that are handled can bubble up the stack or not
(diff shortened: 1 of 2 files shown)


Litarnus:
Hi @melkamar! Can you tell us why you need the channel name in the log attributes? Adding it to every log message would increase the number of bytes per message, so we are a bit reluctant to add it without proper reason

melkamar:
@Litarnus sure, the reason is that I may want to filter for messages sent to a certain channel. For example, if some troubleshooting issues related to async processing, I may want to narrow my search to the `messenger` channel only to see the Symfony messenger logs.

In our application, we use dozens of different channels. I may want to drill down to a specific one. That's what we currently do in Elastic/Kibana and it's very useful. Having the same in Sentry, along with all the other app data, would be great.

I do understand the reluctance of adding data. I also understand that this is truly pointless for use cases where someone uses a single channel (e.g. `app`) to send the logs to. So, could this be an opt-in?

Litarnus:
Thanks for the explanation, I can see that it makes a lot of sense to have that option in your setup.

> So, could this be an opt-in?

Yes, I think that would be best. I'm happy to review the PR if you change it to opt-in

melkamar:
@Litarnus Thanks, I have pushed an update that makes this opt-in

PR #2061, opened 2026-04-09
feat(monolog): add handler to only capture exceptions

### LINKED ISSUE #2060: Add handler to send Monolog exceptions to Sentry
We want to add a new handler that will send Monolog exceptions to Sentry. This functionality already exists in the now deprecated `Handler`, but we want to create a new handler with just this single functionality.

Litarnus:
Adds a Monolog handler that will report exceptions to Sentry. In contrast to the `Handler`, it will not convert log messages with a certain level to errors.

By default it will report exceptions for all levels but it's possible to also restrict it to certain levels.

### DIFF
--- a/src/Monolog/SentryExceptionHandler.php
+++ b/src/Monolog/SentryExceptionHandler.php
@@ -35,23 +35,11 @@ class SentryExceptionHandler extends AbstractHandler
     /**
      * @param array<string, mixed>|LogRecord $record
      */
-    public function isHandling($record): bool
-    {
-        if ($this->getExceptionFromRecord($record) === null) {
-            return false;
-        }
-
-        /** @var LogRecord $record */
-        return parent::isHandling($record);
-    }
-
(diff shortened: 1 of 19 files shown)


src/Monolog/SentryExceptionHandler.php
```diff
@@ -0,0 +1,141 @@
...
+use Monolog\Logger;
+use Monolog\LogRecord;
+use Psr\Log\LogLevel;
+use Sentry\State\HubInterface;
+use Sentry\State\Scope;
+
+/**
+ * This Monolog handler will collect monolog events and send them to sentry.
+ */
+class SentryExceptionHandler extends AbstractHandler
```
Litarnus: 🚲 : Not sure about the name, it's a bit redundant and generic so I'm up for suggestions

stayallive (approved):
🚢

PR #2063, opened 2026-04-10
fix(error): prevent warnings when trying to increase memory for OOM errors

Litarnus:
Adds a defensive check to only increase the memory if it would still be below the currently consumed memory. Also prevents warnings from setting `memory_limit` which might end up in Sentry.

We could consider trying to increase the limit when OOMing to `current_memory + 5MB` instead


stayallive (approved):
Cool fix!

PR #2075, opened 2026-04-28
feat: add handler to convert logs to sentry issues

### LINKED ISSUE #2069: Add handler to capture log lines and send them to Sentry as Issues
### Problem Statement

We want to deprecate `Monolog\Handler` but this means that converting log messages with a certain level to Sentry Issues will no longer work. For people using Sentry logs, this might be a non issue but if someone is not using Logs, the removal feels like they have to do more work to set up alerts and get a similar functionality.

### Solution Brainstorm

Add a handler that just converts log messages with a certain level to issues. Give it a proper name like `LogToSentryIssueHandler` so that it's clear it only does one thing. Avoid adding more functionality and generic names to not repeat the mistake of `Monolog\Handler`.

If `exception -> issue` and `log -> issue` is desired, the solution would be to have two handlers, each handling their respective part

Origin: https://github.com/getsentry/sentry-php/pull/2051#issuecomment-4236397138

Litarnus:
Adds a monolog handler to convert log messages with a configurable severity to Sentry Issues. This acts as a partial replacement for the deprecated `Monolog\Handler` in an effort to split responsibilities properly.

### DIFF
--- a/src/Monolog/Handler.php
+++ b/src/Monolog/Handler.php
@@ -17,8 +17,9 @@ use Sentry\State\Scope;
  * hub instance.
  *
  * @deprecated since version 4.24. To be removed in version 5.0. Use {@see LogsHandler}
- *             with the `enable_logs` SDK option instead for logging. {@see ExceptionToSentryIssueHandler}
- *             to send monolog exceptions to Sentry.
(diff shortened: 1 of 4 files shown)


stayallive (approved):
🥳

PR #2085, opened 2026-05-07
ref: add CodeLocationResolver

Litarnus:
Adds `CodeLocationResolver` into the PHP SDK so that it can be used by the depending SDKs. Currently SQL code location is only available in Laravel and moving this class here makes it available for Symfony too without having to duplicate code.

The code is mostly extracted from https://github.com/getsentry/sentry-laravel/blob/54cb822e74ffbc9a6475cfc0ba1ef677b12d0f0f/src/Sentry/Laravel/Tracing/BacktraceHelper.php#L14 and the Laravel part will be replaced by this in the future

### DIFF
--- a/analysis-baseline.toml
+++ b/analysis-baseline.toml
@@ -439,18 +439,6 @@ message = "Unreachable else clause"
 count = 1
 
 [[issues]]
-file = "src/Integration/ModulesIntegration.php"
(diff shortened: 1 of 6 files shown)


stayallive (approved):
Good plan!

PR #2111, opened 2026-06-09
ci: pin mago and generate baseline to prevent random failures

Litarnus:
Pins `carthage-software/mago` to the newest version and updates the baseline. The reason for this change is that once a new version is released, it might introduce new failures to the CI job. While it's generally good to discover new potential issues, it's annoying if it happens in the middle of completely different changes.


stayallive (approved):
Agreed!

PR #2122, opened 2026-06-15
feat(scope): remove Hub from Transaction

Litarnus:
This PR removes the `Hub` from the transaction and instead replaces it with `ClientInterface` in some instances.

### DIFF
--- a/src/Tracing/Transaction.php
+++ b/src/Tracing/Transaction.php
@@ -9,18 +9,12 @@ use Sentry\EventId;
 use Sentry\Options;
 use Sentry\Profiling\Profiler;
 use Sentry\SentrySdk;
-use Sentry\State\HubInterface;
 
 /**
  * This class stores all the information about a Transaction.
  */
 final class Transaction extends Span
 {
-    /**
-     * @var HubInterface The hub instance
-     */
-    private $hub;
-
     /**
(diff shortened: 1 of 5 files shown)


src/Tracing/Transaction.php
```diff
@@ -45,15 +39,13 @@ final class Transaction extends Span
...
-     * @param HubInterface|null  $hub     Instance of a hub to flush the transaction
      *
      * @internal
      */
-    public function __construct(TransactionContext $context, ?HubInterface $hub = null)
+    public function __construct(TransactionContext $context)
     {
         parent::__construct($context);
 
-        $this->hub = $hub ?? SentrySdk::getCurrentHub();
```
stayallive, line 56: Does it now make sense that this transaction is able to "switch clients" I feel like the AI comments try to point this out with a lot of words but that seems odd that if the global client changes for whatever reason a transaction will be finished on it where it wasn't started. Client or scope. Maybe it should get the scope instead of the hub and hold a ref to it?

stayallive (approved):

src/Tracing/Transaction.php
Litarnus: True that's a good point, I will update the signature to allow a (isolation) Scope

PR #2132, opened 2026-06-23
feat(scope): dont use hub for capturing events anymore

Litarnus:
Extracts logic into `BreadcrumbRecorder` and `EventCatcher` and replaces usage in `functions.php` to no longer use the `Hub`.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -10,6 +10,8 @@ use Sentry\Integration\IntegrationInterface;
 use Sentry\Integration\OTLPIntegration;
 use Sentry\Logs\Logs;
 use Sentry\Metrics\TraceMetrics;
+use Sentry\State\BreadcrumbRecorder;
+use Sentry\State\EventCapturer;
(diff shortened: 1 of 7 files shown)


src/functions.php
```diff
@@ -10,6 +10,8 @@
 use Sentry\Integration\OTLPIntegration;
 use Sentry\Logs\Logs;
 use Sentry\Metrics\TraceMetrics;
+use Sentry\State\BreadcrumbRecorder;
+use Sentry\State\EventCapturer;
```
stayallive, line 14: Capturer is awkward in my head maybe EventRecorder is a nice fit. But highly subjective so feel free to leave as is!

stayallive (approved):

src/functions.php
Litarnus: Yeah that makes more sense, we have `BreadcrumbRecorder` and `SpanRecorder` so that would fit

PR #2133, opened 2026-06-23
feat(scopes): reduce usage of hub further

Litarnus:
Replaces many usages of `SentrySdk::getCurrentHub()` with `SentrySdk::getClient()`


stayallive (approved):
This is starting to look more like it!

PR #2139, opened 2026-06-24
feat(scopes): introduce typed scopes

Litarnus:
Replaces one `Scope` and introduces a total of 3 scopes:
* `GlobalScope` - Cannot store spans but can store user data such as tags, contexts, etc
* `IsolationScope` - Can store spans and is used to store request local informations
* `MergedScope` - Transient scope that is the result of merging the global and isolation scope. Is considered a read-only scope and should only be used to apply data to events

### Why different types of scopes?

With only one type of scopes, the burden of knowing when to merge and which one to use where is passed to the developer all the time. While it would work with any scope and it would apply some informations, it's very easy to forget to merge scopes and only get a fraction of the required informations. With having specific types, the signature can ensure that the correct scopes are being used in the correct context. It's also easier to understand intent.

The scopes also expose different methods, meaning that it's impossible to add data to a merged scope which gets discarded afterwards.

All scopes share a common `ScopeData` object which acts as a data storage to simplify merging/cloning. `ScopeData` is effectively what was the `Scope` before but without event specific methods


src/State/GlobalScope.php
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * The scope holds data that should implicitly be sent with Sentry events. It
+ * can hold context data, extra parameters, level overrides, fingerprints etc.
+ */
+class GlobalScope extends MutableScope
+{
+    public function __construct()
+    {
+        parent::__construct();
+    }
```
stayallive: nit: we can just drop this to make it a little less confusing to read.

stayallive (approved):

src/State/GlobalScope.php
Litarnus: Goot catch

PR #2141, opened 2026-06-29
feat(spans): send gen_ai spans using the span v2 protocol

Litarnus:
Adds support for the SpanV2 protocol for `gen_ai` spans.

When a transaction contains `gen_ai` spans, they will be serialized in a different format to work around the limitations of the old protocol. The envelope will contain two different items, the first one is the transaction with regular spans and the second one is just `gen_ai` spans


giortzisg (approved):
🚀

src/Serializer/EnvelopItems/TransactionItem.php
```diff
@@ -187,4 +223,169 @@ protected static function serializeSpan(Span $span): array
...
+    protected static function serializeSpanV2(Span $span, Event $event): array
+    {
+        $result = [
+            'trace_id' => (string) $span->getTraceId(),
+            'span_id' => (string) $span->getSpanId(),
+            'name' => $span->getDescription() ?? $span->getOp(),
+            'is_segment' => false,
+            'start_timestamp' => $span->getStartTimestamp(),
+            'attributes' => self::collectV2Attributes($span, $event),
+            'status' => 'ok',
```
stayallive, line 250: This looks suspect; default to "ok" unless it's explicitly set the we override to ok/error few lines down?

src/Serializer/EnvelopItems/TransactionItem.php
```diff
@@ -187,4 +223,169 @@ protected static function serializeSpan(Span $span): array
...
+     *
+     * @phpstan-return SerializedSpanV2Attributes
+     *
+     * @mago-ignore analysis:redundant-null-coalesce
+     * @mago-ignore analysis:mixed-assignment
+     */
+    private static function collectV2Attributes(Span $span, Event $event): array
+    {
+        /** @var SerializedSpanV2Attributes $attributes */
+        $attributes = [];
```
stayallive: Shouldn't or couldn't we not use AttributeBag for this?

src/Serializer/EnvelopItems/TransactionItem.php
Litarnus: It has to be either `ok` or `error` based on the spec: https://develop.sentry.dev/sdk/telemetry/spans/span-protocol/#span-properties

This is just the case when the status is `null` and a few lines down we set it to the actual status if it's not null

stayallive: TIL, perfect then!

stayallive (approved):

PR #2152, opened 2026-07-13
docs(integrations): Update parameter type hints in BreadcrumbHandler

jszczypk:
Update PHPDoc to be consistent with PHPStan annotation. 

Missing types in PHPDoc lead to false errors in project using Phan instead of PHPStan.


Litarnus (approved):
Good catch, thank you!

PR #2159, opened 2026-07-17
perf: improves RingBuffer by using native PHP arrays

Litarnus:
Benchmarks showed that using regular PHP arrays compared to `SplFixedArray` improves pushing into the RingBuffer by around 2x and makes `toArray` significantly faster because we can utilise `array_slice` instead of iteration by hand. 

We also get free copy-on-write mechanics since that's the default with PHP arrays.


giortzisg (approved):
:shipit:

PR #2187, opened 2026-08-26
feat: ungate logs and metrics

Litarnus:
Deprecates and changes `enable_logs` and `enable_metrics` to be effectively no-ops. The reason for this change is that those flags were basically just kill switches for the entire feature and disabled by default. 
Users had to 
1. add a log handler or call log functions directly and 
2. enable the log feature to make sure they arrive. 

Metrics are even more explicit since one has to invoke functions to emit metrics and we don't have any integrations.

Once this is merged, adding a handler or calling a log/metric function will be considered as a signal that this type of telemetry is wanted, and we will no longer respect the flag.

For anyone using those flags as kill switches: it's possible to implement kill switches using `before_send_log` and `before_send_metric` and return `null` in those callbacks.

### DIFF
--- a/src/Logs/LogsAggregator.php
+++ b/src/Logs/LogsAggregator.php
@@ -51,16 +51,6 @@ final class LogsAggregator
         $options = $client->getOptions();
         $sdkLogger = $options->getLogger();
 
-        if (!$options->getEnableLogs()) {
-            if ($sdkLogger !== null) {
-                $sdkLogger->info(
-                    'Log will be discarded because "enable_logs" is "false".'
-                );
-            }
-
-            return;
-        }
-
         $formattedMessage = Str::vsprintfOrNull($message, $values);
 
         if ($formattedMessage === null) {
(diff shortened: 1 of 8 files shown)


stayallive (approved):

Branlute:
Hello, could you update the official documentation and the Symfony bundle that still uses this feature?
https://docs.sentry.io/platforms/php/logs/#setup
https://docs.sentry.io/platforms/php/guides/symfony/configuration/options/#enable_logs
https://github.com/getsentry/sentry-symfony/blob/master/src/DependencyInjection/Configuration.php#L91

Litarnus:
> Hello, could you update the official documentation and the Symfony bundle that still uses this feature? https://docs.sentry.io/platforms/php/logs/#setup https://docs.sentry.io/platforms/php/guides/symfony/configuration/options/#enable_logs https://github.com/getsentry/sentry-symfony/blob/master/src/DependencyInjection/Configuration.php#L91

Hi! I updated the docs and we will remove the option in the next major, otherwise we might crash user applications if they still define it and we don't expect it

PR #2191, opened 2026-09-04
feat: allow providing a hub when starting a runtime context

binaryfire:
This is a small follow-up to #2190.

Some async frameworks provide their own Hub because they need Sentry scope state to follow their execution model. [Hypervel](https://github.com/hypervel/components), for example, keeps the Hub’s scope stack in coroutine-local storage. That keeps tags, breadcrumbs, and active spans separate between overlapping requests while still allowing child coroutines to inherit the right state.

When Hypervel starts a runtime context today, the SDK creates a regular Hub and clones its Scope. Hypervel then immediately replaces that Hub with its coroutine-aware one. It works, but the Hub and Scope Sentry just created are never used.

This PR lets a runtime pass its existing Hub directly:

```php
SentrySdk::startContext($hub);
```

Sentry still creates and manages the runtime context, logs, metrics, storage, and flushing. The supplied Hub is used as-is, so the runtime remains responsible for isolating its state.

Nothing changes when a Hub isn’t provided. The SDK creates an isolated Hub exactly as it does today. Nested `startContext()` calls also remain no-ops and won’t replace the active Hub.

I’ve left `withContext()` unchanged because this is intended for frameworks that manage execution boundaries themselves. The existing FrankenPHP and RoadRunner fixtures use `withContext()`, so they continue to exercise the unchanged default path and don’t need updating.

### DIFF
--- a/src/SentrySdk.php
+++ b/src/SentrySdk.php
@@ -104,11 +104,30 @@ final class SentrySdk
         return $hub;
     }
 
-    public static function startContext(): void
+    /**
+     * Starts an isolated context for the current logical execution.
+     *
+     * A provided hub is used as-is, allowing runtimes with their own HubInterface
+     * implementation to manage hub isolation. When no hub is provided, the SDK
+     * creates an isolated hub from the baseline.
+     *
+     * If a context is already active, this method is a no-op and the provided hub
+     * is ignored. Use setCurrentHub() to replace the active context's hub.
+     *
+     * @param HubInterface|null $hub The hub to use for the new context
+     */
+    public static function startContext(?HubInterface $hub = null): void
     {
-        self::getRuntimeContextManager()->startContext();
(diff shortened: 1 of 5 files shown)


Litarnus:
Looks good, thank you! 

One thing to keep in mind is that the `Hub` will be replaced in favour of two different types of Scopes in the near future:
* GlobalScope - used to store data that persists between requests/units of work
* IsolationScope - request/unit of work local data

You can take a look at [this PR](https://github.com/getsentry/sentry-php/pull/2157), it's the result of many stacked PRs. It's not 100% finalised since it needs a bit of real world testing, but it should give you the general gist of how it will work in the future.

If you have any requests for changes, feel free to open an issue/PR and we can discuss it. Since this is aimed towards the new major, you can even suggest breaking changes

Litarnus (approved):

PR #2192, opened 2026-09-04
fix: Avoid PHP 8.5 warnings when serializing INF, NAN and large floats

maksimovic:
A Symfony console command in our app running on PHP 8.5 failed for an unrelated reason, but reporting that exception to Sentry raised a second issue:

`Warning: The float INF is not representable as an int, cast occurred`

...with the serializer at the top of the stack.

The command in question forces a cache recompute via `CacheInterface::get($key, $callback, INF)`, so `INF` was sitting in the frame arguments of the thrown exception.


Litarnus (approved):
Good catch, thanks for opening the PR!

PR #2194, opened 2026-09-04
feat(pii): add data collection for stack frame variables

Litarnus:
Adds data collection support for stack variables

### DIFF
--- a/src/FrameBuilder.php
+++ b/src/FrameBuilder.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Sentry;
 
+use Sentry\DataCollection\KeyValueDataFilter;
 use Sentry\Serializer\RepresentationSerializerInterface;
 use Sentry\Util\PrefixStripper;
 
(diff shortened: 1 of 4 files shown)


src/FrameBuilder.php
```diff
@@ -201,6 +202,15 @@ private function getFunctionArguments(array $backtraceFrame): array
             }
         }
 
+        $dataCollection = $this->options->getDataCollection();
+
+        if ($dataCollection !== null) {
+            $argumentValues = KeyValueDataFilter::filterKeyValueData(
```
mfb, line 208: `KeyValueDataFilter::filterKeyValueData()` is a recursive function.  But `$argumentValues` could include recursive arrays, which causes `KeyValueDataFilter::filterKeyValueData()` to recurse until the memory limit is exceeded.

For example:
```php
$foo['recursion'] =& $foo;
\Sentry\DataCollection\KeyValueDataFilter::filterKeyValueData($foo, ['mode' => 'on', 'terms' => []]);
```

